PR #314, opened 2021-08-04
Use simple tags for golang container images

zjs:

Instead of shared tags (which aggregate buster- and windowsservercore-
based images), use simple tags which use only buster-based images.

This simplifies license and provenance tracking and reporting.

Fixes: #320

Reviewers: I see that the [CONTRIBUTING.md](https://github.com/vmware-tanzu/tanzu-framework/blob/main/CONTRIBUTING.md) indicates that I MUST reference an issue here. I can file one with the same information that's in this PR. However, I noticed that some other PRs don't include issue references, so perhaps that's not required. Either way, I'd be happy to update the CONTRIBUTING.md to make the desired process here clearer — just help me understand what you're expecting.

Unscientific: verified I could pull these images from my machine, ran some affected build targets.

I'm trusting the pre-merge CI job to catch issues with this.

Reviewers: please let me know if there is other testing I should do! This change _should_ be a no-op, but I'm not confident that our existing automated testing is sufficient for all the places I've made changes.

In addition to the two comments above, I noticed two issues as a part of this change:
1. It seems like we are in some places pulling from a Harbor instance that's only reachable from the VMware VPN. (#316)
2. It seems like in one place we are using an end-of-life golang version. (#317)

I've opted not to address these as a part of this PR to limit scope.

**Does this PR introduce a [user-facing](https://github.com/vmware-tanzu/tanzu-framework/blob/main/docs/release/release-notes.md#does-my-pull-request-need-a-release-note) change?**:

```release-note
NONE
```
**New PR Checklist**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Build the manager binary
-FROM golang:1.16 as builder
+FROM golang:1.16-buster as builder
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
--- a/addons/Dockerfile
+++ b/addons/Dockerfile
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Build the manager binary
-FROM harbor-repo.vmware.com/dockerhub-proxy-cache/library/golang:1.16 as builder
+FROM harbor-repo.vmware.com/dockerhub-proxy-cache/library/golang:1.16-buster as builder
 
 WORKDIR /workspace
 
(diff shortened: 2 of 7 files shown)


imikushin:
@zjs is this PR still relevant?

imikushin:
> @zjs is this PR still relevant?

Actually, I see it still is. Let's get it rebased and merged.

vuil:
@imikushin can you followup with updating the change as well. Please also check if there are additional instances where we need to use simple tags.

randomvariable:
This is redundant now.

Issue #616, opened 2021-09-14
Dual-stack support on vSphere


christianang:
**Describe the feature request**

As an admin/operator:

I would like to be able to configure/deploy Kubernetes clusters that support dual stack networking (IPv4/IPv6)
Because:

The routing of workload traffic, both ingress and egress, may be conditional on the type of workload, and thus may also select a specific address family (IPv4 or IPv6) based on the workload type. Therefore we want to configure/deploy Kubernetes clusters that support dual-stack traffic, both ingress and egress.

**Proposal**

An admin/operator can specify `ipv4,ipv6` or `ipv6,ipv4` for the `TKG_IP_FAMILY` field. The first IP family in the list becomes the "primary" ip family of the cluster. This matters in various ways e.g SingleStack Services, default `podIP`, upgrading from a single stack to dual-stack, control plane endpoint VIP.
If specified, the `SERVICE_CIDR` and `CLUSTER_CIDR` must be a list of two CIDRs and must match the order of `TKG_IP_FAMILY`.
If `SERVICE_CIDR` or `CLUSTER_CIDR` is not specified it should default to the default `ipv4` and `ipv6` CIDRs in the correct order.

**Describe alternatives you've considered**

n/a

**Affected product area (please put an X in all that apply)**

- [ ] APIs
- [x] Addons
- [x] CLI
- [x] Docs
- [x] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Additional context**

cc @mcwumbly @tylerschultz @adobley

christianang:
All known items that we think need to be done for dual-stack have been done and dual-stack support has been added. Closing issue.

PR #729, opened 2021-10-01
Provision dual-stack bootstrap cluster when TKG_IP_FAMILY is `ipv4,ipv6`

christianang:

This allows the tanzu cli to automatically provision dual-stack bootstrap clusters when the TKG_IP_FAMILY is `ipv4,ipv6` so a user doesn't have to create their own externally and pass it to the tanzu cli using the `-e` flag.

Part of #616

Created a cluster with TKG_IP_FAMILY `ipv4,ipv6` with `tanzu management-cluster create`, without creating my own bootstrap cluster, and saw it successfully create a management cluster.

**Release note**:

```release-note
NONE
```

**Note:** Wrote `NONE` for the release note because I think the release note on #617 covers it.

**New PR Checklist**

### DIFF
--- a/pkg/v1/tkg/fakes/config/config_ipv4_ipv6.yaml
+++ b/pkg/v1/tkg/fakes/config/config_ipv4_ipv6.yaml
@@ -1,0 +1,20 @@
+providers:
+  - name: cluster-api
+    url: providers/cluster-api/v0.0.0/core-components.yaml
+    type: CoreProvider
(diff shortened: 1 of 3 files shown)


christianang:
After some more investigation it looks like the kind version in the BOM was reverted from 0.11.1 to 0.8.1. Once the BOM is updated to use 0.11.1 then the kindnet error should go away.

PR #752, opened 2021-10-03
Allow properly-annotated resources to halt cluster deletion

### LINKED ISSUE #751: Allow creation of "undeleteable" resources inside of TKG workload clusters

```
As a user or administrator of TKG
I want to mark important resources within my workload clusters as "undeleteable"
So that I don't accidentally delete important resources when an admin runs `tanzu cluster delete`
```

- Writing a wrapper around `tanzu cluster delete` that seeks "undeleteable" resources and prevents `tanzu cluster delete` from running if any were found.

- [ ] APIs
- [ ] Addons
- [X] CLI
- [ ] Docs
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Additional context**

We noticed that EBS volumes created by CAPA would get left behind once the workload cluster was blown away. EBS volumes cost money.

While the docs state that administrators should delete these resources prior to running `tanzu cluster delete`, it would be ideal if they could be marked as resources that halt cluster deletion so that the admin is gently voluntold to do it :)

carlosonunez:

This PR allows users to mark resources within workload clusters as "undeleteable". These resources will prevent workload clusters from being deleted.

Fixes #751 

- Added unit tests for issue
- Verified that annotation is valid by using a local Kind cluster
- Verified that jsonpath query is valid with a local Kind cluster

**Release note**:

```release-note

```
**New PR Checklist**

- [X] If this PR is just an idea or POC, use a [Draft PR](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests) instead of a full PR

### DIFF
--- a/pkg/v1/tkg/client/workload_cluster.go
+++ b/pkg/v1/tkg/client/workload_cluster.go
@@ -4,6 +4,7 @@
 package client
 
 import (
+	"fmt"
 	"path/filepath"
 	"time"
 
@@ -105,6 +106,15 @@ func (c *TkgClient) DeleteWorkloadCluster(options DeleteWorkloadClusterOptions) 
 
 	if options.Namespace == "" {
 		options.Namespace = constants.DefaultNamespace
+	}
+
+	resources, err := clusterClient.ListUndeleteableResources()
+	if err != nil {
+		return errors.Wrap(err, "unable to locate undeleteable resources")
+	}
+	if err == nil && len(resources) > 0 {
+		msg := fmt.Sprintf("undeleteable resources found; please delete them: %+v", resources)
+		return errors.Wrap(err, msg)
 	}
 
 	isPacific, err := clusterClient.IsPacificRegionalCluster()
--- a/pkg/v1/tkg/clusterclient/clusterclient.go
+++ b/pkg/v1/tkg/clusterclient/clusterclient.go
@@ -292,6 +292,8 @@ type Client interface {
 	DeactivateTanzuKubernetesReleases(tkrName string) error
 	// IsClusterRegisteredToTMC returns true if cluster is registered to Tanzu Mission Control
 	IsClusterRegisteredToTMC() (bool, error)
+	// ListUndeleteableResources returns a list of resources that should prevent cluster deletions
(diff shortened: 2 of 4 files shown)


carlosonunez:
Not really. When I created the feature, I had EBS vols created by CAPA in mind. Thereafter, I realized that TKG admins would likely have operators, daemonsets, and other things installed onto workload clusters that they would definitely be unhappy about seeing blown away upon `tanzu cluster delete` getting invoked.  That’s why i’m using `kubectl get all` and filtering on just the annotation instead of using the Kubernetes API directly.

carlosonunez:
That's interesting. How do you figure? I'd be okay with using the Kubernetes API directly; I was just unsure of whether I needed to create another class method to do it (and create tests for it, etc).

pkg/v1/tkg/client/workload_cluster.go
```diff
@@ -107,6 +108,15 @@ func (c *TkgClient) DeleteWorkloadCluster(options DeleteWorkloadClusterOptions)
 		options.Namespace = constants.DefaultNamespace
 	}
 
+	resources, err := clusterClient.ListUndeleteableResources()
+	if err != nil {
+		return errors.Wrap(err, "unable to locate undeleteable resources")
+	}
+	if err == nil && len(resources) > 0 {
+		msg := fmt.Sprintf("undeleteable resources found; please delete them: %+v", resources)
+		return errors.Wrap(err, msg)
```
anujc25, line 117: nit: Instead of using 2 different statements, we can use `errors.Wrapf`

anujc25 (commented):
Thanks for working on the fix @carlosonunez.

It looks like this PR also includes `cmd/cli/plugin/pinniped-auth/asset/pinniped` binary file of ~40MB. Not sure why it is not part of `.gitingore` but this PR should not include this file.

carlosonunez:
Hey, everyone. The fork that this PR was reffed to was detached when `tanzu-framework` went public, and there's no way for me to update that pointer to my public fork. Thus, I went ahead and created a new PR with the updated ref.

Apologies for the confusion!

PR #766, opened 2021-10-04
CLI package plugin support newly added tagSelection field in packageRepository

blc1996:
Kapp controller's [packageRepostitory CR](https://carvel.dev/kapp-controller/docs/latest/packaging/#packagerepository) has added a `tagSelection` field since 0.24.0

This PR supports the use of this newly added `tagSelection` field, while maintains backward compatibility. 

Putting this PR on hold until #329 is merged

Fixes # TanzuNet does not mix well with mutable tag approach for fetching package repository updates, hence lets add support for unique-tag approach

WIP

**Release note**:

```release-note
For `tanzu package repository` commands, auto detects the repository URL input. If the URL doesn't contain a tag, use the latest release semver as tag.
```
**New PR Checklist**


maralavi (approved):

blc1996:
Fore previous review comments, look at https://github.com/vmware-tanzu/tanzu-framework/pull/669.

pkg/v1/tkg/tkgpackageclient/image_utils.go
```diff
@@ -0,0 +1,84 @@
...
+	if pkgr.Spec.Fetch == nil || pkgr.Spec.Fetch.ImgpkgBundle == nil {
+		return "", "", errors.New("failed to find OCI registry URL")
+	}
+
+	repository, tag, err = parseRegistryImageURL(pkgr.Spec.Fetch.ImgpkgBundle.Image)
+	if err != nil {
+		return "", "", errors.Wrap(err, "failed to parse OCI registry URL")
+	}
+
+	if pkgr.Spec.Fetch.ImgpkgBundle.TagSelection != nil && pkgr.Spec.Fetch.ImgpkgBundle.TagSelection.Semver != nil && pkgr.Status.Fetch != nil {
```
cppforlife, line 57: keep in mind that stdout contents might shift over time (this is why its contents are not exposed as first class fields). if you are planning to parse it, you have to have a fallback. throwing examples of cases:
- apiVersion in output might be different
- directories field might change to be different (might have multiple directories, etc.)

in general i would recommend not to expose this information back to the user. if they need it they could evaluate it themselves.

blc1996: cc @vijaykatam @maralavi 
Do we want to show which tag is being used currently?
Or is it good enough to put something general like `latest release tag`?

vijaykatam: +1 to what @cppforlife said, stdout is internal detail of package repository object. Unless explicitly exposed through an API object in status lets show generic detail.

pkg/v1/tkg/tkgpackageclient/image_utils.go
```diff
@@ -0,0 +1,48 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package tkgpackageclient
+
+import (
+	"strings"
+
+	dockerParser "github.com/novln/docker-parser"
```
vijaykatam, line 9: Synced up with @blc1996 offline that we should look at removing introducing of this dependency and see if we can use imgpkg or alternate approach.

blc1996: Switched to use https://github.com/google/go-containerregistry

pkg/v1/tkg/tkgpackageclient/image_utils.go
```diff
@@ -0,0 +1,49 @@
...
+		return "", "", errors.New("failed to find OCI registry URL")
+	}
+
+	repository, tag, err = parseRegistryImageURL(pkgr.Spec.Fetch.ImgpkgBundle.Image)
+	if err != nil {
+		return "", "", errors.Wrap(err, "failed to parse OCI registry URL")
+	}
+
+	if tag == "" {
+		tag = tkgpackagedatamodel.LatestReleaseTag
```
vijaykatam, line 45: We are treating empty tag as `LATEST RELEASE` here and `Constraints:>0.0.0` for creating new repository below. But they are not exactly the same, for instance 0.1.0 can satisfy >0.0.0 but latest is probably another tag. cc @cppforlife

pkg/v1/tkg/tkgpackageclient/image_utils.go
```diff
@@ -0,0 +1,49 @@
...
+	ref, err := name.NewTag(imgURL)
+	if err != nil {
+		return "", "", err
+	}
+
+	repository = ref.Context().String()
+	tag = ref.TagStr()
+	// dockerParser sets the tag to "latest" if not specified, however we want it to be empty
+	if tag == tkgpackagedatamodel.DefaultRepositoryImageTag && !strings.HasSuffix(imgURL, ":"+tkgpackagedatamodel.DefaultRepositoryImageTag) {
+		tag = ""
```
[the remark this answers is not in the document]
blc1996: It actually does

pkg/v1/tkg/tkgpackageclient/image_utils.go
blc1996: Actually I'm not sure what would be the right word to put here.
My logic is that `Constraints:>0.0.0` will select the latest non-prerelease version, so I call the tag `LATEST RELEASE  `.

vijaykatam: Offline sync with @blc1996 and got confirmation that kapp-controller with `Constraints:>0.0.0` will pull a newer version if the current version selected is no longer latest so `Constraints:>0.0.0` ~ `latest`.

PR #774, opened 2021-10-05
CLI configuration holds edition value

### LINKED ISSUE #569: Move edition detection into configuration file (away from build-time flag)

Today, we use `BUILD_EDITION` as an [LDFLAG](https://github.com/vmware-tanzu/tanzu-framework/blob/905b9a665918f3e2b2fd4ab2ed4f124a9e34c928/Makefile#L103) to determine things like:

* Whether we present the TCE kickstart UI / standard kickstart UI
* Defaults (such as CEIP settings)

Near term, we need to continue detection of editions inside of plugins such as `management-cluster`, however, we need to move away from this happening at build time, for the following reasons:

* We need interoperability between tanzu-* and tanzu-community-edition
* TCE should be using Tanzu Framework's published binaries and **not** re-building them from source.

🌟 **Proposal** 🌟

Let's move away from LDFLAGs and add a field to the [CLIOptions struct](https://github.com/vmware-tanzu/tanzu-framework/blob/905b9a665918f3e2b2fd4ab2ed4f124a9e34c928/apis/config/v1alpha1/clientconfig_types.go#L93-L99) that specifies the edition being used.

```go
// CLIOptions are options for the CLI.
type CLIOptions struct {
	// Repositories are the plugin repositories.
	Repositories []PluginRepository `json:"repositories,omitempty" yaml:"repositories"`
	// Edition of tanzu that is being run (e.g. community, standard, advanced).
	Edition string `json:"edition,omitempty" yaml:"edition"`
	// UnstableVersionSelector determined which version tags are allowed
	UnstableVersionSelector VersionSelectorLevel `json:"unstableVersionSelector,omitempty" yaml:"unstableVersionSelector"`
}
```

This would produce a configuration file client-side, that looks as follows (`cat ~/.config/tanzu/config.yaml`).

```yaml
clientOptions:
  cli:
    repositories:
    - gcpPluginRepository:
        bucketName: tanzu-cli
        name: core
    - gcpPluginRepository:
        bucketName: tce-cli-plugins
        name: tce
        rootPath: artifacts
    unstableVersionSelector: none
    edition: community
kind: ClientConfig
metadata:
  creationTimestamp: null
```

none

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Additional context**

none

swalner-vmware:
This PR changes the "edition" (community, standard, advanced) from a build-time parameter value to a run-time configuration value.

Fixes #569 

Deployed the CLI locally and verified that (a) the "tanzu config set edition" command works, (b) the help usage texts are correct, and (c) different values for the edition were pulled from the configuration file and returned from the /edition endpoint

**Release note**:
```release-note
The CLI edition can now be set using the command "tanzu config set edition xxx", or manually in the configuration file using "edition: xxx" under the "cli" section. Valid values are "tkg", "tce" and "tce-standalone"
```
**New PR Checklist**

- [X] If this PR is just an idea or POC, use a [Draft PR](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests) instead of a full PR

### DIFF
--- a/apis/config/v1alpha1/clientconfig.go
+++ b/apis/config/v1alpha1/clientconfig.go
@@ -24,7 +24,13 @@ const (
 const (
 	// FeatureCli allows a feature to be set at the CLI level (globally) rather than for a single plugin
 	FeatureCli string = "cli"
+	// Edition value (in config) affects branding and cluster creation
+	EditionStandard  = "tkg"
+	EditionCommunity = "tce"
 )
+
+// EditionSelector allows selecting edition versions based on config file
+type EditionSelector string
 
 // VersionSelectorLevel allows selecting plugin versions based on semver properties
 type VersionSelectorLevel string
@@ -118,4 +124,22 @@ func (c *ClientConfig) SplitFeaturePath(featurePath string) (string, string, err
 		return "", "", errors.New("unsupported feature config path parameter [" + featuresLiteral + "] (was expecting 'features.<plugin>.<feature>')")
 	}
 	return plugin, flag, nil
+}
+
+// SetEditionSelector indicates the edition of tanzu to be run
+// EditionStandard is the default, EditionCommunity is also available.
+// These values affect branding and cluster creation
+func (c *ClientConfig) SetEditionSelector(edition EditionSelector) {
+	if c.ClientOptions == nil {
+		c.ClientOptions = &ClientOptions{}
+	}
+	if c.ClientOptions.CLI == nil {
+		c.ClientOptions.CLI = &CLIOptions{}
+	}
+	switch edition {
+	case EditionCommunity, EditionStandard:
+		c.ClientOptions.CLI.Edition = edition
+		return
+	}
+	c.ClientOptions.CLI.UnstableVersionSelector = EditionStandard
 }
--- a/pkg/v1/tkg/client/cluster.go
+++ b/pkg/v1/tkg/client/cluster.go
@@ -552,7 +552,7 @@ func (c *TkgClient) ConfigureAndValidateWorkloadClusterConfiguration(options *Cr
 	// BUILD_EDITION is the Tanzu Edition, the plugin should be built for. Its value is supposed be constructed from
 	// cmd/cli/plugin/managementcluster/create.go. So empty value at this point is not expected.
 	if options.Edition == "" {
-		return NewValidationError(ValidationErrorCode, "required config variable 'BUILD_EDITION' is not set")
+		return NewValidationError(ValidationErrorCode, "required config variable 'edition' is not set")
 	}
 	c.SetBuildEdition(options.Edition)
 	c.SetTKGClusterRole(options.ClusterType)
--- a/pkg/v1/tkg/cmd/config_cluster.go
+++ b/pkg/v1/tkg/cmd/config_cluster.go
@@ -83,7 +83,7 @@ func runGenerateClusterConfig(name string, options *createClusterOptions) error 
 }
 
 func buildCreateClusterOption(name, tkrVersion string, options *createClusterOptions) tkgctl.CreateClusterOptions {
-	cfg, _ := config.GetClientConfig()
+	edition, _ := config.GetEdition()
 	return tkgctl.CreateClusterOptions{
 		ClusterConfigFile:           options.clusterConfigFile,
 		ClusterName:                 name,
@@ -102,6 +102,6 @@ func buildCreateClusterOption(name, tkrVersion string, options *createClusterOpt
 		VsphereControlPlaneEndpoint: options.vsphereControlPlaneEndpoint,
 		SkipPrompt:                  options.unattended || skipPrompt,
 		Timeout:                     options.timeout,
-		Edition:                     cfg.ClientOptions.CLI.Edition,
+		Edition:                     edition,
 	}
 }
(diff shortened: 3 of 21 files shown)


pkg/v1/tkg/client/cluster.go
```diff
@@ -520,7 +520,7 @@ func (c *TkgClient) ConfigureAndValidateWorkloadClusterConfiguration(options *Cr
 	// BUILD_EDITION is the Tanzu Edition, the plugin should be built for. Its value is supposed be constructed from
 	// cmd/cli/plugin/managementcluster/create.go. So empty value at this point is not expected.
 	if options.Edition == "" {
-		return NewValidationError(ValidationErrorCode, "required config variable 'BUILD_EDITION' is not set")
+		return NewValidationError(ValidationErrorCode, "required config variable 'edition' is not set")
```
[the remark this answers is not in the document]
imikushin: Thanks for the clarification!

imikushin (approved):
LGTM!

pkg/v1/tkg/tkgctl/create_cluster.go
```diff
@@ -305,9 +306,9 @@ func (t *tkgctl) configureCreateClusterOptionsFromConfigFile(cc *CreateClusterOp
 
 	// set BuildEdition from config variable
 	if cc.Edition == "" {
-		edition, err := t.TKGConfigReaderWriter().Get(constants.ConfigVariableBuildEdition)
+		cfg, err := config.GetClientConfig()
```
anujc25, line 309: We should try not to get an edition from config file at this place as this is part of the library code and we might want the library users to provide an edition with the `BUILD_EDITION` config variable or through the library interface.

If we think we don't want to allow users to specify `BUILD_EDITION` config variable then we need to communicate with library consumers to always pass `edition` as part of the `CreateClusterOptions` or `InitRegionOptions` argument.

pkg/v1/tkg/tkgctl/init.go
```diff
@@ -276,9 +277,9 @@ func (t *tkgctl) configureInitManagementClusterOptionsFromConfigFile(iro *InitRe
 
 	// set BuildEdition from config variable
 	if iro.Edition == "" {
-		edition, err := t.TKGConfigReaderWriter().Get(constants.ConfigVariableBuildEdition)
+		cfg, err := config.GetClientConfig()
```
anujc25, line 280: Same as above. (We should try not to get an edition from config file at this place as this is part of the library code and we might want the users to provide an edition with the `BUILD_EDITION` config variable)

anujc25 (commented):
Overall looks good. Added comment regarding using the tanzu config file(`config.GetClientConfig()`) inside the library interface.

pkg/v1/tkg/tkgctl/create_cluster.go
swalner-vmware: @anujc25 , we'll continue to allow the BUILD_EDITION flag in the cluster config file as before (no change).

pkg/v1/tkg/cmd/config_cluster.go
```diff
@@ -100,5 +102,6 @@ func buildCreateClusterOption(name, tkrVersion string, options *createClusterOpt
 		VsphereControlPlaneEndpoint: options.vsphereControlPlaneEndpoint,
 		SkipPrompt:                  options.unattended || skipPrompt,
 		Timeout:                     options.timeout,
+		Edition:                     edition,
```
tenczar, line 105: nit: this change shouldn't be necessary. This cmd was for the old tkg cli.

tenczar (approved):
LGTM

apis/config/v1alpha1/clientconfig.go
```diff
@@ -16,6 +16,16 @@ const (
 	NoUnstableVersions VersionSelectorLevel = "none"
 )
 
+// Edition value (in config) affects branding and cluster creation
+const (
+	EditionTkg           = "tkg"
+	EditionTce           = "tce"
+	EditionTceStandalone = "tce-standalone"
```
joshrosso, line 23: We decided this is an implementation detail and can be removed, correct?

apis/config/v1alpha1/clientconfig.go
```diff
@@ -16,6 +16,16 @@ const (
 	NoUnstableVersions VersionSelectorLevel = "none"
 )
 
+// Edition value (in config) affects branding and cluster creation
+const (
+	EditionTkg           = "tkg"
+	EditionTce           = "tce"
```
joshrosso, line 22: s/tce/community/g

joshrosso (changes requested):
Aside from small nits, I have one more check I want to run by the framework team *before* we merge this functionality.

As I read it, this PR:

**Sets the edition to `tkg` by default and requires intervention to switch to community**

Is Framework sure this is the approach we want to take? To be clear, we're setting the default edition to that of a downstream product. To me, i'd advise considering:

1. Prompting if edition isn't set (this is why i recommended it be considered for `init`)
- or - 
2. Default to `community` rather than `tkg`

Hope this helps and sorry to keep blocking/delaying this PR @swalner-vmware your patience is super appreciated.

joshrosso:
> I propose that I create two other GitHub issues:

> Change the value of the TCE edition from 'tce' to 'community' (and TKG from 'tkg' to 'standard'), and
> Change the default edition from TKG to TCE

This makes sense to me.

joshrosso (approved):

PR #775, opened 2021-10-05
IPv6 for vSphere support on kickstart UI 

raymondz1:

Fixes #

**Release note**:

```release-note

```
**New PR Checklist**

- [ ] Ensure PR contains only public links or terms


mcwumbly:
I think this is release note worthy so I edited the OP to include one. feel free to edit further.

pspinrad:
+1 for tooltip, and from looking at screenshot above, I'm wondering about **Select version to use throughout this installer** because 1) **version** has so many other meanings elsewhere in UX, could be confusing, and **throughout this installer** implies that you're only selecting IPv4 vs IPv6 to use while the installer itself runs, not permanently for the mgmt and other clusters that you're creating-- unless I'm confused, which is entirely likely...
I do get that installer-internal checks etc. change based on the choice, but from user POV I think they'd see it as a choice that persists beyond the lifetime of the installer.
If so, maybe **Select IP version** or **IP version for internal communication** would be clearer.
WDYT? Thanks!

pspinrad:
Thanks @mcwumbly - that's what I thought, that the IPv4 / IPv6 choice will apply to the mgmt cluster that you're creating, and in turn to the workload clusters that it can create. But when I first read "...to use throughout this installer" it made me think that maybe it actually doesn't, and that I was confused. So I want the label to convey that the choice does indeed persist after the installer itself is done. WDYT would be a good label to make that clear?

Re: guardrails-- do you mean e.g. that someone can specify IPv4 as their IP family choice, but then nothing in the UI prevents them from entering IPv6 addresses for endpoints, CIDRs, etc.-- which won't work, so the "...to use throughout this installer" is a way of telling them not to?

If so, I totally didn't see that interpretation at first, but it makes sense now. To address this, WDYT of simplifying the heading per above and putting something like "This choice dictates the required IP format for addresses and CIDRs in the rest of the installer" in the Tooltip, and then for every IP address and CIDR field tooltip later, add something like "IP format must match the IP Family (IPv4 or IPv6) chosen under **IaaS Provider**"  And then in the docs, we can reinforce this, explicitly say that the values you enter need to be consistent with the IP family you chose?

mcwumbly:
> Re: guardrails-- do you mean e.g. that someone can specify IPv4 as their IP family choice, but then nothing in the UI prevents them from entering IPv6 addresses for endpoints, CIDRs, etc.-- which won't work, so the "...to use throughout this installer" is a way of telling them not to?

No, sorry. _Those_ guardrails exist. In the context of installing a single cluster, this choice applies _and is enforced_ throughout the installer.

What we _don't_ enforce today via the UI or other validation is the following: after deploying an IPv6 management cluster, we don't _prevent_ you from _trying_ to deploy an IPv4 workload cluster. That will fail slowly instead of failing fast.

PR #776, opened 2021-10-05
Fix the template file download issue when CLI is executed first time

prkalle:
Signed-off-by: PremKumar Kalle <pkalle@vmware.com>

This PR fixes the template file downloaad issue when CLI is executed. This is caused because the providers configuration from the TKG config file was not read after the providers are downloaded from the image repository.

Fixes #

1) Deleted the $HOME/.config directory on my test machine and ran `tanzu cluster create testcluster -f <file_with_required_config>` command with `-d` option and It generated the templates successfully
2) Ran the `tanzu cluster create` command ran in the first testcase and noticed that the providers are not downloaded and the templates are generated successfuly.
3) created a cluster and it was successfull

**Release note**:

```release-note
Fix the template file download issue when CLI is executed first time which is caused due to the provider config from TKG config fie is not updated after the providers are downloaded
```
**New PR Checklist**

- [ ] Ensure PR contains only public links or terms


pkg/v1/tkg/tkgctl/client.go
```diff
@@ -144,6 +144,15 @@ func New(options Options) (TKGClient, error) { //nolint:gocritic
...
 		return nil, errors.Wrap(err, "unable to ensure prerequisites")
 	}
+	tkgConfigFile, err := allClients.TKGConfigPathsClient.GetTKGConfigPath()
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to get TKG config file path")
+	}
+	err = tkgClient.TKGConfigReaderWriter().Init(tkgConfigFile)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to initialize the TKG config reader writer")
+	}
```
anujc25, line 154: might be good to add a comment to describe why we are doing this as it is not clear from the function invocation.

PR #793, opened 2021-10-07
add validation to TKG_HTTP_PROXY_ENABLED when HTTP_PROXY is present

gwang550:
add validation to TKG_HTTP_PROXY_ENABLED when HTTP_PROXY is present. Management cluster can still be deployed with proxy without setting TKG_HTTP_PROXY_ENABLED to true, however, the CAP* component no_proxy will miss import IP such as service_cidr/cluster_cidr, etc.

Fixes #

Add testcases, manually deployed the cluster. 

**Release note**:

```release-note

```
**New PR Checklist**


pkg/v1/tkg/client/validate.go
```diff
@@ -1426,11 +1426,16 @@ func (c *TkgClient) EncodeAWSCredentialsAndGetClient(clusterClient clusterclient
...
 	proxyEnabled, err := c.TKGConfigReaderWriter().Get(constants.TKGHTTPProxyEnabled)
-	if err != nil || proxyEnabled != trueString {
+	httpProxy, _ := c.TKGConfigReaderWriter().Get(constants.TKGHTTPProxy)
+	if (err != nil || proxyEnabled != trueString) && httpProxy == "" {
 		return nil
 	}
+	// httpProxy and httpsProxy are presents, check if the TKGHTTPProxyEnabled
+	if err != nil || proxyEnabled != trueString {
+		return errors.Wrapf(err, "cannot get %s", constants.TKGHTTPProxyEnabled)
+	}
```
nicolehanjing, line 1436: just a minor one, maybe we can rewrite it to something like this:
```
if err != nil || proxyEnabled != trueString {
    if httpProxy == "" {
        return nil
    }
    return errors.Wrapf(err, "cannot get %s", constants.TKGHTTPProxyEnabled)
}
```

PR #819, opened 2021-10-08
Provision dual-stack bootstrap cluster when TKG_IP_FAMILY is `ipv4,ipv6`

christianang:

This allows the tanzu cli to automatically provision dual-stack bootstrap clusters when the TKG_IP_FAMILY is ipv4,ipv6 so a user doesn't have to create their own externally and pass it to the tanzu cli using the -e flag. This also enables a user to delete a management cluster without creating a bootstrap cluster as well.

Part of #616 

Created a cluster with TKG_IP_FAMILY `ipv4,ipv6` with `tanzu management-cluster create`, without creating my own bootstrap cluster, and saw it successfully create a management cluster.

Note: This does require the kind version to be 0.11.1 in the BOM. Currently this is not the case and will be required for this to work. I believe there is some parallel work happening to make this true.

I also deleted the same cluster with `tanzu management-cluster delete`, without creating my own bootstrap cluster, and saw it successfully delete the management cluster.

This replaces an earlier closed PR: #729

**Release note**:

```release-note
Bootstrap clusters will be dual-stack when deploying a dual-stack management-cluster.
```

**New PR Checklist**

### DIFF
--- a/pkg/v1/providers/infrastructure-vsphere/v0.7.10/ytt/overlay.yaml
+++ b/pkg/v1/providers/infrastructure-vsphere/v0.7.10/ytt/overlay.yaml
@@ -88,13 +88,11 @@ spec:
     name: #@ "{}-control-plane".format(data.values.CLUSTER_NAME)
   clusterNetwork:
     pods:
-      cidrBlocks:
-      #@overlay/match by=overlay.index(0)
-      - #@ data.values.CLUSTER_CIDR
+      #@overlay/replace
+      cidrBlocks: #@ data.values.CLUSTER_CIDR.split(",")
     services:
-      cidrBlocks:
-      #@overlay/match by=overlay.index(0)
-      - #@ data.values.SERVICE_CIDR
+      #@overlay/replace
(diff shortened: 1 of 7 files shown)


mcwumbly (approved):

mcwumbly:
> Note: This does require the kind version to be 0.11.1 in the BOM. Currently this is not the case and will be required for this to work. I believe there is some parallel work happening to make this true.

This is too bad, but I think it may be best to merge this in the meantime anyway.

In the meantime, one can workaround this by doing the following:

```bash
# dump the provider yaml to disk
tanzu management-cluster create || true

# replace the kind version with  v1.21.5+vmware.1_v0.11.1
sed -i '/kubernetes-sigs_kind/!b;n;c\  - version: v1.21.5+vmware.1_v0.11.1'  ~/.config/tanzu/tkg/bom/tkg-bom-v1.5.0-framework-v0.5.0.yaml

# create your cluster
tanzu management-cluster create -f myconfig.yaml
```

anujc25 (approved):
lgtm. Thanks for the changes.

PR #831, opened 2021-10-11
Enhance docker pre-req checks to verify CPU and Memory settings

miclettej:
Uses docker system info to verify user has minimum 4 CPU allocated and 6GB memory allocated in docker settings

This pre-requisite check has been added to the UI and CLI right after docker daemon is verified to be running

Screen Shot 2021-10-11 at 4 27 11 PM

Screen Shot 2021-10-11 at 4 15 14 PM

Screen Shot 2021-10-11 at 4 15 36 PM

Fixes https://github.com/vmware-tanzu/tanzu-framework/issues/687
Tested using combination of valid and invalid pre-req settings in Docker desktop for both UI and CLI

**Release note**:
```release-note
Docker pre-requisite checks now validate that user has sufficient memory and CPU settings selected. Minimum CPU required is 4 and minimum Total Memory is 6GB.
```
**New PR Checklist**


pkg/v1/tkg/client/common.go
```diff
@@ -140,6 +201,24 @@ func (c *TkgClient) validateDockerPrerequisites() error {
 	if !isDockerDaemonRunning {
 		return errors.New("docker daemon is not running, Please make sure Docker daemon is up and running")
 	}
+	// validate docker allocated CPU and memory against recommended minimums
+	if validateDockerResources {
+		if dockerResourceCpus, err = checkDockerResourceCpu(); err != nil {
+			return errors.Wrap(err, "Failed to check docker minimum number of CPUs")
+		}
+
+		if !(dockerResourceCpus >= 4) {
```
saji-pivotal, line 210: nit-ish: The conditional can be changed to `if dockerResourceCPUs < 4`

pkg/v1/tkg/client/common.go
```diff
@@ -81,6 +82,65 @@ func checkDockerDaemonIsRunning() (bool, error) {
 	return true, nil
 }
 
+func checkDockerResourceCpu() (int, error) {
```
saji-pivotal, line 85: IMO `checkDockerResourceCpu()` and `checkDockerResourceMemory()` can be combined into a single validation function like `checkDockerResource(resourceType string)` and that function can be called from validateDockerPrerequisites as follows 

```
func (c *TkgClient) ValidatePrerequisites {
  ...
  if dockerResourceCpus, err = checkDockerResource("NCPU"); err != nil {
   ... 
 }
 if  dockerResourceTotalMemory, err = checkDockerResource("MemTotal"); err != nil {
  ...
 }
}
```

Please let me know what you think

pkg/v1/tkg/client/common.go
```diff
@@ -140,6 +201,24 @@ func (c *TkgClient) validateDockerPrerequisites() error {
 	if !isDockerDaemonRunning {
 		return errors.New("docker daemon is not running, Please make sure Docker daemon is up and running")
 	}
+	// validate docker allocated CPU and memory against recommended minimums
+	if validateDockerResources {
```
saji-pivotal, line 205: I'm just curious. Why are these checks optional? Why are they not part of the prerequisites?

saji-pivotal (changes requested):
@miclettej Nice work overall! I do have a few questions though? 
Along with the other review comments, is it possible to add some unit tests for this change?

pkg/v1/tkg/client/common.go
miclettej: @saji-pivotal The original checks for docker daemon running serve two purposes. First, we are validating that docker is running prior to executing the CLI create command for any provider. Second, this is repurposed for the UI to check if the daemon is running at the start of the docker workflow...the step referenced in the screen shots. So what we have to do is determine what is the source of triggering this validation to occur. If it is coming from the UI to validate that the daemon is running, then we also want to confirm the user has the appropriate resource minimums set. 

We would want to bypass those memory/cpu resource checks when the create command is run for AWS, vSphere, Azure.

miclettej: Agreed, much better approach. Thank you for the suggestion.

pkg/v1/tkg/client/common.go
```diff
@@ -144,6 +172,32 @@ func (c *TkgClient) validateDockerPrerequisites() error {
 	return nil
 }
 
+func (c *TkgClient) ValidateDockerResourcePrerequisites() error {
+	var dockerResourceCpus, dockerResourceTotalMemory int
+	var err error
+
+	// validate docker allocated CPU and memory against recommended minimums
+	if dockerResourceCpus, err = checkDockerResource("NCPU"); err != nil {
```
saji-pivotal, line 180: nit: Can we initialize "NCPU" and "MemTotal" as string constants, and pass those constants?

pkg/v1/tkg/client/common.go
```diff
@@ -144,6 +172,32 @@ func (c *TkgClient) validateDockerPrerequisites() error {
...
+		return errors.Wrap(err, "Failed to check docker minimum number of CPUs")
+	}
+
+	if !(dockerResourceCpus >= 4) {
```
saji-pivotal, line 184: nit: if dockerResourceCPUs < 4

pkg/v1/tkg/client/common.go
```diff
@@ -144,6 +172,32 @@ func (c *TkgClient) validateDockerPrerequisites() error {
 	return nil
 }
 
+func (c *TkgClient) ValidateDockerResourcePrerequisites() error {
```
saji-pivotal, line 175: Q: I'm not sure if we should add this here or in https://github.com/vmware-tanzu/tanzu-framework/blob/main/pkg/v1/tkg/client/validate.go . IMO it would be worth a quick discussion

pkg/v1/tkg/client/common.go
miclettej: Ah yes, this is a question I forgot to ask when we synched over zoom. I just pushed a commit that initializes these as constants within the function. I meant to ask, is there a better place that we should initialize constants that would be consumed by this function...maybe somewhere we would import constants from. Or is it okay to just initialize within the function as I have in this commit.

PR #864, opened 2021-10-13
[Draft] Addons should upgrade before cluster upgrade

blc1996:

The addons should upgrade prior to cluster upgrade to account for forward compatibility.
i.e. some old addons may not run on the nodes with new k8s version. The new addons will be only scheduled on new nodes, so we don't need to worry about backward compatibility as it's a rolling upgrade

the cluster upgrade process
1. Upgrades cluster-api resources
1. Some special pre upgrade handling
1. Updates management cluster node by patching the version of KubeadmControlPlane cluster-api resource. Once this is patched, a machine and a node with the new version will be reconciled. The CLI wait's for cluster-api's Ready condition before proceeds.
1. Updates workload cluster node by patching the version of MachineDeployment cluster-api resource. It will create a new worker node. CLI waits on MachineDeployment until it's successfully reconciled.
1. Upgrade addons. Including patching kapp-controller, TKR, addons manager.
 
Calico:  The CRD definitions are using **apiextensions.k8s.io/v1beta1** API version. This will make the CRD unusable once the cluster is upgraded to 1.22.0. 
The CLI will wait for the old CNI package running on the new node to mark the node as ready, it will block waits on Calico which will never come up.
Due to the issue of Calico, the process will be stuck at step 3. KubeadmControlPlane will never be ready because it is missing CNI.

Fixes N+1 compatibility of Core addons with k8s 1.22

POC
Manually replicated the same upgrade situation, the upgrade was successful with this order change.

**Release note**:

```release-note

```
**New PR Checklist**

### DIFF
--- a/addons/controllers/addon_reconcilers.go
+++ b/addons/controllers/addon_reconcilers.go
@@ -205,9 +205,9 @@ func (r *AddonReconciler) ReconcileAddonDataValuesSecretNormal(
 			addonDataValuesSecret.Data["imageInfo.yaml"] = imageInfoBytes
 		}
 		// Add kubernetesVersion info
-		TKRVersionBytes, err := util.GetKubernetesVersionInfo(bom)
+		TKRVersionBytes, err := util.GetTKRVersionInfo(bom)
 		if err != nil {
-			log.Error(err, "Error retrieving Kubernetes version info from BOM")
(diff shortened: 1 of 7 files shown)


blc1996:
Migrated and merged in #959

Issue #74, opened 2021-07-07
Umbrella: API-driven discovery of CLI plugins


iancoffey:
**Constituent Issues**
- https://github.com/vmware-tanzu/tanzu-framework/pull/873
- https://github.com/vmware-tanzu/tanzu-framework/issues/945
- https://github.com/vmware-tanzu/tanzu-framework/issues/946
- https://github.com/vmware-tanzu/tanzu-framework/issues/947
- https://github.com/vmware-tanzu/tanzu-framework/issues/958
- https://github.com/vmware-tanzu/tanzu-framework/issues/1035
- https://github.com/vmware-tanzu/tanzu-framework/issues/1110
- https://github.com/vmware-tanzu/tanzu-framework/pull/1011
- https://github.com/vmware-tanzu/tanzu-framework/pull/1072
- https://github.com/vmware-tanzu/tanzu-framework/issues/1150
- https://github.com/vmware-tanzu/tanzu-framework/issues/1169
- https://github.com/vmware-tanzu/tanzu-framework/issues/1164
- https://github.com/vmware-tanzu/tanzu-framework/issues/1168
- https://github.com/vmware-tanzu/tanzu-framework/issues/1204
- https://github.com/vmware-tanzu/tanzu-framework/issues/1208
- https://github.com/vmware-tanzu/tanzu-framework/issues/1264
- https://github.com/vmware-tanzu/tanzu-framework/issues/1376
- https://github.com/vmware-tanzu/tanzu-framework/issues/1500
- https://github.com/vmware-tanzu/tanzu-framework/issues/1671
- https://github.com/vmware-tanzu/tanzu-framework/issues/1764

**Describe the feature request**
In moving toward packaging things as `imgpkg`, the clearest way of packaging up a CLI plugin will be to make an API for it.

`extensions.tanzu.vmware.com`
```go
// CLIPlugin is the Schema for the plugins API
type CLIPlugin struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec   CLIPluginSpec   `json:"spec,omitempty"`
	Status CLIPluginStatus `json:"status,omitempty"`
}

// CLIPluginSpec defines the desired state of Plugin
type CLIPluginSpec struct {
	// Description of the plugin
	Description string `json:"description"`

	// Version of the plugin
	Version string `json:"version"`

	// Artifacts for the plugin
[...]

PR #873, opened 2021-10-13
Add CLIPlugin API interfaces and base implementation

anujc25:

- Use CLI side feature gates introduced with #841 (`tanzu config set features.global.use-context-aware-discovery true`) instead of direct config
- Refactor `pkg/v1/cli/catalog.go` and move not catalog functions to `pkg/v1/cli/cmd.go`
- Add interface and base implementation for catalog
- Add discovery interface
- Add pluginmanager package and base implementation

Related to: #74

- Add unit tests for catalog package
- Add unit/integration tests for pluginmanager package

**Release note**:

```release-note
NONE
```
**New PR Checklist**

### DIFF
--- a/config/crd/bases/config.tanzu.vmware.com_clientconfigs.yaml
+++ b/config/crd/bases/config.tanzu.vmware.com_clientconfigs.yaml
@@ -90,7 +90,8 @@ spec:
                             type: string
                           path:
                             description: Path is a local path pointing to directory
-                              containing a list of CLIPlugin API resources
+                              containing YAML files, each of which contains single
+                              CLIPlugin API resource.
                             type: string
                         required:
                         - name
@@ -104,7 +105,8 @@ spec:
                             description: Image is an OCI compliant image. Which include
                               DNS-compatible registry name, a valid URI path(MAY contain
                               zero or more ‘/’) and a valid tag. E.g., harbor.my-domain.local/tanzu-cli/plugins-manifest:latest
-                              Contains list of CLIPlugin API resources
+                              Contains a directory containing YAML files, each of
+                              which contains single CLIPlugin API resource.
                             type: string
                           name:
                             description: Name is a name of the discovery
@@ -245,7 +247,8 @@ spec:
                           type: string
                         path:
                           description: Path is a local path pointing to directory
-                            containing a list of CLIPlugin API resources
+                            containing YAML files, each of which contains single CLIPlugin
+                            API resource.
                           type: string
                       required:
                       - name
@@ -259,7 +262,8 @@ spec:
                           description: Image is an OCI compliant image. Which include
                             DNS-compatible registry name, a valid URI path(MAY contain
                             zero or more ‘/’) and a valid tag. E.g., harbor.my-domain.local/tanzu-cli/plugins-manifest:latest
-                            Contains list of CLIPlugin API resources
+                            Contains a directory containing YAML files, each of which
+                            contains single CLIPlugin API resource.
                           type: string
                         name:
                           description: Name is a name of the discovery
--- a/pkg/v1/cli/artifact/local.go
+++ b/pkg/v1/cli/artifact/local.go
@@ -14,13 +14,16 @@ import (
 
 // LocalArtifact defines local artifact path
 type LocalArtifact struct {
+	// Path is path to local binary artifact
+	// if path is not an absolute path search under
+	// `xdg.ConfigHome/tanzu-plugin/localPath` directory
 	Path string
 }
 
 // NewLocalArtifact creates Local Artifact object
+// If path is not an absolute path
+// search under `xdg.ConfigHome/tanzu-plugin/distribution` directory
 func NewLocalArtifact(path string) Artifact {
-	// If path is not an absolute path
-	// search under `xdg.ConfigHome/tanzu-plugin/localPath` directory
 	if !filepath.IsAbs(path) {
 		path = filepath.Join(common.DefaultLocalPluginDistroDir, "distribution", path)
 	}
@@ -29,11 +32,11 @@ func NewLocalArtifact(path string) Artifact {
 	}
 }
 
-// Fetch an artifact.
+// Fetch reads the local artifact from its path
 func (l *LocalArtifact) Fetch() ([]byte, error) {
 	b, err := os.ReadFile(l.Path)
 	if err != nil {
-		return nil, errors.Wrapf(err, "error while reading manifest file")
+		return nil, errors.Wrapf(err, "error while reading artifact")
 	}
 	return b, nil
 }
--- a/pkg/v1/cli/artifact/oci.go
+++ b/pkg/v1/cli/artifact/oci.go
@@ -17,5 +17,6 @@ func NewOCIArtifact(image string) Artifact {
 
 // Fetch an artifact.
 func (g *OCIArtifact) Fetch() ([]byte, error) {
+	// TODO(anujc25): implement OCI artifact fetch
 	return nil, nil
 }
--- a/pkg/v1/cli/catalog/catalog.go
+++ b/pkg/v1/cli/catalog/catalog.go
@@ -60,9 +60,9 @@ func NewContextCatalog(context string) (*ContextCatalog, error) {
 }
 
 // Upsert inserts/updates the given plugin.
-func (c *ContextCatalog) Upsert(plugin cliv1alpha1.PluginDescriptor) error { //nolint:gocritic
+func (c *ContextCatalog) Upsert(plugin *cliv1alpha1.PluginDescriptor) error {
 	c.plugins[plugin.Name] = plugin.InstallationPath
-	c.sharedCatalog.IndexByPath[plugin.InstallationPath] = plugin
+	c.sharedCatalog.IndexByPath[plugin.InstallationPath] = *plugin
 
 	if !utils.ContainsString(c.sharedCatalog.IndexByName[plugin.Name], plugin.InstallationPath) {
 		c.sharedCatalog.IndexByName[plugin.Name] = append(c.sharedCatalog.IndexByName[plugin.Name], plugin.InstallationPath)
@@ -71,7 +71,7 @@ func (c *ContextCatalog) Upsert(plugin cliv1alpha1.PluginDescriptor) error { //n
 	return saveCatalogCache(c.sharedCatalog)
 }
 
-// Get looks up the given plugin.
+// Get looks up the descriptor of a plugin given its name.
 func (c *ContextCatalog) Get(plugin string) (cliv1alpha1.PluginDescriptor, bool) {
 	pd := cliv1alpha1.PluginDescriptor{}
 	path, ok := c.plugins[plugin]
@@ -88,6 +88,8 @@ func (c *ContextCatalog) Get(plugin string) (cliv1alpha1.PluginDescriptor, bool)
 }
 
 // List returns the list of active plugins.
+// Active plugin means the plugin that are available to the user
+// based on the current logged-in server.
 func (c *ContextCatalog) List() []cliv1alpha1.PluginDescriptor {
 	pds := make([]cliv1alpha1.PluginDescriptor, 0)
 	for _, installationPath := range c.plugins {
--- a/pkg/v1/cli/common/defaults.go
+++ b/pkg/v1/cli/common/defaults.go
@@ -20,7 +20,9 @@ var (
 	// DefaultLocalPluginDistroDir is the default Local plugin distribution root directory
 	// This directory will be used for local discovery and local distribute of plugins
 	DefaultLocalPluginDistroDir = filepath.Join(xdg.ConfigHome, "tanzu-plugin")
+)
 
-	// DefaultDistro is the core set of plugins that should be included with the CLI.
-	DefaultDistro = []string{"login", "pinniped-auth", "cluster", "management-cluster", "kubernetes-release"}
+const (
+	// IsContextAwareDiscoveryEnabled defines default to use when the user has not configured a value
+	IsContextAwareDiscoveryEnabled = false
 )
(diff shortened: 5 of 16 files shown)


pkg/v1/config/clientconfig.go
```diff
@@ -399,3 +400,31 @@ func EndpointFromServer(s *configv1alpha1.Server) (endpoint string, err error) {
...
+
+	fm, exists := cfg.ClientOptions.Features["global"]
+	if !exists {
+		return contextAwareDiscoveryEnabled
+	}
+
+	contextAwareDiscoveryEnabledString, exists := fm["useContextAwareDiscovery"]
+	if !exists {
+		return contextAwareDiscoveryEnabled
+	}
```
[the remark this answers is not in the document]
anujc25: As mentioned here in L409 the default value is set to `false` for `contextAwareDiscoveryEnabled` variable.

But I will rename the variable to make it more clear.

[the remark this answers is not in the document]
giri-varma: My bad! missed the first line.. LGTM then

giri-varma (approved):
LGTM! Thanks for making the change

config/crd/bases/config.tanzu.vmware.com_clientconfigs.yaml
```diff
@@ -85,16 +85,16 @@ spec:
...
-                          manifestPath:
-                            description: ManifestPath is a local path pointing to
-                              manifest file
-                            type: string
                           name:
                             description: Name is a name of the discovery
                             type: string
+                          path:
+                            description: Path is a local path pointing to directory
+                              containing a list of CLIPlugin API resources
```
zjs, line 93: **nitpick**: It seems imprecise to say that a directory contain a list. Typically, a directory contains files :smile:

I think this supposed to be something like: a directory containing YAML files, each of which contain one or more CLIPlugin API resources? (But I haven't finished reading the PR and there are some potentially important details to note here: can a given file contain multiple CLIPlugin API resources? are non-YAML files ignored? are other resources in the files ignored?)

config/crd/bases/config.tanzu.vmware.com_clientconfigs.yaml
```diff
@@ -103,8 +103,8 @@ spec:
                           image:
                             description: Image is an OCI compliant image. Which include
                               DNS-compatible registry name, a valid URI path(MAY contain
-                              zero or more ‘/’) and a valid tag. Contains a manifest
-                              file E.g., harbor.my-domain.local/tanzu-cli/plugins-manifest:latest
+                              zero or more ‘/’) and a valid tag. E.g., harbor.my-domain.local/tanzu-cli/plugins-manifest:latest
+                              Contains list of CLIPlugin API resources
```
zjs, line 107: **nitpick**: As above, can this be more precise? (How are the CLIPlugin API resources contained within the OCI image?)

pkg/v1/cli/artifact/interface.go
```diff
@@ -0,0 +1,18 @@
...
+
+// Artifact is an interface to download a single plugin binary.
+type Artifact interface {
+	// Fetch the binary for a plugin version.
+	Fetch() ([]byte, error)
+}
+
+// NewURIArtifact creates new artifacts based on the URI
+func NewURIArtifact(uri string) Artifact {
+	// TODO: Support other artifact types
```
zjs, line 16: **nitpick**: Should we track this TODO in an issue?

pkg/v1/cli/artifact/local.go
```diff
@@ -0,0 +1,39 @@
...
+	"path/filepath"
+
+	"github.com/pkg/errors"
+
+	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli/common"
+)
+
+// LocalArtifact defines local artifact path
+type LocalArtifact struct {
+	Path string
```
zjs, line 17: **nitpick**: It seems like we expect this to be an absolute path. Perhaps that's worth noting in the godoc?

pkg/v1/cli/artifact/local.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// NewLocalArtifact creates Local Artifact object
+func NewLocalArtifact(path string) Artifact {
+	// If path is not an absolute path
+	// search under `xdg.ConfigHome/tanzu-plugin/localPath` directory
```
zjs, line 23: **nitpick**: It seems like we allow both absolute and relative paths, and that there's some important information about how a relative path is handled. I'd capture all of that in the method-level comment so that someone looking at the godoc for the method can easily see it (without having to click into the method definition).

pkg/v1/cli/artifact/local.go
```diff
@@ -0,0 +1,39 @@
...
+	if !filepath.IsAbs(path) {
+		path = filepath.Join(common.DefaultLocalPluginDistroDir, "distribution", path)
+	}
+	return &LocalArtifact{
+		Path: path,
+	}
+}
+
+// Fetch an artifact.
```
zjs, line 32: **nitpick**: The comment isn't wrong, but also doesn't feel very useful. Also, I think "fetch" is the right verb to use for the interface, but it feels like a weird verb in the context of local files, so perhaps the comment could help map the idea of "fetch" to "read"?

Something like:
```suggestion
// Fetch reads the local artifact from its path
```

pkg/v1/cli/artifact/local.go
```diff
@@ -0,0 +1,39 @@
...
+func (l *LocalArtifact) Fetch() ([]byte, error) {
+	b, err := os.ReadFile(l.Path)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error while reading manifest file")
```
zjs, line 36: **nitpick**: This isn't reading a manifest file, is it?

pkg/v1/cli/artifact/oci.go
```diff
@@ -0,0 +1,22 @@
...
+// NewOCIArtifact creates OCI Artifact object
+func NewOCIArtifact(image string) Artifact {
+	return &OCIArtifact{
+		Image: image,
+	}
+}
+
+// Fetch an artifact.
+func (g *OCIArtifact) Fetch() ([]byte, error) {
+	// TODO(anujc25): implement OCI artifact fetch
```
zjs, line 20: **nitpick**: Do/should we have an issue tracking this? Could we reference that here?

pkg/v1/cli/catalog/catalog.go
```diff
@@ -0,0 +1,227 @@
...
+	apimachineryjson "k8s.io/apimachinery/pkg/runtime/serializer/json"
+
+	cliv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/cli/v1alpha1"
+	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli/common"
+	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkg/utils"
+)
+
+const (
+	// catalogCacheFileName is the name of the file which holds Catalog cache
+	// TODO: Use the original catalog file instead of using v2 once the feature is enabled by default
```
zjs, line 21: **thought**: This seems like an easy thing to forget to do. Is there some way to more clearly associate it with making the feature enabled by default? (E.g., perhaps leaving a comment over in clientconfig.go referencing this file? Or extracting the default `isContextAwareDiscoveryEnabled` into a const and adding a little unit test that fails if that const becomes `true` and this `catalogCacheFileName` const hasn't been updated?)

pkg/v1/cli/catalog/catalog.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	return &ContextCatalog{
+		sharedCatalog: sc,
+		plugins:       plugins,
+	}, nil
+}
+
+// Upsert inserts/updates the given plugin.
+func (c *ContextCatalog) Upsert(plugin cliv1alpha1.PluginDescriptor) error { //nolint:gocritic
```
zjs, line 63: **nitpick**: I like including reasons with my `//nolint` comments. (E.g., `//nolint:gocritic // reason …`)

I can't actually tell what linter warning you're suppressing here, and I suspect some future readers of the code may not be able to either; if there's a reason, that may be a hint about whether the method has changed enough that the nolint should be removed.

(If you're looking at the code in an IDE, it's always easy to remove the comment and see what happens, but sometimes we read the code from the GitHub UI, and then the reader is left feeling curious. :smile:)

pkg/v1/cli/catalog/catalog.go
```diff
@@ -0,0 +1,227 @@
...
+	c.sharedCatalog.IndexByPath[plugin.InstallationPath] = plugin
+
+	if !utils.ContainsString(c.sharedCatalog.IndexByName[plugin.Name], plugin.InstallationPath) {
+		c.sharedCatalog.IndexByName[plugin.Name] = append(c.sharedCatalog.IndexByName[plugin.Name], plugin.InstallationPath)
+	}
+
+	return saveCatalogCache(c.sharedCatalog)
+}
+
+// Get looks up the given plugin.
```
zjs, line 74: **nitpick**: The comment is accurate, but I don't find it to be super helpful.

Maybe something like:
```suggestion
// Get looks up the descriptor of a plugin given its name.
```

Maybe the interface's godoc could be updated with this info as well.

pkg/v1/cli/catalog/catalog.go
```diff
@@ -0,0 +1,227 @@
...
+
+	pd, ok = c.sharedCatalog.IndexByPath[path]
+	if !ok {
+		return pd, false
+	}
+
+	return pd, true
+}
+
+// List returns the list of active plugins.
```
zjs, line 90: **nitpick**: I think this comment has a really important piece of information in it: that the list is of _active_ plugins. I don't think I would have inferred that from just reading this method's implementation.

I think it'd be even better to say a little more. What determines whether a plugin is "active"? Is "active" a well-defined term in this context? (Is the term defined somewhere already? Could this comment refer to that one?)

Maybe the interface's godoc could be updated with this info as well.

pkg/v1/cli/catalog/catalog.go
```diff
@@ -0,0 +1,227 @@
...
+	pds := make([]cliv1alpha1.PluginDescriptor, 0)
+	for _, installationPath := range c.plugins {
+		pd := c.sharedCatalog.IndexByPath[installationPath]
+		pds = append(pds, pd)
+	}
+	return pds
+}
+
+// Delete deletes the given plugin from the catalog, but it does not delete
+// the installation.
```
zjs, line 101: **nitpick**: This is another case where your comment has key information in it. The distinction between plugins being installed and being included in the catalog seems important. Is there somewhere that should be explained? (E.g., the package-level godoc, in the interface file?)

pkg/v1/cli/common/defaults.go
```diff
@@ -0,0 +1,26 @@
...
+
+	// DefaultPluginRoot is the default plugin root.
+	DefaultPluginRoot = filepath.Join(xdg.DataHome, "tanzu-cli")
+
+	// DefaultLocalPluginDistroDir is the default Local plugin distribution root directory
+	// This directory will be used for local discovery and local distribute of plugins
+	DefaultLocalPluginDistroDir = filepath.Join(xdg.ConfigHome, "tanzu-plugin")
+
+	// DefaultDistro is the core set of plugins that should be included with the CLI.
+	DefaultDistro = []string{"login", "pinniped-auth", "cluster", "management-cluster", "kubernetes-release"}
```
zjs, line 25: **question**: Don't we have a list like this already? Is this duplicative?

pkg/v1/cli/discovery/local.go
```diff
@@ -0,0 +1,126 @@
...
+// LocalDiscovery is a artifact discovery endpoint utilizing a local host os.
+type LocalDiscovery struct {
+	path string
+	name string
+}
+
+// NewLocalDiscovery returns a new local repository.
+func NewLocalDiscovery(name, localPath string) Discovery {
+	// If path is not an absolute path
+	// search under `xdg.ConfigHome/tanzu-plugin/localPath` directory
```
zjs, line 30: **nitpick**: As above, I think it's worth combining this all into one comment.

pkg/v1/cli/discovery/oci.go
```diff
@@ -0,0 +1,49 @@
...
+func (od *OCIDiscovery) List() (plugins []plugin.Discovered, err error) {
+	// TODO(anujc25): implement OCI discovery plugin list
+	return
+}
+
+// Describe a plugin.
+func (od *OCIDiscovery) Describe(name string) (p plugin.Discovered, err error) {
+	// TODO(anujc25): implement OCI discovery plugin describe
+	return
+}
```
zjs, line 39: **nitpick**: Is there an issue tracking this work that you could reference?

pkg/v1/cli/distribution/interface.go
```diff
@@ -0,0 +1,14 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package distribution implements plugin distribution interface
```
zjs, line 4: **nitpick**: I think it'd be worth including some additional context here — perhaps you can copy a paragraph or two straight from your design doc.

pkg/v1/cli/discovery/interface.go
```diff
@@ -0,0 +1,41 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package discovery is implements discovery interface for plugin discovery
```
zjs, line 4: **nitpick**: I think it'd be worth including some additional context here — perhaps you can copy a paragraph or two straight from your design doc.

pkg/v1/cli/catalog/catalog.go
anujc25: Addressed by adding constant and a unit test.

vuil (approved):
LGTM, thanks for changes!

If a couple of the fix-up type commits can be squashed into the more logical ones without too much trouble, that would be nice. But given the size of the change and there are not too many of such commits in the commit chain, I wouldn't block the merge because of it.

PR #906, opened 2021-10-18
Update scale API to operate on node pools

tenczar:
This change updates the scale api to accept a node-pool name so that only
that node pool will be scaled. Tests have been added for this new
feature. Updates the cli to support a new flag to pass the node-pool
name.

This PR updates the scale API to support scaling a node pool by name. 

Fixes #

Added tests for scale API. 

### Release note

```release-note
Update Scale API to accept a node-pool name, which will scale just the desired node-pool to the desired number of replicas. 
```

### PR Checklist

### DIFF
--- a/pkg/v1/tkg/client/scale_test.go
+++ b/pkg/v1/tkg/client/scale_test.go
@@ -4,11 +4,13 @@
 package client_test
 
 import (
+	"fmt"
 	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/cluster-api/api/v1alpha3"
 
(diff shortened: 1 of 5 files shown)


pkg/v1/tkg/client/scale_test.go
```diff
@@ -123,86 +127,181 @@ var _ = Describe("Scale API", func() {
...
+
+						scaleClusterOptions.NodePoolName = md1Name
+						scaleClusterOptions.WorkerCount = 3
+
+						err = tkgClient.DoScaleCluster(regionalClusterClient, &scaleClusterOptions)
+						Expect(err).Should(MatchError(fmt.Sprintf("Could not find node pool with name %s", scaleClusterOptions.NodePoolName)))
+					})
+				})
+				When("the node pool fails to be updated", func() {
+					It("should update the replicas of that node pool", func() {
```
prkalle, line 292: nit: should it be `should not update the replicas of that node pool` or `should return error`?

PR #918, opened 2021-10-19
Allow adding custom root CA certificate for Proxy

### LINKED ISSUE #917: Allow adding custom root CA certificate through configuration variable when using proxy
The user must be able to specify a custom root CA certificate for the proxy if the proxy server performs SSL interception with the custom root CA. 

The CA cert involved with proxy is used to make the connection to the proxy itself, regardless of the purpose of that connection (e.g., containerd pulling an agent image). The use case is that the user's proxy does traffic inspection and so connections from the clusters to the proxy have to use a CA cert that the proxy knows.

We have the `TKG_CUSTOM_IMAGE_REPOSITORY_CA_CERTIFICATE` configuration variable which specifies the CA certificate of the registry that we are using. One option is to use the same for proxy CA certificate.

One issue is that, because CLI itself needs to talk to registry to download image and CLI client can use different proxy configuration than the cluster that we are deploying, we need a different way to provide CA certificate for CLI client and cluster configuration and using `TKG_CUSTOM_IMAGE_REPOSITORY_CA_CERTIFICATE` for does not work.

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [x] User Experience

**Additional context**

To solve this issue we can introduce two more configuration variables `PROXY_CA_CERT` and `TKG_PROXY_CA_CERT` to be used for CLI/Library and Cluster configuration respectively.

anujc25:
Allow adding custom root CA certificate through configuration variable when using a proxy. 

- Introduces two more configuration variables `PROXY_CA_CERT` and `TKG_PROXY_CA_CERT` to be used for CLI/Library and Cluster configuration respectively.

Fixes #917

- Added `TKG_PROXY_CA_CERT` to clustergen tests

### Release note

```release-note
Allow adding custom root CA certificate for proxy environments with `TKG_PROXY_CA_CERT`.  `TKG_PROXY_CA_CERT` takes precedence over `TKG_CUSTOM_IMAGE_REPOSITORY_CA_CERTIFICATE` for the custom registry.
```

### PR Checklist


pkg/v1/tkg/kind/client.go
```diff
@@ -394,3 +376,38 @@ func (k *KindClusterProxy) getIPFamily() (kindv1.ClusterIPFamily, error) {
 		return "", fmt.Errorf("TKG_IP_FAMILY should be one of %s, %s, %s, got %s", kindv1.IPv4Family, kindv1.IPv6Family, kindv1.DualStackFamily, normalisedIPFamily)
 	}
 }
+
+func (k *KindClusterProxy) setupProxyConfigurationForKindCluster() {
+	var httpProxy, httpsProxy, noProxy, tkgHttpProxy, tkgHttpsProxy, tkgNoProxy string
+
+	httpProxy, _ = k.options.Readerwriter.Get(constants.HTTPProxy)
```
vuil, line 383: nit: this is also a behavioral change, I suggest a small comment to say we are respecting the non-"TKG_" proxy settings as well.

pkg/v1/providers/ytt/lib/helpers.star
```diff
@@ -98,7 +98,7 @@ def tkg_image_repo_skip_tls_verify():
 end
 
 def tkg_image_repo_ca_cert():
-  return data.values.TKG_CUSTOM_IMAGE_REPOSITORY_CA_CERTIFICATE
+  return data.values.TKG_PROXY_CA_CERT if data.values.TKG_PROXY_CA_CERT else data.values.TKG_CUSTOM_IMAGE_REPOSITORY_CA_CERTIFICATE
```
vuil, line 101: I suggest adding TKG_PROXY_CA_CERT to some clustergen model file (e.g. optional.model) so the effect of the overlay changes can be seen in the generated yaml outputs.

pkg/v1/tkg/tkgconfigbom/bom.go
```diff
@@ -337,14 +337,29 @@ func (c *client) IsCustomRepositorySkipTLSVerify() bool {
...
-	if err != nil {
-		// return empty content when not specified
+// GetCustomRepositoryCaCertificateForClient returns CA certificate to use with cli client
+// This function reads the CA certificate from following variables in below orders:
+// 1. PROXY_CA_CERT
+// 2. TKG_PROXY_CA_CERT
+// 3. TKG_CUSTOM_IMAGE_REPOSITORY_CA_CERTIFICATE
+func (c *client) GetCustomRepositoryCaCertificateForClient() ([]byte, error) {
+	caCert := ""
+	value1, err1 := c.TKGConfigReaderWriter().Get(constants.ProxyCACert)
```
vuil, line 347: nit: more readable if we name the valueX variables appropriately. proxyCACertValue, tkgProxyCACertValue, etc.

I suggest changing the comment to say that the variables are used in decreasing order of precedence or something like that.

pkg/v1/tkg/tkgconfigbom/client.go
```diff
@@ -51,7 +51,12 @@ type Client interface {
 	GetCurrentTKGVersion() string
 	GetCustomRepository() (string, error)
 	IsCustomRepositorySkipTLSVerify() bool
-	GetCustomRepositoryCaCertificate() ([]byte, error)
+	// GetCustomRepositoryCaCertificateForClient returns CA certificate to use with cli client
+	// This function reads the CA certificate from following variables in below orders:
```
vuil, line 55: See comment above

vuil (commented):
I have a few minor comments.
There is a also a linter comment being really opinionated on how you should write the acronyms in your variables :-)

vijaykatam:
I found [this explanation](https://github.com/vmware-tanzu/tanzu-framework/issues/917#issuecomment-946904472) pretty helpful to understand the variables being introduced. I think we should call out the behavior change in the release note that `TKG_PROXY_CA_CERT` takes precedence over `TKG_CUSTOM_IMAGE_REPOSITORY_CA_CERTIFICATE` for custom registry.

pkg/v1/tkg/kind/client.go
anujc25: Sure. Added description as part of the function.

pkg/v1/tkg/tkgconfigbom/bom.go
anujc25: Updated

pkg/v1/providers/ytt/lib/helpers.star
anujc25: Added clustergen tests. https://storage.googleapis.com/tkg-clustergen/918/20211020044536/clustergen.diff.txt

PR #919, opened 2021-10-19
auto populate defaults when there's only a single choice

raymondz1:
show the option automatically if there is only one choice. 

Fixes # https://github.com/vmware-tanzu/tanzu-framework/issues/692

### Release note

```release-note

```

### PR Checklist


pkg/v1/tkg/web/src/app/views/landing/aws-wizard/node-setting-step/node-setting-step.component.ts
```diff
@@ -289,21 +289,25 @@ export class NodeSettingStepComponent extends StepFormDirective implements OnIni
...
 
                 prodFields.forEach(attr => this.disarmField(attr, true));
+                if (this.nodeAzs && this.nodeAzs.length === 1) {
+                    this.formGroup.get('awsNodeAz1').setValue(this.nodeAzs[0].name);
+                }
 
                 if (this.clusterType !== 'standalone') {
-                    this.resurrectField('workerNodeInstanceType1', [Validators.required]);
+                    this.resurrectField('workerNodeInstanceType1', [Validators.required],
+                        this.azNodeTypes.awsNodeAz1.length === 1 ? this.azNodeTypes.awsNodeAz1[0] : '');
```
swalner-vmware, line 298: Maybe we should say that if there is only one option (which we will automatically select) then there's no need for a validator?

pkg/v1/tkg/web/src/app/views/landing/azure-wizard/vnet-step/vnet-step.component.ts
```diff
@@ -273,11 +279,16 @@ export class VnetStepComponent extends StepFormDirective implements OnInit {
         let filteredSubnets = this.controlPlaneSubnets.filter(s => s.name === this.getSavedValue('controlPlaneSubnet', ''));
         if (filteredSubnets.length > 0) {
             this.formGroup.get('controlPlaneSubnet').setValue(filteredSubnets[0].cidr)
+        } else if (this.controlPlaneSubnets.length === 1) {
+            this.formGroup.get('controlPlaneSubnet').setValue(this.controlPlaneSubnets[0].name)
```
swalner-vmware, line 283: nit: I would tend to start with the `if (this.controlPlaneSubnets.length === 1)` and put the filter and assignment in the `else` block.

pkg/v1/tkg/web/src/app/views/landing/vsphere-wizard/resource-step/resource-step.component.ts
```diff
@@ -116,23 +125,16 @@ export class ResourceStepComponent extends StepFormDirective implements OnInit {
                     }
                     if (source === TkgEventType.GET_VM_FOLDERS) {
                         this.resurrectField('vmFolder',
-                            [Validators.required, this.validationService.isValidNameInList(data.map(vmFolder => vmFolder.name))]);
+                            [Validators.required, this.validationService.isValidNameInList(
+                                data.map(vmFolder => vmFolder.name))], data.length === 1 ? data[0].name : '');
```
swalner-vmware, line 129: nit: this format is a bit dense to parse (for me). I would tend to do the `data.map(vmFolder => vmFolder.name))]` first, and then not even add the validator in the case where `data.length === 1`

pkg/v1/tkg/web/src/app/views/landing/aws-wizard/node-setting-step/node-setting-step.component.ts
raymondz1: @swalner-vmware  we have an empty option in all dropdown menus for unselecting. If we removed it, it won't be consistent design. In addition, the empty option is useful for the optional fields. so I think the current approach is fine.

pkg/v1/tkg/web/src/app/views/landing/azure-wizard/vnet-step/vnet-step.component.ts
raymondz1: @swalner-vmware we have to make sure that there is at least one item in filteredSubnet array. otherwise filteredSubnets[0] will throw error.

PR #935, opened 2021-10-20
Adds default feature flags (when missing) to config file

### LINKED ISSUE #933: Enhance CLI feature gate for existing config files
The current implementation of configuration feature gates is that when there is no config file, we use the defaults hard-coded in the release. However, if there exists a configuration file, we use whatever values are in that file. This means a newly-added hard-coded feature (value of true) would not be activated for a user with a pre-existing config file.
To handle this, we intend to add all default feature gate values to the user's configuration file, unless they already have an entry.

We thought about simply using the default (hard-coded) value when the config file lacks an entry, but this is "invisible" behavior, and would not expose beta features (ie feature gates explicitly set false) because the default value would not make it into a user's existing configuration file.

- [ ] APIs
- [ ] Addons
- [X] CLI
- [ ] Docs
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

swalner-vmware:
The current implementation of configuration feature gates is that when there is no config file, we use the defaults hard-coded in the release. However, if there exists a configuration file, we use whatever values are in that file. This means a newly-added hard-coded feature (value of true) would not be activated for a user with a pre-existing config file.
To handle this, we intend to add all default feature gate values to the user's configuration file, unless they already have an entry.

Fixes #933 

Testing locally to see that default values populate config file

### PR Checklist

### DIFF
--- a/config/crd/bases/config.tanzu.vmware.com_clientconfigs.yaml
+++ b/config/crd/bases/config.tanzu.vmware.com_clientconfigs.yaml
@@ -166,6 +166,14 @@ spec:
                     Plugin API based plugin discovery mechanism
                   type: boolean
               type: object
+            features:
+              additionalProperties:
+                additionalProperties:
+                  type: string
+                description: FeatureMap is simply a hash table, but needs an explicit
+                  type to be an object in another hash map (cf ClientOptions.Features)
+                type: object
+              type: object
           type: object
         current:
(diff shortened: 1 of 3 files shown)


anujc25 (approved):
LGTM. Thanks for putting effort to fix this.

PR #955, opened 2021-10-22
Extend CABPK timeout to 30m to account for etcd cluster settling in vSphere environments.

### LINKED ISSUE #954: Increase the value of CABPK --bootstrap-token-ttl to account for etcd replication time in vSphere clusters
**Bug description**

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**
See https://github.com/kubernetes-sigs/cluster-api/issues/5477

Deploy a 3-node CP cluster using CAPV with kube-vip using PhotonOS (for some reason, it's more likely to occur here) and then set kcp.spec.upgradeAfter to trigger a rollout.

Due to a mixture of kube-vip/kube-vip#214 and the fact we haven't yet implemented etcd learner mode in kubeadm or have full support in kubernetes, etcd leader switches around many times, with kube-vip leader election also rotating. During this time, CAPI controllers are unable to fully reconcile, and neither can kubelet register nodes. Importantly, CABPK is also unable to renew the bootstrap token. Eventually, etcd replication completes but after the 15 minute bootstrap token timeout. kubelet node registration ultimately fails and we end up with an orphaned control plane machine which is a valid member of the etcd cluster, complicating cleanup.

TKG v1.4.0

vSphere 7 U2
Photon v1.21 OVAs

As @fabriziopandini notes, the `--bootstrap-token-ttl` is configurable. I think it makes sense to bump that TTL given vSphere is one of the most common deployments  for Tanzu.

randomvariable:

Fixes #954 

### Release note

```release-note

```

### PR Checklist

### DIFF
--- a/pkg/v1/providers/bootstrap-kubeadm/v0.3.23/bootstrap-components.yaml
+++ b/pkg/v1/providers/bootstrap-kubeadm/v0.3.23/bootstrap-components.yaml
@@ -3924,6 +3924,7 @@ spec:
         - --upstream=http://127.0.0.1:8080/
         - --logtostderr=true
         - --v=10
+        - --bootstrap-token-ttl=30m
         image: registry.tkg.vmware.run/cluster-api/kube-rbac-proxy:${KUBE_RBAC_PROXY_IMAGE_TAG}
         name: kube-rbac-proxy
         ports:


pydctw:
Changes looks good to me. 

@vuil
1) Is there a command to re-trigger failing tests?
2) Looks like I don't have an access to view this file - https://storage.googleapis.com/tkg-clustergen/955/20211022143020/clustergen.diff.tx.t.

alfredthenarwhal:
> @pydctw:  /test
Commit: 9d08a6e

Tests failed! Build no: 955

vuil:
> Changes looks good to me.
> 
> @vuil
> 
> 1. Is there a command to re-trigger failing tests?

You can click the failed check(s), and click "rerun all jobs" on the top right. Are you seeing the option? I just reran the failed checks, and it does look like this change may be responsible for some integration/e2e tests failing

> 2. Looks like I don't have an access to view this file - https://storage.googleapis.com/tkg-clustergen/955/20211022143020/clustergen.diff.txt

It's not that. The UX of the link can possibly but improved, but if the PR does not result in any differences in any cluster configuration generation, the file will show blank today. 
As for why it is blank for _this_ change, this is because clustergen only captures the dry-run output of "cluster create", iow that which is produced via yttprocessor processing the infra-specific cluster template along with the associated overlays.
So changes in yaml deploying actual CAPI components such as this PR are not captured, unfortunately.

pydctw:
> You can click the failed check(s), and click "rerun all jobs" on the top right. Are you seeing the option? I just reran the failed checks, and it does look like this change may be responsible for some integration/e2e tests failing
I don't see "rerun all jobs" option but will keep looking. This is the first time I am reviewing PRs in tanzu-framework so trying to get familiarized with different label/commands.

Thanks for the explanation on the clustergen.diff file. Makes sense.

PR #986, opened 2021-10-26
Service cidr size validation

flawedmatrix:

Kube-apiserver validates the provided service CIDR is not too large.
When provided an invalide service cidr, the kube-apiserver fails
to come up and causes the cluster to hang.

This PR duplicates the service CIDR validation so that cluster creation
with an invalid CIDR fails fast.

The kube-apiserver validation can be seen here:
https://github.com/kubernetes/kubernetes/blob/3c87c43ceff6122637c8d8070601f7271026360e/cmd/kube-apiserver/app/options/validation.go#L52

passes unit tests

### PR Checklist


vuil (approved):
lgtm, thanks!

PR #1013, opened 2021-10-28
add dualstack feature flags

### LINKED ISSUE #920: Add CLI feature gate for dual stack management cluster and workload cluster creation

- Add two flags to control whether dual stack management clusters and workload clusters can be created:
   - `tanzu config set features.management-cluster.dual-stack true`
   - `tanzu config set features.cluster.dual-stack true`

Initial support for creating dual stack management and workload clusters was merged in #617.

But there are outstanding changes needed for the feature (#616) to be considered complete, for example #819. Other work in progress includes updating `cloud-provider-vsphere` to a version that supports dual stack, CoreDNS, and the KinD library, etc.

We'd like to be able to continue to iterate and merge our changes into the main branch to enable users to give it a spin and provide feedback while clearly communicating it being a work in progress.

Currently, our dual stack feature is opt-in because users have to specify `"ipv4,ipv6"` as the value for `TKG_IP_FAMILY` so we feel pretty safe doing this. But now that #841 has been merged, we'd like to leverage CLI feature gates to make this more explicit.

We think it makes sense to add two feature flags to begin with. While management cluster creation and workload cluster creation are somewhat related, they are also independent to some degree and implemented by different plugins. 

We are not sure how likely it is, but it's conceivable that some users may wish to experiment only with dual stack management clusters or workload clusters.

- Add a single global CLI flag to control whether dual stack management clusters and workload clusters can be created:
   - `tanzu config set features.global.dual-stack-cluster true`
       - this might be easier for users but it feels less "correct" since it really only applies to two plugins
- Use a CLI flag for the management cluster, and a [`FeatureGate`](https://github.com/vmware-tanzu/tanzu-framework/blob/main/docs/api-machinery/features-and-featuregates.md) on the management cluster for the workload cluster feature
   -  This might be better in the long run, but seems more risky and the FeatureGate itself may need some design work. We can add this later if desired, and then remove the CLI flag for workload clusters if it is no longer necessary.

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] Installation
- [x] Plugin
- [ ] Security
- [ ] Test and Release
- [x] User Experience

**Additional context**

adobley:
We have added four new feature flags for dualstack:
"features.management-cluster.dual-stack-ipv4-primary"
"features.management-cluster.dual-stack-ipv6-primary"
"features.workload-cluster.dual-stack-ipv4-primary"
"features.workload-cluster.dual-stack-ipv6-primary"

These features default to "false" and will need to be enabled for users
to use dualstack functionality while it is still under development.

Fixes #920 

Tested combinations of the flag settings and `TKG_IP_FAMILY` configurations when running the tanzu commands to validate that the flags work as expected.

### Release note

```release-note
Experimental dual-stack functionality can be enabled with a feature flag.
```

### PR Checklist


adobley:
Fixed a conflict.
@vmware-tanzu/tanzu-framework-reviewers Can someone take a look at this? It would be great to get this flag in.

vuil (commented):
Thanks for the changes. I think the feature flag paths needs an update, but otherwise lgtm.

pkg/v1/config/clientconfig.go
```diff
@@ -24,6 +24,14 @@ const (
 	// to use the new context-aware Plugin API based plugin discovery mechanism
 	// Users can set this featureflag so that we can have context-aware plugin discovery be opt-in for now.
 	FeatureContextAwareDiscovery = "features.global.context-aware-discovery"
+	// DualStack feature flags determine whether it is permitted to create
+	// clusters with a dualstack TKG_IP_FAMILY.  There are separate flags for
+	// each primary, "ipv4,ipv6" vs "ipv6,ipv4", and flags for management vs
+	// workload cluster plugins.
+	FeatureFlagManagementClusterDualStackIPv4Primary = "features.management-cluster.dual-stack-ipv4-primary"
+	FeatureFlagManagementClusterDualStackIPv6Primary = "features.management-cluster.dual-stack-ipv6-primary"
+	FeatureFlagWorkloadClusterDualStackIPv4Primary   = "features.workload-cluster.dual-stack-ipv4-primary"
```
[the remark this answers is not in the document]
adobley: Ohhh, thank you! I'll fix this now.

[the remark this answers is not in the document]
adobley: Fixed and rebased! Thanks!

vuil (approved):
lgtm, thanks!

pkg/v1/tkg/client/validate.go
```diff
@@ -1800,3 +1806,31 @@ func (c *TkgClient) validateNameservers(nameserverConfigVariable string) error {
...
+		if !dualIPv6PrimaryEnabled && ipFamily == constants.DualStackPrimaryIPv6Family {
+			return dualStackFeatureFlagError(ipFamily, config.FeatureFlagClusterDualStackIPv6Primary)
+		}
+	}
+
+	return nil
+}
+
+func dualStackFeatureFlagError(ipFamily, featureFlag string) error {
+	return fmt.Errorf("option TKG_IP_FAMILY is set to %q, but dualstack support is under development. To enable dualstack, set %s to \"true\"", ipFamily, featureFlag)
```
swalner-vmware, line 1835: I would tweak this message slightly to:
"option TKG_IP_FAMILY is set to %q, but dualstack support is not enabled (because it is under development)..."

pkg/v1/tkg/tkgctl/client.go
```diff
@@ -277,3 +280,11 @@ func ensureConfigImages(configDir string, tkgConfigUpdater tkgconfigupdater.Clie
 
 	return tkgConfigUpdater.EnsureConfigImages()
 }
+
+func getFeatureFlagClient() client.FeatureFlagClient {
+	featureFlagClient, err := config.GetClientConfig()
+	if err != nil {
+		featureFlagClient = &configv1alpha1.ClientConfig{}
```
swalner-vmware, line 287: I don't know the context here, but I'm concerned about failing silently if there's an error getting the config. Is there any way to surface a warning as to what this error was?

pkg/v1/tkg/client/validate.go
adobley: Updated! Thanks for the suggestion!

pkg/v1/tkg/tkgctl/client.go
adobley: We thought that it would only get here if the file containing the feature flags didn't exist, so that would mean there are no feature flags to worry about. Though we moved this code into the AllClients bit and now return an error which feels like a safer choice. Thanks for the suggestion!

PR #1020, opened 2021-10-29
increase default poll timeout value

maralavi:
- This PR is to increase default poll timeout value to accommodate better default timeout for packages that install many other packages

Existing UT and integration tests

None

**Release note**:

```release-note
increases default poll timeout value to prevent failure in package installation (due to time out) when a package installs many other packages
```

**New PR Checklist**

### DIFF
--- a/pkg/v1/tkg/tkgpackagedatamodel/constants.go
+++ b/pkg/v1/tkg/tkgpackagedatamodel/constants.go
@@ -11,7 +11,7 @@ const (
 	ClusterRoleName                     = "%s-%s-cluster-role"
 	DefaultAPIVersion                   = "install.package.carvel.dev/v1alpha1"
 	DefaultPollInterval                 = 1 * time.Second
-	DefaultPollTimeout                  = 5 * time.Minute
+	DefaultPollTimeout                  = 15 * time.Minute
 	DefaultRepositoryImageTag           = "latest"
 	DefaultRepositoryImageTagConstraint = ">0.0.0"
 	ErrPackageNotInstalled              = "package install does not exist in the namespace"


shivaani0505 (approved):
LGTM

danniel1205:
thanks for this PR, @maralavi . Do we have an issue linked to this PR ? And could you elaborate a little bit on what problems we will have if not increasing the default poll timeout ? Thanks

maralavi:
> thanks for this PR, @maralavi . Do we have an issue linked to this PR ? And could you elaborate a little bit on what problems we will have if not increasing the default poll timeout ? Thanks

@danniel1205 It was understood as part of tap bash experience. It's local and can not put the links.
As mentioned in the PR description, increasing the default polltimeout helps in cases where a package installs various other packages and normally needs a longer wait for the reconciliation before timing out. Also kapp reconciliation timeout is also 15m

zjs:
>  increases default poll timeout value

Thank you for including a release note for this! I wonder, though, if the average user will be able to understand what this means.

Looking at the rest of your PR description, the phrase "for packages that install many other packages" seems significant. Maybe a more complete release note could include something like that?

I don't have enough context to know exactly what this does, but maybe something like this?

> increases default poll timeout value to prevent installation from failing when a package installs many other packages

maralavi:
> > increases default poll timeout value
> 
> Thank you for including a release note for this! I wonder, though, if the average user will be able to understand what this means.
> 
> Looking at the rest of your PR description, the phrase "for packages that install many other packages" seems significant. Maybe a more complete release note could include something like that?
> 
> I don't have enough context to know exactly what this does, but maybe something like this?
> 
> > increases default poll timeout value to prevent installation from failing when a package installs many other packages

Thanks @zjs for your comment. Your description indeed improves the release note message. I just changed mine, thank you!

PR #1034, opened 2021-11-01
update secret registry readme

maralavi:
- This PR is to update secret registry README to add the workflow for adding a private package repository and a installation of a private package 

Not applicable

None

**Release note**:

```release-note
updates secret registry README to add the workflow for adding a private package repository using package and secret plugins
```

**New PR Checklist**

### DIFF
--- a/cmd/cli/plugin/secret/README.md
+++ b/cmd/cli/plugin/secret/README.md
@@ -171,3 +171,63 @@ the "secret registry" can be used to:
      Updated registry secret 'test-secret' in namespace 'test-ns'
      Unexported registry secret 'test-secret' from all namespaces
    ```
+
+## Workflow for adding a private package repository and installation of a private package
+
+You can add a private package repository and install a private package using the following procedure:
+
+1. First, create the namespace in which the secret is getting added to:
+
+   ```sh
+   kubectl create namespace <NAMESPACE>
+   ```
+
+2. Before adding a private package repository, registry secret should be added to the cluster:
+
+   ```sh
+   tanzu secret registry add <SECRET-NAME> --server <PRIVATE-REGISTRY> --username <USERNAME> --namespace <SECRET-NAMESPACE> --password <PASSWORD> -y
+   ```
+
+3. In case you want to add the private repository in a different namespace than the namespace in which the secret was added to, you need o export the secret to all other namespaces. Please be aware that by doing so, the given secret contents will be available to ALL users in ALL namespaces. Please ensure that included registry credentials allow only read-only access to the registry with minimal necessary scope. You can export the secret to other namespaces by running:
+
+   ```sh
+   tanzu secret registry update <SECRET-NAME> --namespace <SECRET-NAMESPACE> --export-to-all-namespaces=true -y
+   ```
+
+4. Add the private package repository to the target namespace in which you want to install the private package by running:
+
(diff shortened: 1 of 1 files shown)


cmd/cli/plugin/secret/README.md
```diff
@@ -171,3 +171,63 @@ the "secret registry" can be used to:
...
+## Workflow for adding a private package repository and installation of a private package
+
+You can add a private package repository and install a private package using the following procedure:
+
+1. First, create the namespace in which the secret is getting added to:
+
+   ```sh
+   kubectl create namespace <NAMESPACE>
+   ```
+
```
shivaani0505, line 184: Here we are mentioning 2 & 3 separately(only one way). 
We can start by saying there are 2 ways of adding/exporting a secret after Step 1:
1. 2& 3 separately ( where 2 adds secret and 3 updates secret by exporting to a diff namespace)
or
specify a single one line command for doing adding/exporting together

cmd/cli/plugin/secret/README.md
```diff
@@ -171,3 +171,63 @@ the "secret registry" can be used to:
...
+2. Before adding a private package repository, registry secret should be added to the cluster:
+
+   ```sh
+   tanzu secret registry add <SECRET-NAME> --server <PRIVATE-REGISTRY> --username <USERNAME> --namespace <SECRET-NAMESPACE> --password <PASSWORD> -y
+   ```
+
+3. In case you want to add the private repository in a different namespace than the namespace in which the secret was added to, you need o export the secret to all other namespaces. Please be aware that by doing so, the given secret contents will be available to ALL users in ALL namespaces. Please ensure that included registry credentials allow only read-only access to the registry with minimal necessary scope. You can export the secret to other namespaces by running:
+
+   ```sh
+   tanzu secret registry update <SECRET-NAME> --namespace <SECRET-NAMESPACE> --export-to-all-namespaces=true -y
```
shivaani0505, line 194: It would be nice to have the sample output of commands.When the doc writers try it out, they know they are going on right track by comparing the output

shivaani0505:
Since in README , we have the topic heading which says "Workflow for adding a private package repository and installation of a private package", we should probably use something like below for release notes
"updates secret registry README to add the workflow for adding a private package repository using package and secret plugins"

cmd/cli/plugin/secret/README.md
```diff
@@ -171,3 +171,123 @@ the "secret registry" can be used to:
...
+   ```
+
+   The output would be as follows:
+
+   ```sh
+   - Adding registry secret '<SECRET-NAME>'...
+    Added registry secret '<SECRET-NAME>' into namespace '<SECRET-NAMESPACE>'
+   ```
+
+3. In case you want to add the private repository in a different namespace than the namespace in which the secret was added to, you need o export the secret to all other namespaces. Please be aware that by doing so, the given secret contents will be available to ALL users in ALL namespaces. Please ensure that included registry credentials allow only read-only access to the registry with minimal necessary scope. You can export the secret to other namespaces using one of the followings:
```
shivaani0505, line 198: nitpick: o -> to

shivaani0505: nitpick: followings => following

shivaani0505 (approved):
LGTM

Issue #857, opened 2021-10-12
Bump go version to 1.17


stmcginnis:
**Describe the feature request**

Go 1.17 has been out for awhile now, and upstream k8s and others have already transitioned to using it. It would be good for us to come up with an official policy describing how we want to handle go updates. In the meantime, filing this so we can track some of the steps that we need to perform when we do decide to update.

- [ ] Update all `go.mod` files to include `go 1.17.x`
- [ ] Run `go mod tidy -go=1.17` to update our go.mod files to pick up [module graph](https://golang.org/doc/go1.17#tools) changes
- [x] Update our version of golangci-lint to address a [panic in one of the linters](https://github.com/golangci/golangci-lint/issues/2222)

Will add others if they come up.

**Describe alternatives you've considered**

We can stay on 1.16, but there are improvements in newer version that we would miss out on.

benjaminapetersen:
Nice, meanwhile it may be valuable to point out a recommended go version (1.16) in a `Development Setup` heading in the main README or CONTRIBUTING doc.

PR #1050, opened 2021-11-03
Update golangci-lint to 1.43.0

stmcginnis:

This updates the version of golangci-lint we use for code linting to a
version that works with go 1.17. The updated version pulls in various
linter updates, so this also includes fixes for the issues those newer
lint checks were now flagging.

This also drops the use of the gomnd linter for magic number checking.
In general this is a good thing to check, but in nearly all cases where
this was flagging an issue in our code, we would just nolint the error
and keep using the numbers. So rather than enabling it globally and
disabling it locally, this just disables it globally until we decide if
this is something we want to take the effort to address.

Related #857

Ran `make lint` locally and made sure it did not panic with go 1.17 and no linting errors reported.

### Release note

```release-note
NONE
```

### PR Checklist

When we run local checks, we build a local copy of `golangci-lint` that gets stored in `hack/tools/bin`. With this update,
make might still see that the binary is present and not rebuild/update to the newer version. If this is the case, you will
need to run `rm hack/tools/bin/golangci-lint` to clear it out, then run `make lint` again to have it recompiled and lint
checks run with the correct version.


stmcginnis:
Rebased and updated to pick up recent changes.

rajathagasthya (approved):
Thanks!

vuil (approved):
LGTM, thanks!

PR #1072, opened 2021-11-05
Link standalone discovery image information during build time

anujc25:
- This PR implements build-time linking of standalone discovery image
- Allows user to override configuration with runtime environment variables
- Also adds base for Kubernetes Discovery Support
- Unit tests

Build time variables introduced:
- `TKG_DEFAULT_IMAGE_REPOSITORY` (already exists before this change) 
- `DEFAULT_STANDALONE_DISCOVERY_IMAGE_PATH` 
- `DEFAULT_STANDALONE_DISCOVERY_IMAGE_TAG`

Users can override standalone discovery at runtime with below environment variables:
- `TKG_CUSTOM_IMAGE_REPOSITORY`
- `TKG_DEFAULT_STANDALONE_DISCOVERY_IMAGE_PATH`
- `TKG_DEFAULT_STANDALONE_DISCOVERY_IMAGE_TAG`

Related to #74 

- Manually tested the CLI by providing build-time configuration and testing it to determine the default standalone discovery is getting used correctly

### Release note

```release-note
NONE
```

### PR Checklist

### DIFF
--- a/pkg/v1/cli/discovery/interface.go
+++ b/pkg/v1/cli/discovery/interface.go
@@ -32,14 +32,15 @@ type Discovery interface {
 
 // CreateDiscoveryFromV1alpha1 creates discovery interface from v1alpha1 API
 func CreateDiscoveryFromV1alpha1(pd v1alpha1.PluginDiscovery) (Discovery, error) {
-	if pd.GCP != nil {
+	switch {
+	case pd.GCP != nil:
 		return NewGCPDiscovery(pd.GCP.Bucket, pd.GCP.ManifestPath, pd.GCP.Name), nil
-	}
-	if pd.OCI != nil {
+	case pd.OCI != nil:
 		return NewOCIDiscovery(pd.OCI.Name, pd.OCI.Image), nil
-	}
-	if pd.Local != nil {
+	case pd.Local != nil:
 		return NewLocalDiscovery(pd.Local.Name, pd.Local.Path), nil
+	case pd.Kubernetes != nil:
+		return NewKubernetesDiscovery(pd.Kubernetes.Name, pd.Kubernetes.Path, pd.Kubernetes.Context), nil
 	}
-	return nil, errors.New("unknown plugin discovery")
+	return nil, errors.New("unknown plugin discovery source")
 }
--- a/pkg/v1/cli/discovery/kubernetes.go
+++ b/pkg/v1/cli/discovery/kubernetes.go
@@ -1,0 +1,89 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package discovery
+
+import (
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli/common"
+	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli/plugin"
+	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkg/clusterclient"
+)
+
+// KubernetesDiscovery is an artifact discovery utilizing CLIPlugin API in kubernetes cluster
+type KubernetesDiscovery struct {
+	name           string
+	kubeconfigPath string
+	kubecontext    string
+}
+
+// NewKubernetesDiscovery returns a new kubernetes repository
+func NewKubernetesDiscovery(name, kubeconfigPath, kubecontext string) Discovery {
+	return &KubernetesDiscovery{
+		name:           name,
+		kubeconfigPath: kubeconfigPath,
+		kubecontext:    kubecontext,
+	}
+}
+
+// List available plugins.
+func (k *KubernetesDiscovery) List() ([]plugin.Discovered, error) {
+	return k.Manifest()
+}
+
+// Describe a plugin.
+func (k *KubernetesDiscovery) Describe(name string) (p plugin.Discovered, err error) {
+	plugins, err := k.Manifest()
+	if err != nil {
+		return
+	}
+
+	for i := range plugins {
+		if plugins[i].Name == name {
+			p = plugins[i]
+			return
+		}
+	}
+	err = errors.Errorf("cannot find plugin with name '%v'", name)
+	return
+}
+
+// Name of the repository.
+func (k *KubernetesDiscovery) Name() string {
+	return k.name
+}
+
+// Manifest returns the manifest for a local repository.
+func (k *KubernetesDiscovery) Manifest() ([]plugin.Discovered, error) {
+	plugins := make([]plugin.Discovered, 0)
+
+	// Create cluster client
+	clusterClientOptions := clusterclient.Options{GetClientInterval: 2 * time.Second, GetClientTimeout: 5 * time.Second}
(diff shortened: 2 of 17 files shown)


pkg/v1/cli/discovery/kubernetes.go
```diff
@@ -0,0 +1,65 @@
...
+
+// Name of the repository.
+func (k *KubernetesDiscovery) Name() string {
+	return k.name
+}
+
+// Manifest returns the manifest for a local repository.
+func (k *KubernetesDiscovery) Manifest() ([]plugin.Discovered, error) {
+	plugins := make([]plugin.Discovered, 0)
+	// TODO: Implement Kubernetes Discovery
```
anujc25, line 58: This will be implemented as part of separate PR.

saji-pivotal (approved):
Looks good to me. But I would definitely like another person to take a look at this as well

pkg/v1/cli/pluginmanager/manager.go
```diff
@@ -113,15 +113,30 @@ func DiscoverServerPlugins(serverName string) (plugins []plugin.Discovered, err
 		return
```
imikushin, line 113: The returned error here is nil, and I guess that's not intended. This is a consequence of using return vars. 

Generally, you don't need named return vars. Sure, it makes your `return` statements shorter, but that's also a source of confusion for the reader (future you :P). Instead of immediately seeing what's returned you have to traverse back to see what was last assigned to the return var.

Please make your `return` statements explicit. 

And, the tests did *not* catch it, so add a test :)

pkg/v1/config/clientconfig.go
```diff
@@ -143,6 +146,75 @@ func NewClientConfig() (*configv1alpha1.ClientConfig, error) {
...
+		Local: &configv1alpha1.LocalDiscovery{
+			Name: DefaultStandaloneDiscoveryName,
+			Path: DefaultStandaloneDiscoveryLocalPath,
+		},
+	}
+
+	// Prepend default discovery to available discovery sources
+	c.ClientOptions.CLI.DiscoverySources = append([]configv1alpha1.PluginDiscovery{defaultDiscovery}, c.ClientOptions.CLI.DiscoverySources...)
+	return true
+}
```
imikushin, line 192: This function doesn't seem to be covered by unit tests.

pkg/v1/config/defaults.go
```diff
@@ -37,3 +53,24 @@ var DefaultRepositories []configv1alpha1.PluginRepository = []configv1alpha1.Plu
...
+	// Run-time overrides of the configuration
+	if customImageRepo := os.Getenv(constants.ConfigVariableCustomImageRepository); customImageRepo != "" {
+		defaultStandaloneDiscoveryRepository = customImageRepo
+	}
+	if imagePath := os.Getenv(constants.ConfigVariableDefaultStandaloneDiscoveryImagePath); imagePath != "" {
+		defaultStandaloneDiscoveryImagePath = imagePath
+	}
+	if imageTag := os.Getenv(constants.ConfigVariableDefaultStandaloneDiscoveryImageTag); imageTag != "" {
+		defaultStandaloneDiscoveryImageTag = imageTag
+	}
```
imikushin, line 73: All these `if` conditions seem to eval to false when running unit tests, so they're only "half" covered.

imikushin (commented):
A few comments. Sorry for being picky 🙂

pkg/v1/cli/pluginmanager/manager.go
anujc25: This is intended. If the current server provided is incorrect, I believe we should throw a warning indicating the message instead of throwing an error. (added a warning message)

This will allow users to run `tanzu plugin list`, `tanzu plugin sync` etc commands to sync plugins with a warning message which indicates `Error while discovering contexted plugins`.  I feel we should not block users from running other plugin commands if something is misconfigured (especially someone mistakenly misconfigures tanzu config file)

pkg/v1/cli/discovery/interface.go
```diff
@@ -41,5 +41,8 @@ func CreateDiscoveryFromV1alpha1(pd v1alpha1.PluginDiscovery) (Discovery, error)
 	if pd.Local != nil {
 		return NewLocalDiscovery(pd.Local.Name, pd.Local.Path), nil
 	}
+	if pd.Kubernetes != nil {
+		return NewKubernetesDiscovery(pd.Kubernetes.Name, pd.Kubernetes.Path, pd.Kubernetes.Context), nil
+	}
```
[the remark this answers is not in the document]
anujc25: updated and added unit tests

pkg/v1/config/clientconfig.go
anujc25: added unit test

pkg/v1/config/defaults.go
anujc25: added unit tests

pkg/v1/config/clientconfig.go
imikushin: Thanks for adding tests! The `for` loop body is still not covered, and it contains non-trivial logic. Perhaps, it even makes sense to extract it as a separate function.

pkg/v1/config/clientconfig.go
```diff
@@ -507,3 +580,30 @@ func IsFeatureActivated(feature string) bool {
...
+	if server.Type == configv1alpha1.ManagementClusterServerType {
+		defaultClusterK8sDiscovery := configv1alpha1.PluginDiscovery{
+			Kubernetes: &configv1alpha1.KubernetesDiscovery{
+				Name:    fmt.Sprintf("default-%s", serverName),
+				Path:    server.ManagementClusterOpts.Path,
+				Context: server.ManagementClusterOpts.Context,
+			},
+		}
+		discoverySources = append(discoverySources, defaultClusterK8sDiscovery)
+	}
```
imikushin, line 607: The `if` body doesn't seem to be covered by tests.

This method does get called (indirectly) from manager_test.go, but it doesn't have "its own" unit tests.

imikushin (commented):
A couple more comments

pkg/v1/cli/pluginmanager/manager.go
```diff
@@ -102,26 +102,24 @@ func DiscoverStandalonePlugins() (plugins []plugin.Discovered, err error) {
 }
 
 // DiscoverServerPlugins returns the available plugins associated with the given server
-func DiscoverServerPlugins(serverName string) (plugins []plugin.Discovered, err error) {
-	plugins = []plugin.Discovered{}
+func DiscoverServerPlugins(serverName string) ([]plugin.Discovered, error) {
+	plugins := []plugin.Discovered{}
```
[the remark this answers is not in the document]
anujc25: I would prefer not to return `nil` and return an empty list.

pkg/v1/config/clientconfig.go
anujc25: added unit test for this function which includes `if` body.

pkg/v1/cli/pluginmanager/manager.go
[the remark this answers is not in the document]
imikushin: If `plugins == nil` the following code works:
`plugins = append(plugins, &plugin.Discovered{})`

imikushin (approved):
Looks good!

PR #1073, opened 2021-11-05
Fix clustergen tests

tylerschultz:

We saw that optional cluster model values were not getting included in case files.

- generate cases correctly when value is NoneType
- Update constraint check so it matches the current vsphere version
- change "NOTPROVIDED" -> "" in constraints, "NOTPROVIDED" is only handled when
  variable keys start with "--"

ran the clustergen steps and saw case get generated with the optional model data.

### PR Checklist

### DIFF
--- a/pkg/v1/providers/tests/clustergen/param_models/cluster_optional.model
+++ b/pkg/v1/providers/tests/clustergen/param_models/cluster_optional.model
@@ -136,12 +136,12 @@ BUILD_EDITION: "tkg", "tce"
 IDENTITY_MANAGEMENT_TYPE: "none", "oidc", "ldap"
 
 # pass --VSPHERE-CONTROLPLANE-ENDPOINT flag only for vSphere provider
-IF [_INFRA] = "\"vsphere:v0.7.8\"" THEN [--VSPHERE-CONTROLPLANE-ENDPOINT] = "\"10.10.10.10\"";
-IF [_INFRA] <> "\"vsphere:v0.7.8\"" THEN [--VSPHERE-CONTROLPLANE-ENDPOINT] = "\"NOTPROVIDED\"";
+IF [_INFRA] LIKE "\"vsphere:*" THEN [--VSPHERE-CONTROLPLANE-ENDPOINT] = "\"10.10.10.10\"";
+IF [_INFRA] NOT LIKE "\"vsphere:*" THEN [--VSPHERE-CONTROLPLANE-ENDPOINT] = "\"NOTPROVIDED\"";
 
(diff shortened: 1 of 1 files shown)


pkg/v1/providers/tests/clustergen/param_models/cluster_optional.model
```diff
@@ -136,12 +136,12 @@ BUILD_EDITION: "tkg", "tce"
 IDENTITY_MANAGEMENT_TYPE: "none", "oidc", "ldap"
 
 # pass --VSPHERE-CONTROLPLANE-ENDPOINT flag only for vSphere provider
-IF [_INFRA] = "\"vsphere:v0.7.8\"" THEN [--VSPHERE-CONTROLPLANE-ENDPOINT] = "\"10.10.10.10\"";
```
vuil, line 139: thanks for the fix. To be more future-proof, these lines can be changed to:

```
IF [_INFRA] LIKE "\"vsphere:*" THEN ...
```

pkg/v1/providers/tests/clustergen/gencluster_params.py
```diff
@@ -52,7 +52,10 @@ def write_test_cases(params_file, test_dir):
                     if k == "AZURE_CUSTOM_TAGS" and v.startswith('tagKey1='):
                         cfg_args.append('{}: {}'.format(k, 'tagKey1=tagValue1, tagKey2=tagValue2'))
                     elif v != "NA":
-                        cfg_args.append('{}: {}'.format(k, v.replace("<comma>", ",")))
+                        if type(v) == str:
```
vuil, line 55: Can you point to a situation where this is needed? All the pict csv files generated should be a fully formed CSV with
(R+1) x C cells, with quoted strings in the value rows.

pkg/v1/providers/tests/clustergen/param_models/cluster_optional.model
vuil: In fact, after the provider bump from @saimanoj01 , this is necessary now.

pkg/v1/providers/tests/clustergen/gencluster_params.py
tylerschultz: We think we were running into problems with the vsphere variable, and that was our initial fix. We've since updated our PR to wildcard on the vsphere semver, and we're no longer seeing that issue. We've reverted these lines.

vuil (approved):
lgtm, thanks.

pkg/v1/providers/tests/clustergen/gencluster_params.py
tylerschultz: You're right, this wasn't needed.

PR #1074, opened 2021-11-06
Add MachineDeployment specifier to node pool

### LINKED ISSUE #1067: A dangerous assumption is made when creating a new machine deployment that the first MD to be returned can be used as the base
**Bug description**

https://github.com/vmware-tanzu/tanzu-framework/blob/fe23cf4d8a1d33251ebf2599b408419122a34808/pkg/v1/tkg/client/machine_deployment.go#L127

	workerMDs, err := clusterClient.GetMDObjectForCluster(options.ClusterName, options.Namespace)
	baseMD := workerMDs[0]

workerMDs[0] is not necessarily a good machine deployment to use as a template for a new machine deployment.

There used to be a similar assumption made in earlier versions of this code: https://github.com/vmware-tanzu/tanzu-framework/blob/v0.1.0/pkg/v1/tkg/client/machine_deployment.go#L82

This is causing a big problem for us as a TKGM 1.4 customer. If we try and create a new node pool, Tanzu picks up the wrong MD and we end up with a mess. For example, we have an existing dedicated node pool / md for Istio gateway nodes, which has specific labels and whose machines are in a dedicated DMZ network. In one of our environments, if I try to create a new node pool via the tanzu CLI, tanzu bases the new MD on the istio-gateway MD, and our new nodes end up with unwanted labels and are built in the wrong network. We cannot therefore use the node pool abstraction via the tanzu cli and instead have to work directly with machine deployment resources on the management cluster (which works fine).

It seems to me it would be a better idea to explicitly annotate an MD as being the one to use as a template, rather than just using workerMDs[0].

@tenczar 

- [x] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**
I should be able to create additional node pools based on a stable, explicitly defined base machine deployment.

Seen on 1.4.0 and seems to be present on main.

Vsphere 7 TKGM 1.4

tenczar:
Adds a command line flag to specify the MachineDeployment to use as a
base when creating a new node pool

Fixes #1067 

### Release note

```release-note
Adds a new flag to cluster node-pool set `--machine-deployment-base` This flag allows a user to specify an explicit MachineDeployment as the base from which a new node pool will be created.
```

### PR Checklist

### DIFF
--- a/cmd/cli/plugin/cluster/set_node_pool.go
+++ b/cmd/cli/plugin/cluster/set_node_pool.go
@@ -17,8 +17,9 @@ import (
 )
 
 type clusterSetNodePoolCmdOptions struct {
-	FilePath  string
-	Namespace string
+	FilePath              string
+	Namespace             string
+	BaseMachineDeployment string
 }
 
 var setNodePoolOptions clusterSetNodePoolCmdOptions
@@ -32,6 +33,7 @@ var clusterSetNodePoolCmd = &cobra.Command{
 func init() {
 	clusterSetNodePoolCmd.Flags().StringVarP(&setNodePoolOptions.FilePath, "file", "f", "", "The file describing the node pool (required)")
 	clusterSetNodePoolCmd.Flags().StringVar(&setNodePoolOptions.Namespace, "namespace", "default", "The namespace the cluster is found in.")
+	clusterSetNodePoolCmd.Flags().StringVar(&setNodePoolOptions.BaseMachineDeployment, "base-machine-deployment", "", "The machine deployment to use as a base for creating a new node pool (ignored for TKGs)")
 	_ = clusterSetNodePoolCmd.MarkFlagRequired("file")
 	clusterNodePoolCmd.AddCommand(clusterSetNodePoolCmd)
 }
@@ -64,6 +66,7 @@ func SetNodePool(server *v1alpha1.Server, clusterName string) error {
 	if err = yaml.Unmarshal(fileContent, &nodePool); err != nil {
 		return errors.Wrap(err, "Could not parse file contents")
 	}
+	nodePool.BaseMachineDeployment = setNodePoolOptions.BaseMachineDeployment
 
 	options := client.SetMachineDeploymentOptions{
 		ClusterName: clusterName,
--- a/pkg/v1/tkg/client/machine_deployment.go
+++ b/pkg/v1/tkg/client/machine_deployment.go
@@ -55,18 +55,19 @@ type DeleteMachineDeploymentOptions struct {
 
 // NodePool a struct describing a node pool
 type NodePool struct {
-	Name             string                    `yaml:"name"`
-	Replicas         *int32                    `yaml:"replicas,omitempty"`
-	AZ               string                    `yaml:"az,omitempty"`
-	NodeMachineType  string                    `yaml:"nodeMachineType,omitempty"`
-	Labels           *map[string]string        `yaml:"labels,omitempty"`
-	VSphere          VSphereNodePool           `yaml:"vsphere,omitempty"`
-	Taints           *[]corev1.Taint           `yaml:"taints,omitempty"`
-	VMClass          string                    `yaml:"vmClass,omitempty"`
-	StorageClass     string                    `yaml:"storageClass,omitempty"`
-	Volumes          *[]tkgsv1alpha2.Volume    `yaml:"volumes,omitempty"`
-	TKR              tkgsv1alpha2.TKRReference `yaml:"tkr,omitempty"`
-	NodeDrainTimeout *metav1.Duration          `yaml:"nodeDrainTimeout,omitempty"`
+	Name                  string                    `yaml:"name"`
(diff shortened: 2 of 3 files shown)


pkg/v1/tkg/client/machine_deployment.go
```diff
@@ -54,18 +54,19 @@ type DeleteMachineDeploymentOptions struct {
...
+	NodeMachineType       string                    `yaml:"nodeMachineType,omitempty"`
+	Labels                *map[string]string        `yaml:"labels,omitempty"`
+	VSphere               VSphereNodePool           `yaml:"vsphere,omitempty"`
+	Taints                *[]corev1.Taint           `yaml:"taints,omitempty"`
+	VMClass               string                    `yaml:"vmClass,omitempty"`
+	StorageClass          string                    `yaml:"storageClass,omitempty"`
+	Volumes               *[]tkgsv1alpha2.Volume    `yaml:"volumes,omitempty"`
+	TKR                   tkgsv1alpha2.TKRReference `yaml:"tkr,omitempty"`
+	NodeDrainTimeout      *metav1.Duration          `yaml:"nodeDrainTimeout,omitempty"`
+	MachineDeploymentBase string                    `yaml:"-"`
```
imikushin, line 69: Why `yaml:"-"`?

imikushin (commented):
One question. Also, can we add a test for this? (e.g. when mdBase is specified, and 2 cases: when one exists, and when one doesn't)

pkg/v1/tkg/client/machine_deployment.go
tenczar: So that it won't be deserialized if a user does include it in the config file for the node pool

randomvariable: I don't think I get that atm. I would at least put a comment explaining why for future git archeologists.

randomvariable:
@nickperry PTAL

nickperry:
> @nickperry PTAL

LGTM

pkg/v1/tkg/client/machine_deployment.go
imikushin: I don't see any harm if it can get picked up from the config file. As long as the command line param has priority, it's a-ok

swalner-vmware (approved):

cmd/cli/plugin/cluster/set_node_pool.go
```diff
@@ -32,6 +33,7 @@ var clusterSetNodePoolCmd = &cobra.Command{
 func init() {
 	clusterSetNodePoolCmd.Flags().StringVarP(&setNodePoolOptions.FilePath, "file", "f", "", "The file describing the node pool (required)")
 	clusterSetNodePoolCmd.Flags().StringVar(&setNodePoolOptions.Namespace, "namespace", "default", "The namespace the cluster is found in.")
+	clusterSetNodePoolCmd.Flags().StringVar(&setNodePoolOptions.MachineDeploymentBase, "machine-deployment-base", "", "The machine deployment to use as a base for creating a new node pool")
```
prkalle, line 36: We should add in help that machine-deployment-base option is not applicable to TKGS

imikushin (approved):
LGTM

prkalle (approved):
LGTM with one minor nit comment

pkg/v1/tkg/client/machine_deployment.go
```diff
@@ -55,18 +55,19 @@ type DeleteMachineDeploymentOptions struct {
...
+	NodeMachineType       string                    `yaml:"nodeMachineType,omitempty"`
+	Labels                *map[string]string        `yaml:"labels,omitempty"`
+	VSphere               VSphereNodePool           `yaml:"vsphere,omitempty"`
+	Taints                *[]corev1.Taint           `yaml:"taints,omitempty"`
+	VMClass               string                    `yaml:"vmClass,omitempty"`
+	StorageClass          string                    `yaml:"storageClass,omitempty"`
+	Volumes               *[]tkgsv1alpha2.Volume    `yaml:"volumes,omitempty"`
+	TKR                   tkgsv1alpha2.TKRReference `yaml:"tkr,omitempty"`
+	NodeDrainTimeout      *metav1.Duration          `yaml:"nodeDrainTimeout,omitempty"`
+	MachineDeploymentBase string                    `yaml:"machineDeploymentBase,omitempty"`
```
saji-pivotal, line 70: nit: Can this be renamed to `BaseMachineDeployment` or `BaseMachineDeploymentName` ?

saji-pivotal (approved):
One tiny nit. But LGTM otherwise

pkg/v1/tkg/client/machine_deployment.go
tenczar: What about the command line flag?

saji-pivotal: how about `base-machine-deployment` ?

saji-pivotal (approved):
LGTM

PR #1086, opened 2021-11-08
Support set/unset environment variables through tanzu config file

### LINKED ISSUE #1035: Allow configuring generic configuration/environment variable in Tanzu config file
There are use-cases when the user needs to configure environment variables so that tanzu cli can use the configuration to function correctly.
One of the examples is to configure environment variables like `TKG_CUSTOM_IMAGE_REPOSITORY`, `TKG_CUSTOM_IMAGE_REPOSITORY_CA_CERTIFICATE`, `TKG_CUSTOM_IMAGE_REPOSITORY_SKIP_TLS_VERIFY` as environment variables so that CLI can download BoM and providers files from the correct image registry for internet restricted environment.

The proposal is to simplify this by allowing users to configure this type of env variable as part of the Tanzu CLI and let Tanzu CLI configure those as environment variables irrespective of where we run our CLI. Also, use that configuration for cluster creation if applicable. This can be achieved by adding another key `env:` under cli options in the tanzu config file.
```
kind: ClientConfig
clientOptions:
  cli:
    unstableVersionSelector: experimental
  features:
    global:
      context-aware-discovery: "true"
  env:
    FOO: "bar"
```

If there are conflicts, meaning user has set environmental variable outside tanzu config file with `export FOO=foo` as well as variable is mentioned under config file, CLI should try to throw warning about possible conflicts. And CLI should give higher precedence to user configured environment variables to allow user to override config for one off command runs.

Also, add another subkey to be configured with `tanzu config set env.FOO bar` command.

Implement another `tanzu config unset env.FOO` command that can be used to delete the added env configuration.

- Continue to provide this variables as environment variable.

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [x] User Experience

**Additional context**

anujc25:
- The change allows users to configure env variables as part of tanzu
      config file
- Env configuration can be set using `tanzu config set env.<variable> <value>`
      command
- User can unset the already configured settings with new `tanzu config
      unset env.<variable>` command
- In the case of conflicts, meaning user has set environment variable
      outside tanzu config file with `export FOO=bar` and as part of tanzu
      config file, higer precedence is given to user configured environment
      variable to allow user to override config for one off command runs

- Sample configuration file:
```
clientOptions:
  env:
    FOO: bar
    FOO1: baz
```

Fixes #1035

Manual testing
```
- tanzu config set env.FOO bar
- tanzu config set env.FOO1 baz
- tanzu config unset env.FOO
- tanzu config unset env.FOO1
```

### Release note

```release-note
Support set/unset environment variables through tanzu config file
```

### PR Checklist

### DIFF
--- a/apis/config/v1alpha1/clientconfig.go
+++ b/apis/config/v1alpha1/clientconfig.go
@@ -86,6 +86,16 @@ func (c *ClientConfig) IsConfigFeatureActivated(featurePath string) (bool, error
 	return booleanValue, nil
 }
 
+// GetEnvConfigurations returns a map of environment variables to values
+// it returns nil if configuration is not yet defined
+func (c *ClientConfig) GetEnvConfigurations(plugin string) EnvMap {
+	if c.ClientOptions == nil || c.ClientOptions.Env == nil ||
+		c.ClientOptions.Env[plugin] == nil {
+		return nil
+	}
+	return c.ClientOptions.Env[plugin]
+}
+
 // SplitFeaturePath splits a features path into the pluginName and the featureName
 // For example "features.management-cluster.dual-stack" returns "management-cluster", "dual-stack"
 // An error results from a malformed path, including any path that does not start with "features."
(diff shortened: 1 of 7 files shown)


vuil (approved):
lgtm, thanks!

vuil:
@anujc25 please add a release note to mention the update to the command.

PR #1088, opened 2021-11-08
Add logic for asking user content if secret export exists before adding secret of the same name

### LINKED ISSUE #1057: For secret plugin, Sharing of registry secret to all the namespace without user consent
**Bug description**
tanzu cli supports the functionality of adding registry secret via secret plugin. This helps to create a k8s secret in the particular namespace. This plugin supports the flag --export-to-all-namespaces which allows to export the created secret to all the existing and future namespaces in the cluster. When using this flag, user gets a warning and has to explicitly confirm so that the secret can be exported to all the namespaces. --export-to-all-namespaces flag in the backed creates a SecretExport CRD which is responsible for exporting the secret.

The issue here is at the time of creating the secret (without exporting), the plugin doesn't check for the existence of secret's SecretExport CRD. If there is an existing SecretExport CRD with the same name as of the secret, even though --export-to-all-namespaces flag is not used, the secret will be exported to all the namespaces.

Currently on TKG there are many service accounts (SA) with the privilege of creating the SecretExport. Any of the below SAs can create SecretExport CR

ako-tkg-system-sa (tkg-system namespace)
fluentbit-tanzu-package-repo-global-sa (tanzu-package-repo-global namespace)
secretgen-controller-sa (tanzu-system namespace)
tanzu-addons-package-sa (tkg-system namespace)
tkg-addons-app-sa (tkg-system namespace)
Reproduction Steps

tanzu secret registry add test-secret --server projects-stg.registry.vmware.com --username test-user --password test-password
In this case, though the user has just created a namespaced scoped secret, it will be available to all the namespace without user's knowledge/consent.

- [ ] APIs
- [ ] Addons
- [X] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [X] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**
Before creating the secret, make sure the SecretExport with the same name does not exists. If it exists, inform the user accordingly and ask for consent before exporting the secret or delete the existing CR before creating the secret.
There are lot of other service accounts with capability of creating SecretExport functionality. Since this is very critical feature, as a principle of least privilege, try to limit the SAs from creating/modifying the SecretExport CR if not required.

**Step 1**: Using cluster-admin or any of the above SA, create the SecretExport CR directly on the cluster using kubectl

apiVersion: secretgen.carvel.dev/v1alpha1
kind: SecretExport
metadata:
  name: test-secret
  namespace: default
spec:
  toNamespace: '*'

**Step 2**: Using tanzu cli, create a registry secret without exporting

shivaani0505:
Add prompt for asking user content when secret is being added and secretexport of the same name exists before.

Fixes #1057 

```
 >> tanzu secret registry add test-private -n test-ns --server us-east4-docker.pkg.dev --username _json_key --password-file ~/private-auth-plugin/pass-file-5
Warning: SecretExport with the same name exists already, given secret contents will be available 'to all namespaces'.

Are you sure you want to proceed? [y/N]: y

\ Updating registry secret 'test-private'... 
 Updated registry secret 'test-private' in namespace 'test-ns'
```
### Release note

```release-note
Add prompt for asking user content when secret is being added and secretexport of the same name exists before.
```

### PR Checklist

### DIFF
--- a/cmd/cli/plugin/secret/registry_secret_add.go
+++ b/cmd/cli/plugin/secret/registry_secret_add.go
@@ -54,10 +54,7 @@ func init() {
 	registrySecretAddCmd.MarkFlagRequired("username") //nolint
 }
 
-func registrySecretAdd(cmd *cobra.Command, args []string) error {
-
-	var exported string
-
+func registrySecretAdd(cmd *cobra.Command, args []string) error { //nolint:gocyclo
 	registrySecretOp.SecretName = args[0]
 
 	password, err := extractPassword()
@@ -94,25 +91,24 @@ func registrySecretAdd(cmd *cobra.Command, args []string) error {
 		}
 
 		for j := range secretExportList.Items {
-			exported = "not exported"
+			export := ""
 			secretExport := secretExportList.Items[j]
 			if secretExport.Name == registrySecretOp.SecretName {
-
 				ns := &corev1.NamespaceList{}
 
 				err = kc.GetClient().List(context.Background(), ns)
 				if err != nil {
 					return err
 				}
-				
+
 				if findInList(secretExport.Spec.ToNamespaces, "*") || secretExport.Spec.ToNamespace == "*" {
-					exported = "to all namespaces"
+					export = "all namespaces"
 				} else {
-					exported = "to some namespaces"
+					export = "some namespaces"
 				}
 
-				// Ask user consent when secretexport has been created by kubectl and user tries to add secret of the same name as secretexport without using --export-to-all-namespaces flag
-				log.Warningf("Warning: SecretExport with the same name exists already, given secret contents will be available '%s'.\n\n", exported)
+				// Ask user consent when SecretExport has been created by kubectl and user tries to add secret of the same name as SecretExport without using --export-to-all-namespaces flag
+				log.Warningf("Warning: SecretExport with the same name exists already, given secret contents will be available to %s. If you decide not to proceed, you can either delete the SecretExport or specify a different secret name.\n\n", export)
 				if !registrySecretOp.SkipPrompt {
 					if err := cli.AskForConfirmation("Are you sure you want to proceed?"); err != nil {
 						return errors.New("creation of the secret got aborted")
--- a/cmd/cli/plugin/secret/registry_secret_add.go
+++ b/cmd/cli/plugin/secret/registry_secret_add.go
@@ -77,13 +77,15 @@ func registrySecretAdd(cmd *cobra.Command, args []string) error {
 	}
 
 	if registrySecretOp.ExportToAllNamespaces {
-		log.Warning("Warning: By choosing --export-to-all-namespaces, given secret contents will be available to ALL users in ALL namespaces. Please ensure that included registry credentials allow only read-only access to the registry with minimal necessary scope.\n\n")
-		if !registrySecretOp.SkipPrompt {
-			if err := cli.AskForConfirmation("Are you sure you want to proceed?"); err != nil {
-				return errors.New("creation of the secret got aborted")
-			}
+		// Creating a SecretExport resource X that conflicts with a previously defined SecretExport Y that was exported to all namespaces could result in privilege escalation if the user does not access to other namespaces. This check prevents it by trying to list the namespaces
+		if err := checkNamespaceList(kc); err != nil {
+			return err
 		}
-		log.Info("\n")
+		msg := "Warning: By choosing --export-to-all-namespaces, given secret contents will be available to ALL users in ALL namespaces. Please ensure that included registry credentials allow only read-only access to the registry with minimal necessary scope.\n\n"
+		err := issueWarningAndPromptUser(msg)
+		if err != nil {
+			return err
+		}
 	} else if err := checkSecretExportExists(pkgClient, kc); err != nil {
 		return err
 	}
@@ -216,13 +218,13 @@ func checkSecretExportExists(pkgClient tkgpackageclient.TKGPackageClient, kc kap
 	}
 
 	// Ask user consent when SecretExport has been created by kubectl and user tries to add secret of the same name as SecretExport without using --export-to-all-namespaces flag
-	log.Warningf("Warning: SecretExport with the same name exists already, given secret contents will be available to %s. If you decide not to proceed, you can either delete the SecretExport or specify a different secret name.\n\n", export)
-	if !registrySecretOp.SkipPrompt {
-		if err := cli.AskForConfirmation("Are you sure you want to proceed?"); err != nil {
-			return errors.New("creation of the secret got aborted")
-		}
+	msg := "Warning: SecretExport with the same name exists already, given secret contents will be available to %s. If you decide not to proceed, you can either delete the SecretExport or specify a different secret name.\n\n"
+	msg = fmt.Sprintf(msg, export)
+	err = issueWarningAndPromptUser(msg)
+	if err != nil {
+		return err
 	}
-	log.Info("\n")
+
 	return nil
 }
 
@@ -236,4 +238,15 @@ func checkSecretExists(kc kappclient.Client) (bool, error) {
 		}
 	}
 	return notFound, nil
+}
+
+func issueWarningAndPromptUser(msg string) error {
+	log.Warning(msg)
+	if !registrySecretOp.SkipPrompt {
+		if err := cli.AskForConfirmation("Are you sure you want to proceed?"); err != nil {
+			return errors.New("creation of the secret got aborted")
+		}
+	}
+	log.Info("\n")
+	return nil
 }
--- a/cmd/cli/plugin/secret/registry_secret_add.go
+++ b/cmd/cli/plugin/secret/registry_secret_add.go
@@ -55,6 +55,9 @@ func init() {
 }
 
 func registrySecretAdd(cmd *cobra.Command, args []string) error {
+
+	var exported string
+
 	registrySecretOp.SecretName = args[0]
 
 	password, err := extractPassword()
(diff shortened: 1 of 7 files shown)


cmd/cli/plugin/secret/registry_secret_add.go
```diff
@@ -74,16 +87,41 @@ func registrySecretAdd(cmd *cobra.Command, args []string) error {
...
 
-	kc, err := kappclient.NewKappClient(registrySecretOp.KubeConfig)
-	if err != nil {
-		return err
+		for j := range secretExportList.Items {
+			exported = "not exported"
+			secretExport := secretExportList.Items[j]
+			if secretExport.Name == registrySecretOp.SecretName {
+
+				ns := &corev1.NamespaceList{}
```
shivaani0505, line 101: @vijaykatam I am bit confused about the namespace check ( for kubectl get ns), whether it should be a List or Get.
Is this on the same lines you had intended?

piyush-saurabh:
The change looks good to me from security point of view. As a UX feedback, the warning message can be more descriptive.

cmd/cli/plugin/secret/registry_secret_add.go
```diff
@@ -54,7 +54,7 @@ func init() {
 	registrySecretAddCmd.MarkFlagRequired("username") //nolint
 }
 
-func registrySecretAdd(cmd *cobra.Command, args []string) error {
+func registrySecretAdd(cmd *cobra.Command, args []string) error { //nolint:gocyclo
```
vijaykatam, line 57: If we are adding `nolint`, could we make it specific? For example `// nolint:gocyclo,funlen`

cmd/cli/plugin/secret/registry_secret_add.go
```diff
@@ -74,16 +84,40 @@ func registrySecretAdd(cmd *cobra.Command, args []string) error {
...
 		}
 		log.Info("\n")
-	}
-
-	pkgClient, err := tkgpackageclient.NewTKGPackageClient(registrySecretOp.KubeConfig)
-	if err != nil {
-		return err
-	}
+	} else {
+		secretExportList, err := pkgClient.ListSecretExports(registrySecretOp)
```
vijaykatam, line 88: Could we get the SecretExport with the name = registrySecretOp.SecretName instead of listing all? Would remove the need to do a for loop on line 93.

cmd/cli/plugin/secret/registry_secret_add.go
vijaykatam: Probably a code comment explaining this. Something like "Creating a SecretExport resource _X_ that conflicts with a previously defined SecretExport _Y_ that was exported to all namespaces could result in privilege escalation if the user does not access to other namespaces. This check prevents it by trying to list the namespaces"

cmd/cli/plugin/secret/registry_secret_add.go
```diff
@@ -74,16 +84,40 @@ func registrySecretAdd(cmd *cobra.Command, args []string) error {
...
+				}
+
+				if findInList(secretExport.Spec.ToNamespaces, "*") || secretExport.Spec.ToNamespace == "*" {
+					export = "all namespaces"
+				} else {
+					export = "some namespaces"
+				}
+
+				// Ask user consent when SecretExport has been created by kubectl and user tries to add secret of the same name as SecretExport without using --export-to-all-namespaces flag
+				log.Warningf("Warning: SecretExport with the same name exists already, given secret contents will be available to %s. If you decide not to proceed, you can either delete the SecretExport or specify a different secret name.\n\n", export)
```
danniel1205, line 111: nit-pick: we might not want to add `Warning:` since `log.Warningf` should already have that prepended.

cmd/cli/plugin/secret/registry_secret_add.go
```diff
@@ -74,16 +84,40 @@ func registrySecretAdd(cmd *cobra.Command, args []string) error {
...
+				if !registrySecretOp.SkipPrompt {
+					if err := cli.AskForConfirmation("Are you sure you want to proceed?"); err != nil {
+						return errors.New("creation of the secret got aborted")
+					}
+				}
+				log.Info("\n")
```
danniel1205, line 117: Question: is there any specific reason we want to print a change-line via `log.Info` ? The reason why I am asking is that I see we already have two `\n`s from https://github.com/vmware-tanzu/tanzu-framework/pull/1088/files#diff-acb06893c75ae6caca902b2669b6d4d6fd949951490626d59d8241b0b03088b8R111

cmd/cli/plugin/secret/registry_secret_add.go
maralavi: As far as I can tell, ```log.Warningf``` doesn't prepend a "Warning"

cmd/cli/plugin/secret/registry_secret_add.go
maralavi: I respond as I had added the same to the ```registry_secret_add.go``` and ```registry_secret_update.go```. That is to add a separation after the warning message and the rest of the flow, so for aesthetic reasons.

cmd/cli/plugin/secret/registry_secret_add.go
shivaani0505: Output is:
"Warning: SecretExport with the same name exists already, given secret contents will be available to all namespaces. If you decide not to proceed, you can either delete the SecretExport or specify a different secret name."

cmd/cli/plugin/secret/registry_secret_add.go
shivaani0505: Fixed

cmd/cli/plugin/secret/registry_secret_add.go
shivaani0505: it is only gocyclo 
no issue with funlen

cmd/cli/plugin/secret/registry_secret_add.go
```diff
@@ -74,16 +85,41 @@ func registrySecretAdd(cmd *cobra.Command, args []string) error {
...
-	}
+	} else {
+		export := ""
+		secretExport, err := pkgClient.GetSecretExport(registrySecretOp)
 
-	pkgClient, err := tkgpackageclient.NewTKGPackageClient(registrySecretOp.KubeConfig)
-	if err != nil {
-		return err
-	}
+		if err != nil {
```
vijaykatam, line 92:
```
if err != nil && !apierrors.IsNotFound(err) {
return err
}
```

cmd/cli/plugin/secret/registry_secret_add.go
vijaykatam: ^ could you address

cmd/cli/plugin/secret/registry_secret_add.go
danniel1205: Thanks for the explanation. There are many other places have `\n`s constructed within the message string. It might become hard to maintain with the growth of our code base. 

- If the newer log lib has some enhancements on `change-lines`, I as a developer might have a hard time to make modifications accordingly in all places and it also takes time to test all messages.
- I am also thinking about the "consistency" among all plugins. Other plugin owner might not append `\n` within their messages or they might think `\n\n\n` is better. We might be fine in this case as long as users are ok with it.

Based on above reasons(maintainability and consistency), I feel like building an API on top of `log.xxx` might be a good idea. Developers just need to provide the content of a message without worrying about `change-lines`. This is definitely something could be addressed in the future, not a blocker for this PR to get merged.

cmd/cli/plugin/secret/registry_secret_add.go
shivaani0505: Fixed by breaking into functions

cmd/cli/plugin/secret/registry_secret_add.go
shivaani0505: Fixed

cmd/cli/plugin/secret/registry_secret_add.go
```diff
@@ -66,6 +66,16 @@ func registrySecretAdd(cmd *cobra.Command, args []string) error {
...
+		return err
+	}
+
+	kc, err := kappclient.NewKappClient(registrySecretOp.KubeConfig)
+	if err != nil {
+		return err
+	}
+
 	if registrySecretOp.ExportToAllNamespaces {
 		log.Warning("Warning: By choosing --export-to-all-namespaces, given secret contents will be available to ALL users in ALL namespaces. Please ensure that included registry credentials allow only read-only access to the registry with minimal necessary scope.\n\n")
```
vijaykatam, line 80: Lets add `checkNamespacesList()` here as well prior to warning.

cmd/cli/plugin/secret/registry_secret_add.go
```diff
@@ -185,3 +183,57 @@ func extractPassword() (string, error) {
...
+	export := ""
+	secretExport, err := pkgClient.GetSecretExport(registrySecretOp)
+
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+
+	// No error means we found a matching SecretExport
+
+	// Creating a SecretExport resource X that conflicts with a previously defined SecretExport Y that was exported to all namespaces could result in privilege escalation if the user does not access to other namespaces. This check prevents it by trying to list the namespaces
```
vijaykatam, line 207: @shyaamsn - Would you recommend a more terse comment? (I agree the statement can be confusing)

shyaamsn: I dont have a better recommendation either. Lets stick to this and iterate later if we have a better wording.

cmd/cli/plugin/secret/registry_secret_add.go
shivaani0505: After adding this, cyclomatic complexity increased to 16>15, so had to introduce another function issueWarningAndPromptUser to further break the parent function and resolve lint issue

vijaykatam (approved):

pkg/v1/tkg/test/tkgpackageclient/package_plugin_integration_test.go
```diff
@@ -411,6 +413,7 @@ func testHelper() {
 	err = json.Unmarshal(resultImgPullSecret.Stdout.Bytes(), &imgPullSecretOutput)
 	Expect(err).ToNot(HaveOccurred())
 	Expect(len(imgPullSecretOutput)).To(BeNumerically(">=", 1))
+	 */
```
maralavi, line 416: Did this changed pass locally? If not and if the pipeline failed, I'd probably comment till the end of line 437, so that it starts off by adding a repository.

Issue #1103, opened 2021-11-09
Nameservers from DHCP clobber the `CONTROL_PLANE_NODE_NAMESERVERS` or `WORKER_NODE_NAMESERVERS` settings


flawedmatrix:
**Bug description**

 In https://github.com/vmware-tanzu/tanzu-framework/issues/238 the ability to configure custom nameservers for the Control Plane/Worker nodes was added to tanzu-framework, but although the nameservers to get threaded to the Kubernetes nodes, they do not appear to work in certain cases. For instance, if a nameserver with a custom DNS entry for `foo.com` is configured in `CONTROL_PLANE_NODE_NAMESERVERS`, then trying to resolve `foo.com` from the Control Plane Node fails.

**Affected product area (please put an X in all that apply)**

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [x] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [x] User Experience

**Expected behavior**
It is expected that in the provided example, resolving `foo.com` from the Control Plane Node should succeed.

**Steps to reproduce the bug**
1. Set up a vSphere environment that supports both IPv4 and IPv6 and has DHCP configured to provide DNS nameservers as well.
1. Deploy a DNS server with an entry for `some-nonexistent-domain.foo`.
1. In the configuration file for creating a management cluster, configure `CONTROL_PLANE_NODE_NAMESERVERS` with this DNS server.
1. Use an Ubuntu 20.04 image as a template for the control plane node.
1. Run `tanzu management-cluster create` with this configuration file. Wait for the management cluster to create successfully.
1. SSH onto the management cluster control plane node and run DNS utils to attempt to resolve `some-nonexistent-domain.foo`.
1. The entry is likely not resolvable.

**Version** (include the SHA if the version is not obvious)
Tanzu Framework v0.10.0

**Environment where the bug was observed (cloud, OS, etc)**
vSphere 7 with IPv4 and IPv6 support and DHCP configured to provide DNS nameservers
Ubuntu 20.04

**Investigation**

Upon further inspection, running `sudo resolvectl status` shows multiple DNS servers listed, including ones provided by DHCP and the `CONTROL_PLANE_NODE_NAMESERVER` that was configured. However, the `Current DNS Server` likely picks only one provided by DHCP and will not resolve with the `CONTROL_PLANE_NODE_NAMESERVER`.

[...]

PR #1104, opened 2021-11-09
introduce feature flags for custom nameservers

adobley:
Custom nameservers support is not fully functional and may not work in all environments.

Two new feature flags are added with this commit, one for each plugin:
`features.management-cluster.custom-nameservers`
`features.cluster.custom-nameservers`

This is related to #1103, once that issue is resolved, we believe we can remove these flags.

We tested cluster creation with the CONTROL_PLANE_NODE_NAMESERVERS and WORKER_NODE_NAMESERVERS with the new flags set to true and set to false. We verified that the message is displayed to users correctly when the feature is disabled and that the cluster is able to deploy when it is enabled.

### Release note

```release-note
NONE
```

### PR Checklist

### DIFF
--- a/pkg/v1/tkg/client/validate.go
+++ b/pkg/v1/tkg/client/validate.go
@@ -582,7 +582,7 @@ func (c *TkgClient) ConfigureAndValidateManagementClusterConfiguration(options *
 		return NewValidationError(ValidationErrorCode, err.Error())
 	}
 
-	if err = c.ConfigureAndValidateNameserverConfiguration(); err != nil {
+	if err = c.ConfigureAndValidateNameserverConfiguration(TkgLabelClusterRoleManagement); err != nil {
 		return NewValidationError(ValidationErrorCode, err.Error())
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/v1/tkg/client/validate.go
```diff
@@ -1790,6 +1790,20 @@ func (c *TkgClient) validateNameservers(nameserverConfigVariable string) error {
...
+	featureActivated := false
+	switch clusterRole {
+	case TkgLabelClusterRoleManagement:
+		featureActivated = c.IsFeatureActivated(config.FeatureFlagManagementClusterCustomNameservers)
+	case TkgLabelClusterRoleWorkload:
+		featureActivated = c.IsFeatureActivated(config.FeatureFlagClusterCustomNameservers)
+	}
+
+	if !featureActivated {
+		return fmt.Errorf("option %s is set to %q, but custom nameserver support is not enabled (because it is not fully functional). To enable custom nameservers, set features.management-cluster.custom-nameservers to \"true\"",
```
swalner-vmware, line 1802: I suggest setting the error message in the case statement, because the current error message only applies to management name clusters, but it might be a workload case.
Also (a nit) I would suggest making the message more explicit, something like:
`To enable custom nameservers, run the command: tanzu config set features.management-cluster.custom-nameservers true`

adobley: Ah, yes, of course. That would be much better. Thanks!

adobley: Fixed and pushed!

swalner-vmware (approved):

danniel1205 (approved):
LGTM

PR #1105, opened 2021-11-09
Makefile: Use specific versions of controller-gen and kustomize

randomvariable:

Instead of using controller-gen that is in someone's go bin directory or on path, compile and use a local copy from hack/tools/bin. This mirrors what happens upstream in Cluster API, and ensures consistency of generated files.

Fixes #

### Release note

```release-note
NONE
```

### PR Checklist


Makefile
```diff
@@ -148,25 +150,10 @@ manifests: controller-gen ## Generate manifests e.g. CRD, RBAC etc.
 generate-go: $(COUNTERFEITER) ## Generate code via go generate.
 	PATH=$(abspath hack/tools/bin):$(PATH) go generate ./...
 
-generate: controller-gen ## Generate code via controller-gen
+generate: tools ## Generate code via controller-gen
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt",year=$(shell date +%Y) paths="./..."
 	$(MAKE) fmt
 
-controller-gen: ## Download controller-gen
```
rajathagasthya, line 155: Should we remove this from the other Makefiles in this repo? Maybe as part of a followup PR?

randomvariable: Let's do as a follow up and create an issue. Think it might need a bigger refactoring, and realistically not going to have time this week. This PR at least stopped all the main manifests being regenerated with the wrong controller-gen version when i changed a core package in another PR.

imikushin (approved):
LGTM. Thank you!

timothysc:
@randomvariable is this out of date?

Makefile
randomvariable: Removed the other ones as per zoom call.

randomvariable:
@timothysc  this is definitely needed as the developer experience relies on correct versions of binaries being installed in the user's path, this ensures the version we want is always used, and has build caching for the tooling.

randomvariable:
I don't know why the linter is failing. `make lint` works fine locally.

PR #1117, opened 2021-11-09
Update node-pool API to handle vpshere ipv6

### LINKED ISSUE #1116: Node Pool API does not support ipv6 on vSphere
**Bug description**
The node pool api does not properly set DHCP values for ipv4 and ipv6 as desired by an end user. Also node pools currently do not support customizing nameservers in vsphere.

- [x] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**
When a customer specifies a TKG IP Family including ipv6 the relevant properties should be set on nodes in the node pool.

tenczar:
This change updates the vsphere node pool definition to accept the tkg
ip family and nameservers. These properties will only be applied if the
network name is also given.

Fixes #1116 

Added unit tests to confirm properties are set correctly.

### Release note

```release-note
Node pools on vSphere now properly set ip family and nameservers using properties tkgIPFamily and nameservers
```

### PR Checklist

### DIFF
--- a/pkg/v1/tkg/client/machine_deployment.go
+++ b/pkg/v1/tkg/client/machine_deployment.go
@@ -72,18 +72,20 @@ type NodePool struct {
 
 // VSphereNodePool a struct describing properties necessary for a node pool on vSphere
 type VSphereNodePool struct {
-	CloneMode         string `yaml:"cloneMode,omitempty"`
-	Datacenter        string `yaml:"datacenter,omitempty"`
-	Datastore         string `yaml:"datastore,omitempty"`
-	StoragePolicyName string `yaml:"storagePolicyName,omitempty"`
-	Folder            string `yaml:"folder,omitempty"`
-	Network           string `yaml:"network,omitempty"`
-	ResourcePool      string `yaml:"resourcePool,omitempty"`
-	VCIP              string `yaml:"vcIP,omitempty"`
-	Template          string `yaml:"template,omitempty"`
-	MemoryMiB         int64  `yaml:"memoryMiB,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/v1/tkg/client/machine_deployment.go
```diff
@@ -617,10 +619,18 @@ func populateVSphereMachineTemplate(machineTemplate *vsphere.VSphereMachineTempl
 			Devices: []vsphere.NetworkDeviceSpec{
 				{
 					NetworkName: options.VSphere.Network,
-					DHCP4:       true,
 				},
 			},
 		}
+		if len(options.VSphere.Nameservers) > 0 {
+			machineTemplate.Spec.Template.Spec.Network.Devices[0].Nameservers = options.VSphere.Nameservers
```
saji-pivotal, line 626: Why are modifying only Devices[0] here?

tenczar: Our default configuration only has one network device. This code specifies exactly one new device to be set on the VSphereMachineTemplate. I think it would take a significant feature request to expand the number of devices in our preconfigured plans and I'm really not sure how we'd correctly handle setting up the networking in those situations.

saji-pivotal: Cool. Thanks!

saji-pivotal:
@tenczar The changes are fine according to me, but I would rather another person take a look at this as well

prkalle (approved):
LGTM. Thanks!

imikushin (approved):
LGTM

PR #1118, opened 2021-11-09
bump CPI to version 1.22.3

nicolehanjing:
bump CPI to version 1.22.3

Fixes #

generating CPI configs for data values

### Release note

```release-note
bump CPI to version 1.22.3
```

### PR Checklist


alfredthenarwhal:
> @rajaskakodkar:  /test
Commit: 02a8b4f

Tests failed! Build no: 1029

pkg/v1/providers/ytt/vendir/vsphere_cpi/_ytt_lib/addons/packages/vsphere-cpi/1.22.3/bundle/config/overlays/update-daemonset.yaml
```diff
@@ -1,36 +1,43 @@
...
   annotations:
     kapp.k14s.io/disable-default-label-scoping-rules: ""
+    kapp.k14s.io/update-strategy: fallback-on-replace
 
 spec:
   template:
+    metadata:
+      labels:
+        #@overlay/match missing_ok=True
+        vsphere-cpi/data-values-hash: #@ sha256.sum(yaml.encode(values))
```
vijaykatam, line 20: Could we move this to an annotation?

nicolehanjing: @vijaykatam thanks for the review! 
We noticed that CSI also uses label: https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/vsphere-csi/2.4.0/bundle/config/overlays/update-csi-driver.yaml#L77
Is it okay to just use the label? As we are close to FC, if this needs to be changed, can we update it in the next version?

vijaykatam: The issue with sha256 value is it can exceed 63 characters(restriction for label values) so you need to trim it and also cast to string to account for the case where truncated string is all numeric and k8s throws an error complaining integer overflow. Also labels means k8s will index the label. CSI needs to be changed but since this is net new I would encourage to not add tech debt.

nicolehanjing: just want to double check - to update this, we will need to first update in tce repo, and then sync in framework repo and tkg-packages repo, right?

vijaykatam: Yep, also the daemonset will fail with the current code(due to label value validation) so a change would have been required either ways. 
Also acknowledge the pain for upstream and downstream, we are looking to remove CRS for CPI in a future release so you won't need to sync to framework repo. For tce repo and tkg-packages, we will investigate a cayman mirror or other alternatives.

nicolehanjing: gotcha, thanks for the info! We will make the change soon

lubronzhan: Hi @vijaykatam thanks for investigating it. I noticed that Annotation also has  [253 limitation](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set) on the length, it just remit the problem but still we will have issue in the future. 
I think the long term solution is still to watch the secret of config map in controller. Or we gzip the string first then encode it..
Unless this is breaking our current bolt pipeline, I think it's better to at least to get the code in then fix the issue since we don't want to miss the code freeze, since updating all dependent PRs and running pipelines takes much more time and probably will block us from getting our bump in in time

vijaykatam: The ytt function `sha256.sum` will yield a string of 256 characters and it will break downstream pipeline if you try this change as is.

lubronzhan: Hi the sha256 will give out 256 bit not characters. which is 32 characters and we should be fine if we put it inside label

lubronzhan: Ok it's 64 characters in generated by ytt. so annotation should cover it, no need to truncate the sha

lubronzhan: Oh 253 length for label value, so there is no need to change 
https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set @vijaykatam

vijaykatam:
```
Valid label value:

    must be 63 characters or less (can be empty),
```
sha265 ytt function returns more so needs to be trucated

nicolehanjing:
@vijaykatam can you help to review this PR? It's passing all the pipelines

pkg/v1/providers/ytt/vendir/vsphere_cpi/_ytt_lib/addons/packages/vsphere-cpi/1.22.3/bundle/config/overlays/update-daemonset.yaml
lubronzhan: Oh yes you are right so the key prefix could be 253, not the value, I got it wrong. And no length limitation for annotation value

vijaykatam (approved):

vijaykatam:
Just FYI, this change should only target 1.22 TKr.

PR #788, opened 2021-10-06
Separate avi vip network

iXinqi:
Signed-off-by: Xinqi Li lxinqi@vmware.com

**What this PR does / why we need it**:

Add a few new configurations in ako-operator for user to specify the control plane vip network and data vip network as well as control plane SEG and data plane SEG separately. 

The ako in bootstrap cluster will use the vip network and SEG of management cluster control plane.

**New PR Checklist**

- [x] Ensure PR contains only public links or terms
- [x] Use good commit [messages](https://github.com/vmware-tanzu/tanzu-framework/blob/main/CONTRIBUTING.md)
- [x] Squash the commits in this branch before merge to preserve our git history
- [x] If this PR is just an idea or POC, use a [Draft PR](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests) instead of a full PR
- [x] Add appropriate [kind label](../docs/release/kind-labels.md) according to what type of issue is being addressed.

PR #1121, opened 2021-11-10
[Don't merge] test vip network separation

XudongLiuHarold:

Fixes #

### Release note

```release-note

```

### PR Checklist


XudongLiuHarold:
> Why do we need this PR to test the changes same as #788?

Don't want to add mass to that PR since I don't locate the problem yet, will update that one once root cause it.

iXinqi:
> > Why do we need this PR to test the changes same as #788?
> 
> Don't want to add mass to that PR since I don't locate the problem yet and bolt relies on it, will update that one once root cause it.

Okay, please go ahead. I double checked both pr and there is only one difference. https://github.com/vmware-tanzu/tanzu-framework/pull/788/files#diff-fc664bc4e5d59f553d15dddf8c4a986fdd01f596bf335991276bde8e20f41da4L3

XudongLiuHarold:
> > > Why do we need this PR to test the changes same as #788?
> > 
> > 
> > Don't want to add mass to that PR since I don't locate the problem yet and bolt relies on it, will update that one once root cause it.
> 
> Okay, please go ahead. I double checked both pr and there is only one difference. https://github.com/vmware-tanzu/tanzu-framework/pull/788/files#diff-fc664bc4e5d59f553d15dddf8c4a986fdd01f596bf335991276bde8e20f41da4L3

Good point, fixing it.

Issue #578, opened 2021-09-08
version prefix matching should be changed to exact match when updating TKR compatibility condition


prkalle:
**Bug description**
Currently the compatibility condition of the TKRs is updated by using a prefix matching of the management cluster version with the list of `managementClusterVersions` list in BOM compatibility metadata. However it only selects the _last_ managementClusterVersion that satisfies the prefix matching with management cluster version. 

An example tkr-compatiblity file that could cause issue is as shown below
```
version: v5
managementClusterVersions:
- version: v1.3.1-fips.1
  supportedKubernetesVersions:
  - v1.20.5+vmware.2-fips.1-tkg.1
- version: v1.3.1
  supportedKubernetesVersions:
  - v1.19.8+vmware.3-tkg.1
  - v1.20.5+vmware.2-tkg.1
  - v1.18.17+vmware.2-tkg.1
  - v1.20.4+vmware.3-tkg.1
  - v1.19.9+vmware.2-tkg.1
  - v1.18.16+vmware.3-tkg.1
  - v1.17.16+vmware.3-tkg.1
```

If the management cluster version is `v1.3.1-fips.1` the ideal match should be the first match.  But the current implementation only matches the last prefix match (it would match with v1.3.1) which would cause incorrect TKR compatibility conditions.
We should update the prefix matching to exact match unless we uncover any usecases.

There is another issue [#452](https://github.com/vmware-tanzu/tanzu-framework/issues/452) already filed relavent to this. But this issue is filed to see if we can update the version prefix match to exact match.

PR #1139, opened 2021-11-10
version prefix matching changed to exact match when updating TKR compatibility condition

prkalle:
Signed-off-by: PremKumar Kalle <pkalle@vmware.com>

The compatibility condition of the TKRs is updated by using a prefix matching of the management cluster version with the list of managementClusterVersions list in BOM compatibility metadata. Because of this it only selects the last managementClusterVersion that satisfies the prefix matching with management cluster version. This PR changed the prefix matching to exact match.

Fixes # #578 

Deleted the current configmaps and TKrs in "tkr-system' namespace and later tested the TKr controller changes on the cluster. The compatibility conditions are updated successfully.

### Release note

```release-note
version prefix matching changed to exact match when updating TKR compatibility condition
```

### PR Checklist

### DIFF
--- a/pkg/v1/tkr/controllers/source/tkr_source_controller.go
+++ b/pkg/v1/tkr/controllers/source/tkr_source_controller.go
@@ -347,9 +347,7 @@ func (r *reconciler) UpdateTKRCompatibleCondition(ctx context.Context, tkrs []ru
 
 	compatibleReleases := []string{}
 	for _, mgmtVersion := range metadata.ManagementClusterVersions {
-		// Fix before TKG v1.10: what if mgmtClusterVersion is "v1.10" and mgmtVersion.TKGVersion is "v1.1"?
-		// See https://github.com/vmware-tanzu/tanzu-framework/issues/452
-		if strings.HasPrefix(mgmtClusterVersion, mgmtVersion.TKGVersion) {
+		if mgmtClusterVersion == mgmtVersion.TKGVersion {
 			compatibleReleases = mgmtVersion.SupportedKubernetesVersions
 		}
 	}
--- a/pkg/v1/tkr/fakes/boms/metadata.yaml
+++ b/pkg/v1/tkr/fakes/boms/metadata.yaml
@@ -2,10 +2,6 @@ managementClusterVersions:
 - version: v1.0
   supportedKubernetesVersions:
   - v1.17.13+vmware.1
-- version: v1.1
-  supportedKubernetesVersions:
-  - v1.18.10+vmware.1
-  - v1.17.13+vmware.1
 - version: v1.2
   supportedKubernetesVersions:
   - v1.18.10+vmware.1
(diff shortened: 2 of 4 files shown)


pkg/v1/tkr/controllers/source/tkr_source_controller.go
```diff
@@ -347,9 +347,7 @@ func (r *reconciler) UpdateTKRCompatibleCondition(ctx context.Context, tkrs []ru
 
 	compatibleReleases := []string{}
 	for _, mgmtVersion := range metadata.ManagementClusterVersions {
-		// Fix before TKG v1.10: what if mgmtClusterVersion is "v1.10" and mgmtVersion.TKGVersion is "v1.1"?
-		// See https://github.com/vmware-tanzu/tanzu-framework/issues/452
-		if strings.HasPrefix(mgmtClusterVersion, mgmtVersion.TKGVersion) {
+		if mgmtClusterVersion == mgmtVersion.TKGVersion {
```
tenczar, line 350: Do we have any sense of why this was a prefix match previously, and why this wasn't just moved to exact match instead of adding the TODO?

pkg/v1/tkr/fakes/boms/metadata.yaml
```diff
@@ -18,3 +14,11 @@ managementClusterVersions:
   - v1.19.1+vmware.2
   - v1.19.3+vmware.1
   - v1.17.13+vmware.1
+- version: v1.10
+  supportedKubernetesVersions:
+  - v1.18.10+vmware.1
+  - v1.20.5+vmware.1
+- version: v1.1
```
tenczar, line 21: Is it important that the v1.1 version is moved after v1.10? That smells funny.

tenczar (commented):
I'm not comfortable enough with how TKR compatibility works to rubber stamp this. It seems like this should work, but someone with more domain knowledge should definitely look it over.

pkg/v1/tkr/controllers/source/tkr_source_controller.go
anujc25, line 350: Will this work for `dev`, `RC` builds as well?

prkalle: I _think_ this was initial limitation from release team where the management cluster version(fetched from tkg BOM) for dev builds was superset of the version published in tkr-compatibility. But later the release team had removed this limitation.
cc: @ZihanJiang96

prkalle: Yes, release team publishes  tkr-compatibility images with the right management cluster versions. for example for v1.5.0-rc.1 builds a seperate tkr-compatibility image would be published having the below version information

```
- version: v1.5.0-rc.1
  supportedKubernetesVersions:
  - v1.20.12+vmware.1-tkg.1-rc.1
  - v1.21.6+vmware.1-tkg.1-rc.1
  - v1.22.3+vmware.1-tkg.1-rc.1
```

anujc25 (approved):
lgtm

vuil (approved):
lgtm, thanks!

PR #1144, opened 2021-11-11
Enable feature flag for API driven CLIPlugin Discovery

anujc25:
### What this PR does / why we need it
Currently, the feature flag for API-driven CLIPlugin discovery also known as Context-Aware-Discovery is turned off by default.

This PR turns on this feature flag by default.

Changes:
- Updates `Makefile` to add/update targets to build CLI with feature flag enabled
    - Build time configuration variable `ENABLE_CONTEXT_AWARE_PLUGIN_DISCOVERY` to enable/disable context-aware discovery
    - Added config variable `DEFAULT_STANDALONE_DISCOVERY_TYPE` to use local/oci discovery type when building CLI
- Local build for the development purpose, publishes the plugin discovery yamls and distribution artifacts to local discovery and uses that when doing plugin operations
- Test plugins and `Plugin Tests` pipeline still uses non-context-aware discovery for testing 

Bug Fixes:
- Fix the populateStandaloneDiscovery logic (To not add additional default discovery if one already exists)
- Show warning instead of error when a discovery failure. So that `tanzu plugin list` etc operation can report correct status
- Do not throw error when catalog cache file doesn't exist on `tanzu plugin clean`

**NOTE:** Developers using the pre-release CLI, might have the feature flag disabled in the config file and might have to enable it manually by running `tanzu config set features.global.context-aware-discovery true` once. Because the previous default value of the flag was `false`

### Which issue(s) this PR fixes

Fixes #1150

### Describe testing done for PR

- Tested plugins `sync`, `install`, `list` commands with local, oci discovery
- Tested the management-cluster creation with local,oci discovery

### Release note

```release-note
Enabled API-driven CLIPlugin discovery by default. The change makes tanzu CLI context-aware
```

### PR Checklist

- [x] Squash the commits into one or a small number of logical commits
      
- [x] Use good commit [messages](https://github.com/vmware-tanzu/tanzu-framework/blob/main/CONTRIBUTING.md)
- [x] Ensure PR contains terms all contributors can understand and links all contributors can access

PR #1157, opened 2021-11-12
Add install scripts

### LINKED ISSUE #562: Provide a script for installing the CLI from the archive files we publish as a part of releases

As a user, I should be able to run a script that performs the multi-step process of installing the CLI:

- download and untar the appropriate tanzu-framework-(OS)-(ARCH).tar
- `cd` to the tar destination directory:
- move/copy core/v0.1.0/tanzu_core-(OS)-(ARCH) into somewhere in your PATH and rename it `tanzu`
- `export TANZU_CLI_NO_INIT=true`
- `tanzu plugin install --local ./ all`
- `tanzu plugin list` show the following plugins installed: "login", "pinniped-auth", "cluster", "management-cluster", "kubernetes-release", "package"

Community Edition includes scripts that accomplish this:

- https://github.com/vmware-tanzu/community-edition/blob/main/hack/install.bat
- https://github.com/vmware-tanzu/community-edition/blob/main/hack/install.sh

We should also find a better way to distribute the core CLI binary (#3) and provide an API for CLI plugin discovery (#74). However, this is intended as a stop-gap until that's available.

- [ ] APIs
- [ ] Addons
- [x] CLI
- [x] Docs
- [x] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

rajaskakodkar:
This adds install scripts to install tanzu cli on Linux, MacOS and Windows OS. As of now, these scripts need the path where tanzu-framework binary has been extracted to and the release version, as arguments. 

We can drop these arguments in the future once we package these scripts as part of the release.

Signed-off-by: Rajas Kakodkar <rkakodkar@vmware.com>

Fixes #562

These scripts have been tested in 
https://github.com/rajaskakodkar/tanzu-framework/actions/runs/1451790118 -> MacOS and Linux
https://github.com/rajaskakodkar/tanzu-framework/actions/runs/1451789814 -> Windows OS

### Release note

```release-note
NONE
```

### PR Checklist

### DIFF
--- a/hack/install.sh
+++ b/hack/install.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+
+# Copyright 2021 VMware Tanzu Community Edition contributors. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Inspired by - https://github.com/vmware-tanzu/community-edition/blob/main/hack/install.sh
+# Script to install tanzu framework
+# Usage: ./hack/install.sh /path/to/tanzu-framework-binary v0.10.0
+
+# set -o errexit
+set -o nounset
+set -o pipefail
+set -o xtrace
+
+TF_INSTALL_PATH="${1:?Tanzu-Framework path argument empty. Example usage: ./hack/install.sh /path/to/tanzu-framework-binary v0.10.0}"
+VERSION="${2:?Version argument empty. Example usage: ./hack/install.sh /path/to/tanzu-framework-binary v0.10.0}"
+
+ALLOW_INSTALL_AS_ROOT="${ALLOW_INSTALL_AS_ROOT:-""}"
+if [[ "$EUID" -eq 0 && "${ALLOW_INSTALL_AS_ROOT}" != "true" ]]; then
+  echo "Do not run this script as root"
+  exit 1
+fi
+
+MY_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+ARCH="amd64"
+
+case "${OS}" in
+  linux)
+    XDG_DATA_HOME="${HOME}/.local/share"
+    ;;
+  darwin)
+    XDG_DATA_HOME="${HOME}/Library/Application Support"
+    ;;
+  *)
+    echo "${OS} is unsupported"
+    exit 1
+    ;;
+esac
+echo "${XDG_DATA_HOME}"
+
+# check if the tanzu CLI already exists and remove it to avoid conflicts
+TANZU_BIN_PATH=$(command -v tanzu)
+if [[ -n "${TANZU_BIN_PATH}" ]]; then
+  # best effort, so just ignore errors
+  sudo rm -f "${TANZU_BIN_PATH}" > /dev/null
+fi
+
+# set install dir to /usr/local/bin
+TANZU_BIN_PATH="/usr/local/bin"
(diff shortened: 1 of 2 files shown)


hack/install.sh
```diff
@@ -0,0 +1,67 @@
...
+
+# if plugin cache pre-exists, remove it so new plugins are detected
+TANZU_PLUGIN_CACHE="${HOME}/.cache/tanzu/catalog.yaml"
+if [[ -n "${TANZU_PLUGIN_CACHE}" ]]; then
+  echo "Removing old plugin cache from ${TANZU_PLUGIN_CACHE}"
+  rm -f "${TANZU_PLUGIN_CACHE}" > /dev/null
+fi
+
+# install tanzu cli
+install $1/cli/core/$2/tanzu-core-"${OS}"_"${ARCH}" /usr/local/bin/tanzu
```
navidshaikh, line 59: - on Linux, `install` command would require `sudo` AFAIK
- use `TANZU_BIN_PATH` for destination

navidshaikh (changes requested):
Lets sync latest install scripts from TCE repo.

hack/install.sh
rajaskakodkar: I have addressed this in https://github.com/vmware-tanzu/tanzu-framework/pull/1157/commits/db0e4d47f9654288c1a2e9893e069783e1df4580 

thanks!

anujc25:
We might need to change some of this based on #1144 PR.
I am working on updating the documentation and installation guide as part of PR #1188

hack/install.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
+
+# Copyright 2021 VMware Tanzu Community Edition contributors. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Inspired by - https://github.com/vmware-tanzu/community-edition/blob/main/hack/install.sh
+# Script to install tanzu framework
+# Usage: ./hack/install.sh /path/to/tanzu-framework-binary v0.10.0
```
navidshaikh, line 8: `/path/to/tanzu-framework-binary` is actually path to the extracted released artifacts tarball and `$VERSION` is only used in this script to form the path, since this is an intermediate state (we want to get proper install scripts part of the release artifacts), I think we can either
- update the script to expect the exact path to the binary (and not requiring version arg)
- or update the usage

navidshaikh: same for Windows install script

navidshaikh (approved):
lgtm! suggested a small change

hack/install.sh
rajaskakodkar: sure, I'll update the script with the first option. This will eventually get dropped once we have the script as a release artifact.

navidshaikh (approved):
lgtm!

hack/install.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+# Copyright 2021 VMware Tanzu Community Edition contributors. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Inspired by - https://github.com/vmware-tanzu/community-edition/blob/main/hack/install.sh
+# Script to install tanzu framework
+# Usage: ./hack/install.sh /path/to/tanzu-framework/core/binary
+
+# set -o errexit
```
vuil, line 10: Was there a reason for this being commented out?

vuil (approved):
lgtm, but please address the commit in a followup

hack/install.sh
rajaskakodkar: I'll fix it in a follow-up PR. Thanks!

PR #1162, opened 2021-11-15
Add cert-manager templates to providers

### LINKED ISSUE #1114: Seeing errors installing cert-manager v1.5.3 after the CAPI bump
**Bug description**
We are leveraging CAPI to pull the cert-manager deployment manifests when creating/upgrading management cluster. This is resulting in error due to GitHub rate limiting.

```
06:40:09  Checking cert-manager version...
06:40:09  Deleting cert-manager Version="v1.1.0"
06:40:11  Installing cert-manager Version="v1.5.3"
06:40:12  Error: failed to upgrade management cluster providers: failed to apply providers upgrade: failed to get repository client for the  with name cert-manager: error creating the GitHub repository client: failed to get GitHub latest version: failed to get repository versions: failed to get repository versions: rate limit for github api has been reached. Please wait one hour or get a personal API tokens a assign it to the GITHUB_TOKEN environment variable 
```
https://github.com/kubernetes-sigs/cluster-api/blob/main/cmd/clusterctl/client/config/cert_manager_client.go#L69

- [ ] APIs
- [ ] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**

saimanoj01:
In CAPI v1.0.0, it moved away from relying on a packaged version of cert-manager manifests during management cluster creation. Instead it is pulling the cert-manager manifest from github. This change ensures that CAPI uses the cert-manager manifests packaged along with the providers instead of pulling from github.

Fixes #1114

1. Deleted the `~/config/tanzu/tkg` directory and successfully created a management cluster. Made sure that the cert-manager image is being pulled from the tkg registry.
2. Created a management cluster with the config file already populated with cert-manager entry.

### Release note

```release-note
Package cert-manager v1.5.3 manifests along with the providers. 
```

### PR Checklist


prkalle (approved):
LGTM. Thanks!

pkg/v1/tkg/tkgconfigupdater/provider.go
```diff
@@ -46,17 +52,44 @@ func (c *client) defaultProviders() (providers, error) {
...
+	certManagerBytes, err := yaml.Marshal(defaultProviders.CertManager)
+	if err != nil {
+		return errors.Wrap(err, "unable to get cert-manager")
+	}
+
+	certManagerNode := yaml.Node{}
+	err = yaml.Unmarshal(certManagerBytes, &certManagerNode)
+	if err != nil {
+		return errors.Wrap(err, "unable to get cert-manager")
+	}
```
imikushin, line 87: suggestion: Extract as a helper func. It will simplify the code and improve readability. I've also noticed at least two other places where it could be used: see my comment on `EnsureProvidersInConfig()`.

The signature would look something like this:
```go
func copyData(from, to interface{}) error
```

anujc25 (commented):
Is the addition of `pkg/v1/tkg/fakes/providers/providers.zip` file intentional? Is it getting used for unit tests?

Otherwise LGTM

saimanoj01:
@anujc25 Yes, that was intentional. I modified it for the sake of unit tests. Thanks

pkg/v1/tkg/tkgconfigupdater/provider.go
```diff
@@ -75,18 +75,43 @@ func ensureCertManagerInConfig(defaultProviders providers, tkgConfigNode *yaml.N
...
+
+	switch v := to.(type) {
+	case *yaml.Node, *providers:
+		err = yaml.Unmarshal(bytes, v)
+		if err != nil {
+			return err
+		}
+	default:
+		return errors.New("unable to configure providers")
```
imikushin: There's no need for the `switch`. This would work just as well:

```go
 if err := yaml.Unmarshal(bytes, to); err != nil {
 	return err
 }
```

We can wrap the error where we're calling `copyData()`

imikushin (commented):
Thanks for addressing my comments! Just one small detail and it's good to go.

pkg/v1/tkg/tkgconfigupdater/provider.go
saimanoj01: Done!

PR #1229, opened 2021-11-23
Fix the issue where cluster upgrade doesn't timeout and stuck indefin…

prkalle:
…itely

This PR fixes the issue where the cluster create/upgrade doesn't timeout and stuck indefinitely.
This PR introduces a change where in if the cluster keeps changing the state(current wait logic depends on the cluster state change and would wait if the cluster creation/upgrade is in progress) cli would **also** check if the operation exceeds max timeout (which depends on the timeout value provided by the user). If the operation exceeds maxTimeout it would return error

Fixes # #1230 

Create management cluster on Azure and was successfull
Created workload cluster and was success
Upgraded workload cluster and was success

### Release note

```release-note
Fix the issue where where cluster upgrade doesn't timeout and stuck indefinitely
```

### PR Checklist

### DIFF
--- a/pkg/v1/tkg/clusterclient/clusterclient.go
+++ b/pkg/v1/tkg/clusterclient/clusterclient.go
@@ -486,6 +486,9 @@ func (c *client) WaitForClusterInitialized(clusterName, namespace string) error 
 	var lastReason string
 	unchangedCounter := 0
 	interval := 15 * time.Second
+	// maxTimeout to time-bound wait operation to avoid indefinite wait if the cluster state keeps changing
+	maxTimeout := 3 * c.operationTimeout
+	maxTimeoutCounter := 0
 
 	getterFunc := func() (interface{}, error) {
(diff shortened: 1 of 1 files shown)


pkg/v1/tkg/clusterclient/clusterclient.go
```diff
@@ -486,6 +486,8 @@ func (c *client) WaitForClusterInitialized(clusterName, namespace string) error
 	var lastReason string
 	unchangedCounter := 0
 	interval := 15 * time.Second
+	maxTimeout := 3 * c.operationTimeout
```
anujc25, line 489: We should add a comment to explain why we added this timeout.

anujc25 (commented):
Overall looks good.

pkg/v1/tkg/clusterclient/clusterclient.go
prkalle: sure, updated

anujc25 (approved):
lgtm

PR #1241, opened 2021-11-29
Sync calico package updates to the ClusterResourceSet

12345lcr:
Update the calico cni in `/pkg/v1/providers/ytt/vendir/cni/_ytt_lib/addons/packages/calico ` to fetch the newest changes form tce by running vendir sync.

Changes include:
* Dual-stack support
* Disable label scoping rule for calico-kube-controllers
* Make MTU configurable in calico

Fixes: https://github.com/vmware-tanzu/tanzu-framework/issues/1275

### Release note

```release-note
None
```

### PR Checklist

After running `vendir sync`, I excluded the updates that not for calico

### DIFF
--- a/pkg/v1/providers/ytt/vendir/cni/_ytt_lib/addons/packages/calico/3.19.1/bundle/config/overlay/calico_overlay.yaml
+++ b/pkg/v1/providers/ytt/vendir/cni/_ytt_lib/addons/packages/calico/3.19.1/bundle/config/overlay/calico_overlay.yaml
@@ -6,6 +6,16 @@
 #@  return left.startswith(right)
 #@ end
 
+#@overlay/match by=overlay.subset({"kind":"Deployment","metadata":{"name": "calico-kube-controllers"}})
+---
+kind: Deployment
+metadata:
+  name: calico-kube-controllers
+  #@overlay/match missing_ok=True
+  annotations:
+    kapp.k14s.io/disable-default-label-scoping-rules: ""
+    kapp.k14s.io/update-strategy: always-replace
+
 #@overlay/match by=overlay.subset({"kind":"DaemonSet", "metadata": {"name": "calico-node"}})
 ---
 kind: DaemonSet
@@ -20,7 +30,7 @@ spec:
         #@overlay/match by=overlay.subset({"name":"calico-node"})
         - name: calico-node
           env:
-#@ if values.ipFamily == "ipv6":
+            #@ if values.ipFamily == "ipv6":
             #@overlay/match by=overlay.subset({"name":"IP"})
             - value: none
             #@overlay/match by=overlay.subset({"name":"FELIX_IPV6SUPPORT"})
@@ -31,13 +41,13 @@ spec:
             #@overlay/match by=overlay.subset({"name":"CALICO_IPV4POOL_VXLAN"})
             #@overlay/remove
             - {}
-#! Calico v3.19.1 support cluster CIDR auto-detect.
-#! So the cluster CIDR is only appended when configured in values.
-#@ if values.calico.config.clusterCIDR:
+            #! Calico v3.19.1 support cluster CIDR auto-detect.
+            #! So the cluster CIDR is only appended when configured in values.
+            #@ if values.calico.config.clusterCIDR:
             #@overlay/append
             - name: CALICO_IPV6POOL_CIDR
               value: #@ values.calico.config.clusterCIDR
-#@ end
+            #@ end
             #@overlay/append
             - name: IP6
               value: autodetect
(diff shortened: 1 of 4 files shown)


12345lcr:
Wait until these changes merged into bom file and I will test again

pkg/v1/providers/ytt/vendir/cni/_ytt_lib/addons/packages/calico/3.19.1/bundle/config/overlay/calico_overlay.yaml
```diff
@@ -6,6 +6,16 @@
...
 
+#@overlay/match by=overlay.subset({"kind":"Deployment","metadata":{"name": "calico-kube-controllers"}})
+---
+kind: Deployment
+metadata:
+  name: calico-kube-controllers
+  #@overlay/match missing_ok=True
+  annotations:
+    kapp.k14s.io/disable-default-label-scoping-rules: ""
+    kapp.k14s.io/update-strategy: always-replace
```
vijaykatam, line 17: `always-replace` means kapp will delete the deployment and then install instead of an update. What is the reasoning for doing `always-replace` for a critical component like CNI?

12345lcr: Because after we add this `disable-default-label-scoping-rules` annotation, kapp no longer add its `labelSelector` to calico-kube-controllers. But for a cluster update with old version of calico installed who do not have this annotation, means the kapp `labelSelector` already exists. Since this field is immutable, it will leads to an update failure due to not able to remove kapp `labelSelector`.

Calico-kube-controllers is a set of controllers who only have 1 replica, which means only 1 pod and will be restarted during update anyway, and it does not obtain any volume also, so the best and only way to prevent update failure is to replace it during updating.

vijaykatam: If I am reading https://carvel.dev/kapp/docs/latest/faq/#migrating-from-kubectl-apply-to-kapp correctly, just the annotation should be enough to tell kapp to not touch the labels. Looping in @cppforlife for input.

Also consider using `fallback-on-replace` instead of `always-replace`

12345lcr: I was also considered `fallback-on-replace`, based on [kapp doc](https://carvel.dev/kapp/docs/latest/apply/#kappk14sioupdate-strategy), this fallback is used for service rolling updates who have multiple pods to prevent downtime, or pods who obtain volumes to prevent data loss, which none of them meets calico-kube-controller.

And when doing an update to an deployment, some field change(for example label or container environment variable) will not lead to a pod restart or recreate, I was worried that it may leads to any unexpected behavior in the future if we have any manifest changes but do not replace it.

vijaykatam: > And when doing an update to an deployment, some field change

Field changes always trigger an update by kapp, if there is an actual issue lets get that logged to kapp and get it triaged.

Going back to original issue, does just adding `disable-default-label-scoping-rules` not fix the issue you are seeing?

12345lcr: @vijaykatam No, we firstly added `disable-default-label-scoping-rules`, it will solve the **management cluster install** error when pkgi adopt ClusterResourceSet calico componentes. But after this annotation is added, latter we found all the  **workload cluster upgrading** is breaking with the error below.
```
[2021-11-30T03:19:47.656Z] kapp: Error: Applying update deployment/calico-kube-controllers (apps/v1) namespace: kube-system:
[2021-11-30T03:19:47.656Z]   Updating resource deployment/calico-kube-controllers (apps/v1) namespace: kube-system:
[2021-11-30T03:19:47.656Z]     Deployment.apps "calico-kube-controllers" is invalid: spec.selector:
[2021-11-30T03:19:47.656Z]       Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"k8s-app":"calico-kube-controllers"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable (reason: Invalid), retrying
```
From the log you can see that it is trying to update a "clean" LabelSelector, which do not contain kapp LabelSelectors. This error showed up because the previous version of calico that installed in workload cluster do not have the `disable-default-label-scoping-rules`, that means it contains kapp LabelSelector, and after we add `disable-default-label-scoping-rules` in new version of calico, during update it will try to remove those kapp LabelSelector, which is immutable.

Then we added this `update-strategy` annotation to let it replace the deployment of calico-kube-controllers who has kapp LabelSelector, and the problem is fixed. If we do not add this, the upgrade of workload cluster becomes a breaking change which will always fail.

vijaykatam: I see, in that case can we change update-strategy to fallback-on-replace?

12345lcr: https://github.com/projectcalico/calico/blob/master/getting-started/kubernetes/installation/calico-kube-controllers.yaml#L19-L25
Hey @vijaykatam , I found in calico manifest that calico-kube-controllers deployment already set the strategy to `Recreate` due to only one single instance of the this pod should be active at a time. Otherwise there will be potential problems like reconciling at the same time and creating duplicate resources.

This means during the update, the pod will be deleted first, then a new one is recreated. There is no `RollingUpdate` situation that we need to take care of.

So from the pod level, basically there is no difference between `fallback-on-replace` and `always-replace`, the pod will always be deleted and recreated no matter which update-strategy we choose. Since the `always-replace` strategy is fully tested and we are approaching release date, I suggest we keep this as it is to follow the `Recreate` pattern. And if anything unexpected happened let's create another issue to track, triage and fix. What do you think?

vijaykatam: For pod level there is no difference but the deployment could have other objects associated to it, in which case there is difference between kapp delete deployment object on `always-replace` and deployment deleting pod object on `Recreate`. You are going for nuclear options always but looks like `calico-kube-controllers` doesn't have any other objects such as configmaps or secrets associated to it so in this case it is identical. 

It looks like you did a bunch of testing and want to get this merged and I am not opposed to it but in the future have me review upstream TCE PR so that we can have it resolved there before you do a bunch of downstream integration testing.

vijaykatam (approved):

pkg/v1/providers/ytt/vendir/cni/_ytt_lib/addons/packages/calico/3.19.1/bundle/config/overlay/calico_overlay.yaml
12345lcr: Totally agree. Currently in tce there is also a concept to standardize the update strategy.
https://github.com/vmware-tanzu/community-edition/pull/2277

PR #1242, opened 2021-11-29
add packaging API check for non-tanzu cluster

### LINKED ISSUE #340: tanzu package plugin takes too long to respond on non tanzufied clusters
**Bug description**
If you execute `tanzu package available list` on a cluster that does not have kapp-controller 0.20+ installed, it takes between 6 and 10 seconds to get an answer with the following error, followed by the help description:
```
tanzu package available list
\ Retrieving available packages... Error: no matches for kind "PackageMetadata" in version "data.packaging.carvel.dev/v1alpha1"
Usage:
  tanzu package available list or list PACKAGE_NAME [flags]

Examples:

    # List available packages across all namespaces
    tanzu package available list -A

    # List all versions for available package from specified namespace
    tanzu package available list contour.tanzu.vmware.com --namespace test-ns

Flags:
  -h, --help   help for list

Global Flags:
  -A, --all-namespaces      If present, list packages across all namespaces.
      --kubeconfig string   The path to the kubeconfig file, optional
      --log-file string     Log file path
  -n, --namespace string    Namespace of packages, optional (default "default")
  -o, --output string       Output format (yaml|json|table)
      --verbose int32       Number for the log level verbosity(0-9)

Error: exit status 1
```
The error message `Error: no matches for kind "PackageMetadata" in version "data.packaging.carvel.dev/v1alpha1"` is buried in the output, not intuitive for many users and the help message aftwards does not make sense given it's an incompatibility with the cluster and not that the provided CLI command is wrong.

- [x] APIs
- [ ] Addons
- [ ] CLI
- [ ] Docs
- [ ] Installation
- [X] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**
I would expect the CLI to error in less time, with a more "general user" friendly message (e.g. This cluster does not have Tanzu Package capabilities installed) and without the error message afterwards.

- Select a kube context of a non tanzuified cluster or a cluster with kapp-controller less than 0.20
- Execute `tanzu package available list` (time the command)

```
tanzu package info
{"name":"package","description":"Tanzu package management","version":"v1.4.0-pre-alpha-2","buildSHA":"5a42a2b","group":"Run","docURL":"","completionType":0}
```

macOS

Since there's no --verbose output for this plugin or --log-file, nothing can be provided.

ggpaue:
If cluster doesn't have Carvel packaging API available, package plugin will return error immediately

Fixes #340 

Passed package plugin integration test and manually tested with a cluster without Carvel packaging api-resources

### Release note

```release-note
check Carvel packaging API before package plugin cmd executes
```

### PR Checklist

### DIFF
--- a/cmd/cli/plugin/package/main.go
+++ b/cmd/cli/plugin/package/main.go
@@ -12,6 +12,9 @@ import (
 	cliv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/cli/v1alpha1"
 	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli/command/plugin"
 	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli/component"
+	capdiscovery "github.com/vmware-tanzu/tanzu-framework/pkg/v1/sdk/capabilities/discovery"
+	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkg/kappclient"
+	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkg/tkgpackagedatamodel"
 )
 
 var descriptor = cliv1alpha1.PluginDescriptor{
@@ -51,4 +54,23 @@ func getOutputFormat() string {
 		format = string(component.ListTableOutputType)
 	}
 	return format
+}
+
+func isPackagingAPIAvailable(kubeCfgPath string) (bool, error) {
+	cfg, err := kappclient.GetKubeConfig(kubeCfgPath)
+	if err != nil {
+		return false, err
+	}
+	clusterQueryClient, err := capdiscovery.NewClusterQueryClientForConfig(cfg)
+	if err != nil {
+		log.Error(err, "failed to create a new instance of the cluster query builder")
+		return false, err
+	}
+
+	apiGroup1 := capdiscovery.Group("packageMetadateAPIQuery", tkgpackagedatamodel.DataPackagingAPIName).WithVersions(tkgpackagedatamodel.PackagingAPIVersion).WithResource("packagemetadatas")
+	apiGroup2 := capdiscovery.Group("packageAPIQuery", tkgpackagedatamodel.DataPackagingAPIName).WithVersions(tkgpackagedatamodel.PackagingAPIVersion).WithResource("packages")
+	apiGroup3 := capdiscovery.Group("packageRepositoryAPIQuery", tkgpackagedatamodel.PackagingAPIName).WithVersions(tkgpackagedatamodel.PackagingAPIVersion).WithResource("packagerepositories")
+	apiGroup4 := capdiscovery.Group("packageInstallAPIQuery", tkgpackagedatamodel.PackagingAPIName).WithVersions(tkgpackagedatamodel.PackagingAPIVersion).WithResource("packageinstalls")
+
+	return clusterQueryClient.Query(apiGroup1, apiGroup2, apiGroup3, apiGroup4).Execute()
 }
--- a/cmd/cli/plugin/package/package_available.go
+++ b/cmd/cli/plugin/package/package_available.go
@@ -4,6 +4,9 @@
 package main
 
 import (
+	"fmt"
+
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
 	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkg/tkgpackagedatamodel"
@@ -12,14 +15,27 @@ import (
 var packageAvailableOp = tkgpackagedatamodel.NewPackageAvailableOptions()
 
 var packageAvailableCmd = &cobra.Command{
-	Use:       "available",
-	ValidArgs: []string{"list", "get"},
-	Short:     "Manage available packages",
-	Args:      cobra.RangeArgs(1, 2),
+	Use:               "available",
+	ValidArgs:         []string{"list", "get"},
+	Short:             "Manage available packages",
+	Args:              cobra.RangeArgs(1, 2),
+	PersistentPreRunE: packagingAvailabilityCheck,
 }
 
 func init() {
 	packageAvailableCmd.PersistentFlags().StringVarP(&packageAvailableOp.KubeConfig, "kubeconfig", "", "", "The path to the kubeconfig file, optional")
 	packageAvailableCmd.PersistentFlags().StringVarP(&packageAvailableOp.Namespace, "namespace", "n", "default", "Namespace of packages, optional")
 	packageAvailableCmd.PersistentFlags().StringVarP(&outputFormat, "output", "o", "", "Output format (yaml|json|table), optional")
+}
+
+func packagingAvailabilityCheck(_ *cobra.Command, _ []string) error {
+	found, err := isPackagingAPIAvailable(packageAvailableOp.KubeConfig)
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("failed to check for the availability of '%s' API", tkgpackagedatamodel.PackagingAPIName))
+	}
+	if !found {
+		return errors.New(fmt.Sprintf(tkgpackagedatamodel.PackagingAPINotAvailable, tkgpackagedatamodel.PackagingAPIName, tkgpackagedatamodel.PackagingAPIVersion))
+	}
+
+	return nil
 }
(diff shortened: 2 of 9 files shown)


cmd/cli/plugin/package/main.go
```diff
@@ -52,3 +54,48 @@ func getOutputFormat() string {
...
+	if err != nil {
+		return false, err
+	}
+	clusterQueryClient, err := capdiscovery.NewClusterQueryClientForConfig(cfg)
+	if err != nil {
+		log.Error(err, "failed to create a new instance of the cluster query builder")
+		return false, err
+	}
+
+	apiGroup := capdiscovery.Group("packageMetadateAPIQuery", "data.packaging.carvel.dev").WithVersions("v1alpha1").WithResource("packagemetadatas")
```
vijaykatam, line 69: I think you can combine all three queries into one with discovery client. cc @rajathagasthya

rajathagasthya: Yep, you can combine by adding multiple arguments to `Query()`.

```go
q1 := capdiscovery.Group("packageMetadateAPIQuery", "data.packaging.carvel.dev").WithVersions("v1alpha1").WithResource("packagemetadatas")
q2 := capdiscovery.Group("packageRepositoryAPIQuery", "packaging.carvel.dev").WithVersions("v1alpha1").WithResource("packagerepositories")
q3 := capdiscovery.Group("packageInstallAPIQuery", "packaging.carvel.dev").WithVersions("v1alpha1").WithResource("packageinstalls")

clusterQueryClient.Query(q1, q2, q3).Execute()
```

cmd/cli/plugin/package/package_install.go
```diff
@@ -25,8 +25,9 @@ var packageInstallCmd = &cobra.Command{
 	
     # Install package contour with kubeconfig flag and waiting for package reconciliation to complete	
     tanzu package install contour-pkg --package-name contour.tanzu.vmware.com --namespace test-ns --version 1.15.1-tkg.1-vmware1 --kubeconfig path/to/kubeconfig`,
-	RunE:         packageInstall,
-	SilenceUsage: true,
+	RunE:              packageInstall,
+	SilenceUsage:      true,
+	PersistentPreRunE: packageMetadataAvailabilityCheck,
```
maralavi, line 30: I think ```"package install"``` and ```"package installed"``` operation should use the same availability call (they need both ```"packagemetadatas"``` and ```"packageinstalls"``` api to be available). But right now, ```"package install"``` is using ```"packageMetadataAvailabilityCheck"``` and ```"package installed"``` operations are using ```"packageInstallAvailabilityCheck"```.

cmd/cli/plugin/package/main.go
ggpaue: thanks for the review @vijaykatam , I'm wondering should we combine all the API checks together?

for example, `package available` only needs packageMetadateAPI check, `package repository` only needs packageRepositoryAPI check

ggpaue: bring @maralavi into this discussion

maralavi: @ggpaue It won't hurt to check the availability of all apis for all commands. Anyways whoever is going to use package plugin needs all of those be available (eventhough when one of those available, almost always the remaining related apis for packaging are also available). I assume the discovery calls are quite fast and you could merge all these 3 into one query and use this combined approach as suggested above, and you could call that combined query for all commands.

cmd/cli/plugin/package/package_install.go
ggpaue: thanks for the review @maralavi , now we put all the checks together

cmd/cli/plugin/package/main.go
ggpaue: updated the code to put all API queries together

cmd/cli/plugin/package/main.go
```diff
@@ -52,3 +54,22 @@ func getOutputFormat() string {
 	}
 	return format
 }
+
+func isCarvelAPIAvailable(kubeCfgPath string) (bool, error) {
```
maralavi, line 58: Isn't ```isPackagingAPIAvailable``` a slightly better naming?

cmd/cli/plugin/package/package_available.go
```diff
@@ -12,14 +15,27 @@ import (
...
+	PersistentPreRunE: carvelAvailabilityCheck,
 }
 
 func init() {
 	packageAvailableCmd.PersistentFlags().StringVarP(&packageAvailableOp.KubeConfig, "kubeconfig", "", "", "The path to the kubeconfig file, optional")
 	packageAvailableCmd.PersistentFlags().StringVarP(&packageAvailableOp.Namespace, "namespace", "n", "default", "Namespace of packages, optional")
 	packageAvailableCmd.PersistentFlags().StringVarP(&outputFormat, "output", "o", "", "Output format (yaml|json|table), optional")
 }
+
+func carvelAvailabilityCheck(_ *cobra.Command, _ []string) error {
```
maralavi, line 31: how about ```packagingAPIAvailabilityCheck```?

cmd/cli/plugin/package/main.go
ggpaue: thanks for the naming advice @maralavi 
updated the function name accordingly

maralavi (approved):

danniel1205:
Questions:
- Shall we update the release notes in description ? It looks like belongs to `Tests done` section. 
- Shall we add unit/integ tests to cover the newly added function, especially for negative cases ?

cmd/cli/plugin/package/main.go
```diff
@@ -52,3 +54,22 @@ func getOutputFormat() string {
...
+	clusterQueryClient, err := capdiscovery.NewClusterQueryClientForConfig(cfg)
+	if err != nil {
+		log.Error(err, "failed to create a new instance of the cluster query builder")
+		return false, err
+	}
+
+	apiGroup1 := capdiscovery.Group("packageMetadateAPIQuery", "data.packaging.carvel.dev").WithVersions("v1alpha1").WithResource("packagemetadatas")
+	apiGroup2 := capdiscovery.Group("packageAPIQuery", "data.packaging.carvel.dev").WithVersions("v1alpha1").WithResource("packages")
+	apiGroup3 := capdiscovery.Group("packageRepositoryAPIQuery", "packaging.carvel.dev").WithVersions("v1alpha1").WithResource("packagerepositories")
+	apiGroup4 := capdiscovery.Group("packageInstallAPIQuery", "packaging.carvel.dev").WithVersions("v1alpha1").WithResource("packageinstalls")
```
danniel1205, line 72: question: Is there a way we could discover resources without hardcoding the version? A use case I could think of is:

* I have v1alpha1 carvel with my current cluster
* I upgrade my cluster which has v1beta1 carvel now
* My current package plugin will break because there will be no `v1alpha1` anymore

ggpaue: not sure how to discover the API version but I believe the upgrade case will also affect secretGen API version

https://github.com/vmware-tanzu/tanzu-framework/blob/main/cmd/cli/plugin/secret/main.go

cc @maralavi

danniel1205: By looking at this [code](https://github.com/vmware-tanzu/tanzu-framework/blob/978797e5abd67ee1a8b0e1a087340f87e4ae3dd4/pkg/v1/sdk/capabilities/discovery/cluster_gvr.go#L51), I dont think `WithVersions` is required. Could you help test it out ?

danniel1205:
Adding `do-not-merge/hold` because this [thread](https://github.com/vmware-tanzu/tanzu-framework/pull/1242#discussion_r761528548) needs to be addressed.

ggpaue:
> Questions:
> 
> * Shall we update the release notes in description ? It looks like belongs to `Tests done` section.
> * Shall we add unit/integ tests to cover the newly added function, especially for negative cases ?

thanks @danniel1205 , the packaging API check is in each command `PersistentPreRunE`, so the current integration test will cover this check. I will add another case which will delete kapp then see if it can return error

cmd/cli/plugin/package/main.go
ggpaue: yeah I've deleted `WithVersions` and it will fall into this error:
https://github.com/vmware-tanzu/tanzu-framework/blob/978797e5abd67ee1a8b0e1a087340f87e4ae3dd4/pkg/v1/sdk/capabilities/discovery/cluster_gvr.go#L53
`
Error: failed to check for the availability of 'packaging.carvel.dev/v1alpha1' API: cannot check for group and resource existence without version info; use WithVersion method`

ggpaue: added constants to represent API name and version instead of hardcoding

danniel1205 (approved):
LGTM

danniel1205:
A github [issue](https://github.com/vmware-tanzu/tanzu-framework/issues/1294) has been filed to follow up on this discussion: https://github.com/vmware-tanzu/tanzu-framework/pull/1242#discussion_r761528548. We might want to switch to new discover API once it allows us to discover resources without specifying a version.

blc1996 (approved):
LGTM

PR #1255, opened 2021-11-29
Makes the bootstrap token ttl configurable

srm09:
This patch changes the bootstrap token ttl value to a customizable value which can be overridden by setting the env variable in the config file.
The default ttl value is set to 15m which is same as the default value of the flag in the kubeadm bootstrap provider.

Fixes # n/a

### Release note

```release-note
Makes the bootstrap kubeadm token TTL value configurable
```

### PR Checklist


vuil (approved):
lgtm, thanks!

PR #1278, opened 2021-12-02
Support installing plugins with local filesystem when context-aware-cli-for-plugins feature is enabled

### LINKED ISSUE #1264: Support installing plugins with local filesystem when `context-aware-cli-for-plugins` feature is enabled

With the implementation of #1150, we have enabled the feature flag for context-aware-cli plugins discovery by default. With this, the tanzu CLI can discover and install plugins from OCI based image registry for official builds and from the local filesystem only during development.

As we currently only support local discovery and distribution for the development purpose because the local plugin discovery and distribution currently rely on absolute filesystem paths to be defined as part of CLIPlugin resources making it not usable for distribution and sharing among the different users.

The request is to propose a way to allow users to install standalone plugins from the local file system that can be distributed and downloaded as a tarball from a verified source and make them available to the tanzu cli when `context-aware-cli-for-plugins` feature is enabled.

Users can disable `context-aware-cli-for-plugins` the feature entirely and install the plugins from the local source however, as we made the feature enabled by default this doesn't seem like the right thing to do long term.

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [x] Plugin
- [ ] Security
- [ ] Test and Release
- [x] User Experience

**Additional context**

anujc25:

Changes:
- Return installed plugins even if they are not discovered from any discovery source
- When generating CLIPlugin resource yaml for local discovery, include the relative path to the local CLIPlugin definitions instead of using an absolute path. This will allow us to distribute it as a tarball
- Add discoveryType to discovered object and give precedence to locally
- Install plugins with `tanzu plugin install all --local <path>` command

**UX Proposal:**
https://github.com/vmware-tanzu/tanzu-framework/issues/1264#issuecomment-984304211 
https://github.com/vmware-tanzu/tanzu-framework/issues/1264#issuecomment-986982087

Fixes #1264

### Release note

```release-note
Support installing plugins with local filesystem when context-aware-cli-for-plugins feature is enabled
The `--local` flag with the `tanzu plugin` command now only takes the single input directory instead of multiple directories as an input
```

### PR Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,7 @@ ifndef ENABLE_CONTEXT_AWARE_PLUGIN_DISCOVERY
 ENABLE_CONTEXT_AWARE_PLUGIN_DISCOVERY = "true"
 endif
 ifndef DEFAULT_STANDALONE_DISCOVERY_IMAGE_PATH
-DEFAULT_STANDALONE_DISCOVERY_IMAGE_PATH = "packages/management/standalone-cliplugins"
+DEFAULT_STANDALONE_DISCOVERY_IMAGE_PATH = "packages/standalone/standalone-plugins"
 endif
 ifndef DEFAULT_STANDALONE_DISCOVERY_IMAGE_TAG
 DEFAULT_STANDALONE_DISCOVERY_IMAGE_TAG = "${BUILD_VERSION}"
@@ -125,7 +125,8 @@ ARTIFACTS_DIR ?= ./artifacts
 ARTIFACTS_ADMIN_DIR ?= ./artifacts-admin
 
 XDG_CACHE_HOME := ${HOME}/.cache
-XDG_CONFIG_HOME :=${HOME}/.config
+XDG_CONFIG_HOME := ${HOME}/.config
+TANZU_PLUGIN_PUBLISH_PATH ?= $(XDG_CONFIG_HOME)/tanzu-plugins
 
 export XDG_DATA_HOME
 export XDG_CACHE_HOME
@@ -240,6 +241,9 @@ CLI_ADMIN_JOBS_OCI_DISCOVERY := $(addprefix build-plugin-oci-,${ENVS})
 CLI_JOBS_LOCAL_DISCOVERY := $(addprefix build-cli-local-,${ENVS})
 CLI_ADMIN_JOBS_LOCAL_DISCOVERY := $(addprefix build-plugin-admin-local-,${ENVS})
 
+LOCAL_PUBLISH_PLUGINS_JOBS := $(addprefix publish-plugins-local-,$(ENVS))
+
+
 RELEASE_JOBS := $(addprefix release-,${ENVS})
 
 .PHONY: build-cli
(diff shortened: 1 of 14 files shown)


pkg/v1/cli/command/core/plugin_manager.go
```diff
@@ -222,21 +231,43 @@ var describePluginCmd = &cobra.Command{
 var installPluginCmd = &cobra.Command{
 	Use:   "install [name]",
 	Short: "Install a plugin",
```
tenczar, line 233:
```suggestion
	Short: "Install a plugin",
	Args:  cobra.ExactArgs(1),
```

instead of the explicit check at line 226

pkg/v1/cli/pluginmanager/manager.go
```diff
@@ -154,16 +168,60 @@ func AvailablePlugins(serverName string) ([]plugin.Discovered, error) {
...
+func getInstalledButNotDiscoveredStandalonePlugins(availablePlugins []plugin.Discovered, installedPluginDesc []cliv1alpha1.PluginDescriptor) []plugin.Discovered {
+	var newPlugins []plugin.Discovered
+	for i := range installedPluginDesc {
+		found := false
+		for j := range availablePlugins {
+			if installedPluginDesc[i].Name == availablePlugins[j].Name {
+				found = true
+				// If plugin is installed but marked as not installed as part of availablePlugins list
+				// mark the plugin as installed
+				// This is possible if use has used --local mode to install the plugin which is also
```
tenczar, line 186: nit: use -> user

pkg/v1/cli/pluginmanager/manager.go
```diff
@@ -154,16 +168,60 @@ func AvailablePlugins(serverName string) ([]plugin.Discovered, error) {
...
+	}
+	return newPlugins
+}
+
+// DiscoveredFromPluginDescriptor returns discovered plugin object from k8sV1alpha1
+func DiscoveredFromPluginDescriptor(p *cliv1alpha1.PluginDescriptor) plugin.Discovered {
+	dp := plugin.Discovered{
+		Name:               p.Name,
+		Description:        p.Description,
+		RecommendedVersion: p.Version,
```
tenczar, line 208: Should this include DiscoveryType, or is it even knowable here?

pkg/v1/cli/pluginmanager/manager.go
```diff
@@ -425,6 +483,80 @@ func SyncPlugins(serverName string) error {
...
+	plugins, err := DiscoverPluginsFromLocalSource(localPath)
+	if err != nil {
+		return err
+	}
+
+	found := false
+
+	errList := make([]error, 0)
+	for idx := range plugins {
+		if pluginName == plugins[idx].Name || pluginName == cli.AllPlugins {
```
tenczar, line 501: I would switch the order of this `||`. If it is AllPlugins, you may as well short circuit the rest of the if.

pkg/v1/cli/pluginmanager/manager.go
```diff
@@ -425,6 +483,80 @@ func SyncPlugins(serverName string) error {
...
+}
+
+// DiscoverPluginsFromLocalSource returns the available plugins that are discovered from the provided local path
+func DiscoverPluginsFromLocalSource(localPath string) ([]plugin.Discovered, error) {
+	if localPath == "" {
+		return nil, nil
+	}
+
+	// Set default local plugin distro to localpath as while installing the plugin
+	// from local source we should take t
```
tenczar, line 526: incomplete comment?

pkg/v1/cli/pluginmanager/manager.go
```diff
@@ -425,6 +483,80 @@ func SyncPlugins(serverName string) error {
...
+
+	plugins, err := discoverPlugins(pds)
+	if err != nil {
+		return nil, err
+	}
+
+	for i := range plugins {
+		plugins[i].Scope = common.PluginScopeStandalone
+		plugins[i].Status = common.PluginStatusNotInstalled
+	}
```
tenczar, line 555: would it be meaningful to set the DiscoveryType here?

tenczar (approved):
just a couple questions and nits.

pkg/v1/cli/pluginmanager/manager.go
anujc25: Thanks. Yes, we should. Added it here.

pkg/v1/cli/pluginmanager/manager.go
anujc25: No. We will not get the `DiscoveryType` information from the plugin descriptor.

Makefile
```diff
@@ -240,6 +241,9 @@ CLI_ADMIN_JOBS_OCI_DISCOVERY := $(addprefix build-plugin-oci-,${ENVS})
 CLI_JOBS_LOCAL_DISCOVERY := $(addprefix build-cli-local-,${ENVS})
 CLI_ADMIN_JOBS_LOCAL_DISCOVERY := $(addprefix build-plugin-admin-local-,${ENVS})
 
+LOCAL_PUBLISH_PLUGINS_JOBS := $(addprefix publish-plugins-local-,${ENVS})
```
navidshaikh, line 244:
```suggestion
LOCAL_PUBLISH_PLUGINS_JOBS := $(addprefix publish-plugins-local-,$(ENVS))
```

lets use single notation for vars with parenthesis, we've a PR open streamlining this across the Makefile

vuil:
Some nits: but lgtm in general. 
The is a behavior change on the handling of multiple --local flags into a single one. It is reasonable simplication and even warranted given how the installation flow has evolved. We should documented the change in the docs.

vuil (approved):
LGTM, thanks!

PR #883, opened 2021-10-13
Make Capability GVR query's group field optional

rajathagasthya:
### What this PR does / why we need it

Changes GVR query's `group` field to be optional. Optional field lets you query core k8s resources as they don't have an
API group name.

### Which issue(s) this PR fixes

Fixes #882 

### Describe testing done for PR

Apply new CRD and create a Capability CR:

```
apiVersion: run.tanzu.vmware.com/v1alpha1
kind: Capability
metadata:
  name: capability-sample
spec:
  queries:
    - name: "q1"
      groupVersionResources:
        - name: "fooGVR"
          group: ""
          versions:
            - v1
          resource: "pods"
```

Verify CR is successfully created and controller reconciles result.

### Release note

```release-note
The Capability CRD now allows specifying an empty group name in the GVR query to check for core Kubernetes resources.
```

### PR Checklist

- [x] Squash the commits into one or a small number of logical commits
      
- [x] Use good commit [messages](https://github.com/vmware-tanzu/tanzu-framework/blob/main/CONTRIBUTING.md)
- [x] Ensure PR contains terms all contributors can understand and links all contributors can access

PR #1281, opened 2021-12-02
Make group field optional in Capability GVR query

### LINKED ISSUE #882: Capability CRD's GVR query should allow empty API group
**Bug description**

Currently, GVR query in the `Capability` CRD has the `group` field as required with a minimum length of 1. This does not allow users to query for core k8s resources as they don't have an API group. The CRD needs to be changed to make this field optional.

- [x] APIs
- [ ] Addons
- [ ] CLI
- [ ] Docs
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**

GVR query allows a GVR check with no group specified.

Create Capability CRD to query for a core k8s resource such as `pods`:

```
apiVersion: run.tanzu.vmware.com/v1alpha1
kind: Capability
metadata:
  name: capability-sample
spec:
  queries:
    - name: "q1"
      groupVersionResources:
        - name: "fooGVR"
          group: ""
          versions:
            - v1
          resource: "pods"
```

The CRD's openAPI schema does not allow this resource to be created:

```
The Capability "capability-sample" is invalid: spec.queries.groupVersionResources.group: Invalid value: "": spec.queries.groupVersionResources.group in body should be at least 1 chars long
```

rajathagasthya:

Follow up to #883. This changes the actual CRD used during installation.

Fixes #882

Apply new CRD and create a Capability CR:

```
apiVersion: run.tanzu.vmware.com/v1alpha1
kind: Capability
metadata:
  name: capability-sample
spec:
  queries:
    - name: "q1"
      groupVersionResources:
        - name: "fooGVR"
          group: ""
          versions:
            - v1
          resource: "pods"
```

Verify CR is successfully created and controller reconciles result.

### Release note

```release-note

```

### PR Checklist

### DIFF
--- a/pkg/v1/providers/ytt/03_customizations/capabilities/add_capabilities.yaml
+++ b/pkg/v1/providers/ytt/03_customizations/capabilities/add_capabilities.yaml
@@ -60,7 +60,6 @@ spec:
                             group:
                               description: Group is the API group to check for in the
                                 cluster.
-                              minLength: 1
                               type: string
                             name:
                               description: Name is the unique name of the query.
@@ -79,7 +78,6 @@ spec:
                                 type: string
                               type: array
                           required:
-                            - group
                             - name
                           type: object
                         type: array


yharish991 (approved):
lgtm, thanks for taking care of this :)

pkg/v1/providers/ytt/03_customizations/capabilities/add_capabilities.yaml
```diff
@@ -60,7 +60,6 @@ spec:
                             group:
                               description: Group is the API group to check for in the
                                 cluster.
-                              minLength: 1
```
jmoroski, line 63: Is this correct? I think required and minLength validate 2 different things. This YAML used to say that the group field MUST be present and it MUST contain at least 1 character.

I just want to make sure we intend to say "group is no longer required and if it is present it can be empty", compared to "group is no longer required, but if you specify one it still has a min length"

jmoroski: Or I guess there's a 3rd option, "group is a required field but it can be empty". This seems closest to what the original issue asked for.

rajathagasthya: I played around with `minLength` and `omitempty` and I still think the current code makes sense.

If we enforce `minLength=1`, it doesn't allow authors to explicitly set `group: ""` to refer to k8s core API group and they have to rely on omitting the field itself as way to set use core group. I think that's confusing; I'd expect both the absence of the field and setting it to empty string to be the same.

With the current code, the behavior is pretty straightforward:

1. If `group` field is not present or is set to `""`, it refers to core group.
2. If `group` field is present and is set to ay value, it refers to a different API group.

I think length validation is not necessary at all. 

Let me know what you think!

rajathagasthya:
This was actually inadvertently merged when capabilities CRD was vendir-ed in #2257

PR #1283, opened 2021-12-03
Fix upgrade exits prematurely when upgrade does not involve a change in Kubernetes version

### LINKED ISSUE #1282: Control plane state polling during upgrade exits prematurely when upgrade does not involve a change in Kubernetes version.
**Bug description**

https://github.com/vmware-tanzu/tanzu-framework/blob/e2263794b578f3e596f3f008ea967c8aefc4cbad/pkg/v1/tkg/clusterclient/clusterclient.go#L640

```
[2021-12-02T19:48:57.239Z] Waiting for kubernetes version to be updated for control plane nodes
   (no wait at all)
[2021-12-02T19:48:57.239Z] Upgrading worker nodes...
...
```

The wait loop to check if the control plane returns to a ready state after a rollout has a race that cause it to exit immediately when there is no kubernetes version change in the upgrade operation because the first iteration of the wait loop could be seeing the preexisting ready state of the control plane prior to it being upgraded.

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [x] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**

There is proper polling that the control plane of the cluster transitions to updating then eventually back to ready.

upgrade cluster from a TKR version to another where there is no change in the Kubernetes version.
(v1.21.2+vmware.1-tkg.1 -> v1.21.2+vmware.1-tkg.2-rc.2)

0.12.0

vSphere, but potentially any others as well.

anujc25:
Wait for 60 seconds before we start the polling to check the upgrade status
This is done to account for the CP/MD upgrade to start rolling out
because it was noticed that CP/MD upgrade rollout take few seconds
to update the conditions which we rely on to check the upgrade status
ClusterReady is already present and wait logic to come out of the poll
loop assuming the upgrade is complete whereas the upgrade has not been
started yet

Fixes #1282

### Release note

```release-note
Fixed upgrade exits prematurely when upgrade does not involve a change in Kubernetes version
```

### PR Checklist

### DIFF
--- a/pkg/v1/tkg/clusterclient/clusterclient.go
+++ b/pkg/v1/tkg/clusterclient/clusterclient.go
@@ -78,7 +78,13 @@ const (
 	kubectlApplyRetryTimeout  = 30 * time.Second
 	kubectlApplyRetryInterval = 5 * time.Second
 	// DefaultKappControllerHostPort is the default kapp-controller port for it's extension apiserver
-	DefaultKappControllerHostPort = 10100
+	DefaultKappControllerHostPort           = 10100
+	waitPeriodBeforePollingForUpgradeStatus = 60 * time.Second
+)
+
+var (
+	// Sleep implements time.Sleep and used indirectly to mock for unit tests
+	Sleep = time.Sleep
 )
 
 // Client provides various aspects of interaction with a Kubernetes cluster provisioned by TKG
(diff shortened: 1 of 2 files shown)


pkg/v1/tkg/clusterclient/clusterclient.go
```diff
@@ -860,6 +860,15 @@ func (c *client) waitK8sVersionUpdateGeneric(clusterName, namespace, newK8sVersi
...
 	}
 
+	// Wait for 90 seconds before we start the polling to check upgrade status
+	// This is done to account for the CP/MD upgrade to start rolling out
+	// because it was noticed that CP/MD upgrade rollout take few seconds
+	// to update the conditions which we rely on to check the upgrade status
+	// ClusterReady is already present and wait logic comes out of the poll
+	// loop assuming the upgrade is complete whereas the upgrade has not been
+	// started yet
+	time.Sleep(90 * time.Second)
```
vuil, line 870: good to define a constant for the wait duration. 

This means we are applying a wait before checking on both the control plane and worker nodes. But given the current state is pretty harmless to do.

vuil (approved):
Thanks for the quick turnaround.
LGTM

alfredthenarwhal:
> @vuil:  /test upgrade-vc7
Commit: feedf95

Tests failed! Build no: 1151

vuil (changes requested):
Nit: The commit subject should be updated.
Also, I noticed there are some clusterclient unit tests that need to be updated.

alfredthenarwhal:
> @anujc25:  /test upgrade-vc7
Commit: 8c28862

Tests failed! Build no: 1152

vuil (approved):

PR #1297, opened 2021-12-06
Initial Mixed Windows Linux Workload Cluster Support

### LINKED ISSUE #1167: Add Support for mixed windows and linux clusters

Add the ability to configure a cluster with both windows and linux machine deployments for vSphere.

Writing a custom YTT overlay and add custom YTT values to achieve this.

- [ ] APIs
- [ ] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Additional context**
adding @jayunit100

vrabbi:

Add support for vSphere Workload Clusters to have a windows machine deployment and a linux machine deployment.

Fixes #1167 

Created and Ran Unit Tests

### Release note

```release-note
Added Support for workload clusters on vSphere to have both windows and linux worker nodes.
```

### PR Checklist


vrabbi:
@knabben ive rebased the PR and fixed the merge conflicts.

timothysc:
FWIW I view mix-mode clusters as an anti-pattern given how trivial we've made it to manage individual clusters w/CAPI.  Mixed-mode clusters come with all sorts of issues that are obviated when you tailor your cluster to be purpose-built. 

All that said, I think it's fine for folks to want to create them on their own for community purposes, but I'm wondering if this change will be obviated by the ClusterClass work...? @vuil @vincepri

vincepri:
There isn't anything technically stopping users to create mixed linux/windows clusters today from a Cluster API perspective. Definitely echoing @timothysc's comment above that managing mixed workload clusters can be seen as an anti-pattern especially given that the configuration of these workload clusters can vary wildly between operating systems. 

ClusterClass makes it even easier to create clusters without the need of all the ytt templating, you can have classes of clusters (windows, linux, etc) and manage the inventory in an easier way.

In general though, we shouldn't be mixing configuration and this might be something we can also discuss in Cluster API to avoid users issues down the line.

timothysc:
@vrabbi Thank you for the PR but with the impending shift to cluster class you will be able to do this server-side, with your own ClusterClass so, therefore, I'm going to close this one.

vrabbi:
With cluster class it is indeed very easy to do so glad that cluster class is coming. I understand the idea that mixed clusters may be an anti pattern in some cases but i think it has some use cases we see by customers that are very understandable. For example multiple customers of ours are migrating now both to k8s and also from dotnet to dotnet core. As part of that they want to first migrate to k8s, and then slowly microservice by microservice move to dotnet core on linux. This is a common pattern we see. In these cases having a single cluster with windows amd linux is huge as an application can be partially migrated and partially still dotnet on windows and with full transparency workloads can be migrated to linux nodes.

PR #1302, opened 2021-12-06
Fix panic on creating nodepools

tenczar:
This change fixes a panic that occurs when trying to create a node pool
in an environment with multiple files applied to the
kubeadmconfigtemplate.

Fixes #1302 

Created workload clusters in Azure and vSphere (with AVI). Confirmed that the azure secret is applied correctly in Azure environment. Confirmed that node-pool set does not panic on vpshere with AVI.

### Release note

```release-note
Fixed a panic in node-pool set when extra files are applied to kubeadmconfigtemplates
```

### PR Checklist

### DIFF
--- a/pkg/v1/tkg/client/machine_deployment.go
+++ b/pkg/v1/tkg/client/machine_deployment.go
@@ -172,9 +172,7 @@ func DoSetMachineDeployment(clusterClient clusterclient.Client, options *SetMach
 		options.Name = fmt.Sprintf("%s-%s", options.ClusterName, options.Name)
 		machineTemplateName := fmt.Sprintf("%s-mt", options.Name)
 		kcTemplate.Name = fmt.Sprintf("%s-kct", options.Name)
-		if kcTemplate.Spec.Template.Spec.Files != nil && len(kcTemplate.Spec.Template.Spec.Files) > 0 {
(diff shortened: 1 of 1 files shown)


saji-pivotal (approved):

vuil (commented):
Can we update the DoSetMachineDeployment unit test to account for this?

pkg/v1/tkg/client/machine_deployment.go
```diff
@@ -172,9 +172,7 @@ func DoSetMachineDeployment(clusterClient clusterclient.Client, options *SetMach
 		options.Name = fmt.Sprintf("%s-%s", options.ClusterName, options.Name)
 		machineTemplateName := fmt.Sprintf("%s-mt", options.Name)
 		kcTemplate.Name = fmt.Sprintf("%s-kct", options.Name)
-		if kcTemplate.Spec.Template.Spec.Files != nil && len(kcTemplate.Spec.Template.Spec.Files) > 0 {
```
vuil, line 175: Preexisting: should such manipulation logic be done after checking for the right InfrastructureRef.Kind?

vuil (approved):
LGTM. But if the review suggestions make sense, please see about a followup commit.

PR #1308, opened 2021-12-07
Add retry with polling when patching Cluster resource after successfu…

### LINKED ISSUE #1311: Upgrade fails intermittently while patching cluster resource if webhook controller is not running
**Bug description**
In some of the test pipelines, we are observing that patching the Cluster object just after a successful upgrade is causing the issue and throws an error `Internal error occurred: failed calling webhook "default.cluster.cluster.x-k8s.io": Post "https://capi-webhook-service.capi-webhook-system.svc:443/mutate-cluster-x-k8s-io-v1alpha3-cluster?timeout=30s": dial tcp 100.64.234.120:443: connect: connection refused`
This is possible because the pods like the webhook controller are not ready yet and CLI only tries to patch the resource only once. 

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**
The upgrade should be successful.

- Run `tanzu management-cluster upgrade`

anujc25:
…l upgrade

- In some of the test pipelines, we are observing that patching the `Cluster` object just after a successful upgrade is causing the issue and throws an error `Internal error occurred: failed calling webhook "default.cluster.cluster.x-k8s.io": Post "https://capi-webhook-service.capi-webhook-system.svc:443/mutate-cluster-x-k8s-io-v1alpha3-cluster?timeout=30s": dial tcp 100.64.234.120:443: connect: connection refused`
- This is possible because the pods like the webhook controller are not ready yet and CLI only tries to patch the resource only once. This change uses polling to patch the resource multiple times before throwing error.

Fixes #1311

### Release note

```release-note
Fix cluster patch failing after successful k8s nodes upgrade for management-cluster
```

### PR Checklist

### DIFF
--- a/pkg/v1/tkg/client/upgrade_cluster.go
+++ b/pkg/v1/tkg/client/upgrade_cluster.go
@@ -328,7 +328,8 @@ func (c *TkgClient) addKubernetesReleaseLabel(regionalClusterClient clusterclien
 		}
 	}`
 	patchLabel := fmt.Sprintf(patchFormat, utils.GetTkrNameFromTkrVersion(options.TkrVersion))
-	err := regionalClusterClient.PatchClusterObject(options.ClusterName, options.Namespace, patchLabel)
+	pollOptions := clusterclient.PollOptions{Interval: upgradePatchInterval, Timeout: upgradePatchTimeout}
+	err := regionalClusterClient.PatchClusterObjectWithPollOptions(options.ClusterName, options.Namespace, patchLabel, &pollOptions)
 	if err != nil {
 		return errors.Wrap(err, "unable to patch the cluster object with TanzuKubernetesRelease label")
 	}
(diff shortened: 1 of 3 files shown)


vuil:
LGTM, thanks! Please add a release note entry to the PR description as well.

pkg/v1/tkg/client/upgrade_cluster.go
```diff
@@ -328,7 +328,8 @@ func (c *TkgClient) addKubernetesReleaseLabel(regionalClusterClient clusterclien
 		}
 	}`
 	patchLabel := fmt.Sprintf(patchFormat, utils.GetTkrNameFromTkrVersion(options.TkrVersion))
-	err := regionalClusterClient.PatchClusterObject(options.ClusterName, options.Namespace, patchLabel)
+	pollOptions := clusterclient.PollOptions{Interval: 30 * time.Second, Timeout: 5 * time.Minute}
```
vuil, line 331: upgradePatchInterval, upgradePatchTimeout?

vuil (approved):
lgtm.

PR #1330, opened 2021-12-09
Add network separation feature gate

iXinqi:

We have added five new variables for avi vip separation feature:
"AVI_MANAGEMENT_CLUSTER_SERVICE_ENGINE_GROUP"
"AVI_CONTROL_PLANE_NETWORK"
"AVI_CONTROL_PLANE_NETWORK_CIDR"
"AVI_MANAGEMENT_CLUSTER_CONTROL_PLANE_VIP_NETWORK_NAME"
"AVI_MANAGEMENT_CLUSTER_CONTROL_PLANE_VIP_NETWORK_CIDR"

These variables default to "" and warmings are added for users
to fill in this variables and enable avi vip separation functionality while it is still under development.

Signed-off-by: Xinqi Li <lxinqi@vmware.com>

### PR Checklist

### DIFF
--- a/pkg/v1/config/clientconfig.go
+++ b/pkg/v1/config/clientconfig.go
@@ -37,6 +37,11 @@ const (
 	// when creating a cluster.
 	FeatureFlagManagementClusterCustomNameservers = "features.management-cluster.custom-nameservers"
 	FeatureFlagClusterCustomNameservers           = "features.cluster.custom-nameservers"
+	// Network Separation feature flags determine whether it is permitted to
+	// provide the AVI_MANAGEMENT_CLUSTER_SERVICE_ENGINE_GROUP, AVI_CONTROL_PLANE_NETWORK, AVI_CONTROL_PLANE_NETWORK_CIDR,
+	// AVI_MANAGEMENT_CLUSTER_CONTROL_PLANE_VIP_NETWORK_NAME and AVI_MANAGEMENT_CLUSTER_CONTROL_PLANE_VIP_NETWORK_CIDR
+	// when creating a cluster.
+	FeatureFlagManagementClusterNetworkSeparation = "features.management-cluster.network-separation-beta"
 )
 
 // DefaultCliFeatureFlags is used to populate an initially empty config file with default values for feature flags.
@@ -63,6 +68,7 @@ var (
 		FeatureFlagClusterDualStackIPv6Primary:                false,
 		FeatureFlagManagementClusterCustomNameservers:         false,
 		FeatureFlagClusterCustomNameservers:                   false,
+		FeatureFlagManagementClusterNetworkSeparation:         false,
 	}
 )
 
--- a/pkg/v1/tkg/client/validate.go
+++ b/pkg/v1/tkg/client/validate.go
@@ -586,6 +586,10 @@ func (c *TkgClient) ConfigureAndValidateManagementClusterConfiguration(options *
 		return NewValidationError(ValidationErrorCode, err.Error())
 	}
 
+	if err = c.ConfigureAndValidateNetworkSeparationConfiguration(TkgLabelClusterRoleManagement); err != nil {
+		return NewValidationError(ValidationErrorCode, err.Error())
+	}
+
 	isProdPlan := options.Plan == constants.PlanProd
 	_, workerMachineCount := c.getMachineCountForMC(options.Plan)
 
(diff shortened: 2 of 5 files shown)


pkg/v1/config/clientconfig.go
```diff
@@ -37,6 +37,10 @@ const (
 	// when creating a cluster.
 	FeatureFlagManagementClusterCustomNameservers = "features.management-cluster.custom-nameservers"
 	FeatureFlagClusterCustomNameservers           = "features.cluster.custom-nameservers"
+	// Custom Nameserver feature flags determine whether it is permitted to
```
christianang, line 40: Copy paste error? I don't think you want this comment specifically as it is specific to custom nameservers.

iXinqi: Good catch! It's fixed now.

christianang:
Looks good to me. I noticed one code comment that looked like it shouldn't be there and left a comment.

Also wondering if any checks need to be done for workload clusters or is the feature management-cluster only? Maybe another way to ask is if those environment variables are set on a workload cluster would it do anything? If not it looks good to me.

christianang (approved):

pkg/v1/tkg/client/validate.go
```diff
@@ -1767,6 +1771,38 @@ func getDockerBridgeNetworkCidr() (string, error) {
...
+	return nil
+}
+
+func (c *TkgClient) validateNetworkSeparation(networkSeparationConfigVariable, clusterRole string) error {
+	// if not set Get will return an empty string
+	configValue, err := c.TKGConfigReaderWriter().Get(networkSeparationConfigVariable)
+	if err != nil {
+		return nil
+	}
+	if clusterRole == TkgLabelClusterRoleManagement && !c.IsFeatureActivated(config.FeatureFlagManagementClusterNetworkSeparation) {
```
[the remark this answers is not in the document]
iXinqi: Nothing happened. I tried it out today and the workload cluster was created successfully. These variables are only used in management cluster. https://github.com/vmware-tanzu/tanzu-framework/blob/main/pkg/v1/providers/ytt/02_addons/avi/ako_operator_secret.yaml#L67

iXinqi:
> Looks good to me. I noticed one code comment that looked like it shouldn't be there and left a comment.
> 
> Also wondering if any checks need to be done for workload clusters or is the feature management-cluster only? Maybe another way to ask is if those environment variables are set on a workload cluster would it do anything? If not it looks good to me.

Thanks for review! Yes, it's management-cluster only and not related to workload cluster.

pkg/v1/tkg/client/validate.go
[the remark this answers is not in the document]
lubronzhan: Ok good

lubronzhan (approved):
Thanks!

pkg/v1/config/clientconfig.go
```diff
@@ -37,6 +37,11 @@ const (
 	// when creating a cluster.
 	FeatureFlagManagementClusterCustomNameservers = "features.management-cluster.custom-nameservers"
 	FeatureFlagClusterCustomNameservers           = "features.cluster.custom-nameservers"
+	// Network Separation feature flags determine whether it is permitted to
+	// provide the AVI_MANAGEMENT_CLUSTER_SERVICE_ENGINE_GROUP, AVI_CONTROL_PLANE_NETWORK, AVI_CONTROL_PLANE_NETWORK_CIDR,
+	// AVI_MANAGEMENT_CLUSTER_CONTROL_PLANE_VIP_NETWORK_NAME and AVI_MANAGEMENT_CLUSTER_CONTROL_PLANE_VIP_NETWORK_CIDR
+	// when creating a cluster.
+	FeatureFlagManagementClusterNetworkSeparation = "features.management-cluster.network-separation"
```
anujc25, line 44: Should we add this flag with a beta flag notation? cc @swalner-vmware 

@iXinqi are we expecting to enable this feature to `true` by default sometime in the future when we have the support?

iXinqi: @anujc25 yes, this feature will be true by default in the future, after it's been fully tested.
What is beta flag notation?

swalner-vmware: @iXinqi , the idea of a "beta" flag is to avoid this scenario:
A user installs the CLI with your flag set to false (by default); let's use the example of `feature.cluster.foobar`. They never change the value. Later when they install a newer release where `feature.cluster.foobar` is true by default, we have no way of overriding the false setting in their config file, because it appears identical to the case where they intentionally set the file false (and we should NOT override the value in that case). (We have no installation script that would help avoid this problem, so we'd have to tell the user to run a specific command to override their previous value.)
To avoid this issue we recommend having two flags. The "beta" flag, e.g. `feature.cluster.foobar-beta` is installed during the beta period with a value of false. Later, when the feature is ready to be released with a default of value of true, the developer changes the definition of the flag to `feature.cluster.foobar`. This is a "new" flag in terms of the config file, so when the new software runs, the old `feature.cluster.foobar-beta` flag in the config file is totally ignored, and now we can set the default of `feature.cluster.foobar` to true (when we don't find it in the config file).

I hope the above is clear, but I'm happy to chat directly about it (if that would help).

iXinqi: @swalner-vmware Do you think we must add the beta flag right now? Since Tanzu RC4 is due sometime this week, is it possible to merge this pr and then we'll submit a new pr for the bata flag after Eretria RTM?

swalner-vmware: My suggestion is simply to change (in this PR):
`FeatureFlagManagementClusterNetworkSeparation = "features.management-cluster.network-separation"`
to 
`FeatureFlagManagementClusterNetworkSeparation = "features.management-cluster.network-separation-beta"`

swalner-vmware: If you have an installation script that touches this flag, or documentation of it, then those things would have to change as well. Otherwise, you should be fine.

iXinqi: Thanks @swalner-vmware, I updated the flag. May I have your review again?

anujc25 (approved):
Looks good.

Issue #1099, opened 2021-11-08
Uprev Kickstart UI Angular and Clarity versions


miclettej:
Upgrade Kickstart UI Angular library to v13.x and Clarity v5.x (latest

**Affected product area (please put an X in all that apply)**

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Additional context**

PR #1359, opened 2021-12-13
Kickstart UI: upgrade angular and clarity library versions

raymondz1:
upgrade angular to v12.2.13
upgrade clarity to v5.6.0

Fixes # https://github.com/vmware-tanzu/tanzu-framework/issues/1099

### PR Checklist


miclettej:
Great job on this Raymond. I will check out your branch to test it today, but this looks good, very few workarounds which is awesome.

swalner-vmware (approved):
LGTM.
My only thought is that it might be nice to have convenience wrapper functions that set the options (rather than having them sprinkled throughout the code), and then the wrapper function could have the comment that explains what these `onlySelf` and `emitEvent` options actually do. As it is, that explanation occurs only in one place, whereas the options appear in many places.

miclettej:
> 

I agree with this. Much easier to have a well documented wrapper function than if we were to try to paste the same doc/comment around every usage. 

We also need to get the pipelines running with node v16 to get them passing. I can push a commit to your branch.

pkg/v1/tkg/web/src/app/views/landing/aws-wizard/node-setting-step/node-setting-step.component.ts
```diff
@@ -162,18 +162,18 @@ export class NodeSettingStepComponent extends StepFormDirective implements OnIni
...
             }
         }
         this.setControlWithSavedValue(AwsField.NODESETTING_BASTION_HOST_ENABLED, true);
-        this.formGroup.addControl(
+        this.formGroupWithEmition.addControl(
             AwsField.NODESETTING_MACHINE_HEALTH_CHECKS_ENABLED,
             new FormControl(true, [])
         );
-        this.formGroup.addControl(
+        this.formGroupWithEmition.addControl(
```
miclettej, line 176: The class is named FormGroupWithEmission, but in all usage you have it spelled with a 't' as Emition. Please change to correct spelling of usages.

pkg/v1/tkg/web/src/app/views/landing/wizard/shared/utils/form-group-with-emission.ts
```diff
@@ -0,0 +1,15 @@
+import { AbstractControl, FormGroup } from "@angular/forms";
+
+export class FormGroupWithEmission {
+    formGroup: FormGroup;
+    constructor(formGroup: FormGroup) {
+        this.formGroup = formGroup;
+    }
+    addControl(name: string, control: AbstractControl) {
+        // Clarity 5 has a new logic to check statuschanges. if the panel status is changed from valid to
```
miclettej, line 9: Please change comment to something like:

Extends default FormGroup.addControl method to add form controls and set the emitEvent property to false by default. This may be used to bypass Clarity 5's status change logic, which triggers form validation to occur immediately after form controls are created. Setting emitEvent to false avoids the Clarity 'triggerAllFormControlValidation' method from being executed.

@swalner-vmware please feel free to add your input on this

pkg/v1/tkg/web/src/app/views/landing/aws-wizard/node-setting-step/node-setting-step.component.ts
raymondz1: fixed

swalner-vmware (approved):
Excellent!

miclettej (approved):

PR #1365, opened 2021-12-13
vSphere upgrade rely on moid instead of name

### LINKED ISSUE #1396: Tanzu upgrade isn't idempotent on vSphere when VM template is moved/renamed
**Bug description**
The upgrade operation will trigger a repave of a cluster in the unlikely event that it is called after the VM template for the current version has been moved or renamed. 

- [x] APIs
- [ ] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**
The upgrade operation should remain idempotent when the VM template underlying the cluster is moved/renamed.

Create a cluster on vSphere. Move/rename the VM template the cluster was just created from and trigger and upgrade. This will cause the cluster's nodes to be repaved. 

tenczar:
This change updates the upgrade logic so that a change to the name(path)
of the virtual machine template does not cause VSphereMachineTemplates
to be regenerated.

It is possible that the upgrade operation will not be idempotent in vSphere clusters when the name/path of the VM template has changed. This PR updates this logic so that the VM template can be renamed without causing a repaving of the cluster.

Fixes #1396 

Added unit tests to verify new templates are only created when the vm moid changes. Tested upgrade in vcenter after changing vm template name. Verified that new machine templates were not generated

### Release note

```release-note
Fixes a rare issue where cluster machines would be incorrectly rotated when performing an upgrade where the only change to the cluster was the name or path of the VM template.
```

### PR Checklist


vuil:
Thanks for the changes @tenczar . Please file an issue to description the problem and reference it in this PR. We will likely need to continue discussion there on the most appropriate way to address it.

pkg/v1/tkg/client/upgrade_cluster.go
```diff
@@ -947,7 +949,7 @@ func isNewVSphereTemplateRequired(machineTemplate *capvv1beta1.VSphereMachineTem
 		return true
 	}
 	// If vm template given is the same as we already have in VSphereMachineTemplate
-	if machineTemplate.Spec.Template.Spec.Template != clusterUpgradeConfig.UpgradeComponentInfo.VSphereVMTemplateName {
```
vuil, line 950: two things:
- (nit) the comment does not match the code any more
- what happens when upgrading existing clusters that to not have this annotation set?

tenczar: > what happens when upgrading existing clusters that to not have this annotation set?
In this case a user presumably intends for an upgrade. For example, existing 1.5 clusters don't have this annotation. When a user performs an upgrade using the 1.6 cli they should be expecting an upgrade. This might trip up some main branch users immediately after this gets added to the tree, but for end users this upgrade should be expected.

steven-zou:
@tenczar / @vuil 

When is this PR available?

anujc25 (approved):
Looks good

imikushin (approved):
LGTM

PR #1395, opened 2021-12-16
Update generated data files

stmcginnis:

There are a few files that are generated and have changes that have not
been merged in to the repo. This is causing problems with some PRs since
we have a `verify` target that is run that makes sure there aren't any
changed files after performing a build.

This includes changes related to newer versions of go, but does not
remove the compatibility for the official 1.16 version we declare. This
shouldn't cause any harm for now, and will make for less issues if/when
we decide to update to go 1.17 in the future.

Performed local build and ran `make verify`.

### Release note

```release-note
NONE
```


stmcginnis:
Just testing, others have a better handle on this now.

PR #1399, opened 2021-12-16
Run all lint checks as part of the lint Action

stmcginnis:

The current GitHub Action for linting only runs golangci-lint and the
doc-lint target on the matrix of local package paths. We have additional
linting checks that need to be run like the check for license headers.

The current state also requires providing the paths to each local module
in both the Makefile target and the Action definition.

To make this simpler and make sure all of our defined lint checks are
run, this updates the linting Action definition to just call our `make
lint` target so everything is controlled from the one place.

This also addresses a couple licence header issues that were identified
by running the full lint check.

Ran `make lint` locally and made sure output showed everything was being run and there were no issues.

### Release note

```release-note
NONE
```

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -83,26 +83,18 @@ jobs:
         make configure-bom
         make test
 
-  golangci-lint:
+  check:
     name: Lint
     runs-on: ubuntu-latest
-    strategy:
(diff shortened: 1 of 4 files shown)


vuil (approved):
lgtm, thanks for the cleanup!

PR #1401, opened 2021-12-16
Revert "Kickstart UI: fix naming of contextual help service"

miclettej:
Reverts vmware-tanzu/tanzu-framework#1387

May have caused issue with zz generated bindata


stmcginnis:
Failure in `Main / Build` doesn't appear to be related. I'll kick that off again.

Seems like there has to be a way to rename that while getting the tests to pass. But in the interest of getting tests to pass while we sort that out, I think this makes sense to revert it.

miclettej:
> Failure in `Main / Build` doesn't appear to be related. I'll kick that off again.
> 
> Seems like there has to be a way to rename that while getting the tests to pass. But in the interest of getting tests to pass while we sort that out, I think this makes sense to revert it.

Thanks Sean, yeah this is a bizarre situation. The commit I merged yesterday did not generate a new bindata file and it did not cause an error in its own PR, but now it has gotten the main build Verify step into a jam. This causes a bit of a chicken and egg situation if there is in fact a failing test in main build that needs a PR to fix it. That being said, I agree we should just revert this and clear up the bindata error that's causing Verify to fail.

anujc25 (approved):

PR #1409, opened 2021-12-17
add CAPD integration tests for telemetry

maralavi:
- This PR is to add CAPD integration tests for telemetry

Added integration tests as part of the PR

None

**Release note**:

```release-note
added CAPD integration tests for telemetry
```

**New PR Checklist**

### DIFF
--- a/pkg/v1/tkg/test/tkgctl/docker/docker_ceip_test.go
+++ b/pkg/v1/tkg/test/tkgctl/docker/docker_ceip_test.go
@@ -1,0 +1,23 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// nolint:typecheck,nolintlint
+package docker
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo"
+
+	. "github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkg/test/tkgctl/shared"
+)
+
+var _ = Describe("CEIP tests for capd clusters", func() {
+	E2ECEIPSpec(context.TODO(), func() E2ECEIPSpecInput {
+		return E2ECEIPSpecInput{
+			E2EConfig:       e2eConfig,
+			ArtifactsFolder: artifactsFolder,
+			Cni:             "antrea",
+		}
+	})
+})
--- a/pkg/v1/tkg/test/tkgctl/docker/docker_ceip_test.go
+++ b/pkg/v1/tkg/test/tkgctl/docker/docker_ceip_test.go
@@ -1,7 +1,6 @@
 // Copyright 2021 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-// nolint:typecheck,nolintlint
 package docker
 
 import (
--- a/pkg/v1/tkg/test/tkgctl/shared/ceip.go
+++ b/pkg/v1/tkg/test/tkgctl/shared/ceip.go
@@ -13,14 +13,25 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/api/batch/v1beta1"
-	"k8s.io/apimachinery/pkg/types"
-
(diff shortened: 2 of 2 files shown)


pkg/v1/tkg/test/tkgctl/shared/ceip.go
```diff
@@ -118,3 +133,83 @@ func verifyTelemetryJobURL(context context.Context, url string, mcProxy *framewo
...
+		return true, nil
+	}); err != nil {
+		return err
+	}
+	if len(pods.Items) == 0 {
+		return errors.New("no telemetry pod is running")
+	}
+
+	// check to make sure that the telemetry pod does not have "Failed" status
+	if pods.Items[0].Status.Phase == "Failed" {
```
vijaykatam, line 200: I would check if the pod phase is running and this should use a wait with poll or preferably gomega's `Eventually` assertion.

pkg/v1/tkg/test/tkgctl/docker/docker_ceip_test.go
```diff
@@ -0,0 +1,23 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// nolint:typecheck,nolintlint
```
vijaykatam, line 4: (question): why do we need these disabled?

pkg/v1/tkg/test/tkgctl/shared/ceip.go
danniel1205, line 200: question: do we want to check the status for all pods ?

pkg/v1/tkg/test/tkgctl/shared/ceip.go
```diff
@@ -118,3 +133,83 @@ func verifyTelemetryJobURL(context context.Context, url string, mcProxy *framewo
...
+
+	scheme := mcProxy.GetScheme()
+	batchv1.AddToScheme(scheme)
+
+	cronJob := &v1beta1.CronJob{}
+	if err = client.Get(context, types.NamespacedName{Name: telemetryName, Namespace: telemetryNamespace}, cronJob); err != nil {
+		return err
+	}
+	// updating the telemetry cron job schedule to "* * * * *" so that the cronjob can be scheduled to run within the next 59 seconds
+	cronJob.Spec.Schedule = "* * * * *"
```
danniel1205, line 154: Suggestion: please deepCopy and make changes to the copied object instead. It is risky to change the pointer directly. The same to L205 as well.

pkg/v1/tkg/test/tkgctl/docker/docker_ceip_test.go
danniel1205: had the same question.

maralavi: I just added this file similar to other test files under tkgctl/docker added by the CLI team. I'll check what would happen with these removed and will let you know in a bit

pkg/v1/tkg/test/tkgctl/shared/ceip.go
maralavi: @danniel1205 The cluster and cronjob is created just then, and so the job is created just then and there is only one pod running, as I return the schedule to create telemetry job to once every six hours right after the test.

pkg/v1/tkg/test/tkgctl/shared/ceip.go
maralavi: Thanks for this. Will fix

pkg/v1/tkg/test/tkgctl/shared/ceip.go
maralavi: Sure, will change to only allow either of  'running' or 'succeeded' statuses

vijaykatam (approved):
LGTM

danniel1205 (approved):

PR #1412, opened 2021-12-20
Handle `context-aware-cli-for-plugin` feature flag with `tanzu init` …

### LINKED ISSUE #1411: `tanzu init` and `tanzu update` command doesn't respect `context-aware-cli-for-plugins` feature flag
**Bug description**

`tanzu init` and `tanzu update` command doesn't respect `context-aware-cli-for-plugins` feature flag. 

This allows unpredictable behavior of tanzu cli because running this command tries to install old plugins sometimes with `tanzu init` from GCP bucket depending on the `repositories` configuration under config file.

Running `tanzu update` tries to download tanzu-cli from the GCP bucket.

As when `context-aware-cli-for-plugins` feature flag is enabled, tanzu cli doesn't rely on GCP bucket or the `repositories` configuration under config.yaml file, it should not download and install plugins from GCP bucket. 

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**

- `tanzu init` will sync/update all available plugins for the user based on `context-aware-cli-for-plugin` feature
- `tanzu update` should not rely on the GCP bucket when `context-aware-cli-for-plugin` feature flag is enabled

- Run below commands:
- `tanzu init`
- `tanzu update`

v0.14.0

anujc25:
…and `tanzu update` commands

- `tanzu init` will sync/update all available plugins for the user based on `context-aware-cli-for-plugin` feature
- `tanzu update` command is not supported when `context-aware-cli-for-plugin` feature flag is enabled. This is to avoid relying on the GCP bucket to download the core and plugin binaries as we don't want to rely on the GCP bucket when `context-aware-cli-for-plugins` feature flag is enabled.
- 

Fixes #1411

### Release note

```release-note
Fix `tanzu init` and `tanzu update` command not respecting `context-aware-cli-for-plugins` feature flag
```

### PR Checklist

### DIFF
--- a/pkg/v1/cli/command/core/update.go
+++ b/pkg/v1/cli/command/core/update.go
@@ -4,6 +4,7 @@
 package core
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
@@ -14,6 +15,7 @@ import (
 	cliv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/cli/v1alpha1"
 	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/buildinfo"
 	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli"
+	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/config"
 )
 
 var yesUpdate bool
@@ -31,6 +33,10 @@ var updateCmd = &cobra.Command{
 		"group": string(cliv1alpha1.SystemCmdGroup),
 	},
 	RunE: func(cmd *cobra.Command, args []string) error {
+		if config.IsFeatureActivated(config.FeatureContextAwareCLIForPlugins) {
+			return errors.New("CLI self-update is currently not supported. For updating plugins please use the `tanzu plugin sync` command")
+		}
+
 		// clean the catalog cache when updating the cli
 		if err := cli.CleanCatalogCache(); err != nil {
 			log.Debugf("Failed to clean the Plugin descriptors cache %v", err)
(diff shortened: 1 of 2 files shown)


pkg/v1/cli/command/core/update.go
```diff
@@ -31,6 +32,12 @@ var updateCmd = &cobra.Command{
 		"group": string(cliv1alpha1.SystemCmdGroup),
 	},
 	RunE: func(cmd *cobra.Command, args []string) error {
+		if config.IsFeatureActivated(config.FeatureContextAwareCLIForPlugins) {
+			// TODO: Implement inline update of tanzu CLI when `context-aware-cli-for-plugins` feature flag is enabled
+			log.Warning("Updating CLI and plugins is not supported when `context-aware-cli-for-plugins` feature flag is enabled")
```
vuil, line 37: The log message may not make sense to users especially now that the feature is activated by default.
Maybe something like:

log.Warning("CLI self-update is currently not support. For updating plugins please use the `plugin sync` command")

pkg/v1/cli/command/core/update.go
```diff
@@ -31,6 +32,11 @@ var updateCmd = &cobra.Command{
 		"group": string(cliv1alpha1.SystemCmdGroup),
 	},
 	RunE: func(cmd *cobra.Command, args []string) error {
+		if config.IsFeatureActivated(config.FeatureContextAwareCLIForPlugins) {
+			log.Warning("CLI self-update is currently not support. For updating plugins please use the `tanzu plugin sync` command")
```
vuil, line 36: nit: support -> supported.
also, should it return an error instead of nil?

vuil (approved):
lgtm, thanks!

alfredthenarwhal:
> @vuil:  /test install-vc7
Commit: a748ef5

Tests passed! Build no: 1264

PR #1415, opened 2021-12-21
DO NOT MERGE - troubleshooting bindata file issue

swalner-vmware:
Regenerates the bindata file
### PR Checklist


stmcginnis:
Looks like the same failures @miclettej  was hitting. Is this from some sort of mismatch in package versions causing different bindata formats? Or stale data?

I'm wondering if we need to just disable this check for now to get things updated.

miclettej:
> Looks like the same failures @miclettej was hitting. Is this from some sort of mismatch in package versions causing different bindata formats? Or stale data?
> 
> I'm wondering if we need to just disable this check for now to get things updated.

Hi @stmcginnis yes this is becoming problematic. We have tried several different times to get this bindata file check passing in numerous PRs. What Shimon and I are noticing is that the CI build is generating compiled Angular webapp files which are named differently than the files generated locally when we do a make ui-build. Angular has a hashing strategy that appends a hash to filenames for cache busting. That hash only changes when there are actual changes in the UI code. This must mean that somehow the compiled Angular code in the CI main build is slightly different than when we build locally. We noticed that there is a slight difference in the number of npm modules that are included in the webapp when building locally vs in the CI main build. We have confirmed that we are all running the same Go, Node.js, and npm versions as the CI but still seeing this issue. It will take some time to debug and Shimon and I agree, we should disable the check for now. I'm not yet certain why this recently started happening, but something has changed making this issue consistent. Do you know if there were any changes to the Github runners?

stmcginnis:
> Do you know if there were any changes to the Github runners?

Not that I know of, but I suppose it could be possible.

PR #1273, opened 2021-12-01
Fix private registry testcases for package plugin integration test

blc1996:
Signed-off-by: Lucheng Bao <luchengb@vmware.com>

### What this PR does / why we need it

Use a in cluster private registry for package plugin authentication test

### Which issue(s) this PR fixes

Fixes #1141 

### Release note

```release-note
None
```

### PR Checklist

- [x] Squash the commits into one or a small number of logical commits
      
- [x] Use good commit [messages](https://github.com/vmware-tanzu/tanzu-framework/blob/main/CONTRIBUTING.md)
- [x] Ensure PR contains terms all contributors can understand and links all contributors can access

PR #1422, opened 2021-12-22
Add the ability to create defaults based on package schema

vijaykatam:
* Create generic openapischema helpers for defaults based on schema
* Package available get exposes new flag --generate-default-values-file to automatically create default values file

1. Allow package plugin users to generate default values using schema as opposed to looking at documentation
2. Generic helpers for defaulting based on schema to be used for addon APIs

1. Unit tests
2. Test with package plugin and packages
```
artifacts/darwin/amd64/cli/package/v0.15.0-dev/tanzu-package-darwin_amd64 available get contour.tanzu.vmware.com/1.18.2+vmware.1-tkg.1-rc.1 --generate-default-values-file
- Retrieving package details for contour.tanzu.vmware.com/1.18.2+vmware.1-tkg.1-rc.1... I1221 17:10:46.089849   58510 request.go:665] Waited for 1.034035038s due to client-side throttling, not priority and fairness, request: GET:https://10.182.8.147:6443/apis/cli.tanzu.vmware.com/v1alpha1?timeout=32s


NAME:                             contour.tanzu.vmware.com
VERSION:                          1.18.2+vmware.1-tkg.1-rc.1
RELEASED-AT:                      2021-10-04 17:00:00 -0700 PDT
DISPLAY-NAME:                     contour
SHORT-DESCRIPTION:                An ingress controller
PACKAGE-PROVIDER:                 VMware
MINIMUM-CAPACITY-REQUIREMENTS:    Varies significantly based on number of Services, Ingresses/HTTPProxies, etc. A starting point is 128MB RAM and 0.5 CPU for each Contour and Envoy pod, but this can and should be tuned based on observed usage.
LONG-DESCRIPTION:                 An Envoy-based ingress controller that supports dynamic configuration updates and multi-team ingress delegation. See https://projectcontour.io for more information.
MAINTAINERS:                      [{Steve Kriss} {Steve Sloka} {Nick Young} {Sunjay Bhatia} {Nicholas Seemiller}]
RELEASE-NOTES:                    contour 1.18.2 https://github.com/projectcontour/contour/releases/tag/v1.18.2
LICENSE:                          [VMware’s End User License Agreement (Underlying OSS license: Apache License 2.0)]
SUPPORT:                          Support provided by VMware for deployment on TKG 1.4+ clusters. Best-effort support for deployment on any conformant Kubernetes cluster. Contact support by opening a support request via VMware Cloud Services or my.vmware.com.
CATEGORY:                         [ingress]

Created default values file at /Users/vkatam/go/src/github.com/vmware-tanzu/tanzu-framework/contour-default-values.yaml
```

### Release note

```release-note
Add a flag --generate-default-values-file to command tanzu package available get
```

### PR Checklist

### DIFF
--- a/cmd/cli/plugin/package/README.md
+++ b/cmd/cli/plugin/package/README.md
@@ -56,7 +56,7 @@ Use "tanzu package repository [command] --help" for more information about a com
 
 1. Create a management cluster using latest tanzu cli
 
-1. Use package commands to:
+2. Use package commands to:
    * add a repository
    * list a repository
    * get a repository status
@@ -75,14 +75,14 @@ Use "tanzu package repository [command] --help" for more information about a com
 
    Here is an example workflow
 
-1. Add a repository
+3. Add a repository
 
    ```sh
    >>> tanzu package repository add standard-repo --url projects-stg.registry.vmware.com/tkg/test-packages/standard-repo:v1.0.0 -n test-ns --create-namespace
    Added package repository 'standard-repo'
    ```
 
-1. Get repository status
+4. Get repository status
 
    ```sh
    >>> tanzu package repository get standard-repo -n test-ns
@@ -93,14 +93,14 @@ Use "tanzu package repository [command] --help" for more information about a com
    REASON:
    ```
 
-1. Update a repository
+5. Update a repository
 
    ```sh
    >>> tanzu package repository update standard-repo --url projects-stg.registry.vmware.com/tkg/test-packages/standard-repo:v1.0.0 -n test-ns
    Updated package repository 'standard-repo' in namespace 'test-ns'
    ```
 
-1. List the repository
+6. List the repository
 
    ```sh
    >>> tanzu package repository list -A
@@ -109,7 +109,7 @@ Use "tanzu package repository [command] --help" for more information about a com
    repo           projects-stg.registry.vmware.com/tkg/test-packages/test-repo:v1.0.0      Reconcile succeeded           test-ns
    ```
 
-1. Get information of a package
+7. Get information of a package
 
    Example 1: Get detailed information of a package
 
@@ -146,7 +146,53 @@ Use "tanzu package repository [command] --help" for more information about a com
     namespace                   external-dns  string   The namespace in which to deploy ExternalDNS.
    ```
 
-1. Install a package
+   Example 3: Generate default values.yaml for a package
+
+   ```sh
+   >>> tanzu package available get contour.tanzu.vmware.com/1.18.2+vmware.1-tkg.1-rc.1 --generate-default-values-file
+    NAME:                             contour.tanzu.vmware.com
+    VERSION:                          1.18.2+vmware.1-tkg.1-rc.1
+    RELEASED-AT:                      2021-10-04 17:00:00 -0700 PDT
+    DISPLAY-NAME:                     contour
+    SHORT-DESCRIPTION:                An ingress controller
+    PACKAGE-PROVIDER:                 VMware
+    MINIMUM-CAPACITY-REQUIREMENTS:    Varies significantly based on number of Services, Ingresses/HTTPProxies, etc. A starting point is 128MB RAM and 0.5 CPU for each Contour and Envoy pod, but this can and should be tuned based on observed usage.
+    LONG-DESCRIPTION:                 An Envoy-based ingress controller that supports dynamic configuration updates and multi-team ingress delegation. See https://projectcontour.io for more information.
+    MAINTAINERS:                      [{Steve Kriss} {Steve Sloka} {Nick Young} {Sunjay Bhatia} {Nicholas Seemiller}]
+    RELEASE-NOTES:                    contour 1.18.2 https://github.com/projectcontour/contour/releases/tag/v1.18.2
+    LICENSE:                          [VMware’s End User License Agreement (Underlying OSS license: Apache License 2.0)]
+    SUPPORT:                          Support provided by VMware for deployment on TKG 1.4+ clusters. Best-effort support for deployment on any conformant Kubernetes cluster. Contact support by opening a support request via VMware Cloud Services or my.vmware.com.
+    CATEGORY:                         [ingress]
+
+    Created default values file at /home/contour-default-values.yaml
+   ```
+
+   contour-default-values.yaml
+
+   ```yaml
+    certificates:
+      duration: 8760h
+      renewBefore: 360h
+    contour:
+      logLevel: info
+      replicas: 2
+      useProxyProtocol: false
+    envoy:
+      hostNetwork: false
+      hostPorts:
+        enable: true
+        http: 80
+        https: 443
+      logLevel: info
+      service:
+        aws:
+          LBType: classic
+        externalTrafficPolicy: Cluster
+      terminationGracePeriodSeconds: 300
+    namespace: tanzu-system-ingress  
+   ```
+
+8. Install a package
 
    Example 1: Install the specified version for package name "fluent-bit.tkg-standard.tanzu.vmware", while providing the values.yaml file and without waiting for package reconciliation to complete
 
@@ -188,7 +234,7 @@ Use "tanzu package repository [command] --help" for more information about a com
      Added installed package 'contour-pkg' in namespace 'test-ns'
     ```
 
-1. Get information of an installed package
+9. Get information of an installed package
 
    Example 1: Get information of an installed package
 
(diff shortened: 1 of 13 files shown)


vrabbi:
How will this work for packages like grafana that dont have all default values in the schema as some are long yaml files and to get the defaults currently one needs to pull the imgpkg bundle and look at the values.yaml within that bundle?

vrabbi:
It would alsi be helpful if the description fjelds in the openapiv3 schema from the pkgi would be added as comments in the generated default values file to make the file self explaining and not require multiple commands to understand what a specific value is used for or what values are possible

cmd/cli/plugin/package/package_available_get.go
```diff
@@ -156,3 +169,50 @@ func getValuesSchemaForPackage(namespace, name, version string, kc kappclient.Cl
...
+
+	if len(pkg.Spec.ValuesSchema.OpenAPIv3.Raw) == 0 {
+		log.Warningf("package '%s/%s' does not have any user configurable values in the '%s' namespace", pkgName, pkgVersion, packageAvailableOp.Namespace)
+		return nil
+	}
+
+	fileNamePrefix := ""
+	pkgNameTokens := strings.Split(pkgName, ".")
+	if len(pkgNameTokens) >= 1 {
+		fileNamePrefix = fmt.Sprintf("%s-", pkgNameTokens[0])
```
blc1996, line 187: Nit pick: Is it better to just use `pkgNameTokens[0]` here and move the `-` to line 189?

cmd/cli/plugin/package/package_available_get.go
```diff
@@ -156,3 +169,50 @@ func getValuesSchemaForPackage(namespace, name, version string, kc kappclient.Cl
...
+	}
+	valuesFileName := fmt.Sprintf("%sdefault-values.yaml", fileNamePrefix)
+	var valuesFileToUse *os.File
+
+	if len(valuesFile) > 0 && valuesFile[0] != nil {
+		// caller is responsible for closing file
+		valuesFileToUse = valuesFile[0]
```
blc1996, line 194: Question: Why we need this `valuesFile` param, I don't see it being used. Is it just for testing purposes?

pkg/v1/openapischema/schema_default.go
```diff
@@ -0,0 +1,121 @@
...
+	if err := yamlEncoder.Close(); err != nil {
+		return nil, err
+	}
+
+	return b.Bytes(), nil
+}
+
+// schemaDefault does defaulting of x depending on default values in s.
+// This is adopted from k8s.io/apiextensions-apiserver/pkg/apiserver/schema/defaulting with 2 changes
+// 1. Prevent deep copy of int as it panics
```
blc1996, line 46: Is this a bug in upstream https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apiserver/schema/defaulting/algorithm.go ?

vijaykatam: Possibly but not sure.

cmd/cli/plugin/package/package_available_get.go
vijaykatam: Yeah valuesFile is not required. Used by unit test to assert output of the file since there is no easy way to mock os.File.

blc1996:
The error in package plugin test pipeline `Error: failed to check for the availability of 'packaging.carvel.dev' API: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable` is because init() functions a not executed as expected, which makes `kubeconfig` param not correctly populated.
The issue is addressed with this commit: https://github.com/vmware-tanzu/tanzu-framework/pull/1273/commits/118901cd50e2b3e44d481f14e3288f54daa49eb3

maralavi:
> The error in package plugin test pipeline `Error: failed to check for the availability of 'packaging.carvel.dev' API: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable` is because init() functions a not executed as expected order, which makes `kubeconfig` param not correctly populated. The issue is addressed with this commit: [118901c](https://github.com/vmware-tanzu/tanzu-framework/commit/118901cd50e2b3e44d481f14e3288f54daa49eb3)

@blc1996 This issue had not been surfaced before. Would it be possible if you fix this for secret plugin as well as part of your fix? it needs change in only one file. Thanks!

vijaykatam:
> It would alsi be helpful if the description fjelds in the openapiv3 schema from the pkgi would be added as comments in the generated default values file to make the file self explaining and not require multiple commands to understand what a specific value is used for or what values are possible

Thanks for feedback @vrabbi, I experimented with [comment support](https://pkg.go.dev/gopkg.in/yaml.v3?utm_source=godoc#Node.HeadComment) in yaml v3 and ran into [this issue](https://github.com/go-yaml/yaml/issues/610) where the comment for a property is added below the key rather than above. 

For example contour-schema 
```
certificates:
  duration: 8760h
  # If using cert-manager, how long the certificates should be valid for. If useCertManager is false, this field is ignored.
  renewBefore: 360h
  # If using cert-manager, how long before expiration the certificates should be renewed. If useCertManager is false, this field is ignored.
  useCertManager: false
  # Whether to use cert-manager to provision TLS certificates for securing communication between Contour and Envoy. If false, the upstream Contour certgen job will be used to provision certificates. If true, the cert-manager addon must be installed in the cluster.
contour:
  logLevel: info
  # The Contour log level. Valid options are info and debug.
  replicas: 2
  # How many Contour pod replicas to have.
  useProxyProtocol: false
  # Whether to enable PROXY protocol for all Envoy listeners.
envoy:
  hostNetwork: false
  # Whether to enable host networking for the Envoy pods.
  hostPorts:
    enable: false
    # Whether to enable host ports. If false, http and https are ignored.
    http: 80
    # If enable == true, the host port number to expose Envoy's HTTP listener on.
    https: 443
    # If enable == true, the host port number to expose Envoy's HTTPS listener on.
  logLevel: info
  # The Envoy log level.
  service:
    externalTrafficPolicy: Local
    # The external traffic policy for the Envoy service.
    type: LoadBalancer
    # The type of Kubernetes service to provision for Envoy.
  terminationGracePeriodSeconds: 300
  # The termination grace period, in seconds, for the Envoy pods.
namespace: projectcontour
# The namespace in which to deploy Contour and Envoy.
```

vijaykatam:
@vrabbi - I found adding the schema description as an inline comment quite brittle and requires explicitly assembling the a yaml Doc with individual yaml.Node(yaml.MappingNode, yaml.SequenceNode and yaml.ScalarNode) elements rather than a map type for it to work reliably so that objects also get inline comments(not just leaf elements). Additionally due to the bug in yaml v3 lib the comment is added below the element. I propose comments be a separate feature.

> How will this work for packages like grafana that dont have all default values in the schema as some are long yaml files and to get the defaults currently one needs to pull the imgpkg bundle and look at the values.yaml within that bundle?

I added an example test case for grafana that shows the default values. The long yaml files(for config) aren't included because the current schema doesn't default them so they will show up as empty. We need to update grafana package so that yaml block is defined as default.

cmd/cli/plugin/package/README.md
```diff
@@ -146,7 +146,53 @@ Use "tanzu package repository [command] --help" for more information about a com
...
+    MINIMUM-CAPACITY-REQUIREMENTS:    Varies significantly based on number of Services, Ingresses/HTTPProxies, etc. A starting point is 128MB RAM and 0.5 CPU for each Contour and Envoy pod, but this can and should be tuned based on observed usage.
+    LONG-DESCRIPTION:                 An Envoy-based ingress controller that supports dynamic configuration updates and multi-team ingress delegation. See https://projectcontour.io for more information.
+    MAINTAINERS:                      [{Steve Kriss} {Steve Sloka} {Nick Young} {Sunjay Bhatia} {Nicholas Seemiller}]
+    RELEASE-NOTES:                    contour 1.18.2 https://github.com/projectcontour/contour/releases/tag/v1.18.2
+    LICENSE:                          [VMware’s End User License Agreement (Underlying OSS license: Apache License 2.0)]
+    SUPPORT:                          Support provided by VMware for deployment on TKG 1.4+ clusters. Best-effort support for deployment on any conformant Kubernetes cluster. Contact support by opening a support request via VMware Cloud Services or my.vmware.com.
+    CATEGORY:                         [ingress]
+
+    Created default values file at /home/contour-default-values.yaml
```
blc1996, line 167: The `/home` here probably means current directory?

blc1996 (approved):
LGTM

cmd/cli/plugin/package/README.md
vijaykatam: yes, it prints out absolute path.

cmd/cli/plugin/package/package_available_get.go
```diff
@@ -156,3 +169,50 @@ func getValuesSchemaForPackage(namespace, name, version string, kc kappclient.Cl
...
 	return nil
 }
+
+func generateDefaultValuesForPackage(pkg *kapppkg.Package, valuesFile ...*os.File) error {
+	if pkg == nil {
+		// should not happen
+		return errors.New("pkg is nil")
+	}
+
+	if len(pkg.Spec.ValuesSchema.OpenAPIv3.Raw) == 0 {
```
danniel1205, line 179: question: will there be an nil exception when `ValueSchema` does not exist in `Spec` ?

vijaykatam: If `ValuesSchema` is empty when creating a package CR it gets defaulted to the following and `len(pkg.Spec.ValuesSchema.OpenAPIv3.Raw) == 0` is true. Verified with a dummy package.
```
  valuesSchema:
    openAPIv3: null
```

danniel1205 (approved):

PR #1439, opened 2022-01-01
Fix file count error returned by OCIArtifact.Fetch

mayankbh:

When the OCI bundle for a CLI plugin contains several files, the OCIArtifact.Fethc()
method bails out (as it only expects one file).

Since errors.Wrapf returns nil if the err passed in was nil, this would
manifest as a (nil []byte, nil error) return from the Fetch() function,
and the plugin installation would end up writing an empty file to disk,
finally leading to a failed plugin installation when the 'info'
subcommand was executed, but with no obvious error message about why.

Omitted, I considered this a trivial fix.

Added a unit test for this case.

`make test` succeeded.

### Release note

N/A

### PR Checklist

### Additional information

To add tests, I extended the OCIArtifact struct to allow injecting the Carvel helper
function to construct the file map.

N/A

### DIFF
--- a/pkg/v1/cli/artifact/oci.go
+++ b/pkg/v1/cli/artifact/oci.go
@@ -1,9 +1,10 @@
-// Copyright 2021 VMware, Inc. All Rights Reserved.
+// Copyright 2021-2022 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package artifact
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


anujc25:
Thanks for the fix @mayankbh. Change looks good. but can you please check whether you have signed the contribution license or not? 
More details https://github.com/vmware-tanzu/tanzu-framework/blob/main/CONTRIBUTING.md#contributor-license-agreement

mayankbh:
Yeah, my CLA got rejected when I tried submitting a patch to pinniped :( - let me try and chase down what I'm doing wrong.

mayankbh:
I'm in the vmware-tanzu org now, so I'll try closing and opening a new PR. (I don't think the CLA bot can detect changes to org membership post PR)

PR #1440, opened 2022-01-02
Kickstart UI: Refactor step description

### LINKED ISSUE #1429: Kickstart UI: refactor step description updates to use events
Currently, a wizard asks each step for its dynamic description, and that description might change during a JavaScript cycle, causing a ExpressionChangedAfterItHasBeenCheckedError to be thrown (during dev operations).
We propose that instead, each step broadcasts a STEP_DESCRIPTION_CHANGE event (with the step name and the new description) which the wizard would consume to keep the step descriptions up-to-date.
This would also eliminate the need for the wizard to keep a reference to the step component instance.

- [ ] APIs
- [ ] Addons
- [X] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**

swalner-vmware:
Refactors step descriptions to use events to communicate changes from steps to wizard

Fixes #1429 

### Release note

```release-note

```

### PR Checklist


pkg/v1/tkg/web/src/app/views/landing/aws-wizard/aws-wizard.component.ts
```diff
@@ -332,7 +336,7 @@ export class AwsWizardComponent extends WizardBaseDirective implements OnInit {
         return this.getOsImageForm(AwsOsImageStepComponent);
     }
     get AwsNetworkForm(): FormDataForHTML {
-        return FormUtility.formOverrideDescription(this.NetworkForm, 'Specify the cluster Pod CIDR');
+        return FormUtility.formWithOverrides(this.NetworkForm, {description: 'Specify the cluster Pod CIDR'});
     }
     //
```
miclettej, line 341: Not your doing, but we could remove the comments in lines 341-342. No longer any template related convenience methods.

pkg/v1/tkg/web/src/app/views/landing/aws-wizard/node-setting-step/node-setting-step.component.ts
```diff
@@ -564,6 +565,12 @@ export class NodeSettingStepComponent extends StepFormDirective implements OnIni
         }
     }
 
+    // Because we use clusterTypeDescriptor in our description, we need to trigger a step description change whenever
```
miclettej, line 568: This all works really well. Maybe a separate PR but it would be very helpful I think if we enhance some of the docs for our methods to include some kind of breadcrumb trail. For example, something as simple as "See StepFormDirective.setClusterTypeDescriptor for usage." That might not be a great example but it would help us immediately go to StepFormDirective setClusterTypeDescriptor method to see where 'onChangeClusterTypeDescriptor' is being called, and then it's easy to work back to see that this is triggered by a branding change event. Maybe this type of exercise would be good to have Raymond and myself take on to get us more familiar with the refactored code. Curious to hear your thoughts on this.

pkg/v1/tkg/web/src/app/views/landing/aws-wizard/vpc-step/vpc-step.fieldmapping.ts
```diff
@@ -1,9 +1,10 @@
+// App imports
 import { AwsField, VpcType } from '../aws-wizard.constants';
 import { StepMapping } from '../../wizard/shared/field-mapping/FieldMapping';
 
 export const AwsVpcStepMapping: StepMapping = {
     fieldMappings: [
-        { name: AwsField.VPC_TYPE, required: true, defaultValue: VpcType.NEW },
+        { name: AwsField.VPC_TYPE, required: true, defaultValue: VpcType.EXISTING },
```
miclettej, line 7: Is this change in default intentional? Sorry if I'm forgetting a conversation about this. I feel like we may have talked about it but not remembering why.

swalner-vmware: Yes, this was intentional; part of the feedback on how folks use the UI

miclettej (approved):

Issue #1066, opened 2021-11-04
Suppress sending ansi code if the output is not attached to TTY


kdvolder:
**Bug description**

When you use tanzu cli in combination with `watch` it spits out a bunch of ansi control characters that aren't supported in that context.

To reproduce do something like this in a terminal:

```
watch tanzu package installed list -A
```

You'll get output that looks like this:

```
Every 2.0s: tanzu package installed list -A                                                                                           dellicioso: Thu Nov  4 13:32:11 2021

^[K | Retrieving installed packages... ^[K / Retrieving installed packages... ^[K - Retrieving installed packages... ^[K \ Retrieving installed packages... ^[K | Retrievi
ng installed packages... ^[K / Retrieving installed packages... ^[K - Retrieving installed packages... ^[K \ Retrieving installed packages... ^[K | Retrieving installed p
ackages... ^[K / Retrieving installed packages... ^[K - Retrieving installed packages... ^[K \ Retrieving installed packages... ^[K | Retrieving installed packages... ^[K
 / Retrieving installed packages... ^[K - Retrieving installed packages... ^[K \ Retrieving installed packages... ^[K | Retrieving installed packages... ^[K / Retrieving
installed packages... ^[K - Retrieving installed packages... ^[K \ Retrieving installed packages... ^[K | Retrieving installed packages... ^[K / Retrieving installed pack
ages... ^[K - Retrieving installed packages... ^[K \ Retrieving installed packages... ^[K | Retrieving installed packages... ^[K / Retrieving installed packages... ^[K -
Retrieving installed packages... ^[K \ Retrieving installed packages... ^[K | Retrieving installed packages... ^[K / Retrieving installed packages... ^[K - Retrieving ins
talled packages... ^[K \ Retrieving installed packages... ^[K | Retrieving installed packages... ^[K / Retrieving installed packages... ^[K - Retrieving installed package
s... ^[K \ Retrieving installed packages... ^[K
  ^[1mNAME                              ^[0m  ^[1mPACKAGE-NAME                                       ^[0m  ^[1mPACKAGE-VERSION^[0m  ^[1mSTATUS             ^[0m  ^[1mNAMES
PACE  ^[0m
  appliveview                         appliveview.tanzu.vmware.com                         0.3.0-build6     Reconcile succeeded  tap-install
  buildservice                        buildservice.tanzu.vmware.com                        1.3.1            Reconciling          tap-install
[...]

Issue #352, opened 2021-08-10
When tanzu cli is used on Microsoft Windows, unhandled characters are displayed in output


cantbewong:
**Bug description**
When tanzu cli is used on Microsoft Windows, unhandled (virtuual terminal formatting?) characters are displayed in output
Example:

```
C:\ISOs\tce\tce-windows-amd64-v0.7.0-rc.1.tar\tce-windows-amd64-v0.7.0-rc.1\tce-windows-amd64-v0.7.0-rc.1>tanzu --help
←[1mTanzu CLI←[0m

←[1mUsage:←[0m
  tanzu [command]

←[1mAvailable command groups:←[0m

  ←[1mRun←[0m
    cluster                 Kubernetes cluster operations
    kubernetes-release      Kubernetes release operations
    management-cluster      Kubernetes management cluster operations
    package                 Tanzu package management
    standalone-cluster      Create clusters without a dedicated management cluster

  ←[1mSystem←[0m
    completion              Output shell completion code
    config                  Configuration for the CLI
    init                    Initialize the CLI
    login                   Login to the platform
    plugin                  Manage CLI plugins
    update                  Update the CLI
    version                 Version information


←[1mFlags:←[0m
  -h, --help   help for tanzu

Use "tanzu [command] --help" for more information about a command.

Not logged in
```

**Affected product area (please put an X in all that apply)**

- [ ] APIs
- [ ] Addons
- [ X] CLI
- [ ] Docs
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**
I suspect these are virtual terminal formating characters that do something useful or attractive in a standard Linux shell. They should either be removed on the Windows version of the CLI, or perhaps converted to something that works
See: [https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences?redirectedfrom=MSDN]([](https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences?redirectedfrom=MSDN)) 

**Steps to reproduce the bug**
Install CLI on Windows 10. run `tanzu --help`
Other Tanzu CLI commands exhibit similar behavior. Result ranges from just looking ugly to making output really difficult to read.

**Version** (include the SHA if the version is not obvious)
version: v1.4.0-pre-alpha-2
buildDate: 2021-08-06
sha: 5a42a2b

**Environment where the bug was observed (cloud, OS, etc)**
[...]

PR #1465, opened 2022-01-10
Remove bold formatting from table output

stmcginnis:

The current output formatting for table output bolds the header text.
This has been shown to cause issues when running commands under the
`watch` command or in environments where the terminal is not a full tty.

This change removes the bold formatting. Many other command lines that
emit table output, such as kubectl, do not bold the headers. So the
easiest solution for us is to just stop doing so.

Related: #352 
Related: #1066 
Related: https://github.com/vmware-tanzu/community-edition/pull/2376#issuecomment-1009002218

Ran local tests.

### Release note

```release-note
Header lines in table output will no longer be formatted bold.
```

### DIFF
--- a/pkg/v1/cli/component/output.go
+++ b/pkg/v1/cli/component/output.go
@@ -220,11 +220,6 @@ func renderTable(ow *outputwriter) {
 	table.SetColWidth(colWidth)
 	table.SetTablePadding("\t\t")
 	table.SetHeader(ow.keys)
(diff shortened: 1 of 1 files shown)


joshrosso:
Heavily agree with this change.

stmcginnis:
Linter job failure is being addressed in https://github.com/vmware-tanzu/tanzu-framework/pull/1467

PR #1487, opened 2022-01-12
default to the correct cidrs for KinD for ipv6,ipv4

adobley:
When configured for `ipv6,ipv4` dualstack the CLI will now default to the correct CIDRs and IPFamily when creating a KinD bootstrap cluster.

Without this PR users would need to configure the CIDRs or create a bootstrap cluster themselves.

Unit tests were written to cover the cases.

### Release note

```release-note
Fix issue with automatic bootstrap cluster creation when TKG_IP_FAMILY is set to "ipv6,ipv4"
```

### PR Checklist

### DIFF
--- a/pkg/v1/tkg/kind/client.go
+++ b/pkg/v1/tkg/kind/client.go
@@ -342,7 +342,10 @@ func (k *KindClusterProxy) getKindNetworkingConfig() kindv1.Networking {
 	podSubnet, err := k.options.Readerwriter.Get(constants.ConfigVariableClusterCIDR)
 	if err != nil {
 		switch ipFamily {
-		case constants.DualStackPrimaryIPv4Family:
+		// We expect the IPv4,IPv6 order for the CIDR in both cases for KinD until
+		// we bump to a version of the kind library that includes a fix for this issue:
+		// https://github.com/kubernetes-sigs/kind/issues/2484
+		case constants.DualStackPrimaryIPv4Family, constants.DualStackPrimaryIPv6Family:
 			podSubnet = constants.DefaultDualStackPrimaryIPv4ClusterCIDR
 		case constants.IPv6Family:
 			podSubnet = constants.DefaultIPv6ClusterCIDR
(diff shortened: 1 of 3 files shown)


pkg/v1/tkg/kind/client.go
```diff
@@ -342,7 +342,8 @@ func (k *KindClusterProxy) getKindNetworkingConfig() kindv1.Networking {
 	podSubnet, err := k.options.Readerwriter.Get(constants.ConfigVariableClusterCIDR)
 	if err != nil {
 		switch ipFamily {
-		case constants.DualStackPrimaryIPv4Family:
+		// We expect the IPv4,IPv6 order for the CIDR in both cases for KinD
```
mcwumbly, line 345:
```suggestion
		// We expect the IPv4,IPv6 order for the CIDR in both cases for KinD until
		// we bump to a version of the kind library that includes a fix for this issue:
		// https://github.com/kubernetes-sigs/kind/issues/2484
```

pkg/v1/tkg/kind/client.go
```diff
@@ -353,7 +354,8 @@ func (k *KindClusterProxy) getKindNetworkingConfig() kindv1.Networking {
 	serviceSubnet, err := k.options.Readerwriter.Get(constants.ConfigVariableServiceCIDR)
 	if err != nil {
 		switch ipFamily {
-		case constants.DualStackPrimaryIPv4Family:
+		// We expect the IPv4,IPv6 order for the CIDR in both cases for KinD
```
mcwumbly, line 357:
```suggestion
		// We expect the IPv4,IPv6 order for the CIDR in both cases for KinD until
		// we bump to a version of the kind library that includes a fix for this issue:
		// https://github.com/kubernetes-sigs/kind/issues/2484
```

mcwumbly (commented):
If I understand correctly, this sets the CIDRs to the bootstrap cluster this change is needed because the version of the kind library currently used for bootstrap clusters does not yet include [this fix in KinD](https://github.com/kubernetes-sigs/kind/issues/2484). 

Is that correct?

I think it may be worth updating the comments here to help leave ourselves or future contributors a breadcrumb.

I also updated the original post text to include a release note. Hope that makes sense.

mcwumbly (approved):
lgtm, @vmware-tanzu/tanzu-framework-reviewers I think this is ready for review and should be `ok-to-merge`

vuil (approved):
lgtm, thanks!

PR #1494, opened 2022-01-13
Enable antrea endpointslice feature flag

tylerschultz:
Co-authored-by: Tyler Schultz <tschultz@vmware.com>

Dual stack `Services` need this configuration when running with antrea CNI. 
Dual stack `Services` are only reachable by the primary IP Family IP without this change.
Adding this configuration to Antrea makes it possible to reach services by both ip families.

Setting ANTREA_ENDPOINTSLICES to `true` means that `antrea-proxy` will watch `EndpointSlices`. `EndpointSlices` were made available in 1.21. This repo pulls in antrea the v1.2.3 package, which is only configured to run with Kubernetes 1.22. 

Dual stack `Services` only write the primary IP into `Endpoints`. Dual stack services write both the IPv4 and IPv6 `EndpointSlices`.

Built and deployed a cluster with this configuration default change and saw the Services were reachable from both IP families.

### Release note

```release-note
Add support for dual stack ClusterIP Services when Antrea.
```

### PR Checklist


mcwumbly (approved):

mcwumbly:
> `EndpointSlices` were made available in 1.21. This repo pulls in antrea the v1.2.3 package, which is only configured to run with Kubernetes 1.22.

Wait, now I'm a bit confused. Aren't TKr's with version 1.20.x still supported? What happens if someone tries to deploy a workload cluster with a 1.20.x TKr? Do they need to override this value to be `false` in that case? If so, perhaps this is still the right default going forward, but do we need to document that somewhere?

mcwumbly (approved):

mcwumbly:
@vmware-tanzu/tanzu-framework-reviewers I think this should be `ok-to-merge` and ready to review/merge.

christianang:
> Wait, now I'm a bit confused. Aren't TKr's with version 1.20.x still supported? What happens if someone tries to deploy a workload cluster with a 1.20.x TKr? Do they need to override this value to be false in that case? If so, perhaps this is still the right default going forward, but do we need to document that somewhere?

Just to further clarify 1.20.x uses an older version of Antrea, which wouldn't use this provider change anyways.

vuil (approved):
lgtm, thanks!

PR #1495, opened 2022-01-13
vSphere upgrade rely on moid instead of name

### LINKED ISSUE #1396: Tanzu upgrade isn't idempotent on vSphere when VM template is moved/renamed

tenczar:
This change updates the upgrade logic so that a change to the name(path)
of the virtual machine template does not cause VSphereMachineTemplates
to be regenerated.

Fixes #1396 

Created prod management cluster and dev workload cluster on vsphere. Verified the annotation is set currently in the metadata.

### Release note

```release-note
Fixes a rare issue where cluster machines would be incorrectly rotated when performing an upgrade where the only change to the cluster was the name or path of the VM template.
```

### PR Checklist


anujc25 (approved):
Looks good

PR #1501, opened 2022-01-14
Remove patch version check from workload create

### LINKED ISSUE #1508: Patch version differences cause workload cluster creation validation fail
**Bug description**
A validation error is throw when using a version of the cli that differs from the management cluster in just the patch version. Our compatibility stance is that the cli can support previous patch versions of the same minor, i.e, the patch version of the cli need not match the managment-cluster

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**

tenczar:
Loosens the validation requirement that cli patch version matches the
management cluster version.

This PR allows a tanzu cli to create clusters using a management cluster that does not match the patch version of the cli.

Fixes #1508 

### Release note

```release-note
Loosens version check when creating workload clusters. Management cluster and cli need not match in patch version.
```

### PR Checklist

### DIFF
--- a/pkg/v1/tkg/client/cluster.go
+++ b/pkg/v1/tkg/client/cluster.go
@@ -720,8 +720,7 @@ func (c *TkgClient) ValidateManagementClusterVersionWithCLI(regionalClusterClien
 	}
 
 	if curMCSemVersion.Major() != defaultTKGSemVersion.Major() ||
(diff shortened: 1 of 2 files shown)


anujc25 (approved):
LGTM

PR #1511, opened 2022-01-18
add calico config API spec

### LINKED ISSUE #1577: CalicoConfig API and Controller

Create Core Addons Config API CRD - Create CalicoConfig CRD and controller to create data values secret

maralavi:

This PR adds CalicoConfigSpec CRD and API types

Fixes https://github.com/vmware-tanzu/tanzu-framework/issues/1577

NA

### Release note

```release-note
package-based-lcm: Add CalicoConfigSpec CRD and API types
```

### PR Checklist

### Additional information
The steps that I took for this PR are as follows:
```
kubebuilder create api --group cni --version v1alpha1 --kind CalicoConfig
```
Pressed ```y``` for ```"Create Resource"``` and ```"Create Controller"```.
This generates some files and directories under ```apis/cni```, ```controllers/```, ```config/crd/```, ```config/rbac/```, ```config/smaples/```. Some of which are auto generated code/manifests and the remaining are place holder code which we should fill in.

Completed the  definition of the API spec in ```apis/cni/v1alpha1/calicoconfig_types.go``` and sample values in ```config/samples/cni_v1alpha1_calicoconfig.yaml```.  Currently we only write basic validations in the form of [kubebuilder declaratives](https://book.kubebuilder.io/reference/markers/crd-validation.html).  Later on, we might add validation webhooks and add more go logic for writing the validations. 
For now, as the webhooks are not added yet, didn't submit the generated ```config/crd/patches/cainjection_in_calicoconfigs.yaml``` and ```config/crd/patches/webhook_in_calicoconfigs.yaml``` files.

Run the following make target:
```
make manifests
```

From the produced files, verify the content of the generated CRD to be what we want for the field types, their defaults, etc in ```config/crd/bases/cni.tanzu.vmware.com_calicoconfigs.yaml ```

run ```make generate``` after making changes to ```apis/cni/v1alpha1/calicoconfig_types.go``` so that the deep copy modifications will get generated. (in my case, I had to add header ```//go:build !ignore_autogenerated``` manually to the file to prevent CI pipeline diff issue)

Move the controllers for cni to the addons folder, which we will fill the logic of the controller in subsequent PRs.
```
git mv controllers/ addons/controllers/calico
```


config/samples/cni_v1alpha1_calicoconfig.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: cni.tanzu.vmware.com/v1alpha1
+kind: CalicoConfig
+metadata:
+  name: calicoconfig-sample
+spec:
+  Namespace: "kube-system"
+  InfraProvider: vsphere
+  IpFamily: ipv4
+  ClusterCIDR: "192.168.0.0/16,fd00:100:96::/48"
+  VethMTU: 0
```
maralavi, line 10: @vijaykatam Do you think it matters if the spec be like this (without calico.config nesting) as opposed to the following as in the values.yaml in the[ upstream TCE](https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/calico/3.19.1/bundle/config/values.yaml)?

```
spec:
  Namespace: "kube-system"
  InfraProvider: vsphere
  IpFamily: ipv4
  calico:
    config:
       ClusterCIDR: "192.168.0.0/16,fd00:100:96::/48"
       VethMTU: 0
```

blc1996: Personally, I think the format should match in both these 2 places, Otherwise, it will introduce trouble when doing conversion. We either change the schema in TCE or update the formats in here.

blc1996:
Hi @maralavi , could you note down the steps it takes to create this MR? i.e. what kubebuilder commands did you run?

config/samples/cni_v1alpha1_calicoconfig.yaml
vijaykatam: We should not expose clusterCIDR as a parameter as it will be derived from [ClusterNetwork](https://github.com/kubernetes-sigs/cluster-api/blob/2f975793bf79f8e40173eecd0999cab2d8f17b9d/api/v1beta1/cluster_types.go#L182).

addons/controllers/cni/calicoconfig_controller.go
```diff
@@ -0,0 +1,50 @@
+// Copyright YEAR VMware, Inc. All Rights Reserved.
```
blc1996, line 1: Could we relocate `addons/controllers/cni/calicoconfig_controller.go` to something like `addons/controllers/calico/calicoconfig_controller.go`. I assume Antrea and Calico will have different folders to use.

config/samples/cni_v1alpha1_calicoconfig.yaml
blc1996: I think we need to have some conversion logic that is shared and used by all the config controllers. This logic will convert fields one by one correspondingly from CRD to the yaml used in data values secret.
On top of that, we add special logic to each controller to handle things like `ClusterNetwork` that Vijay mentioned.

addons/controllers/cni/calicoconfig_controller.go
maralavi: @vijaykatam If relocating controllers, do you think we should define ```calico``` as a separate type for the api as well?  So remove the cni grop totally, like: ```apis/calico/v1alpha1/calicoconfig_types.go``` and ```apis/antrea/v1alpha1/antreaconfig_types.go```?

config/samples/cni_v1alpha1_calicoconfig.yaml
shivaani0505: @blc1996 , Is this logic for converting fields from CRD to yaml in data values secret the same thing what @vijaykatam was mentioning, that it should be automatically generated from CRD?

addons/controllers/cni/calicoconfig_controller.go
vijaykatam: yeah lets move the controller to `addons/controllers`. API still lives at `apis/cni`

maralavi: @vijaykatam I've already moved it to ```addons/controllers``` in my current code. The question is should I move it under ```addons/controllers/cni```, under ```addons/controllers/cni/calico/``` or just under ```addons/controllers/calico/```?

maralavi: Aaah I see. I guess you mean to remove the ```cni``` layering and move it directly under ```addons/controllers/```

maralavi: @blc1996 moved it under ```addons/controllers/calico/```.

config/samples/cni_v1alpha1_calicoconfig.yaml
blc1996: Firstly, if we want to use the [yam.marshall/yaml.unmarshall](https://pkg.go.dev/gopkg.in/yaml.v2#Unmarshal) functions provided by GO to generates the content we put into data values secret

```
Namespace: "kube-system"
InfraProvider: vsphere
IpFamily: ipv4
calico:
    config:
	ClusterCIDR: "192.168.0.0/16,fd00:100:96::/48"
	VethMTU: 0
```

We would need to define some struct similar to the following

```
type CalicoConfigYaml struct {

	Namespace string `yaml:"namespace,omitempty"`

	InfraProvider string `yaml:"infraProvider"`

	IPFamily string `yaml:"ipFamily,omitempty"`

	Config struct `yaml:"vethMTU,omitempty"` {
		VethMTU      int64 `yaml:"vethMTU,omitempty"`
                ClusterCIDR string `yaml:"clusterCIDR,omitempty"`
	}
}
```

It is pretty similar to the struct defined for Kubebuilder

```
// CalicoConfigSpec defines the desired state of CalicoConfig
type CalicoConfigSpec struct {

	// The namespace in which calico is deployed
	// +kubebuilder:validation:Optional
	// +kubebuilder:default:=kube-system
	Namespace string `json:"namespace,omitempty"`

	// Infrastructure provider in use
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:Enum="aws";"azure";"vsphere";"docker"
	InfraProvider string `json:"infraProvider"`

	// The IP family calico should be configured with
	// +kubebuilder:validation:Optional
	// +kubebuilder:validation:Enum="ipv4";"ipv6";"ipv4,ipv6";"ipv6,ipv4"
	// +kubebuilder:default:=ipv4
	IPFamily string `json:"ipFamily,omitempty"`

	// Maximum transmission unit setting
	// +kubebuilder:validation:Optional
	// +kubebuilder:validation:Minimum=0
	// +kubebuilder:default:=0
	VethMTU int64 `json:"vethMTU,omitempty"`
}
```

We would need to do a mapping of values between `CalicoConfigYaml` and `CalicoConfigSpec`. If we keep them as similar as possible, it would be easier for the conversion.

maralavi: Changed to use nested structs fo ```VethMTU``` to align with [upstream TCE](https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/calico/3.19.1/bundle/config/values.yaml) and easier conversion as suggested above

apis/cni/v1alpha1/calicoconfig_types.go
```diff
@@ -0,0 +1,70 @@
...
+type Calico struct {
+	Config Config `json:"config,omitempty"`
+}
+
+type Config struct {
+	// Maximum transmission unit setting
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default:=0
+	VethMTU int64 `json:"vethMTU,omitempty"`
```
vijaykatam, line 41: suggestion: add comment that `"0" as default means MTU will be auto detected`

config/crd/patches/cainjection_in_calicoconfigs.yaml
```diff
@@ -0,0 +1,8 @@
+# The following patch adds a directive for certmanager to inject CA into the CRD
```
vijaykatam, line 1: Webhook files are not needed until we add one. Lets remove this and webhook_in_calicoconfigs.yaml

maralavi: Removed these 2 files.

vijaykatam (approved):
LGTM

blc1996 (approved):
The changes LGTM.
Another thing to start considering is that how are we are going to integrate these config reconcilers. (e.g. we add addon reconciler into main.go, do we want to add config reconcilers into the same manager and build into the same image?)
We should finalize this when we start implementing the config controllers logic, as we need entry points to test the controllers.

PR #1522, opened 2022-01-19
Use log.Infof for formatted log string

### LINKED ISSUE #1521: Management cluster creation emits "WARNING: key is not a string" errors during creation
**Bug description**

When deploying a new management cluster, the terminal output contains:

```
WARNING: key is not a string: %!s(int=1)WARNING: key is not a string: %!s(int=1)
```

This is do to using `log.Info()` where `log.Infof()` was intended.

- [ ] APIs
- [ ] Addons
- [X] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**

No errors output to terminal.

Run `tanzu management-cluster create`.

stmcginnis:

There were a few cases where log.Info() was being used, while the
intention was to use log.Infof(). The Info() call can take a variable
number of arguments, but they are expected to be a set of key/value
pairs that will be formatted and printed. If the number of formatting
arguments is odd, or if the expected key is not a string type, this
would case a somewhat confusing error to be emitted stating "key is not
a string".

This updates the few found cases so they get the expected formatting of
output.

Fixes #1521 

Ran through `tanzu management-cluster create` and verified the expected output to the terminal.

### Release note

```release-note

```

### DIFF
--- a/pkg/v1/tkg/client/utils.go
+++ b/pkg/v1/tkg/client/utils.go
@@ -346,15 +346,15 @@ func (c *TkgClient) getMachineCountForMC(plan string) (int, int) {
 		if cpc%2 == 1 {
 			controlPlaneMachineCount = cpc
 		} else {
-			log.Info("Using default value for CONTROL_PLANE_MACHINE_COUNT= %d. Reason: Provided value is an even number", controlPlaneMachineCount)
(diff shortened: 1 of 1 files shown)


vuil (approved):
lgtm

PR #1523, opened 2022-01-19
Trivial grammatical fix for "does not exists"

stmcginnis:

There were several cases where we had comments or strings stating "does
not exists". This corrects the grammar to "does not exist".


vuil (approved):
lgtm

PR #1525, opened 2022-01-19
Add cluster audit logging

### LINKED ISSUE #1524: Audit logging to help with cluster bootstrapping

Currently, when deploying a new management and workload cluster, the terminal output is to the console and if used, to the UI. This is done at a default verbosity that does not capture all activity.

If the deployment fails, sometimes a higher verbosity is needed to trace through what happened.

It's also possible the user has already closed the UI and their terminal has been reset or scrolled off the buffer, so even the default verbosity output is lost. This makes troubleshooting more difficult.

To improve this situation, it would be great if we captured an audit log of all activity. On failure, this log could be shared to help someone step through what happened and diagnose where the failure occurred.

Some information can be retrieved from the kind cluster or the failed management cluster, but that is not as thorough as the initial deployment logs.

- [ ] APIs
- [ ] Addons
- [X] CLI
- [ ] Docs
- [ ] IAM
- [X] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [X] User Experience

**Additional context**

This is part of a proposed effort from TCE to improve the bootstrap user experience, though this is an improvement that would be beneficial whether [that proposal](https://github.com/vmware-tanzu/community-edition/pull/2769) is accepted or not.

stmcginnis:

This adds new functionality where an audit log will be create for all
management clusters on create. This will capture all logging activity,
regardless of set log verbosity. This can be used to help troubleshoot
bootstrap and other failures by having everything that happened during
deployment available in a timestamped log.

If the management cluster is upgraded, we will attempt to capture
upgrade logging activity to the log so it is captured in case it is
needed later.

This file is kept around until deleted by the user or until management
cluster deletion. The delete operation will log everything to this audit
log in case of deletion failure. If that fails, the user and/or support
person can review the log file to help troubleshoot. On successful
deletion of the cluster the audit log file will be removed.

Updated to also include audit logging for workload cluster creation.

Fixes #1524

- Deployed management clusters and observed content being captured in the audit log.
- Deployed workload cluster and observed content being captured in the log.
- Delete workload cluster, verified cluster's audit log captured process output and was removed on successful completion.
- Deleted management clusters and verified deletion activity was logged, and that audit log file was removed on successful completion of delete operation.

### Release note

```release-note
Management  and workload cluster creation activity is now captured in $HOME/.config/tanzu/tkg/logs/$CLUSTER_NAME for later troubleshooting or review. This file will be removed on cluster deletion.
```

### PR Checklist

### DIFF
--- a/pkg/v1/tkg/tkgconfigpaths/filepaths.go
+++ b/pkg/v1/tkg/tkgconfigpaths/filepaths.go
@@ -132,6 +132,7 @@ func (c *client) GetConfigDefaultsFilePath() (string, error) {
 		return "", err
 	}
 	return filepath.Join(tkgDir, constants.TKGConfigDefaultFileName), nil
+}
 
 // GetLogDirectory returns the directory path where log files should be stored by default.
 func (c *client) GetLogDirectory() (string, error) {
--- a/pkg/v1/tkg/tkgconfigpaths/filepaths.go
+++ b/pkg/v1/tkg/tkgconfigpaths/filepaths.go
@@ -132,4 +132,12 @@ func (c *client) GetConfigDefaultsFilePath() (string, error) {
 		return "", err
 	}
 	return filepath.Join(tkgDir, constants.TKGConfigDefaultFileName), nil
+
+// GetLogDirectory returns the directory path where log files should be stored by default.
+func (c *client) GetLogDirectory() (string, error) {
+	tkgDir, err := c.GetTKGDirectory()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(tkgDir, constants.LogFolderName), nil
 }
--- a/pkg/v1/tkg/constants/files.go
+++ b/pkg/v1/tkg/constants/files.go
@@ -34,4 +34,6 @@ const (
 	TKGClusterConfigFileDirForUI           = "clusterconfigs"
 	TKGRegistryCertFile                    = "registry_certs"
 	TKGRegistryTrustedRootCAFileForWindows = ".registry_trusted_root_certs_win"
+
+	LogFolderName = "logs"
 )
--- a/pkg/v1/tkg/log/interface.go
+++ b/pkg/v1/tkg/log/interface.go
@@ -140,6 +140,12 @@ func SetFile(fileName string) {
 	logWriter.SetFile(fileName)
 }
 
+// SetAuditLog sets the log file that should capture all logging activity. This
+// file will contain all logging regardless of set verbosity level.
+func SetAuditLog(fileName string) {
+	logWriter.SetAuditLog(fileName)
+}
+
 // SetChannel sets the channel to writer
 // if channel is set, writer will forward log messages to this log channel
 func SetChannel(channel chan<- []byte) {
--- a/pkg/v1/tkg/log/writer.go
+++ b/pkg/v1/tkg/log/writer.go
@@ -26,6 +26,10 @@ type Writer interface {
 	// write the logs to this file
 	SetFile(fileName string)
 
+	// SetAuditLog sets the log file that should capture all logging activity. This
+	// file will contain all logging regardless of set verbosity level.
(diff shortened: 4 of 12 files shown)


vuil:
Thanks for the changes! Will give it a look in the next day or two.

pkg/v1/tkg/tkgconfigpaths/filepaths.go
```diff
@@ -124,3 +124,12 @@ func GetRegistryTrustedCACertFileForWindows() (string, error) {
...
 	return path.Join(home, constants.TKGRegistryTrustedRootCAFileForWindows), nil
 }
+
+// GetLogDirectory returns the directory path where log files should be stored by default.
+func (c *client) GetLogDirectory() (string, error) {
+	tkgDir, err := c.GetTKGDirectory()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(tkgDir, constants.LogFolderName), nil
```
navidshaikh, line 134: Q: Should we consider env var`XDG_STATE_HOME` if set, else use the default (base) dir `$HOME/.local/state` ?

stmcginnis: I think at least for now it should be the same path as `GetTKGDirectory` provides. We can decide later if we would like to customize things, but at least initially I think it makes sense to keep everything contained to the same directory tree.

That's my take on things, but maybe a good topic for discussion. I'm certainly open to hearing what others think.

stmcginnis:
Test failure doesn't look related to this change, but I will test locally some more to see if I can figure out any cause for it.

@vuil would love your thoughts on this. Especially related to @navidshaikh's suggestion on location.

stmcginnis:
Looks like failing test may no longer be valid?
```
Msg: "fail: autoscaler is not supported for k8s version v1.23.3+vmware.1"
```

vuil:
@saimanoj01 PTAL at the autoscaler issue. Looks like we need a new version compatible for 1.23. If it means we have skip the test for said version let's file and issue to track reinstating it.

vuil (approved):
LGTM, thanks! This will go some ways towards tracking down bootstrap issues.

anujc25:
Is there a way for some users to disable this audit logging?

This make complete sense for the CLI users but these functionalities are used as a library by other products and considering they capture and store logs differently based on the use-case, Writing the logs always might not be the thing needed for library consumers. 

Should we consider providing a switch to disable this logging incase anyone wants to disable this functionality as they do not want to store this logs on their system.

stmcginnis:
> Should we consider providing a switch to disable this logging incase anyone wants to disable this functionality as they do not want to store this logs on their system.

@vuil any thoughts on this?

My initial thought is no, that we should enforce that this is a consistent behavior for any consumers of `tanzu-framework`. But if that is a requirement for some consumers, it should be possible to introduce a flag that controls whether the log file is created or not. That would complicate things a bit. It's easy enough to have this flag on create operations, but for things like upgrades, etc. we would either also need to provide a way to pass in this flag, or do some detection up front to see if the file exists or not and base on continued logging on its presence.

Unless we know for sure of a consumer that this is going to cause some issue for, I'd rather get this introduced, then if/when we find a consumer that has a problem with it to introduce the added complexity.

vuil:
nit: should the original issue and the testing done be updated to also apply to workload clusters given the update?

> Unless we know for sure of a consumer that this is going to cause some issue for, I'd rather get this introduced, then if/when we find a consumer that has a problem with it to introduce the added complexity.
I would be okay with this approach unless there is already a known library use case where this breaks.

stmcginnis:
> nit: should the original issue and the testing done be updated to also apply to workload clusters given the update?

Good catch - updated the issue wording and expanded on the testing that was done with the final changes.

stmcginnis:
> I would be okay with this approach unless there is already a known library use case where this breaks.

👍

PR #1540, opened 2022-01-24
Use k8s default namespace from kubeconfig

### LINKED ISSUE #1351: `tanzu package` (and possibly other commands) do not respect namespace in kubeconfig
**Bug description**

If you set your kubeconfig context like so:

```shell
kubectl config set-context --current --namespace somewhere
```

Running commands like `tanzu package installed` without a `--namespace` option will operate on the `default` namespace, rather than the `somewhere` namespace. This means that for long sequences / scripts, you need to repeat the `--namespace` argument every time.

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [x] Plugin
- [ ] Security
- [ ] Test and Release
- [x] User Experience

**Expected behavior**

`tanzu package` (and other plugins) respect the namespace in the `KUBECONFIG` the same way that they respect the auth and cluster address there.

Run the following on a TKG cluster:

```shell
$ kubectl config set-context --current --namespace tkg-system
$ tanzu package installed list
<no output>
$ kubectl get pkgi
<list of TKG packages installed>
```

0.10.0

TCE

adduarte:
If contexts are being used in the kubeconfig,
the default value of --namespace flag for
package install should respect the one define in the
kubernetes context

If the a kubernetes context is set in the kubeconfig, the namespace defined
for that context will be used as the default value (instead of "default")
in the case the --namespace flag is not provided by the user.  

Fixes #1351

Verified that if no kubeconfig context is set  default value is "default"
Verified that if  context is set, the context.namespace is used as default value.

### Release note

```release-note
If the a kubernetes context is set in the kubeconfig, the namespace defined
for that context will be used as the default value (instead of "default")
in the case the --namespace flag is not provided by the user.  

```

### PR Checklist

### DIFF
--- a/cmd/cli/plugin/package/main.go
+++ b/cmd/cli/plugin/package/main.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/aunum/log"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"k8s.io/client-go/tools/clientcmd"
 
 	cliv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/cli/v1alpha1"
 	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli/command/plugin"
@@ -22,6 +23,8 @@ var descriptor = cliv1alpha1.PluginDescriptor{
 	Description: "Tanzu package management",
 	Group:       cliv1alpha1.RunCmdGroup,
 }
+
+const defaultString = "default"
 
 var logLevel int32
 var outputFormat string
@@ -75,4 +78,22 @@ func isPackagingAPIAvailable(kubeCfgPath string) (bool, error) {
 	apiGroup4 := capdiscovery.Group("packageInstallAPIQuery", tkgpackagedatamodel.PackagingAPIName).WithVersions(tkgpackagedatamodel.PackagingAPIVersion).WithResource("packageinstalls")
 
 	return clusterQueryClient.Query(apiGroup1, apiGroup2, apiGroup3, apiGroup4).Execute()
+}
+
+func getNamespaceFromKubeconfig(originalValue string) string {
+	if originalValue != defaultString {
+		return originalValue
+	}
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	loadingRules.ExplicitPath = kubeConfig
+	configOverrides := &clientcmd.ConfigOverrides{}
+	kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
+	namespace, _, err := kubeConfig.Namespace()
+	if err != nil {
+		log.Error("Failed to read a valid context namespace from the available kubeconfig. Using 'default' value for namespace.")
+	}
+	if err != nil || namespace == "" {
+		return defaultString
+	}
+	return namespace
 }
--- a/cmd/cli/plugin/package/package_install.go
+++ b/cmd/cli/plugin/package/package_install.go
@@ -31,7 +31,7 @@ func init() {
 	packageInstallCmd.Flags().StringVarP(&packageInstallOp.PackageName, "package-name", "p", "", "Name of the package to be installed")
 	packageInstallCmd.Flags().StringVarP(&packageInstallOp.Version, "version", "v", "", "Version of the package to be installed")
 	packageInstallCmd.Flags().BoolVarP(&packageInstallOp.CreateNamespace, "create-namespace", "", false, "Create namespace if the target namespace does not exist, optional")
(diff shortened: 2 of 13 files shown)


cmd/cli/plugin/package/package_install.go
```diff
@@ -34,7 +34,7 @@ func init() {
 	packageInstallCmd.Flags().StringVarP(&packageInstallOp.PackageName, "package-name", "p", "", "Name of the package to be installed")
 	packageInstallCmd.Flags().StringVarP(&packageInstallOp.Version, "version", "v", "", "Version of the package to be installed")
 	packageInstallCmd.Flags().BoolVarP(&packageInstallOp.CreateNamespace, "create-namespace", "", false, "Create namespace if the target namespace does not exist, optional")
-	packageInstallCmd.Flags().StringVarP(&packageInstallOp.Namespace, "namespace", "n", "default", "Namespace indicates the location of the repository from which the package is retrieved")
+	packageInstallCmd.Flags().StringVarP(&packageInstallOp.Namespace, "namespace", "n", getDefaultNameSpace(), "Namespace indicates the location of the repository from which the package is retrieved")
```
adduarte, line 37: Having to do this everywhere points to a posiblity of moving the definition of this flag to a higher level where it only gets executed once. All the files are part of the main package so we shoudl be able to move it up

shyaamsn:
@adduarte Kubeconfig can be set as an env var or passed in to cmd using the flag . e.g tanzu package install --kubeconfig <path_to_kubeconfig>. Have we tested that both these scenarios work?

cmd/cli/plugin/package/main.go
```diff
@@ -76,3 +77,15 @@ func isPackagingAPIAvailable(kubeCfgPath string) (bool, error) {
...
 	return clusterQueryClient.Query(apiGroup1, apiGroup2, apiGroup3, apiGroup4).Execute()
 }
+
+func getDefaultNameSpace() string {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	loadingRules.ExplicitPath = kubeConfig
+	configOverrides := &clientcmd.ConfigOverrides{}
+	kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
+	namespace, _, err := kubeConfig.Namespace()
+	if err != nil || namespace == "" {
```
danniel1205, line 87: question: Do we want to some logging here in the case of error? And a little bit curious on what is the user experience if `kubeconfig` provided does not exist(`loadingRules.ExplicitPath = kubeConfig`).

cmd/cli/plugin/package/main.go
```diff
@@ -4,6 +4,7 @@
 package main
 
 import (
+	"k8s.io/client-go/tools/clientcmd"
```
maralavi, line 7: The  imports should be grouped  in three sets: std lib and go0internal imports, external imports, local vmware-tanzu imports, and each set should be separated by an empty line. (In this case, ```"k8s.io/client-go/tools/clientcmd"``` can be moved to where ```"github.com/aunum/log"``` is, followed by a ```make fmt``` to fix the formatting

cmd/cli/plugin/package/main.go
adduarte: ^^ yes. The error does not require process to stop but the user should be informed because it means there might be something bad with their kubeconfig

cmd/cli/plugin/package/main.go
```diff
@@ -76,3 +77,15 @@ func isPackagingAPIAvailable(kubeCfgPath string) (bool, error) {
 
 	return clusterQueryClient.Query(apiGroup1, apiGroup2, apiGroup3, apiGroup4).Execute()
 }
+
+func getDefaultNameSpace() string {
```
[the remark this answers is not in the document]
adduarte: done

cmd/cli/plugin/package/main.go
adduarte: thanks. done

cmd/cli/plugin/package/package_install.go
adduarte: Best to leave as is since it will only make the code harder to read and there might be situations where a particulare subcommand wants a different behavior.  either way spelling it out in every case makes it more clear.

cmd/cli/plugin/package/main.go
adduarte: Added error log pointing out failue to obtain a valid context namespace from the available kubeconfigs.

adduarte:
> @adduarte Kubeconfig can be set as an env var or passed in to cmd using the flag . e.g tanzu package install --kubeconfig <path_to_kubeconfig>. Have we tested that both these scenarios work?

currently this patch does not handle the scenario where --kubeconfig is specifed. We need to fix that before we can think about merging it.

PR #1541, opened 2022-01-24
Add API types and controller for TanzuClusterBootstrap

### LINKED ISSUE #1581: TanzuBootstrapCluster APIs and Controller
Adds API types TanzuClusterBootstrapTemplate and TanzuClusterBootstrap and Controller logic

vijaykatam:
* Adds API types TanzuClusterBootstrapTemplate and TanzuClusterBootstrap
* Controller clones TanzuClusterBootstrapTemplate for a cluster
* Demonstrate controller handling of providers and their generated secrets for additionalPackages

Add API types TanzuClusterBootstrapTemplate and TanzuClusterBootstrap and controller for package management via TKR

Fixes #1581

1. Local controller testing by connecting to management cluster. Controller not enabled until BOM updated with new addons package
2. envtest for controller 
### Release note

```release-note
package-based-lcm: Add API types and controller for TanzuClusterBootstrap
```

### PR Checklist


config/crd/bases/config.tanzu.vmware.com_clientconfigs.yaml
```diff
@@ -140,6 +140,9 @@ spec:
                           type: object
                       type: object
                     type: array
+                  edition:
```
vijaykatam, line 143: ^ @anujc25 this sync the clientconfigs CRD

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,588 @@
...
+
+		tanzuClusterBootstrap.Status.ResolvedTKR = tkrName
+		if err := r.Status().Update(r.context, tanzuClusterBootstrap); err != nil {
+			return nil, err
+		}
+		r.Log.Info("cloned tanzuclusterbootstrap", "tanzuclusterbootstrap", tanzuClusterBootstrap)
+		return tanzuClusterBootstrap, nil
+	}
+
+	// TODO upgrade needs patch (update versions of all packages, merge configs, add additional packages and remove packages that don't exist anymore)
```
imikushin, line 310: suggestion: Can we add/find an issue and refer to it here and in the PR?

Please disregard if this is still WIP.

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,588 @@
...
+	return dataValuesSecret, nil
+}
+
+//nolint:funlen
+//gocyclo:ignore
+// cloneSecretsAndProviders clones linked secrets and providers into the same namespace as TanzuClusterBootstrap
+func (r *tanzuClusterBootstrapReconciler) cloneSecretsAndProviders(cluster *clusterapiv1beta1.Cluster, bootstrap *runtanzuv1alpha3.TanzuClusterBootstrap,
+	templateNS string, log logr.Logger) ([]*corev1.Secret, []*unstructured.Unstructured, error) {
+
+	updateValuesFromSecret := func(valuesFrom *runtanzuv1alpha3.ValuesFrom, packageShortName string) (*corev1.Secret, error) {
```
imikushin, line 387: The func value can be defined as its own method.

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,588 @@
...
+			if err := r.Create(r.context, newSecret); err != nil {
+				return nil, err
+			}
+			valuesFrom.SecretRef = newSecret.Name
+		}
+
+		return newSecret, nil
+	}
+
+	updateValuesFromProvider := func(valuesFrom *runtanzuv1alpha3.ValuesFrom, packageShortName string) (*unstructured.Unstructured, error) {
```
imikushin, line 416: The func value can be defined as its own method.

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,588 @@
...
+
+	return createdSecrets, createdProviders, nil
+}
+
+// ensureOwnerRef will ensure the provided OwnerReference onto the secrets and provider objects
+func (r *tanzuClusterBootstrapReconciler) ensureOwnerRef(ownerRef *metav1.OwnerReference, secrets []*corev1.Secret, providers []*unstructured.Unstructured) error {
+	for _, secret := range secrets {
+		secret.OwnerReferences = clusterapiutil.EnsureOwnerRef(secret.OwnerReferences, *ownerRef)
+		// TODO patch
+		if err := r.Client.Update(r.context, secret); err != nil {
```
imikushin, line 534: suggestion: Use PatchSet (defined in pkg/v1/util/patchset/patchset.go). For usage example, see pkg/v1/tkr/controllers/source/tkr_source_controller.go

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,589 @@
...
+	_, err := controllerutil.CreateOrPatch(r.context, clusterClient, dataValuesSecret, dataValuesSecretMutateFn)
+	if err != nil {
+		log.Error(err, "Error creating or patching addon data values secret")
+		return nil, err
+	}
+	return dataValuesSecret, nil
+}
+
+//nolint:funlen
+//gocyclo:ignore
```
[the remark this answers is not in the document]
vijaykatam: I refactored to the maximum and gocyclo still complains that complexity 18 > 15. I am going to keep the lint comment and proposed a better complexity checker https://github.com/vmware-tanzu/tanzu-framework/issues/1563

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,589 @@
...
+			return nil, nil, err
+		} else { //nolint:revive
+			createdSecrets = append(createdSecrets, secret)
+		}
+		if provider, err := updateValuesFromProvider(bootstrap.Spec.Kapp.ValuesFrom, packageShortName(bootstrap.Spec.Kapp.RefName)); err != nil {
+			return nil, nil, err
+		} else { //nolint:revive
+			createdProviders = append(createdProviders, provider)
+		}
+	}
```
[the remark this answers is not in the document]
vijaykatam: yes updating

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,589 @@
...
+
+		return newProvider, nil
+	}
+	var createdProviders []*unstructured.Unstructured
+	var createdSecrets []*corev1.Secret
+
+	if bootstrap.Spec.CNI != nil && bootstrap.Spec.CNI.ValuesFrom != nil {
+		if secret, err := updateValuesFromSecret(bootstrap.Spec.CNI.ValuesFrom, packageShortName(bootstrap.Spec.CNI.RefName)); err != nil {
+			return nil, nil, err
+		} else { //nolint:revive
```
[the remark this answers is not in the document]
vijaykatam: These are all gone now. Resolving similar comments as well.

addons/main.go
```diff
@@ -146,6 +148,20 @@ func main() {
 		os.Exit(1)
 	}
 
+	// TODO add after addons-manager package that contains tanzuclusterbootstrap* CRDs is integrated into the BOM
```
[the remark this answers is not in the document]
vijaykatam: We have issues in https://github.com/orgs/vmware-tanzu/projects/15/views/22, not going to link in code because board is private.

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,588 @@
...
+		return err
+	}
+	r.dynamicClient = dynClient
+	r.providerWatches = make(map[string]client.Object)
+
+	r.providerGVR = make(map[schema.GroupKind]*schema.GroupVersionResource)
+	clientset := kubernetes.NewForConfigOrDie(mgr.GetConfig())
+	r.cachedDiscoveryClient = cacheddiscovery.NewMemCacheClient(clientset.Discovery())
+
+	ticker := time.NewTicker(constants.DiscoveryCacheInvalidateInterval)
```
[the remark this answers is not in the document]
vijaykatam: updated

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,588 @@
...
+	if tanzuClusterBootstrap == nil {
+		return ctrl.Result{}, nil
+	}
+
+	remoteClient, err := util.GetClusterClient(r.context, r.Client, r.Scheme, clusterapiutil.ObjectKey(cluster))
+	if err != nil {
+		log.Error(err, "Error getting remote cluster client")
+		return ctrl.Result{}, err
+	}
+	// TODO handle kapp as a remote packageinstall from mgmt cluster
```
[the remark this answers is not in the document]
vijaykatam: Issues in https://github.com/orgs/vmware-tanzu/projects/15/views/22.

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,588 @@
...
+	// TODO handle cni, cpi, csi
```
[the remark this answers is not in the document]
vijaykatam: ^ same https://github.com/orgs/vmware-tanzu/projects/15/views/22

addons/controllers/tanzuclusterbootstrap_controller.go
vijaykatam: Same as earlier, see https://github.com/orgs/vmware-tanzu/projects/15/views/22. We discussed when reviewing within team to add TODOs to make it easier for other people to pickup.

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,607 @@
...
+	valuesFrom *runtanzuv1alpha3.ValuesFrom, templateNS string, packageShortName string, log logr.Logger) (*unstructured.Unstructured, error) {
+
+	var newProvider *unstructured.Unstructured
+	if valuesFrom.ProviderRef != nil {
+		gvr, err := r.getGVR(schema.GroupKind{Group: *valuesFrom.ProviderRef.APIGroup, Kind: valuesFrom.ProviderRef.Kind})
+		if err != nil {
+			log.Error(err, "failed to getGVR")
+			return nil, err
+		}
+		provider, err := r.dynamicClient.Resource(*gvr).Namespace(templateNS).Get(r.context, valuesFrom.ProviderRef.Name, metav1.GetOptions{})
```
ankeesler, line 580: question: remind me again how these initial `providerRef`'s get onto the cluster? is this with the initial tkg-management package install?

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,607 @@
...
+	remoteClient, err := util.GetClusterClient(r.context, r.Client, r.Scheme, clusterapiutil.ObjectKey(cluster))
+	if err != nil {
+		log.Error(err, "Error getting remote cluster client")
+		return ctrl.Result{}, err
+	}
+	// TODO handle kapp as a remote packageinstall from mgmt cluster
+	// TODO handle cni, cpi, csi
+	// handle additionalPackages
+	for _, additionalPkg := range tanzuClusterBootstrap.Spec.AdditionalPackages {
+		// TODO packageinstall (rbac, serviceaccount, package, packageinstall)
```
ankeesler, line 186: question: what happens if user has already created legacy addon `Secret` and so there is already a `PackageInstall` for that package? (i know this is outside the bounds of this PR, it was just a question I had when trying to figure out the pinniped flow)

addons/controllers/tanzuclusterbootstrap_controller.go
vijaykatam: The CR objects should be part of the TKR package.

ankeesler: cool. so like they'd be a "template" `PinnipedConfig` from which the rest of the `PinnipedConfig` CRs would get created (correct me if i am wrong)

is there guidance on whether we should prevent a controller from reconciling this "template" `PinnipedConfig` CR into a child `Secret`?

this is what i am imagining happening:
1. tkg-management package installed (including `PinnipedConfig` "template" CR and pinniped-config-controller)
2. pinniped-config-controller starts running and sees `PinnipedConfig` "template" CR
3. pinniped-config-controller creates addon `Secret` with data values from `PinnipedConfig` "template" CR
4. this extra addon `Secret` is never used by anything on the cluster

addons/controllers/tanzuclusterbootstrap_controller.go
vijaykatam: Short answer: We'll use annotations on secrets to figure out conflicts. 

For upgrade in general there are various aspects to upgrade. I think the one to address will be upgrade from TKGm 1.6 to uTKG for AWS and Azure. For vsphere it will be a migration. @danniel1205 has started collecting some thoughts on a doc so it would be good to chat in a sync meeting with @vuil and @imikushin on their thoughts as well.

addons/controllers/tanzuclusterbootstrap_controller.go
vijaykatam: The template object is never associated to a cluster(via ownerReferences or naming convention). If you look at https://github.com/vmware-tanzu/tanzu-framework/pull/1555 the secretRef should be set on status only if there is an  associated cluster object.

addons/controllers/tanzuclusterbootstrap_controller.go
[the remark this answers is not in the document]
imikushin: I know, gocyclo is pretty dumb, but if it's complaining, then the code is definitely not simple enough :)

From a closer look at the method, I can see it could definitely be simpler.

`secrets` and `providers` are actually not two completely different things. In fact, their purpose is exactly the same. If we could capture this purpose in an interface type, then we could collect and return one list of objects, not two. 

Sure, `*Secret` and `*Unstructured` don't directly do what we need. But we can wrap each in a struct implementing exactly our intended interface. In this particular case, `EnsureOwnerRef(*metav1.OwnerReference)` (I didn't find anything else). Our interface would look something like:

```go
type configProvider interface {
  EnsureOwnerRef(*metav1.OwnerReference) error
}
```

This would allow us to avoid creating an inline func (`aggregateSecretsProviders()` with 6 args!) mutating the two slices. Instead, `r.updateValues()` would return only two values: `(configProvider, error)`. This dramatically simplifies collecting and returning these to something like:

```go
configProvider, err := r.updateValues(/* args */)
if err != nil {
  return err
}
configProviders = append(configProviders, configProvider)
```

Also, since `CNI`, `CPI`, `CSI`, `Kapp` fields are of the same type, we can simply use a for loop over them:

```go
bootstrapPackages := append([]*runtanzuv1alpha3.TanzuClusterBootstrapPackage{
  bootstrap.Spec.CNI,
  bootstrap.Spec.CPI,
  bootstrap.Spec.CSI,
  bootstrap.Spec.Kapp,
}, bootstrap.Spec.AdditionalPackages...)

for _, bootstrapPackage := range bootstrapPackages {
  if bootstrapPackage == nil {
    continue
  }
  configProvider, err := r.updateValues(/* args */)
  if err != nil {
    return err
  }
  configProviders = append(configProviders, configProvider)
}
```

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,607 @@
...
+	}
+
+	if bootstrap.Spec.Kapp != nil && bootstrap.Spec.Kapp.ValuesFrom != nil {
+		if err := aggregateSecretsProviders(cluster, bootstrap, bootstrap.Spec.Kapp.ValuesFrom, templateNS, packageShortName(bootstrap.Spec.Kapp.RefName), log); err != nil {
+			return nil, nil, err
+		}
+	}
+
+	for _, additionalPackage := range bootstrap.Spec.AdditionalPackages {
+		if err := aggregateSecretsProviders(cluster, bootstrap, additionalPackage.ValuesFrom, templateNS, packageShortName(additionalPackage.RefName), log); err != nil {
```
imikushin, line 428: Potential nil pointer dereference here: additionalPackage is a pointer.

addons/controllers/tanzuclusterbootstrap_controller.go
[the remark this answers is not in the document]
vijaykatam: List of `runtanzuv1alpha3.TanzuClusterBootstrapPackage` is a good suggestion, updating(guess needed a different perspective :). I am not sure about configProvider interface though.

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,588 @@
...
+		// TODO packageinstall (rbac, serviceaccount, package, packageinstall)
+		secret, err := r.createOrPatchPackageInstallSecret(cluster, additionalPkg, remoteClient, tanzuClusterBootstrap.Namespace, log)
+		if err != nil {
+			log.Error(err, "failed to createOrPatchPackageInstallSecret")
+			return ctrl.Result{}, err
+		}
+		if secret != nil {
+			log.Info("created secret for package in cluster", "secret", secret)
+		}
+		// set watches on provider objects in additional packages if not already set
```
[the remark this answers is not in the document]
vijaykatam: yes, I created a seperate watch func since it will be needed for other packages as well.

addons/controllers/tanzuclusterbootstrap_controller.go
vijaykatam: not an issue anymore since this is refactored

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,589 @@
...
+	dynamicClient dynamic.Interface
+	// on demand dynamic watches for provider refs
+	providerWatches map[string]client.Object
+	// discovery client for looking up api-resources and preferred versions
+	cachedDiscoveryClient discovery.CachedDiscoveryInterface
+	// cache for resolved api-resources so that look up is fast (cleared periodically)
+	providerGVR map[schema.GroupKind]*schema.GroupVersionResource
+}
+
+//nolint:revive
```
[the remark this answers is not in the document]
imikushin: We do need a reason to keep the `//nolint`

addons/controllers/tanzuclusterbootstrap_controller.go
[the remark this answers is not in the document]
imikushin: We should strive to have a TODO refer ideally to one issue (maybe a couple). 

The sentence following `// TODO` makes sense to me as a legitimate issue title.

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,590 @@
...
+	if err != nil {
+		return nil, err
+	}
+	for _, apiResource := range apiResourceList {
+		gv, err := schema.ParseGroupVersion(apiResource.GroupVersion)
+		if err != nil {
+			return nil, err
+		}
+		if gv.Group == gk.Group {
+			for _, res := range apiResource.APIResources { //nolint:gocritic
```
imikushin, line 457: `//nolint` needs a reason

addons/controllers/tanzuclusterbootstrap_controller.go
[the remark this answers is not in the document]
imikushin: `configProvider` or another name () - doesn't matter. Maybe even `client.Object` could possibly fit here. The idea is: `createdSecrets` and `createdProviders` are only separate because they have a different element types. 

Their purpose in life is about the same thing though. If we could make that purpose into an interface type, we would further simplify the code by reducing "accidental" complexity.

addons/controllers/tanzuclusterbootstrap_controller.go
imikushin: I'm all for TODOs when we know something needs to be done! I just want these things to be surfaced as issues. I also don't want them to be forgotten. And we, as humans, can forget about them in a couple different ways:

1. Forget what the TODO means (multiple examples to this, right in this repo)
2. Forget TO DO it: because there's no issue, and there's a lot of other issues to address, we address them and forget about the TODO which often leads to (1).

apis/run/v1alpha3/tanzuclusterbootstrap_types.go
```diff
@@ -0,0 +1,45 @@
...
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:path=tanzuclusterbootstraps,shortName=tcb,scope=Namespaced
+// +kubebuilder:printcolumn:name="CNI",type="string",JSONPath=".spec.cni.refName",description="CNI package name and version"
+// +kubebuilder:printcolumn:name="CSI",type="string",JSONPath=".spec.csi.refName",description="CSI package name and version"
+// +kubebuilder:printcolumn:name="CPI",type="string",JSONPath=".spec.cpi.refName",description="CPI package name and version"
+// +kubebuilder:printcolumn:name="Kapp",type="string",JSONPath=".spec.kapp.refName",description="Kapp package name and version"
+// +kubebuilder:printcolumn:name="Additional Packages",type="string",JSONPath=".spec.additionalPackages[*].refName",description="Additional packages",priority=10
+// +kubebuilder:printcolumn:name="Resolved_TKR",type="string",JSONPath=".spec.status.resolvedTKR",description="Resolved TKR name"
+
```
maralavi, line 24: Should for this one it be : ```JSONPath=".status.resolvedTKR"```?

vijaykatam: good catch, updating

addons/controllers/tanzuclusterbootstrap_handlers.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil
+	}
+
+	log := r.Log.WithValues(constants.TanzuClusterBootstrapNameLogKey, bootstrap.Name)
+
+	log.V(4).Info("Mapping TanzuClusterBootstrap to cluster")
+
+	cluster := &clusterv1beta1.Cluster{}
+
+	// TODO use ownerref instead
```
danniel1205, line 65: Question: I am assuming this means that we get the cluster `objectKey` from `bootstrap.Metadata.OwnerReferences` ? And do we want to address it now or have an issue to track ?

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,586 @@
...
+	// if found and resolved tkr is the same, return found object as the TKR is supposed to be immutable
+	// also preserves any user changes
+	if err == nil && tkrName == tanzuClusterBootstrap.Status.ResolvedTKR {
+		return tanzuClusterBootstrap, nil
+	}
+
+	if !apierrors.IsNotFound(err) {
+		return nil, err
+	}
+	if tanzuClusterBootstrap.UID == "" {
```
danniel1205, line 222: question: will this condition always be True ?

addons/controllers/tanzuclusterbootstrap_handlers.go
vijaykatam: Updated to use owneref

addons/controllers/tanzuclusterbootstrap_controller.go
vijaykatam: True when object is not found on line 212

danniel1205: is it equivalent to ?
```
if !apierrors.IsNotFound(err) {
    return nil, err
} else {
    log.Info("tanzuClusterBootstrap for cluster does not exist, cloning from template")
    ......

}
```

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,586 @@
...
+			log.Error(err, "failed to getGVR")
+			return nil, err
+		}
+		provider, err := r.dynamicClient.Resource(*gvr).Namespace(templateNS).Get(r.context, valuesFrom.ProviderRef.Name, metav1.GetOptions{})
+		if err != nil {
+			log.Error(err, "unable to fetch provider", "provider", valuesFrom.ProviderRef, "gvr", gvr)
+			return nil, err
+		}
+		newProvider = provider.DeepCopy()
+		unstructured.RemoveNestedField(newProvider.Object, "metadata")
```
danniel1205, line 529: question: This line makes me think if we want to completely remove `metadata` from `newProvider` OR we could skip this line to preserve something like `annotations` under metadata ? I believe the code below still overwrites `OwnerRef`, `Name` and `Namespace`.

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,586 @@
...
+		return tanzuClusterBootstrap, nil
+	}
+
+	// TODO upgrade needs patch (update versions of all packages, merge configs, add additional packages and remove packages that don't exist anymore)
+	// https://github.com/vmware-tanzu/tanzu-framework/issues/1584
+	if tkrName != tanzuClusterBootstrap.Status.ResolvedTKR {
+		log.Info("TODO handle upgrade")
+		return nil, nil
+	}
+	return nil, errors.New("should not happen")
```
danniel1205, line 277: nit: In theory that our code should not reach this line, but just in case if it does, do we want to provide some meaningful error message ?

addons/controllers/tanzuclusterbootstrap_controller.go
vijaykatam: Not just annotations, there could be labels as well and they probably don't mean anything for cloned objects. By clearing out we aren't making any implicit guarantees for consumers relying on annotations if any. I am thinking this could be improved by doing a deepCopy of just the spec field instead since that's what I want.

addons/controllers/tanzuclusterbootstrap_controller.go
vijaykatam: probably the same but our linters complain

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -0,0 +1,586 @@
...
+		return err
+	}
+	r.dynamicClient = dynClient
+	r.providerWatches = make(map[string]client.Object)
+
+	r.providerGVR = make(map[schema.GroupKind]*schema.GroupVersionResource)
+	clientset := kubernetes.NewForConfigOrDie(mgr.GetConfig())
+	r.cachedDiscoveryClient = cacheddiscovery.NewMemCacheClient(clientset.Discovery())
+
+	go r.periodicGVRCachesClean()
```
danniel1205, line 114: unrelated:  I was expecting client-go handles the cache invalidation to make sure data consistency.

vijaykatam: This is for calling Invalidate() on CachedDiscovery which has a helper for getting preferred apiResources that the server has, since getting all api-resources is an expensive call the cached client helps, with the caveat that it needs to be invalidated periodically if new CRD's are installed and need to be discovered.

Nothing to do with controller-runtime cache for getting specific instance of a resource.

danniel1205 (approved):
LGTM

imikushin (approved):
LGTM! Minor comments can be addressed in a follow-up PR

addons/controllers/tanzuclusterbootstrap_controller.go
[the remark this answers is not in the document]
vijaykatam: Had a chat with Ivan and this is being tracked in https://github.com/vmware-tanzu/tanzu-framework/issues/1610

PR #1554, opened 2022-01-26
Add CPIConfig spec and CRD for vsphere-cpi package.

HanFa:

This PR adds CPIConfig CRD, API types and samples for vsphere-cpi package.

It also implements the CPIConfig controller that reconciles data values for the vsphere-cpi package.

Fixes # https://github.com/vmware-tanzu/tanzu-framework/issues/1578

### Release note

```release-note
package-based-lcm: Add CPIConfig spec and CRD for vsphere-cpi package.
```

### PR Checklist


addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,48 @@
...
+
+//+kubebuilder:rbac:groups=cpi.tanzu.vmware.com,resources=cpiconfigs,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=cpi.tanzu.vmware.com,resources=cpiconfigs/status,verbs=get;update;patch
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the CPIConfig object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
```
danniel1205, line 33: suggestion: lets remove those boilerplate code?

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,48 @@
...
+func (r *CPIConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	_ = context.Background()
+	_ = r.Log.WithValues("cpiconfig", req.NamespacedName)
+
+	// your logic here
```
danniel1205, line 38: question: do we plan to add the controller logic separately ?

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+type ExtraArgs struct {
+	// External arguments for cloud provider
+	// +kubebuilder:validation:Optional
+	TLSCipherSuites string `json:"tls-cipher-suites"`
+}
+
+type VSphereCPI struct {
+	// The vSphere mode. Either `vsphereCPI` or `vsphereParavirtualCPI`. Default value is `vsphereCPI`
```
blc1996, line 86: I think we probably don't need to include default value in the description. It already has a line follows that specifies default. In generated openAPI schema, the default will also be populated.

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,205 @@
...
+	Datacenter string `json:"datacenter"`
+
+	// Username used to access a vSphere endpoint
+	// +kubebuilder:validation:Required
+	Username string `json:"username"`
+
+	// Password used to access a vSphere endpoint
+	// +kubebuilder:validation:Required
+	Password string `json:"password"`
+
```
blc1996, line 107: Looks like `region` and `zone` are missing

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,205 @@
...
+	ClusterAPIVersion string `json:"clusterAPIVersion"`
+
+	// Used in vsphereParavirtual mode, defines the Cluster kind. Default: Cluster.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=Cluster
+	ClusterKind string `json:"clusterKind"`
+
+	// Used in vsphereParavirtual mode, defines the Cluster name. Default: test-cluster.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=test-cluster
```
blc1996, line 163: Probably no default for this one

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,205 @@
...
+	// +kubebuilder:validation:Optional
+	HTTPSProxy string `json:"https_proxy"`
+
+	// No-proxy setting
+	// +kubebuilder:validation:Optional
+	NoProxy string `json:"no_proxy"`
+
+	// Used in vsphereParavirtual mode, defines the Cluster API versions. Default: run.tanzu.vmware.com/v1alpha2.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=run.tanzu.vmware.com/v1alpha2
```
blc1996, line 153: Please update the default

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,205 @@
...
+
+	// HTTPS proxy setting
+	// +kubebuilder:validation:Optional
+	HTTPSProxy string `json:"https_proxy"`
+
+	// No-proxy setting
+	// +kubebuilder:validation:Optional
+	NoProxy string `json:"no_proxy"`
+
+	// Used in vsphereParavirtual mode, defines the Cluster API versions. Default: run.tanzu.vmware.com/v1alpha2.
```
blc1996, line 151: No default is needed in description

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,205 @@
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=run.tanzu.vmware.com/v1alpha2
+	ClusterAPIVersion string `json:"clusterAPIVersion"`
+
+	// Used in vsphereParavirtual mode, defines the Cluster kind. Default: Cluster.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=Cluster
+	ClusterKind string `json:"clusterKind"`
```
blc1996, line 159: Question: Why we need a param for clusterKind? Wouldn't it always be `Cluster`?

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,205 @@
...
+type NsxT struct {
+	// A flag that enables pod routing
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=false
+	PodRoutingEnabled bool      `json:"podRoutingEnabled"`
+	Routes            NsxTRoute `json:"routes"`
+
+	// The username used to access NSX-T
+	// +kubebuilder:validation:Optional
+	Username string `json:"username"`
```
blc1996, line 32: Validation could be added to `nsxt.username` and `nsxt.password` if nsxt is enabled when implementing the controller.

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+type NsxTRoute struct {
+	// NSX-T T0/T1 logical router path
+	// +kubebuilder:validation:Optional
+	RouterPath string `json:"routerPath"`
+
+	// Cluster CIDR
+	// +kubebuilder:validation:Optional
+	ClusterCidr string `json:"clusterCidr"`
```
blc1996, line 20: This could be a derived field.

addons/controllers/cpi/cpiconfig_controller.go
danniel1205: Suggestion: Do we have a separate issue to track this TODO ? If so, could you help put the issue link as a comment in code ? If not, could you help file one issue and add a ref ?

addons/controllers/cpi/cpiconfig_controller.go
HanFa: yes, I will remove this boilerplate once I finish the controller login in another separate PR.

addons/controllers/cpi/cpiconfig_controller.go
HanFa: We are adding the logic in a separate PR 

Issue created: https://github.com/vmware-tanzu/tanzu-framework/issues/1569

apis/cpi/v1alpha1/cpiconfig_types.go
HanFa: Good catch. Adding it

apis/cpi/v1alpha1/cpiconfig_types.go
HanFa: changed to `cluster.x-k8s.io/v1beta1`

apis/cpi/v1alpha1/cpiconfig_types.go
HanFa: removed in the description.

apis/cpi/v1alpha1/cpiconfig_types.go
HanFa: there might be other CR kind for example `TanzuKubernetesCluster`

apis/cpi/v1alpha1/cpiconfig_types.go
HanFa: sure, drop the validation here for now

apis/cpi/v1alpha1/cpiconfig_types.go
HanFa: removed this field, will derived cluster CIDR from the controller

apis/cpi/v1alpha1/cpiconfig_types.go
HanFa: default is removed

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,204 @@
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=false
+	PodRoutingEnabled bool      `json:"podRoutingEnabled"`
+	Routes            NsxTRoute `json:"routes"`
+
+	// The username used to access NSX-T
+	Username string `json:"username"`
+
+	// The password used to access NSX-T
+	Password string `json:"password"`
```
vijaykatam, line 30: password needs to be a ref to a secret

addons/controllers/cpi/cpiconfig_controller.go
danniel1205: trying to understand a little more on what is blocking us from having above cleaned up within this PR?

HanFa: removed the boilerplate. adds a TODO with issue link.

HanFa:
thanks for the feedback! @danniel1205  @vijaykatam   I’ve
* removed the boilerplate command and put on a TODO with issue link
* change plaintext username/password with credentials secretRef
Could you PTAL?

apis/cpi/v1alpha1/cpiconfig_types.go
HanFa: changed to use secret ref for vsphere & nsxt credentials

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,202 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CPIConfigSpec defines the desired state of CPIConfig
+type CPIConfigSpec struct {
+	VSphereCPI VSphereCPI `json:"vsphereCPI"`
+}
+
+type NsxTRoute struct {
```
lubronzhan, line 16: nit For acronyms, it's better to use same case

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,202 @@
...
+	// +kubebuilder:validation:Optional
+	RouterPath string `json:"routerPath"`
+}
+
+type NsxT struct {
+	// A flag that enables pod routing
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=false
+	PodRoutingEnabled bool      `json:"podRoutingEnabled"`
+	Routes            NsxTRoute `json:"routes"`
```
lubronzhan, line 27: missing comment?

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,202 @@
...
+
+// CPIConfigSpec defines the desired state of CPIConfig
+type CPIConfigSpec struct {
+	VSphereCPI VSphereCPI `json:"vsphereCPI"`
+}
+
+type NsxTRoute struct {
+	// NSX-T T0/T1 logical router path
+	// +kubebuilder:validation:Optional
+	RouterPath string `json:"routerPath"`
```
lubronzhan, line 19: Missing clusterCIDR
https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/vsphere-cpi/1.22.4/bundle/config/values.yaml#L23

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,202 @@
...
+type ExtraArgs struct {
+	// External arguments for cloud provider
+	// +kubebuilder:validation:Optional
+	TLSCipherSuites string `json:"tls-cipher-suites"`
+}
+
+type VSphereCPI struct {
+	// The vSphere mode. Either `vsphereCPI` or `vsphereParavirtualCPI`.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=vsphereCPI;vsphereParavirtualCPI
```
lubronzhan, line 83: TLSThumbprint?

lubronzhan:
@HanFa I assume the CPIConfig should only contain parameters for paravirtual CPI? Do you still have the google doc that we collect all necessary parameters for paravirtual CPI? That should have everything

apis/cpi/v1alpha1/cpiconfig_types.go
HanFa: I think the clusterCIDR can be infered from the cluster by the controller.

lubronzhan:
@blc1996 ^^ Please confirm that we don't need the non-paravirtual fields

vijaykatam:
@lubronzhan we need to continue to support non-paravirtual package until TKGm support goes away.

zjs:
>we need to continue to support non-paravirtual package until TKGm support goes away

To elaborate: there are a few prerequisites for transitioning away from support for TKG on vSphere without the Supervisor, including a deprecation period — which we can't start until we have feature parity (at the earliest). Expect support for the non-paravirtual mode to stick around for a while.

lubronzhan:
Thanks @vijaykatam @zjs.
If we want to support non-paravirtual, do we support using the new API or just old kapp controller. Because this seems a lot of effort if we want to support non-paravirtual using new API. 
For example, for AVI, TKGs has their own way of deploying AKO in supervisor cluster, while they don't deploy AKO in guest cluster. If we want to support the AVI with new Addon API, we will have to implement a new AVIConfig and controller that would never be used in uTKG, which is not worthy IMO

apis/cpi/v1alpha1/cpiconfig_types.go
lubronzhan: Yes, it's inferred. Let's expose it and leave a comment, to make it consistent.

blc1996: Our current practice is to not include a field if it can be inferred. This is meant to minimize the user input for configuration.

HanFa:
will continue work on this PR to add the controller logic

lubronzhan:
You need to create this  providerserviceaccount CR as well for non-paravirtual, like https://gitlab.eng.vmware.com/core-build/guest-cluster-controller/-/blob/master/controllers/addons/addons_controller.go#L487-521
Currently the CRD should live in CAPV

lubronzhan (commented):
From today's meeting I learnt that we will also have CPI for AWS and Azure, so we need to rename current name `CPIConfig` to `VSphereCPIConfig`.
Also file name needs to be changed.
You can do this in follow up PR

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+// It retrieves TLSThumbprint and vSphere server endpoint from VSphereCluster object.
+// It reads vSphere credentials from the cluster secret.
+// It reads the NSX-T credentials from the NSX-T secret.
+// Other values are filled from the specified fields in the CPIConfig CRD.
+func (r *CPIConfigReconciler) mapCPIConfigToDataValuesNonParavirtual(ctx context.Context, cpiConfig *cpiv1alpha1.CPIConfig, cluster *clusterapiv1beta1.Cluster) (*CPIDataValues, error) {
+	d := &CPIDataValues{}
+	c := cpiConfig.Spec.VSphereCPI.NonParavirtualConfig
+	d.VSphereCPI.Mode = CPINonParavirtualMode
+
+	vsphereCluster := &capvv1beta1.VSphereCluster{}
```
lubronzhan, line 186: We get v1beta1 VSphere Cluster here just for getting some common configurations.
Question for addon team, if we need to infer parameters from VSphereCluster, we need to be careful because we don't know about which VSphereCluster version we are deploying with. Is there a better place to get the vsphere credentials?
Otherwise CPI will probably be aware of all the possible VSphereCluster the current CPI is supported.

lubronzhan: VSphere Cluster is only exist in management cluster.
This logic will break in Workload cluster. We still need some secret in workload cluster, or passing the credentials directly to workload cluster

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+		if err != nil {
+			r.Log.Error(err, "Error occurs when mutate CPIConfig data values")
+			return err
+		}
+		yamlBytes, err := yaml.Marshal(cpiConfigSpec)
+		if err != nil {
+			r.Log.Error(err, "Error marshaling CPIConfig to Yaml")
+			return err
+		}
+		dataValueBytes := append([]byte(constants.TKGDataValueFormatString), yamlBytes...)
```
blc1996, line 152: We don't need to include `TKGDataValueFormatString` anymore, as we are not going to add overlay support, and it was used for overlay support.

addons/controllers/cpi/cpiconfig_controller.go
blc1996: Looks like this is not a concern any more?

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+		return r.mapCPIConfigToDataValuesParavirtual(ctx, cpiConfig, cluster)
+	default:
+		break
+	}
+	return nil, errors.Errorf("Invalid CPI mode %s, must either be %s or %s", cpiConfig.Spec.VSphereCPI.Mode, CPIParavirtualMode, CPINonParavirtualMode)
+}
+
+// getOwnerCluster verifies that the CPIConfig has a cluster as its owner reference,
+// and returns the cluster
+func (r *CPIConfigReconciler) getOwnerCluster(ctx context.Context, cpiConfig *cpiv1alpha1.CPIConfig) (*clusterapiv1beta1.Cluster, error) {
```
blc1996, line 287: This function can be moved to cpi_util

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,210 @@
...
+	// No-proxy setting
+	// +kubebuilder:validation:Optional
+	NoProxy string `json:"no_proxy"`
+}
+
+type ParavirtualConfig struct {
+	// Used in vsphereParavirtual mode, defines the Cluster API versions.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=cluster.x-k8s.io/v1beta1
+	ClusterAPIVersion string `json:"clusterAPIVersion"`
```
vijaykatam, line 146: why do you need this variable?

apis/cpi/v1alpha1/cpiconfig_types.go
vijaykatam: Could you describe the use case for this?

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,210 @@
...
+	SecretNamespace string `json:"secretNamespace"`
+}
+
+type ExtraArgs struct {
+	// External arguments for cloud provider
+	// +kubebuilder:validation:Optional
+	TLSCipherSuites string `json:"tls-cipher-suites"`
+}
+
+type NonParavirtualConfig struct {
```
vijaykatam, line 83: NSXT config is the only API spec that should be exposed, rest all should be derived.

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,210 @@
...
+
+	// The namespace of secret that stores CPI configuration
+	// +kubebuilder:validation:Optional
+	SecretNamespace string `json:"secretNamespace"`
+}
+
+type ExtraArgs struct {
+	// External arguments for cloud provider
+	// +kubebuilder:validation:Optional
+	TLSCipherSuites string `json:"tls-cipher-suites"`
```
vijaykatam, line 80: Default this var to current ciphersuites. Also inconsistent json naming. Should be camelCase.

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,210 @@
...
+	// +kubebuilder:validation:Optional
+	Routes *NSXTRoute `json:"routes"`
+
+	// A secret reference that contains Nsx-T login credentials to access NSX-T
+	// consists of the field username and password
+	NSXTCredentialsRef v1.SecretReference `json:"nsxTCredentialsRef"`
+
+	// The NSX-T server
+	// +kubebuilder:validation:Optional
+	Host string `json:"host"`
```
vijaykatam, line 38: I am wondering why this is optional. Can NSXT config work without host info?

apis/cpi/v1alpha1/cpiconfig_types.go
vijaykatam: +1

config/crd/patches/webhook_in_cpiconfigs.yaml
```diff
@@ -0,0 +1,17 @@
+# The following patch enables conversion webhook for CRD
```
vijaykatam, line 1: Remove generated webhook configs

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,210 @@
...
+// CPIConfigStatus defines the observed state of CPIConfig
+type CPIConfigStatus struct {
+	// Name of the data value secret created by cpi controller
+	//+ kubebuilder:validation:Optional
+	SecretRef v1.SecretReference `json:"secretRef,omitempty"`
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+
```
vijaykatam, line 189: Missing shortName, resourcePaths and printcolumns

vijaykatam (dismissed):
Requires considerable changes

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+			clusterName = ownerRef.Name
+			break
+		}
+	}
+
+	if err := r.Client.Get(ctx, types.NamespacedName{Namespace: cpiConfig.Namespace, Name: clusterName}, cluster); err != nil {
+		if apierrors.IsNotFound(err) {
+			r.Log.Info(fmt.Sprintf("Cluster resource '%s/%s' not found", cpiConfig.Namespace, clusterName))
+		}
+		r.Log.Error(err, fmt.Sprintf("Unable to fetch cluster '%s/%s'", cpiConfig.Namespace, clusterName))
```
maralavi, line 302: As you have made ```getOwnerCluster()``` a dedicated function and you are checking for nilness of the cluster here, you do not need to do the same check in the caller. You can return ```nil,nil``` for the case where the cluster is not found and simply check if the returned cluster object is nil in the caller. Otherwise you are duplicating this check in both places

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+
+	cpiConfig = cpiConfig.DeepCopy()
+	cluster, err := r.getOwnerCluster(ctx, cpiConfig)
+	if apierrors.IsNotFound(err) {
+		return ctrl.Result{}, nil
+	} else if err != nil {
+		return ctrl.Result{}, err
+	}
+	if res, err := r.ReconcileCPIConfig(ctx, req, cpiConfig, cluster); err != nil {
```
maralavi, line 80: Why do you need to pass ```req``` to this function?

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+			r.Log.Error(err, "Error patching CPIConfig")
+			retErr = err
+		}
+		r.Log.Info("Successfully patched CPIConfig")
+	}()
+
+	if !cpiConfig.GetDeletionTimestamp().IsZero() {
+		return r.ReconcileCPIConfigDelete(ctx, req, cpiConfig, cluster)
+	}
+	res, err = r.ReconcileCPIConfigNormal(ctx, req, cpiConfig, cluster)
```
maralavi, line 106: Don't need to pass ```req``` as argument to these functions. Please remove

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+	if err != nil {
+		r.Log.Error(err, "Error reconciling CPIConfig to create/patch data values secret")
```
maralavi, line 108: You should return ````ctrl.Result{}, err``` here

addons/controllers/cpi/cpiconfig_controller.go
maralavi, line 106: Lets be consistent with other config objects, remove the return type ```ctrl.Result``` from the return value of ```ReconcileCPIConfigNormal```.

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	r.Log.Info("Successfully reconciled CPIConfig")
```
maralavi, line 110: The way this code is written, it will log both error and successful message in case of ```err```

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+		return nil
+	}
+	result, err := controllerutil.CreateOrPatch(ctx, r.Client, secret, mutateFn)
+	if err != nil {
+		r.Log.Error(err, "Error creating or patching CPIConfig data values secret")
+		return ctrl.Result{}, err
+	}
+	r.Log.Info(fmt.Sprintf("Resource '%s' data values secret '%s'", constants.CPIAddonName, result))
+	// update the secret reference in CPIConfig status
+	cpiConfig.Status.SecretRef = v1.SecretReference{
```
maralavi, line 163: ```SecretRef``` should just be a string holding the name. Please see other config objects as the reference

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+		Name:      secret.Name,
+		Namespace: secret.Namespace,
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// ReconcileCPIConfigDelete triggers when a CPIConfig is deleting, currently do nothing
+func (r *CPIConfigReconciler) ReconcileCPIConfigDelete(context.Context, ctrl.Request, *cpiv1alpha1.CPIConfig, *clusterapiv1beta1.Cluster) (res ctrl.Result, retErr error) {
+	return ctrl.Result{}, nil
```
maralavi, line 173: Please just return this on the caller and remove this function

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+	d.VSphereCPI.ClusterName = c.ClusterName
+	d.VSphereCPI.ClusterUID = c.ClusterUID
+	d.VSphereCPI.SupervisorMasterEndpointIP = c.SupervisorMasterEndpointIP
+	d.VSphereCPI.SupervisorMasterPort = c.SupervisorMasterPort
+
+	return d, nil
+}
+
+// mapCPIConfigToDataValues maps CPIConfig CR to data values
+func (r *CPIConfigReconciler) mapCPIConfigToDataValues(ctx context.Context, cpiConfig *cpiv1alpha1.CPIConfig, cluster *clusterapiv1beta1.Cluster) (*CPIDataValues, error) {
```
maralavi, line 273: please move these logic to ```cpiconfig_utils.go```

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+		r.Log.Info("Patching CPIConfig")
+		if err := patchHelper.Patch(ctx, cpiConfig); err != nil {
+			r.Log.Error(err, "Error patching CPIConfig")
+			retErr = err
+		}
+		r.Log.Info("Successfully patched CPIConfig")
+	}()
+
+	if !cpiConfig.GetDeletionTimestamp().IsZero() {
+		return r.ReconcileCPIConfigDelete(ctx, req, cpiConfig, cluster)
```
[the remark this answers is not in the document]
HanFa: if it's not used in addon lifecycle management, I will remove it =)

apis/cpi/v1alpha1/cpiconfig_types.go
lubronzhan: It should be required https://github.com/kubernetes/cloud-provider-vsphere/blob/master/pkg/nsxt/config/config_yaml.go#L80

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+			return ctrl.Result{}, nil
+		}
+		r.Log.Error(err, "Unable to fetch CPIConfig resource")
+		return ctrl.Result{}, err
+	}
+
+	cpiConfig = cpiConfig.DeepCopy()
+	cluster, err := r.getOwnerCluster(ctx, cpiConfig)
+	if apierrors.IsNotFound(err) {
+		return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
HanFa: sure

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+	} else if err != nil {
+		return ctrl.Result{}, err
```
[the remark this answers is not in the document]
HanFa: sure

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+	r.Log = r.Log.WithValues("CPIConfig", req.NamespacedName)
+
+	r.Log.Info("Start reconciliation for CPIConfig")
+
+	// fetch CPIConfig resource
+	cpiConfig := &cpiv1alpha1.CPIConfig{}
+	if err := r.Client.Get(ctx, req.NamespacedName, cpiConfig); err != nil {
+		if apierrors.IsNotFound(err) {
+			r.Log.Info("CPIConfig resource not found")
+			r.Log.Info("CPIConfig resource not found")
```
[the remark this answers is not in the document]
HanFa: removed

apis/cpi/v1alpha1/cpiconfig_types.go
HanFa: so here I am not including `clusterCIDR`, which can be inferred from `cluster.Spec.ClusterNetwork.Pods.CIDRBlocks[0]`

apis/cpi/v1alpha1/cpiconfig_types.go
lubronzhan: This is used for ownerRef
Do we still need to support TKGs? They use `TanzuKubernetesCluster` to deploy Cluster. hm, But in that case, they still has Cluster object.
I guess it's just for better visibility when using `TKC`. But yeah, we can also directly refer to Cluster

addons/controllers/cpiconfig_controller_test.go
```diff
@@ -0,0 +1,158 @@
...
+	cpiv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/cpi/v1alpha1"
+
+	clusterapiv1beta1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/vmware-tanzu/tanzu-framework/addons/testutil"
+)
```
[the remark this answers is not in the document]
HanFa: fixed

config/crd/patches/cainjection_in_cpiconfigs.yaml
```diff
@@ -0,0 +1,8 @@
+# The following patch adds a directive for certmanager to inject CA into the CRD
```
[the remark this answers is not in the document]
HanFa: file removed

addons/controllers/cpi/cpiconfig_utils.go
```diff
@@ -0,0 +1,37 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func getSecret(ctx context.Context, c client.Client, namespace, name string) (*v1.Secret, error) {
+	secret := &v1.Secret{}
+	if err := c.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, secret); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, errors.Errorf("Secret %s/%s not found", namespace, name)
+		}
+		return nil, errors.Errorf("Secret %s/%s cannot be get, error %v", namespace, name, err)
```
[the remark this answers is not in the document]
HanFa: fixed

addons/controllers/cpi/cpiconfig_controller.go
HanFa: removed

addons/controllers/cpi/cpiconfig_controller.go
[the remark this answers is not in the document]
HanFa: the log is handled inside getOwnerCluster

addons/controllers/cpi/cpiconfig_controller.go
HanFa: done

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+		Type: v1.SecretTypeOpaque,
+	}
+	secret.SetOwnerReferences([]metav1.OwnerReference{ownerReference})
+
+	mutateFn := func() error {
+		secret.Data = make(map[string][]byte)
+		cpiConfigSpec, err := r.mapCPIConfigToDataValues(ctx, cpiConfig, cluster)
+		r.Log.Info("Mutate CPIConfig data values", "data values", cpiConfigSpec)
+		if err != nil {
+			r.Log.Error(err, "Error occurs when mutate CPIConfig data values")
```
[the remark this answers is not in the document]
HanFa: changed to `Error occurs when mapping CPIConfig to data values`

addons/controllers/cpi/cpiconfig_controller.go
HanFa: removed

apis/cpi/v1alpha1/cpiconfig_types.go
HanFa: fixed

config/crd/patches/webhook_in_cpiconfigs.yaml
HanFa: removed

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,210 @@
...
+	// +kubebuilder:default:=false
+	PodRoutingEnabled bool `json:"podRoutingEnabled"`
+
+	// Route configuration for NSXT
+	// +kubebuilder:validation:Optional
+	Routes *NSXTRoute `json:"routes"`
+
+	// A secret reference that contains Nsx-T login credentials to access NSX-T
+	// consists of the field username and password
+	NSXTCredentialsRef v1.SecretReference `json:"nsxTCredentialsRef"`
```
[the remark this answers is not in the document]
HanFa: fixed

addons/controllers/cpi/cpiconfig_controller.go
HanFa: moved

addons/controllers/cpi/cpiconfig_controller.go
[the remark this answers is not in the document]
HanFa: removed

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+// Package controllers implements k8s controller functionality for vsphere-cpi.
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	clusterapiutil "sigs.k8s.io/cluster-api/util"
```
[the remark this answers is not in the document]
HanFa: fixed

addons/controllers/cpi/cpiconfig_controller.go
HanFa: fixed

addons/controllers/cpi/cpiconfig_controller.go
HanFa: removed

addons/controllers/cpi/cpiconfig_controller.go
HanFa: fixed

addons/controllers/cpi/cpiconfig_controller.go
HanFa: fixed

addons/controllers/cpi/cpiconfig_controller.go
HanFa: fixed

addons/controllers/cpi/cpiconfig_controller.go
HanFa: fixed

addons/controllers/cpi/cpiconfig_controller.go
HanFa: fixed

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	return nil, errors.Errorf("Invalid CPI mode %s, must either be %s or %s", cpiConfig.Spec.VSphereCPI.Mode, CPIParavirtualMode, CPINonParavirtualMode)
+}
+
+// getOwnerCluster verifies that the CPIConfig has a cluster as its owner reference,
+// and returns the cluster
+func (r *CPIConfigReconciler) getOwnerCluster(ctx context.Context, cpiConfig *cpiv1alpha1.CPIConfig) (*clusterapiv1beta1.Cluster, error) {
+	cluster := &clusterapiv1beta1.Cluster{}
+	clusterName := cpiConfig.Name
+
```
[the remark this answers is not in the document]
HanFa: added comments for readability

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+			r.Log.Info("CPIConfig resource not found")
+			return ctrl.Result{}, nil
+		}
+		r.Log.Error(err, "Unable to fetch CPIConfig resource")
+		return ctrl.Result{}, err
+	}
+
+	cpiConfig = cpiConfig.DeepCopy()
+	cluster, err := r.getOwnerCluster(ctx, cpiConfig)
+	if apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
HanFa: fixed, removed the duplicated checks

apis/cpi/v1alpha1/cpiconfig_types.go
HanFa: from the discussion on Friday:

for NSXT related configurations, it should be optional at the API level but required for the CPI package. If user doesn't specify NSXT in the CPIConfig, the controller should read it from the cluster topology, defined at https://github.com/vmware-tanzu/tanzu-framework/blob/fc958cc58c52fd8a178135541fce70ccf39276eb/addons/controllers/testdata/test-cluster-bootstrap-1.yaml#L31

@vijaykatam @blc1996  could you plz confirm?

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,314 @@
...
+}
+
+// getOwnerCluster verifies that the CPIConfig has a cluster as its owner reference,
+// and returns the cluster
+func (r *CPIConfigReconciler) getOwnerCluster(ctx context.Context, cpiConfig *cpiv1alpha1.CPIConfig) (*clusterapiv1beta1.Cluster, error) {
+	cluster := &clusterapiv1beta1.Cluster{}
+	clusterName := cpiConfig.Name
+
+	for _, ownerRef := range cpiConfig.GetOwnerReferences() {
+		if ownerRef.Kind == "Cluster" {
```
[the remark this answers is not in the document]
HanFa: fixed

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,210 @@
...
+	*NonParavirtualConfig `json:""`
+
+	*ParavirtualConfig `json:""`
+}
+
+// CPIConfigStatus defines the observed state of CPIConfig
+type CPIConfigStatus struct {
+	// Name of the data value secret created by cpi controller
+	//+ kubebuilder:validation:Optional
+	SecretRef v1.SecretReference `json:"secretRef,omitempty"`
```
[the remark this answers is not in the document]
HanFa: fixed

apis/cpi/v1alpha1/cpiconfig_types.go
vijaykatam: Technically CPI doesn't need NSXT, it is only required for generating data value when NSXT config is provided via cluster variables.

apis/cpi/v1alpha1/cpiconfig_types.go
vijaykatam: We discussed this config deriving from control plane machine template if not provided via user input.

HanFa:
I've made some changes on this PR.

* derive more fields from cluster variables or annotations
* override the derived values if the optional fields on CPIConfig are specified
* add shortname, columnname and resource path for the CPIConfig CR

Could you PTAL at your convenience? Thanks. @vijaykatam @blc1996 @lubronzhan

addons/controllers/testdata/infrastructure_vspherecluster.yaml
```diff
@@ -0,0 +1,1723 @@
+apiVersion: apiextensions.k8s.io/v1
```
blc1996, line 1: Looks like this is external CRD. You can try to add additional CRDs in https://github.com/vmware-tanzu/tanzu-framework/blob/main/addons/controllers/suite_test.go#L92, so you don't need to port in the whole file.

addons/controllers/testdata/test-cpi-paravirtual.yaml
```diff
@@ -0,0 +1,26 @@
...
+---
+apiVersion: cpi.tanzu.vmware.com/v1alpha1
+kind: CPIConfig
+metadata:
+  name: test-cluster-cpi-paravirtual
+  namespace: default
+spec:
+  vsphereCPI:
+    mode: vsphereParavirtualCPI
+    clusterAPIVersion: "cluster.x-k8s.io/v1beta1"
```
blc1996, line 21: Have you found the way to consolidate these paravirtual related field as derived fields?

addons/controllers/testdata/test-cpi.yaml
```diff
@@ -0,0 +1,109 @@
+apiVersion: cluster.x-k8s.io/v1beta1
```
blc1996, line 1: Can we rename this file to something like `test-cpi-nonparavirtual.yaml` that corresponds to `test-cpi-paravirtual.yaml` above?

addons/pkg/util/cluster_variable_util.go
```diff
@@ -11,33 +11,44 @@ import (
...
+	_, result, err := parseClusterVariable(cluster, variableName)
+	return result, err
+}
+
 func ParseClusterVariableString(cluster *clusterapiv1beta1.Cluster, variableName string) (string, error) {
+	result, _, err := parseClusterVariable(cluster, variableName)
+	return result, err
+}
+
+func parseClusterVariable(cluster *clusterapiv1beta1.Cluster, variableName string) (string, bool, error) {
```
blc1996, line 24: Wondering if we could use interface{} type for return value, and use reflect to handle it?

blc1996: Having multiple return types looks a bit weird.

apis/cpi/v1alpha1/cpiconfig_types.go
blc1996: `NSXT_MANAGER_HOST` should be available in cluster variable

apis/cpi/v1alpha1/cpiconfig_types.go
```diff
@@ -0,0 +1,227 @@
...
+	PodRoutingEnabled bool `json:"podRoutingEnabled"`
+
+	// Route configuration for NSXT
+	// +kubebuilder:validation:Optional
+	Routes *NSXTRoute `json:"routes"`
+
+	// A secret reference that contains Nsx-T login credentials to access NSX-T
+	// consists of the field username and password
+	// +kubebuilder:validation:Optional
+	NSXTCredentialsRef *v1.SecretReference `json:"nsxtCredentialsRef"`
```
blc1996, line 39: Question: Who creates this credential secret?

addons/controllers/suite_test.go
```diff
@@ -130,6 +134,12 @@ var _ = BeforeSuite(func(done Done) {
 	err = cniv1alpha1.AddToScheme(scheme)
 	Expect(err).NotTo(HaveOccurred())
 
+	err = cpiv1alpha1.AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	err = capvv1beta1.AddToScheme(scheme)
```
blc1996, line 140: Cluster-API resources are already added in line 96

addons/controllers/cpi/cpiconfig_utils.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+// controlPlaneName returns the control plane name for a cluster name
+func controlPlaneName(clusterName string) string {
+	return fmt.Sprintf("%s-control-plane", clusterName)
+}
+
+// tryParseClusterVariableBool tries to parse a boolean cluster variable,
+// info any error that occurs
+func (r *CPIConfigReconciler) tryParseClusterVariableBool(_ context.Context, cluster *clusterapiv1beta1.Cluster, variableName string) bool {
```
blc1996, line 291: Why include `context.Context` in the param list but it's not actually used?

addons/controllers/cpi/cpiconfig_utils.go
```diff
@@ -0,0 +1,307 @@
...
+	res, err := util.ParseClusterVariableBool(cluster, variableName)
+	if err != nil {
+		r.Log.Info(fmt.Sprintf("cannot parse cluster variable with key %s", variableName))
+	}
+	return res
+}
+
+// tryParseClusterVariableString tries to parse a string cluster variable,
+// info any error that occurs
+func (r *CPIConfigReconciler) tryParseClusterVariableString(_ context.Context, cluster *clusterapiv1beta1.Cluster, variableName string) string {
```
blc1996, line 301: Same question as above.

addons/controllers/cpi/cpiconfig_utils.go
```diff
@@ -0,0 +1,307 @@
...
+	if err := c.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, secret); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, errors.Errorf("Secret %s/%s not found", namespace, name)
+		}
+		return nil, errors.Errorf("Secret %s/%s could not be fetched, error %v", namespace, name, err)
+	}
+	return secret, nil
+}
+
+func getUsernameAndPasswordFromSecret(_ context.Context, _ client.Client, s *v1.Secret) (string, string, error) {
```
blc1996, line 272: Why we have `_ context.Context, _ client.Client` but they are not used?

addons/controllers/cpi/cpiconfig_utils.go
```diff
@@ -0,0 +1,307 @@
...
+	}, vsphereCluster); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, errors.Errorf("VSphereCluster %s/%s not found", cluster.Namespace, cluster.Name)
+		}
+		return nil, errors.Errorf("VSphereCluster %s/%s could not be fetched, error %v", cluster.Namespace, cluster.Name, err)
+	}
+	return vsphereCluster, nil
+}
+
+func getSecret(ctx context.Context, c client.Client, namespace, name string) (*v1.Secret, error) {
```
blc1996, line 261: Question: Why getVSphereCluster can be a receiver function of `CPIConfigReconciler`, while getSecret need to be a standalone function?

addons/controllers/cpi/cpiconfig_utils.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+// mapCPIConfigToDataValuesParavirtual generates CPI data values for paravirtual modes
+func (r *CPIConfigReconciler) mapCPIConfigToDataValuesParavirtual(_ context.Context, cpiConfig *cpiv1alpha1.CPIConfig, _ *clusterapiv1beta1.Cluster) (*CPIDataValues, error) {
+	d := &CPIDataValues{}
+	c := cpiConfig.Spec.VSphereCPI
+
+	d.VSphereCPI.Mode = CPIParavirtualMode
+	d.VSphereCPI.ClusterAPIVersion = c.ClusterAPIVersion
+	d.VSphereCPI.ClusterKind = c.ClusterKind
```
blc1996, line 199: question: Can these non-paravirtual fields be derived?

addons/controllers/cpi/cpiconfig_utils.go
```diff
@@ -0,0 +1,307 @@
...
+	// derive IP family or proxy related settings from cluster annotations
+	if cluster.Annotations != nil {
+		d.VSphereCPI.IPFamily = cluster.Annotations[pkgtypes.IPFamilyConfigAnnotation]
+		d.VSphereCPI.HTTPProxy = cluster.Annotations[pkgtypes.HTTPProxyConfigAnnotation]
+		d.VSphereCPI.HTTPSProxy = cluster.Annotations[pkgtypes.HTTPSProxyConfigAnnotation]
+		d.VSphereCPI.NoProxy = cluster.Annotations[pkgtypes.NoProxyConfigAnnotation]
+	}
+
+	// derive nsxt related configs from cluster variable
+	d.VSphereCPI.Nsxt.PodRoutingEnabled = r.tryParseClusterVariableBool(ctx, cluster, "NSXT_POD_ROUTING_ENABLED")
```
blc1996, line 84: Can we define these cluster variable names as constants and store somewhere?

addons/controllers/cpi/cpiconfig_utils.go
```diff
@@ -0,0 +1,307 @@
...
+	pkgtypes "github.com/vmware-tanzu/tanzu-framework/addons/pkg/types"
+	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/util"
+	cpiv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/cpi/v1alpha1"
+)
+
+// mapCPIConfigToDataValuesNonParavirtual generates CPI data values for non-paravirtual modes
+func (r *CPIConfigReconciler) mapCPIConfigToDataValuesNonParavirtual( // nolint
+	ctx context.Context,
+	cpiConfig *cpiv1alpha1.CPIConfig, cluster *clusterapiv1beta1.Cluster) (*CPIDataValues, error,
+) { // nolint:whitespace
```
blc1996, line 29: Better not to use `nolint` unless we have no other choice

addons/controllers/cpi/cpiconfig_datavalues.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+type CPIDataValuesVsphereCPI struct {
+	Mode                               string      `yaml:"mode"`
+	TLSThumbprint                      string      `yaml:"tlsThumbprint"`
+	Server                             string      `yaml:"server"`
+	Datacenter                         string      `yaml:"datacenter"`
+	Username                           string      `yaml:"username"`
+	Password                           string      `yaml:"password"`
+	Region                             interface{} `yaml:"region"`
```
blc1996, line 18: Question: What's the reason of using `interface{}` here?

addons/controllers/cpi/cpiconfig_utils.go
HanFa: it will complain about that the cyclomatic complexity for this function is too large.

addons/controllers/cpi/cpiconfig_datavalues.go
HanFa: to match the package value definition here https://github.com/vmware-tanzu/community-edition/blob/8e903a7fe266432e6c5515441ecf56cfdcb9dc76/addons/packages/vsphere-cpi/1.22.5/bundle/config/values.yaml#L11

addons/controllers/cpi/cpiconfig_utils.go
HanFa: I will put them in `cpiconfig_const.go`

addons/controllers/cpi/cpiconfig_datavalues.go
blc1996: I think it's fine to use string type even the default values.yaml had it as null. Did you see any error when using string?

addons/controllers/cpi/cpiconfig_utils.go
HanFa: for the paravirtual mode, yes they should be derived and at least have a default value.

I would suggest do that in a separate PR since it involves other changes as well. For example, the CPIConfig controller need to generate accounts & rbac rules etc. https://gitlab.eng.vmware.com/core-build/guest-cluster-controller/-/blob/master/controllers/addons/addons_controller.go#L486-522

addons/controllers/cpi/cpiconfig_utils.go
HanFa: let's remove them. it was just for the placeholder

addons/controllers/suite_test.go
HanFa: oh I am adding the scheme for cpi and capv here. not CAPI   =)

apis/cpi/v1alpha1/cpiconfig_types.go
HanFa: API user can optionally provider this secret. If they do so, it will override the username and password that are derived from the cluster variable.

blc1996: OK, we probably should have it documented somewhere. Otherwise people would be confused about the usage of these fields.

addons/controllers/cpi/cpiconfig_utils.go
HanFa: change it to receiver function as well, to be consistent ...

addons/controllers/cpi/cpiconfig_utils.go
HanFa: removed

addons/controllers/cpi/cpiconfig_utils.go
HanFa: removed

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,159 @@
...
+			Namespace: cpiConfig.Namespace,
+		},
+		Type: v1.SecretTypeOpaque,
+	}
+	secret.SetOwnerReferences([]metav1.OwnerReference{ownerReference})
+
+	mutateFn := func() error {
+		secret.Data = make(map[string][]byte)
+		cpiConfigSpec, err := r.mapCPIConfigToDataValues(ctx, cpiConfig, cluster)
+		r.Log.Info("Mutate CPIConfig data values", "data values", cpiConfigSpec)
```
maralavi, line 129: This log message is out of place here

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,159 @@
...
+		if err != nil {
+			r.Log.Error(err, "Error occurs when mapping CPIConfig to data values")
```
maralavi, line 131:
```suggestion
r.Log.Error(err, "Error while mapping CPIConfig to data values")
```

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,159 @@
...
+
+	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/constants"
+	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/util"
+	cpiv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/cpi/v1alpha1"
+)
+
+const (
+	CPINonParavirtualMode = "vsphereCPI"
+	CPIParavirtualMode    = "vsphereParavirtualCPI"
+)
```
maralavi, line 33: Could you please define these constants in cpiconfig_utils.go where the are used?

addons/pkg/util/cluster_variable_util.go
HanFa: we can use golang reflection to handle that

apis/cpi/v1alpha1/cpiconfig_types.go
HanFa: agreed

apis/cpi/v1alpha1/cpiconfig_types.go
HanFa: yes, CPI is currently deriving it from `NSXT_MANAGER_HOST` cluster variable. This CPIConfig field is merely a way for API user to override the derived field.

addons/controllers/cpi/cpiconfig_utils.go
```diff
@@ -0,0 +1,307 @@
...
+func (r *CPIConfigReconciler) tryParseClusterVariableBool(_ context.Context, cluster *clusterapiv1beta1.Cluster, variableName string) bool {
+	res, err := util.ParseClusterVariableBool(cluster, variableName)
+	if err != nil {
+		r.Log.Info(fmt.Sprintf("cannot parse cluster variable with key %s", variableName))
+	}
+	return res
+}
+
+// tryParseClusterVariableString tries to parse a string cluster variable,
+// info any error that occurs
```
maralavi, line 300: Why do we need to have a separate ```tryParseClusterVariableString()``` function where we could directly call ```util.ParseClusterVariableString()```?

addons/pkg/util/cluster_variable_util.go
maralavi: I agree on this

addons/controllers/cpi/cpiconfig_utils.go
blc1996: A separate PR sounds ok to me. cc @vijaykatam

addons/controllers/cpi/cpiconfig_datavalues.go
HanFa: no, I didn't observe any error for now. Let me change it to string

addons/controllers/cpi/cpiconfig_controller.go
HanFa: sure

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,159 @@
...
+		return ctrl.Result{}, err
+	}
+
+	cpiConfig = cpiConfig.DeepCopy()
+	cluster, err := r.getOwnerCluster(ctx, cpiConfig)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if res, err := r.ReconcileCPIConfig(ctx, cpiConfig, cluster); err != nil {
+		r.Log.Error(err, "Fail to reconcile CPIConfig")
```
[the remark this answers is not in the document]
HanFa: fixed

addons/controllers/cpi/cpiconfig_utils.go
HanFa: don't wanna make the function bloated with too many `if ... then print`. let's keep it this way.

addons/controllers/cpi/cpiconfig_controller.go
HanFa: let's move it few lines below

addons/controllers/cpi/cpiconfig_controller.go
HanFa: changed

addons/controllers/cpi/cpiconfig_utils.go
```diff
@@ -0,0 +1,307 @@
...
+	// retrieve the owner cluster for the CPIConfig object
+	for _, ownerRef := range cpiConfig.GetOwnerReferences() {
+		if strings.EqualFold(ownerRef.Kind, constants.ClusterKind) {
+			clusterName = ownerRef.Name
+			break
+		}
+	}
+	if err := r.Client.Get(ctx, types.NamespacedName{Namespace: cpiConfig.Namespace, Name: clusterName}, cluster); err != nil {
+		if apierrors.IsNotFound(err) {
+			r.Log.Info(fmt.Sprintf("Cluster resource '%s/%s' not found", cpiConfig.Namespace, clusterName))
```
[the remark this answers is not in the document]
HanFa: sure, print a log it instead if cluster is not found

addons/controllers/testdata/test-cpi-paravirtual.yaml
HanFa: let's consolidate it in a separate PR, before we making this one even larger =)

addons/controllers/cpi/cpiconfig_datavalues.go
HanFa: changed to string

addons/pkg/util/cluster_variable_util.go
HanFa: changed, added a unit test for boolean as well, could you plz check? @maralavi

addons/controllers/testdata/test-cpi.yaml
HanFa: changed

addons/controllers/testdata/infrastructure_vspherecluster.yaml
HanFa: that's a good idea. I've removed the CR file and it works.

addons/controllers/cpi/cpiconfig_utils.go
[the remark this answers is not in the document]
maralavi: @HanFa I believe this is still not fixed. We should return ```nil``` instead of err when the cluster is ```NotFound```, cause we do not want to requeue again for reconciliation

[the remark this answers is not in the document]
maralavi: Thanks for having fixed

blc1996 (approved):
LGTM if we want to merge, and have a follow up MR to address paravirtual configs.

maralavi (dismissed):

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -61,11 +56,11 @@ func (r *CPIConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 	cpiConfig = cpiConfig.DeepCopy()
 	cluster, err := r.getOwnerCluster(ctx, cpiConfig)
-	if err != nil {
-		return ctrl.Result{}, err
+	if cluster == nil {
+		return ctrl.Result{}, err // no need to requeue if cluster is not found
```
maralavi, line 60: Oops, error is left unchecked here

HanFa: it is checked. If err is not nil, then the returned cluster  will be nil.

maralavi: hmm, you seem to just check the cluster for nilness. Am I missing something?
```
	cluster, err := r.getOwnerCluster(ctx, cpiConfig)
	if cluster == nil {
		return ctrl.Result{}, err // no need to requeue if cluster is not found
	}
	if res, err := r.ReconcileCPIConfig(ctx, cpiConfig, cluster); err != nil {
		r.Log.Error(err, "Failed to reconcile CPIConfig")
```

HanFa: `getOwnerCluster` ensures that it returns cluster as nil whenever

* the cluster is not found
* other error occurs

So by checking the cluster's nilness, we are handling both scenarios above.

maralavi (approved):

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+		r.Log.Info("Successfully patched CPIConfig")
+	}()
+
+	if !cpiConfig.GetDeletionTimestamp().IsZero() {
+		return ctrl.Result{}, nil
+	}
+	if err = r.ReconcileCPIConfigNormal(ctx, cpiConfig, cluster); err != nil {
+		r.Log.Error(err, "Error reconciling CPIConfig to create/patch data values secret")
+	} else {
```
lubronzhan, line 90: This else is not needed if you just return in line 89

addons/controllers/cpi/cpiconfig_controller.go
```diff
@@ -0,0 +1,154 @@
...
+		r.Log.Info("Successfully reconciled CPIConfig")
+	}
+	return ctrl.Result{}, err
+}
+
+// ReconcileCPIConfigNormal triggers when a CPIConfig is not being deleted
+// it ensures the owner reference of the CPIConfig and generates the data values secret for CPI
+func (r *CPIConfigReconciler) ReconcileCPIConfigNormal(ctx context.Context,
```
lubronzhan, line 98: If it's only used as a helper function, it can start with lowercase

addons/controllers/cpi/cpiconfig_controller.go
HanFa: I think it will be needed, since r.Log.Error doesn't return. Am I missing something?

addons/controllers/cpi/cpiconfig_controller.go
HanFa: Here CPIConfig controller is trying to be consistent with AntreaConfig controller. They have function such as `ReconcileAntreaConfigNormal`

config/rbac/cpiconfig_editor_role.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - cpi.tanzu.vmware.com
+  resources:
+  - cpiconfigs/status
+  verbs:
```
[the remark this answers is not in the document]
HanFa: no, they are auto-generated by the kubebuilder. Other config-editor role has the same rbac, such as antreaconfig_editor https://github.com/vmware-tanzu/tanzu-framework/blob/c20eafff0b08f9a52447edeff2b8ecd8c1fedd36/config/rbac/antreaconfig_editor_role.yaml#L24

Not sure why they mismatch with the kubebuilder tag.

addons/controllers/cpi/cpiconfig_utils.go
```diff
@@ -0,0 +1,308 @@
...
+		return nil, err
+	}
+
+	// get the control plane machine template
+	cpMachineTemplate := &capvv1beta1.VSphereMachineTemplate{}
+	if err := r.Client.Get(ctx, types.NamespacedName{
+		Namespace: cluster.Namespace,
+		Name:      controlPlaneName(cluster.Name),
+	},
+		cpMachineTemplate); err != nil {
```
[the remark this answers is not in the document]
HanFa: hmm, will fix that. not sure how it passes the linting..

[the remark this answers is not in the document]
HanFa: fixed

addons/controllers/cpi/cpiconfig_controller.go
lubronzhan: I mean if you add a return function here that's a more common pattern

HanFa: will change

addons/controllers/cpi/cpiconfig_controller.go
HanFa: will change

addons/controllers/cpi/cpiconfig_controller.go
HanFa: fixed

addons/controllers/cpi/cpiconfig_controller.go
HanFa: fixed

addons/controllers/cpi/cpiconfig_utils.go
```diff
@@ -0,0 +1,307 @@
...
+	// derive IP family or proxy related settings from cluster annotations
+	if cluster.Annotations != nil {
+		d.VSphereCPI.IPFamily = cluster.Annotations[pkgtypes.IPFamilyConfigAnnotation]
+		d.VSphereCPI.HTTPProxy = cluster.Annotations[pkgtypes.HTTPProxyConfigAnnotation]
+		d.VSphereCPI.HTTPSProxy = cluster.Annotations[pkgtypes.HTTPSProxyConfigAnnotation]
+		d.VSphereCPI.NoProxy = cluster.Annotations[pkgtypes.NoProxyConfigAnnotation]
+	}
+
+	// derive nsxt related configs from cluster variable
+	d.VSphereCPI.Nsxt.PodRoutingEnabled = r.tryParseClusterVariableBool(cluster, NsxtPodRoutingEnabledVarName)
```
vijaykatam, line 82: Not a blocker for this change but we should follow up in a separate change. I had a sync with @tvs and @sidharthsurana and we agreed that cluster variables should be used only when there is an impact to the topology of cluster or cluster providers but not configs related to addons. Proxy and certs for registry are the only ones so far that permeates both cluster topology and addon configs

vijaykatam (approved):
No blockers to get this merged but we need to follow up on cluster variable usage and deriving paravirt configs.

HanFa:
next steps:
* change cluster variable usage: cluster variables should be used only when there is an impact to the topology of cluster or cluster providers but not configs related to addons
* derive paravirt fields
* reconcile ProviderServiceAccount for paravirt https://gitlab.eng.vmware.com/core-build/guest-cluster-controller/-/blob/master/controllers/addons/addons_controller.go#L486

PR #1555, opened 2022-01-26
Convert CalicoConfig API to data values secret and add CalicoConfig c…

### LINKED ISSUE #1577: CalicoConfig API and Controller

maralavi:
Co-authored-by: Lucheng Bao<luchengb@vmware.com>

This PR converts CalicoConfig API spec to data values secret and adds CalicoConfig controller

Fixes: #1577

Added kubebuilder suite tests as part of the PR

### Release note

```release-note
package-based-lcm: convert CalicoConfig API spec to data values secret and add CalicoConfig controller
```

### PR Checklist


addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -6,16 +6,34 @@ package controllers
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/constants"
+	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/util"
 	cniv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/cni/v1alpha1"
 )
 
-// CalicoConfigReconciler reconciles a CalicoConfig object
+const contextTimeout = 30 * time.Second
```
danniel1205, line 34: question: could this timeout shared by all config controllers or it is `CalicoConfig` only ?

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -27,15 +45,39 @@ type CalicoConfigReconciler struct {
...
+	ctxCancel, cancel := context.WithTimeout(ctx, contextTimeout)
+	defer cancel()
+	log := r.Log.WithValues("CalicoConfig", req.NamespacedName)
+
+	log.Info("Start reconciliation")
+
+	// fetch CalicoConfig resource, ignore not-found errors
+	calicoConfig := &cniv1alpha1.CalicoConfig{}
+	if err := r.Client.Get(ctxCancel, req.NamespacedName, calicoConfig); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
```
danniel1205, line 58: question: would it be a good idea to do some logging here ? For example, if `calicoConfig` is not found, we would have a better idea on debugging.

addons/controllers/calico/calicoconfig_controller.go
vijaykatam: Do we need a timeout for controllers if all this is doing an k8s API request call. I think context passed by manager should be enough as is as it supports cancellation on `SIGHUP` or `SIGKILL`

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -27,15 +45,39 @@ type CalicoConfigReconciler struct {
...
+	}
+
+	// config resources are expected to have the same name as the cluster. However, we ideally try to read the cluster name from the owner reference of the addon config object
```
danniel1205, line 61: thumb-up: Nice comments.

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -27,15 +45,39 @@ type CalicoConfigReconciler struct {
...
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
-// TODO(user): Modify the Reconcile function to compare the state specified by
-// the CalicoConfig object against the actual cluster state, and then
-// perform operations to make the cluster state reflect the state specified by
-// the user.
 func (r *CalicoConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = context.Background()
-	_ = r.Log.WithValues("calicoconfig", req.NamespacedName)
+	ctxCancel, cancel := context.WithTimeout(ctx, contextTimeout)
```
danniel1205, line 49: nit: will `ctxWithTimeout` a better name ?

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -27,15 +45,39 @@ type CalicoConfigReconciler struct {
...
+		if ownerRef.APIVersion == clusterapiv1beta1.GroupVersion.String() && ownerRef.Kind == "Cluster" {
+			clusterName = ownerRef.Name
+			break
+		}
+	}
+
+	// verify that the cluster with the same name is present
+	cluster := &clusterapiv1beta1.Cluster{}
+	if err := r.Client.Get(ctxCancel, types.NamespacedName{Namespace: req.Namespace, Name: clusterName}, cluster); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
```
danniel1205, line 73: question: do we want to do some logging here?

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -46,3 +88,181 @@ func (r *CalicoConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		Complete(r)
 }
+
+// ReconcileCalicoConfig reconciles CalicoConfig data values secrets
+func (r *CalicoConfigReconciler) ReconcileCalicoConfig(
+	ctx context.Context,
+	cluster *clusterapiv1beta1.Cluster,
+	calicoConfig *cniv1alpha1.CalicoConfig) (_ ctrl.Result, retErr error) {
+
+	var patchCRD bool
```
danniel1205, line 98: question: `patchCRD` seems a little bit odd, I am assuming you meant `patchCR` or `patchConfig`

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -46,3 +88,181 @@ func (r *CalicoConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	patchHelper, err := clusterapipatchutil.NewHelper(calicoConfig, r.Client)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	// patch CalicoConfig before returning the function
+	defer func() {
+		// patchCRD will be true if finalizer or owner reference is added/removed
```
danniel1205, line 106: question: will `patchCRD` be true on other cases as well? like CalicoConfig gets updated?

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -46,3 +88,181 @@ func (r *CalicoConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		if patchCRD {
+			r.Log.Info("Patching CalicoConfig")
+			if err := patchHelper.Patch(ctx, calicoConfig.DeepCopy()); err != nil {
```
danniel1205, line 109: question: From my understanding, we do not need deep copy here. Are there any specific reason of having it here ?

addons/pkg/util/addon_util.go
```diff
@@ -218,3 +229,57 @@ func IsPackageInstallPresent(ctx context.Context,
...
+	var patchAddonSecret bool
+
+	// remove finalizer from addon secret
+	if controllerutil.ContainsFinalizer(configCRD, addontypes.AddonFinalizer) {
+		log.Info("Removing finalizer from addon secret", constants.AddonNameLogKey, addonName)
+		controllerutil.RemoveFinalizer(configCRD, addontypes.AddonFinalizer)
+		patchAddonSecret = true
+	}
+
+	return patchAddonSecret
```
danniel1205, line 262: question: the naming is a little bit confusing, because L250 says `return true if finalizer is removed`, but here we are indicating if `patchAddonSecret`.

addons/pkg/util/addon_util.go
```diff
@@ -218,3 +229,57 @@ func IsPackageInstallPresent(ctx context.Context,
...
+		controllerutil.AddFinalizer(configCRD, addontypes.AddonFinalizer)
+		patchAddonSecret = true
+	}
+
+	return patchAddonSecret
+}
+
+// RemoveFinalizerFromCRD removes finalizer from the config CRD if present and
+// returns true if finalizer is removed
+func RemoveFinalizerFromCRD(log logr.Logger, addonName string, configCRD client.Object) bool {
```
danniel1205, line 251: nit: I am assuming you meant `addonConfigCR` or `addonConfig` ?

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -46,3 +88,181 @@ func (r *CalicoConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+}
+
+// ReconcileCalicoConfigDelete reconciles CalicoConfig deletion by deleting data values secret
+func (r *CalicoConfigReconciler) ReconcileCalicoConfigDelete(
+	ctx context.Context,
+	calicoConfig *cniv1alpha1.CalicoConfig,
+	patchCRD *bool) (retErr error) {
+
+	// remove the reference to the data values secret from CalicoConfig Status field
+	calicoConfig.Status.SecretRef = &cniv1alpha1.DataValuesSecretRef{}
```
danniel1205, line 245: suggestion: `calicoConfig` here is the original pointer assigned by `r.Client.Get`. Making changes directly to the original pointer is risky. The caller of this function should make a deep copy at the very beginning and pass the copy down to `ReconcileCalicoConfigDelete`

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -46,3 +88,181 @@ func (r *CalicoConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	// add finalizer to CalicoConfig
+	*patchCRD = util.AddFinalizerToCRD(r.Log, constants.CalicoAddonName, calicoConfig)
+
+	// add owner reference to CalicoConfig if not already added by TanzuClusterBootstrap Controller
+	ownerReference := metav1.OwnerReference{
+		APIVersion:         clusterapiv1beta1.GroupVersion.String(),
+		Kind:               cluster.Kind,
+		Name:               cluster.Name,
+		UID:                cluster.UID,
+		Controller:         pointer.BoolPtr(true),
```
danniel1205, line 158: question: any specific reason we want to set this field ?

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -46,3 +88,181 @@ func (r *CalicoConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		BlockOwnerDeletion: pointer.BoolPtr(true),
```
danniel1205, line 159: Question: Setting this field to be true means we block workload cluster deletion until CalicoConfig resource gets deleted. That makes me whether it is better to NOT block a workload cluster from deleting, calicoconfig_controller could be smart to do the cleanup if the referenced cluster does not exist. We check the cluster existence in L71.

maralavi: Thanks for the comment. Removed this and added logic to remove orphaned CalicoConfig resource on line 75. Does that look good or garbage collector itself would take care of orphaned CalicoConfig resource deletion?

addons/controllers/calico/calicoconfig_controller.go
maralavi: Done

addons/pkg/util/addon_util.go
maralavi: changed to ```addonCR``` for a more generic name. How do you think @danniel1205 ?

addons/pkg/util/addon_util.go
maralavi: Agreed, removed.

addons/controllers/calico/calicoconfig_controller.go
maralavi: It would be true only if finalizer or owner reference are added or removed for CalicoConfig. As of now we don't have logic for the modification of those fields for CalicoConfig.

addons/controllers/calico/calicoconfig_controller.go
maralavi: changed to ```patchCalicoConfig```

addons/controllers/calico/calicoconfig_controller.go
maralavi: Agreed, Done.

addons/controllers/calico/calicoconfig_controller.go
maralavi: I removed the context with timeout and changed to use the context as it is

addons/controllers/calico/calicoconfig_controller.go
maralavi: Agreed, Done.

addons/controllers/calico/calicoconfig_controller.go
maralavi: Ok, removed ti use the context passed by manager. Thanks for the comment.

addons/controllers/calico/calicoconfig_controller.go
danniel1205: suggestion: usually we will do a deep copy here and pass down the copied `calicoConfig`. So that you do not need to do multiple deep copy at L109 and L115

addons/controllers/calico/calicoconfig_controller.go
danniel1205: How about upgrade? Like a version bump on calico, there could be modifications on CalicoConfig ? Or maybe I am wrong. I am trying to think what are the cases we do not want to patch/update the enqueued `CalicoConfig` request. Few lines below, the logic seems always update the status.

maralavi: If CalicoConfig is getting updated, the finalizer and owner reference still remain the same, so the patch variable here will not be true and the ```patchHelper.Patch()``` won't get called. 
But for the Status().Update() we update it on Create/Update/Delete of CalicoConfig to refer to the reference to CalicoConfig or remove the reference. One might argue it is not required to update the status for when CalicoConfig is getting updated as the secret reference still remains the same, but I didn't want to make the logic more complicated by adding another check.

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -6,43 +6,269 @@ package controllers
...
+			r.Log.Info(fmt.Sprintf("CalicoConfig resource '%v' not found", req.NamespacedName))
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	// config resources are expected to have the same name as the cluster. However, we ideally try to read the cluster name from the owner reference of the addon config object
+	clusterName := req.NamespacedName.Name
+	for _, ownerRef := range calicoConfig.GetOwnerReferences() {
+		if ownerRef.APIVersion == clusterapiv1beta1.GroupVersion.String() && ownerRef.Kind == "Cluster" {
```
blc1996, line 64: We probably don't need to check `APIVersion` here. I think we just have to check `owner.Kind == cluster.Kind`

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -6,43 +6,269 @@ package controllers
...
+			clusterName = ownerRef.Name
+			break
+		}
+	}
+
+	// verify that the cluster with the same name is present
+	cluster := &clusterapiv1beta1.Cluster{}
+	if err := r.Client.Get(ctx, types.NamespacedName{Namespace: req.Namespace, Name: clusterName}, cluster); err != nil {
+		if apierrors.IsNotFound(err) {
+			r.Log.Info(fmt.Sprintf("Cluster resource '%s' not found. Cleaning up orphaned resource CalicoConfig '%s'", clusterName, calicoConfig.Name))
```
blc1996, line 74: In the event of cluster not found, I don't think the controller is responsible for deleting the CalicoConfig CRD. The user should take care of the creation and deletion of CRD, the controller is only responsible of reconciling it.

blc1996: If it's desired, you could set the status of the CalicoConfig CRD saying something like `reconciliation fialed, cluster not found`, but shouldn't delete CalicoConfig CRD.

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -6,43 +6,269 @@ package controllers
...
+				return ctrl.Result{}, err
+			}
+			r.Log.Info(fmt.Sprintf("Deleted CalicoConfig '%s'", calicoConfig.Name))
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	// reconcile CalicoConfig resource
+	if retResult, err := r.ReconcileCalicoConfig(cluster, calicoConfig.DeepCopy()); err != nil {
```
blc1996, line 85: Why do we need `Deepcopy` here?

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -6,43 +6,269 @@ package controllers
...
+		if clusterNetwork == nil {
+			err := errors.New("cluster.Spec.ClusterNetwork is not set")
+			r.Log.Error(err, fmt.Sprintf("Could not find ClusterNetwork in cluster '%s' Spec", cluster.Name))
+			return err
+		}
+		clusterCIDR := clusterNetwork.Pods
+		if clusterCIDR == nil || len(clusterCIDR.CIDRBlocks) == 0 {
+			clusterCIDR = clusterNetwork.Services
+		}
+		if clusterCIDR == nil || len(clusterCIDR.CIDRBlocks) == 0 {
```
blc1996, line 215: You should not look for serviceCIDR here. The ClusterCIDR is specifically `CLUSTER_CIDR` which is [pod CIDR](https://github.com/vmware-tanzu/tanzu-framework/blob/main/pkg/v1/providers/ytt/02_addons/cni/calico/calico_addon_data.lib.yaml#L14).

addons/controllers/calico/calicoconfig_controller_test.go
```diff
@@ -0,0 +1,154 @@
...
+				}
+				secret := &v1.Secret{}
+				if err := k8sClient.Get(ctx, secretKey, secret); err != nil {
+					return false
+				}
+				Expect(secret.Type).Should(Equal(v1.SecretTypeOpaque))
+				secretData := string(secret.Data["values.yaml"])
+				Expect(strings.Contains(secretData, "IPFamily: ipv4")).Should(BeTrue())
+				return true
+			}, waitTimeout, pollingInterval).Should(BeTrue())
```
blc1996, line 115: Could you add the test case to check secretRef after successful reconciliation?

addons/controllers/calico/calicoconfig_controller_test.go
```diff
@@ -0,0 +1,154 @@
...
+	JustBeforeEach(func() {
+		// create cluster resources
+		By("Creating a cluster and a CalicoConfig")
+		f, err := os.Open(clusterResourceFilePath)
+		Expect(err).ToNot(HaveOccurred())
+		defer f.Close()
+		err = testutil.CreateResources(f, cfg, dynamicClient)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Creating kubeconfig for cluster")
```
blc1996, line 45: This is not needed any more because we don't have remote client.

addons/controllers/calico/calicoconfig_controller_test.go
```diff
@@ -0,0 +1,154 @@
...
+	})
+
+	AfterEach(func() {
+		By("Deleting cluster and CalicoConfig")
+		f, err := os.Open(clusterResourceFilePath)
+		Expect(err).ToNot(HaveOccurred())
+		defer f.Close()
+		testutil.DeleteResources(f, cfg, dynamicClient, true)
+
+		By("Deleting kubeconfig for cluster")
```
blc1996, line 56: Not needed.

addons/pkg/constants/constants.go
```diff
@@ -97,4 +97,10 @@ const (
 
 	// CRDWaitPollTimeout is poll timeout for checking server resources
 	CRDWaitPollTimeout = time.Minute * 10
+
+	// DataValueFileName is the name of data value file
+	DataValueFileName = "values.yaml"
+
+	// CalicoAddonName is the name of Calico addon
+	CalicoAddonName = "calico"
```
blc1996, line 105: Nit: I think this constant can be put under `calicoconfig_controller.go`. This is because we'll hand over the controller to other teams, let's separate the code as much as possible.

addons/pkg/util/addon_util.go
```diff
@@ -218,3 +229,51 @@ func IsPackageInstallPresent(ctx context.Context,
...
+		log.Info("Removing finalizer from addon custom resource", constants.AddonNameLogKey, addonName)
+		controllerutil.RemoveFinalizer(addonCR, addontypes.AddonFinalizer)
+		return true
+	}
+
+	return false
+}
+
+// get paths for external CRDs by introspecting versions of the go dependencies
+func GetExternalCRDPaths(externalDeps map[string][]string) ([]string, error) {
```
blc1996, line 260: This is a test util function, I don't think we should put it here.

apis/cni/v1alpha1/calicoconfig_types.go
```diff
@@ -34,16 +34,22 @@ type Calico struct {
...
 
 // CalicoConfigStatus defines the observed state of CalicoConfig
-type CalicoConfigStatus struct{}
+type CalicoConfigStatus struct {
+	SecretRef *DataValuesSecretRef `json:"secretRef,omitempty"`
+}
+
+type DataValuesSecretRef struct {
+	Name      string `json:"name,omitempty"`
+	Namespace string `json:"namespace,omitempty"`
```
blc1996, line 51: Looking at how kapp controller creates [secretRef](https://carvel.dev/kapp-controller/docs/latest/packaging/#package-install), they don't have a namespace. I think we shouldn't include Namespace here, as the created data value secret will alway be in the same namespace as the config CRD.

addons/controllers/calico/calicoconfig_controller_test.go
```diff
@@ -0,0 +1,154 @@
...
+				}
+				return true
+			}, waitTimeout, pollingInterval).Should(BeTrue())
+
+			Eventually(func() bool {
+				config := &cniv1alpha1.CalicoConfig{}
+				if err := k8sClient.Get(ctx, key, config); err != nil {
+					return false
+				}
+
```
blc1996, line 93: Please check finalizer and ownerRef

addons/controllers/calico/calicoconfig_controller.go
danniel1205: If you look into the implementation of patchHelper.Patch(), it calculates the changes `before` and `after`, intelligently do the patch/update if there are changes and if the obj needs to be patched.  In addition to above, patchHelper.Patch() also patches status subresource.

My thinking is, if we could make thing simple: a) no need to update and check patchCRD bool b) just call patchHelper.Patch() on each reconciliation loop ?

addons/controllers/calico/calicoconfig_controller.go
danniel1205: Update: I was wrong before, the orphan CalicoConfig would be cleaned up automatically since it has ownerRef field set.

addons/controllers/calico/calicoconfig_controller.go
danniel1205: Per discussion offline, the garbage collection of orphan CalicoConfig will be handled by K8S because of the ownerRef is set.

addons/controllers/calico/calicoconfig_controller.go
maralavi: Removed ```patchCRD``` and now only calling ````patchHelper.Patch()``` on the caller ``ReconcileCalicoConfig()```

addons/controllers/calico/calicoconfig_controller.go
maralavi: Removed from here, thanks.

addons/controllers/calico/calicoconfig_controller.go
maralavi: Don't need it and just removed.

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -6,43 +6,269 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 
+	"github.com/fatih/structs"
```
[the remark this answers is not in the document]
maralavi: Some more to-be-derived fields got introduced, so removed the usage of the above library for the sake of cleaner code

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -6,43 +6,269 @@ package controllers
...
+	secretNamespacedName := types.NamespacedName{
+		Name:      util.GenerateDataValueSecretName(calicoConfig.Name, constants.CalicoAddonName),
+		Namespace: calicoConfig.Namespace,
+	}
+	if err := r.ReconcileDataValuesSecret(cluster, calicoConfig, secretNamespacedName); err != nil {
+		return err
+	}
+
+	// add the reference to the data values secret into CalicoConfig Status field
+	calicoConfig.Status.SecretRef = &cniv1alpha1.DataValuesSecretRef{
```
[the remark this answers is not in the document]
maralavi: removed ```patchCalicoConfig``` per our discussion and now do the patch anyway.

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -6,43 +6,269 @@ package controllers
...
+		// patchCalicoConfig will be true if finalizer or owner reference is added/removed
+		if patchCalicoConfig {
+			r.Log.Info("Patching CalicoConfig")
+			if err := patchHelper.Patch(r.Ctx, calicoConfig); err != nil {
+				r.Log.Error(err, "Error patching CalicoConfig")
+				retErr = err
+			}
+		}
+
+		if err := r.Client.Status().Update(r.Ctx, calicoConfig); err != nil {
```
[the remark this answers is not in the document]
maralavi: Removed ```Status().Update()``` as ```patch()``` internally calls ```Status().Update()```

addons/controllers/calico/calicoconfig_controller.go
maralavi: Per our discussion, moved  deepcopy up after initially reading it from api server

addons/pkg/util/addon_util.go
maralavi: Agreed, moved it to testutil directory instead

addons/controllers/calico/calicoconfig_controller_test.go
maralavi: thanks, added

addons/controllers/calico/calicoconfig_controller_test.go
maralavi: Thanks, this time it worked. I guess the Deepcopy() I had  initially added was overwriting it undesirably

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -6,43 +6,220 @@ package controllers
...
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/constants"
+	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/util"
 	cniv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/cni/v1alpha1"
 )
 
-// CalicoConfigReconciler reconciles a CalicoConfig object
+const CalicoAddonName = "calico"
```
vijaykatam, line 31: nitpick: we have a constants package in addons where this could potentially go to

apis/cni/v1alpha1/calicoconfig_types.go
```diff
@@ -34,26 +23,31 @@ type Calico struct {
...
 	// +kubebuilder:validation:Minimum=0
 	// +kubebuilder:default:=0
-	// "0" as default means MTU will be auto detected
 	VethMTU int64 `json:"vethMTU,omitempty"`
 }
 
 // CalicoConfigStatus defines the observed state of CalicoConfig
-type CalicoConfigStatus struct{}
+type CalicoConfigStatus struct {
+	SecretName string `json:"secretName,omitempty"`
```
vijaykatam, line 35: I think we use SecretRef as the convention for linking to a name. Can we use that instead?

addons/pkg/util/addon_util.go
```diff
@@ -218,3 +225,29 @@ func IsPackageInstallPresent(ctx context.Context,
 
 	return true, nil
 }
+
+// AddFinalizerToCR adds finalizer to the custom resource if not present and
+// returns true if finalizer is added
+func AddFinalizerToCR(log logr.Logger, addonName string, addonCR client.Object) bool {
```
vijaykatam, line 231: We could remove these unused functions for finalizers.

addons/controllers/calico/calicoconfig_controller.go
maralavi: I had moved this over here per @blc1996 review comment to make different addons code a separate as possible as the ownership of these will get in to different teams later on. But I can revert per your suggestion

apis/cni/v1alpha1/calicoconfig_types.go
maralavi: So ```SecretRef``` as a name without pointing to an object? And being a ```name``` instead?

vijaykatam: Yeah just SecretRef string that is the name. Could you make sure the other Config APIs are using the same pattern?

```SecretRef string `json:"secretRef,omitempty"```

File: addons/controllers/calico/calicoconfig_controller.go [hunk 31]
```diff
@@ -6,43 +6,220 @@ package controllers
...
+	var patchConfig bool
+
+	patchHelper, err := clusterapipatchutil.NewHelper(calicoConfig, r.Client)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	// patch CalicoConfig before returning the function
+	defer func() {
+		// patchConfig will be true if owner reference or secretRef are added
+		if patchConfig {
```
vijaykatam, line 117: Don't need separate bool to check if it was changed, patchHelper already should do the diff checking

maralavi: cc: @blc1996

apis/cni/v1alpha1/calicoconfig_types.go
```diff
@@ -15,45 +15,39 @@ type CalicoConfigSpec struct {
...
+}
 
-//+kubebuilder:object:root=true
-//+kubebuilder:subresource:status
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:path=calicoconfigs,scope=Namespaced
+// +kubebuilder:printcolumn:name="Namespace",type="string",JSONPath=".spec.cni.namespace",description="The namespace in which calico is deployed"
+// +kubebuilder:printcolumn:name="vethMTU",type="string",JSONPath=".spec.cni.calico.config.vethMTU",description="Maximum transmission unit setting. '0' as default means MTU will be auto detected"
+// +kubebuilder:printcolumn:name="SecretName",type="string",JSONPath=".status.secretName",description="Name of the Calico data values secret"
```
blc1996, line 43: `SecretName` here is not updated

maralavi:
@vijaykatam pipelines have passed and have addressed all review comments.

vijaykatam (approved):
LGTM

danniel1205 (approved):

PR #1560, opened 2022-01-27
Add dependabot configuration

stmcginnis:

This adds configuration for the GitHub Dependabot. With this
configuration, Dependabot will automatically propose updates for
dependencies. We can then choose whether to merge those updates, or if
there is some reason to wait, keep the PR around or close it. At least
this way we are aware of possible updates so we can be less reactionary
towards security and other update needs.

Based on work @jpmcb has done for the `community-edition` repo.

More details and thorough documentation can be found here:
https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/enabling-and-disabling-dependabot-version-updates

Unfortunately not easy to test locally. Once merged, this will run daily dependabot checks. If we find there are dependencies missing, we can update the dependabot.yaml file to include them. If we find this is causing more noise than the benefit it provides, we can quickly revert this change.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,53 @@
+# Dependabot config for go mod version upgrades
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "gomod"
+    directory: "addons/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "gomod"
+    directory: "addons/pinniped/post-deploy/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "gomod"
+    directory: "hack/packages/kbld-image-replace/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "gomod"
+    directory: "hack/tools/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "gomod"
+    directory: "pkg/v1/providers/tests/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "npm"
+    directory: "pkg/v1/tkg/web/e2e/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "npm"
+    directory: "pkg/v1/tkg/web/node-server/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "npm"
+    directory: "pkg/v1/tkg/web/"
+    schedule:
+      interval: "daily"


stmcginnis:
@vui - would love to get your thoughts on this change.

PushkarJ:
@stmcginnis you are probably familiar with this, so apologies if this is me missing context and repeating things :) 

Cluster API project tried using dependabot in the past, but have had limited success with it. Please see here: https://github.com/kubernetes-sigs/cluster-api/pull/4730 followed by https://github.com/kubernetes-sigs/cluster-api/pull/4881 and https://github.com/kubernetes-sigs/cluster-api/issues/4879

OTOH, we had reasonable success with building our own deps (and image scanning) in k/k. More info can be found here: https://github.com/kubernetes/sig-security/tree/main/sig-security-tooling/vulnerability-mgmt . I am hoping we can do the same in tanzu-framework and community-edition. Happy to work with you on a design doc if you are interested :)

stmcginnis:
Thanks for adding that background @PushkarJ. I vaguely remember some of that going on, but wasn't following it closely. It's too bad some of those revert patches didn't details a little better what the issues were when this was enabled.

We could run with this in the `community-edition` repo for awhile and see what kind of results we get there. Or give this a shot in `tanzu-framework` to see if we're significantly simpler or a better repo setup for something like this to work well. I would be happy to quickly put up a revert patch if we find it isn't working or causing any kind of noise for us.

Definitely open to feedback from anyone else that has tried this. From a high level, this seems like it could be useful thing, but there very well could be a gap between theory and practice here.

vuil:
>Definitely open to feedback from anyone else that has tried this. From a high level, this seems like it could be useful thing, but there very well could be a gap between theory and practice here.

I am of similar minds here. What say we leave this PR open, and watch how well it works in community-edition repo for a few weeks before making the call. And if the situation there doesn't reflect the kind of updates we might be expected to do, we can still enable merge this change as a trial to see how it works for this repo.

jpmcb:
> We could run with this in the community-edition repo for awhile and see what kind of results we get there.

Like Sean said, we've enabled this in TCE and are still waiting on results to see how it fits into our developer workflows. Might be a good proving ground for this in `tanzu-framework` but I've found it to be an extremely useful tool and keeps dependencies really fresh.

Happy to provide anymore insight since I've worked alot with it in the past on cloud-foundry as well

PushkarJ:
@jpmcb that is amazing to hear. I was sort of bummed about dependabot not working for k/k and cluster api but happy that we are seeing positive results for TCE. It may actually end up being useful after all for tanzu-framework!! Agree with you all that we can wait and watch for a bit and then revisit this PR :)

stmcginnis:
We resolved a typo issue in the dependabot config in TCE, so we finally started getting PRs from dependabot.

The initial update was a little... jarring... with how many PRs suddenly came in. This can be controlled by limiting the number of PRs the bot is allowed to submit at a time, but I think once we got over the initial shock it was actually a good thing.

There was some catch up work to be done to check the proposed updates and get them all merged, but now that we've caught up, the dependencies in the repo are more up to date and it should be easier to handle the new updates as they come in.

PushkarJ:
That so great to hear @stmcginnis . Definitely +1 for enabling this in `tanzu-framework`. Will leave final decision to @vuil

vuil:
Thanks, given we have had a good experience on community-edition I'd say let's go for it!

vuil (approved):
lgtm, thanks!

PushkarJ:
Thanks @vuil !! @stmcginnis time to merge and party 🎉 🥳

PR #1598, opened 2022-02-01
Configure Kind subnets for dualstack ipv6 primary

tylerschultz:
Kind currently only supports ipv4 primary dualstack. To deploy ipv6
primary dualstack management clusters we need to configure Kind subnets to use
ipv4 primary ordering.

These changes can be reverted when the next version of kind is released and integrated into TF.

Fixes #

Manually built the cli and plugins and deployed a management cluster.

```release-note
NONE
```

### PR Checklist

### DIFF
--- a/pkg/v1/tkg/kind/client.go
+++ b/pkg/v1/tkg/kind/client.go
@@ -339,42 +339,12 @@ func (k *KindClusterProxy) getKindNetworkingConfig() kindv1.Networking {
 		// ignore this error as TKG_IP_FAMILY is optional
 		ipFamily = ""
 	}
-	podSubnet, err := k.options.Readerwriter.Get(constants.ConfigVariableClusterCIDR)
-	if err != nil {
-		switch ipFamily {
-		// We expect the IPv4,IPv6 order for the CIDR in both cases for KinD until
-		// we bump to a version of the kind library that includes a fix for this issue:
-		// https://github.com/kubernetes-sigs/kind/issues/2484
-		case constants.DualStackPrimaryIPv4Family, constants.DualStackPrimaryIPv6Family:
-			podSubnet = constants.DefaultDualStackPrimaryIPv4ClusterCIDR
-		case constants.IPv6Family:
-			podSubnet = constants.DefaultIPv6ClusterCIDR
-		default:
-			podSubnet = constants.DefaultIPv4ClusterCIDR
-		}
-	}
-	serviceSubnet, err := k.options.Readerwriter.Get(constants.ConfigVariableServiceCIDR)
-	if err != nil {
-		switch ipFamily {
-		// We expect the IPv4,IPv6 order for the CIDR in both cases for KinD until
-		// we bump to a version of the kind library that includes a fix for this issue:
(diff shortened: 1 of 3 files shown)


adobley (approved):
LGTM

pkg/v1/tkg/kind/client.go
```diff
@@ -352,7 +352,14 @@ func (k *KindClusterProxy) getKindNetworkingConfig() kindv1.Networking {
 		default:
 			podSubnet = constants.DefaultIPv4ClusterCIDR
 		}
+	} else if ipFamily == constants.DualStackPrimaryIPv6Family {
+		// We expect the IPv4,IPv6 order for the CIDR in both cases for KinD until
+		// we bump to a version of the kind library that includes a fix for this issue:
+		// https://github.com/kubernetes-sigs/kind/issues/2484
+		podSubnets := strings.Split(podSubnet, ",")
+		podSubnet = fmt.Sprintf("%s,%s", podSubnets[1], podSubnets[0])
```
imikushin, line 360: Please factor out `podSubnet` calculation into its own method:
```go
podSubnet := k.podSubnet(ipFamily)
```

In that method, please avoid using `else if` (e.g. by returning from the `if` body).

pkg/v1/tkg/kind/client.go
```diff
@@ -366,6 +373,12 @@ func (k *KindClusterProxy) getKindNetworkingConfig() kindv1.Networking {
 		default:
 			serviceSubnet = constants.DefaultIPv4ServiceCIDR
 		}
+	} else if ipFamily == constants.DualStackPrimaryIPv6Family {
+		// We expect the IPv4,IPv6 order for the CIDR in both cases for KinD until
+		// we bump to a version of the kind library that includes a fix for this issue:
+		// https://github.com/kubernetes-sigs/kind/issues/2484
+		serviceSubnets := strings.Split(serviceSubnet, ",")
+		serviceSubnet = fmt.Sprintf("%s,%s", serviceSubnets[1], serviceSubnets[0])
```
imikushin, line 381: Please factor out `serviceSubnet` calculation into its own method:
```go
serviceSubnet := k.serviceSubnet(ipFamily)
```

In that method, please avoid using `else if` (e.g. by returning from the `if` body).

When you're done, the body of `getKindNetworkingConfig()` should become linear!

pkg/v1/tkg/kind/client.go
christianang: Sounds good. We ended up reading the ipFamily within the podSubnet/serviceSubnet functions and now the `getKindNetworkingConfig` function is very simple.

pkg/v1/tkg/kind/client.go
christianang: Done.

imikushin (approved):
Thanks! LGTM

imikushin:
Actually, please fix the lint error: it spotted code duplication. Once it’s done, it’s going to be good to merge

tylerschultz:
Thanks for all the feedback Ivan. We think we've addressed the valuable feedback and the linter. PTAL when you've got a min.

Issue #1259, opened 2021-11-30
Framework should not inject `TKGVERSION` annotation into newly created management clusters


joshrosso:
**Bug description**

Since TKG is only 1 distribution of framework, newly created management clusters should **not** be injecting a TKG version annotation on the Cluster object. Specifying management cluster version (per edition/distro and its version) is high-value. So this should be pluggable based on edition:version detection.

You'll find this logic present here: https://github.com/vmware-tanzu/tanzu-framework/blob/166919ff9c388fbc68f2717c7e6dd48811fcf853/pkg/v1/tkg/clusterclient/clusterclient.go#L968-L975

**Affected product area (please put an X in all that apply)**

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

danniel1205:
Thanks for filing this issue. TKG Telemetry currently reads the version from the `TKGVersionKey` of CAPI cluster object. We need to find an alternative for telemetry if we decide to take TKG version out.

anujc25:
This value is heavily used to determine the version of TKG management clusters. Currently, **this value is determined from the TKG BoM file** and has been added as an annotation on the management cluster to denote the version of the management cluster. 

Currently, This value gets used in many places including upgrade validation, plugin version compatibility.

Sample TKG BoM file:
```
apiVersion: run.tanzu.vmware.com/v1alpha2
default:
  k8sVersion: v1.22.3+vmware.1-tkg.2-zshippable
release:
  version: v1.4.0 # This value is used to set TKGVersionKey as part of the annotation
```

> Specifying management cluster version (per edition/distro and its version) is high-value.

If we would like to do this, maybe one possible way to do this is to use a different BoM file for each edition and specify a different version for that edition?

vuil:
Yeah the value of the TKGVERSION is specific to the framework (assuming there is no customization to what that version of framework understands to be the TKG BOM), and whose values carries some implications to telemetry, upgrade and version compatibility checks. None of those implications cannot be addressed, some hard than others, but if it is all the same, @joshrosso would introducing another annotation to set the EDITION work as well?

joshrosso:
> TKGVERSION is specific to the framework

Seems like this should be specific to TKG, not framework, right?

Framework exists to bootstrap different distributions/editions/etc, of which examples are:

[...]

PR #1599, opened 2022-02-01
Add edition as annotation to managemenet cluster

joshrosso:

This commit adds the edition as an annotation to management clusters
that are created. This is important for projects/products that might
introspect these management clusters and want to make decisions based on
whether the cluster was bootstrapped as TCE or Tanzu Standard cluster.

Fixes: https://github.com/vmware-tanzu/tanzu-framework/issues/1259

TODO(joshrosso)

### Release note

```release-note
Adds annotation (edition: ${EDITION}) to management cluster.
```

### DIFF
--- a/pkg/v1/tkg/client/init.go
+++ b/pkg/v1/tkg/client/init.go
@@ -34,10 +34,11 @@ const (
 )
 
 const (
-	statusRunning       = "running"
-	statusFailed        = "failed"
-	statusSuccessful    = "successful"
-	vsphereVersionError = "the minimum vSphere version supported by Tanzu Kubernetes Grid is vSphere 6.7u3, please upgrade vSphere and try again"
+	editionAnnotationKey = "edition"
+	statusRunning        = "running"
+	statusFailed         = "failed"
+	statusSuccessful     = "successful"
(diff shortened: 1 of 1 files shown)


berndtj:
Will this depend on users explicitly setting the `cli.edition` property or will it be set by default?

joshrosso:
> Will this depend on users explicitly setting the `cli.edition` property or will it be set by default?

TCE users will have this value set. I'm not imposing a default as that's a tanzu-framework / TKG decision to make.

tanzu-framework reviewers, please let me know if you'd like a default.

Personally, I'd highly prefer the default **was** TCE.

joshrosso:
I may have been wrong, appears there *is* a default, it is `tkg`:

https://github.com/vmware-tanzu/tanzu-framework/blob/9569fa38908513ebb043cee8a1fcafa4c9836c24/pkg/v1/config/defaults.go#L36-L37

berndtj:
As long as it correctly identifies TCE clusters, I don't really have an issue.  Thanks!

vuil (approved):
lgtm, thanks!

Issue #1602, opened 2022-02-01
`TKG_DEFAULT_IMAGE_REPOSITORY` and `TKG_DEFAULT_COMPATIBILITY_IMAGE_PATH` should come from `~/.config/tanzu/config.yaml` and not from build time 


joshrosso:
(This is used to request new product features)

**Describe the feature request**

Currently, we set the compatibility file at build time:

https://github.com/vmware-tanzu/tanzu-framework/blob/9569fa38908513ebb043cee8a1fcafa4c9836c24/Makefile#L60-L67

Baking this into the binary means that we need to produce a TCE-specific binary and TKG-specific binary. This should be moved into the `config.yaml` and configurable via `tanzu config XXX` command.

**Describe alternatives you've considered**

None

**Affected product area (please put an X in all that apply)**

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Additional context**

None

PR #1605, opened 2022-02-02
Make compatibility file resolution happen via config

joshrosso:
## What this PR does / why we need it

It is critical that tanzu `management-cluster` and `cluster` plugins can work against different editions without re-compiling binaries. Today, they require re-compilation due to setting default via LD_FLAGS. This PR continues to support defaulting via LD_FLAGS, but propagates these values into `${HOME}/.config/tanzu/config.yaml`. This enables the switching of values based on setting editions.

In `config.yaml` the following `clientOptions.cli` fields are introduced:

```yaml
clientOptions:                                                                                                                                 
  cli:                                                                                                                                         
    bomRepo: projects-stg.registry.vmware.com/tkg                                                                                              
    compatibilityFilePath: framework-zshippable/tkg-compatibility 
```

`bomRepo` is the repository (prefix) for resolving the compatibility file and (TKG) BOM.
`compatibilityFilePath` is the path (project) where the compatibility file is found. The full URI will be resolved as: `bomRepo` + "/" + `compatibilityFilePath`.

## Describe testing done for PR

1. Run `make build-install-cli-all ENVS=linux-amd64`.
2. Run `cat ~/.config/tanzu/config.yaml`

    ```yaml
    $ cat ~/.config/tanzu/config.yaml 
    clientOptions:
      cli:
        bomRepo: projects-stg.registry.vmware.com/tkg
        compatibilityFilePath: framework-zshippable/tkg-compatibility
        discoverySources:
        - local:
            name: default-local
            path: standalone
        - local:
            name: admin-local
            path: admin
        edition: tkg
        repositories:
        - gcpPluginRepository:
            bucketName: tanzu-cli-framework
            name: core
        unstableVersionSelector: none
      features:
        cluster:
          custom-nameservers: "false"
        global:
        management-cluster:
          custom-nameservers: "false"
          export-from-confirm: "true"
          import: "false"
          network-separation-beta: "false"
    kind: ClientConfig
    metadata:
      creationTimestamp: null
    ```

3. Verify correct files are downloaded.

    ```sh
    Downloading TKG compatibility file from 'projects-stg.registry.vmware.com/tkg/framework-zshippable/tkg-compatibility'
    Downloading the TKG Bill of Materials (BOM) file from 'projects-stg.registry.vmware.com/tkg/tkg-bom:v1.6.0-zshippable'
    Downloading the TKr Bill of Materials (BOM) file from 'projects-stg.registry.vmware.com/tkg/tkr-bom:v1.22.5_vmware.1-tkg.2-zshippable'
    ```

### Switch the edition to TCE

1. Set the edition.

    ```sh
    $ tanzu config set cli.edition tce
    ℹ  Edition changed. Clearing compatibility file at /home/josh/.config/tanzu/tkg/compatibility/tkg-compatibility.yaml
    ```

1. Examine `~/.config/tanzu/config.yaml`.

    ```yaml
    clientOptions:
      cli:
        bomRepo: projects.registry.vmware.com/tce
        discoverySources:
        - local:
            name: default-local
            path: standalone
        - local:
            name: admin-local
            path: admin
        edition: tce
        repositories:
        - gcpPluginRepository:
            bucketName: tanzu-cli-framework
            name: core
        unstableVersionSelector: none
      features:
        cluster:
          custom-nameservers: "false"
        global:
        management-cluster:
          custom-nameservers: "false"
          export-from-confirm: "true"
          import: "false"
          network-separation-beta: "false"
    kind: ClientConfig
    metadata:
      creationTimestamp: null
    ```

1. Verify correct files are downloaded.

    ```sh
    Downloading TKG compatibility file from 'projects.registry.vmware.com/tce/tkg-compatibility'
    Downloading the TKG Bill of Materials (BOM) file from 'projects.registry.vmware.com/tce/tkg-bom:v1.4.1'
    Downloading the TKr Bill of Materials (BOM) file from 'projects.registry.vmware.com/tkg/tkr-bom:v1.21.2_vmware.1-tkg.2'
    the old providers folder /home/josh/.config/tanzu/tkg/providers is backed up to /home/josh/.config/tanzu/tkg/providers-20220207061309-gzu5wz8j
    ```

Fixes: https://github.com/vmware-tanzu/tanzu-framework/issues/1602

### Release note

```release-note
Compatibility file can now be set via config value rather than exclusively at build time.
```

### DIFF
--- a/apis/config/v1alpha1/clientconfig.go
+++ b/apis/config/v1alpha1/clientconfig.go
@@ -6,8 +6,14 @@ package v1alpha1
 import (
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
+
+	"github.com/aunum/log"
+
+	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkg/tkgconfigpaths"
 )
 
 const (
@@ -19,6 +25,12 @@ const (
 	ExperimentalUnstableVersions VersionSelectorLevel = "experimental"
 	// NoUnstableVersions allows no unstable plugin versions, format major.minor.patch only
 	NoUnstableVersions VersionSelectorLevel = "none"
+	// DefaultTCEBomRepo is OCI repository containing the BOM for TCE
+	DefaultTCEBomRepo = "projects.registry.vmware.com/tce"
+	// DefaultTKGBomRepo is OCI repository containing the BOM for TKG
+	DefaultTKGBomRepo = "projects.registry.vmware.com/tkg"
+	// DefaultCompatibilityPath the path (project) of the compatibility file
+	DefaultCompatibilityPath = "tkg-compatibility"
 )
 
 const (
@@ -142,4 +154,56 @@ func (c *ClientConfig) SetEditionSelector(edition EditionSelector) {
 		return
 	}
 	c.ClientOptions.CLI.UnstableVersionSelector = EditionStandard
+}
+
+// SetCompatibilityFile changes the compatibility file for the edition.
+func (c *ClientConfig) SetCompatibilityFile(edition EditionSelector) error {
+	// TODO(joshrosso): do something with error
+	configPath, err := getTKGConfigDir()
+	if err != nil {
+		return err
+	}
+	path, err := tkgconfigpaths.New(configPath).GetTKGCompatibilityConfigPath()
+	if err != nil {
+		return err
+	}
+
+	// best effort attempt to remove, do not throw error on failure
+	log.Infof("Edition changed. Clearing compatibility file at %s", path)
+	_ = os.Remove(path)
+
+	if c.ClientOptions == nil {
+		c.ClientOptions = &ClientOptions{}
+	}
+	if c.ClientOptions.CLI == nil {
+		c.ClientOptions.CLI = &CLIOptions{}
+	}
+
+	switch edition {
+	case EditionCommunity:
+		c.ClientOptions.CLI.BOMRepo = DefaultTCEBomRepo
+		c.ClientOptions.CLI.CompatibilityFilePath = DefaultCompatibilityPath
+	case EditionStandard:
+		c.ClientOptions.CLI.BOMRepo = DefaultTKGBomRepo
+		c.ClientOptions.CLI.CompatibilityFilePath = DefaultCompatibilityPath
+	}
+
+	return nil
+}
+
+func getTKGConfigDir() (string, error) {
+	tanzuConfigDir, err := localDirPath(".config/tanzu")
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(tanzuConfigDir, "tkg"), nil
+}
+
+func localDirPath(dirname string) (path string, err error) {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return path, err
+	}
+	path = filepath.Join(home, dirname)
+	return
 }
--- a/pkg/v1/cli/command/core/config.go
+++ b/pkg/v1/cli/command/core/config.go
@@ -182,9 +182,14 @@ func setEdition(cfg *configv1alpha1.ClientConfig, edition string) error {
 	editionOption := configv1alpha1.EditionSelector(edition)
 
 	switch editionOption {
-	case configv1alpha1.EditionCommunity,
-		configv1alpha1.EditionStandard:
+	case configv1alpha1.EditionCommunity, configv1alpha1.EditionStandard:
 		cfg.SetEditionSelector(editionOption)
+		// when community edition is set, configure the compatibility file to use
+		// community edition's.
+		err := cfg.SetCompatibilityFile(editionOption)
+		if err != nil {
+			return err
+		}
 	default:
 		return fmt.Errorf("unknown edition: %s; should be one of [%s, %s]", editionOption, configv1alpha1.EditionStandard, configv1alpha1.EditionCommunity)
 	}
--- a/pkg/v1/config/clientconfig.go
+++ b/pkg/v1/config/clientconfig.go
@@ -16,6 +16,7 @@ import (
 
 	configv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/config/v1alpha1"
 	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli/common"
+	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkg/tkgconfigpaths"
 
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
@@ -203,6 +204,26 @@ func addEdition(c *configv1alpha1.ClientConfig, edition configv1alpha1.EditionSe
 	c.ClientOptions.CLI.Edition = edition
 }
 
+func addCompatabilityFile(c *configv1alpha1.ClientConfig, compatibilityFilePath string) {
+	if c.ClientOptions == nil {
+		c.ClientOptions = &configv1alpha1.ClientOptions{}
+	}
+	if c.ClientOptions.CLI == nil {
+		c.ClientOptions.CLI = &configv1alpha1.CLIOptions{}
+	}
+	c.ClientOptions.CLI.CompatibilityFilePath = compatibilityFilePath
+}
+
+func addBomRepo(c *configv1alpha1.ClientConfig, repo string) {
+	if c.ClientOptions == nil {
+		c.ClientOptions = &configv1alpha1.ClientOptions{}
+	}
+	if c.ClientOptions.CLI == nil {
+		c.ClientOptions.CLI = &configv1alpha1.CLIOptions{}
+	}
+	c.ClientOptions.CLI.BOMRepo = repo
+}
+
 // ClientConfigNotExistError is thrown when a tanzu config cannot be found.
 type ClientConfigNotExistError struct {
 	s string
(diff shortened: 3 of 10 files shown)


joshrosso:
Will update testing/fakes and get the merge conflict sorted today.

pkg/v1/cli/command/core/config.go
```diff
@@ -182,9 +182,14 @@ func setEdition(cfg *configv1alpha1.ClientConfig, edition string) error {
 	editionOption := configv1alpha1.EditionSelector(edition)
 
 	switch editionOption {
-	case configv1alpha1.EditionCommunity,
-		configv1alpha1.EditionStandard:
+	case configv1alpha1.EditionCommunity, configv1alpha1.EditionStandard:
 		cfg.SetEditionSelector(editionOption)
+		// when community edition is set, configuration the compatibility file to use
```
nrb, line 187: I think this comment needs to be updated; `configuration` -> `configure`, maybe?

pkg/v1/config/clientconfig.go
```diff
@@ -200,6 +202,16 @@ func addEdition(c *configv1alpha1.ClientConfig, edition configv1alpha1.EditionSe
 	c.ClientOptions.CLI.Edition = edition
 }
 
+func addCompatabilityFile(c *configv1alpha1.ClientConfig, comapatibilityFileLocation string) {
```
nrb, line 205: This function is called in exactly one spot - could we inline it to `addDefaultEditonIfMissing`?

pkg/v1/tkg/tkgconfigbom/bom.go
```diff
@@ -451,18 +456,23 @@ var errorDownloadingTKGCompatibilityFile = `failed to download the TKG Compatibi
 If this is an internet-restricted environment please refer to the documentation to set TKG_CUSTOM_IMAGE_REPOSITORY and related configuration variables in %s
```
vuil, line 456: I think the air-gap/internet-restricted scenario would be broken unless the compatibilityFileLocation provided actually accounted for the repo change that would have been provided via TKG_CUSTOM_IMAGE_REPOSITORY.

If the expectation is that it is the case, the means during the initialization of this compatibilityFileLocation (when no edition exists) or when it is set on an edition switch, TKG_CUSTOM_IMAGE_REPOSITORY will have to be accounted for.

If the expectation is that one can override the config with the 'right' value as a workaround, this error message needs to convey said workaround in the situation where the compatibilityFileLocation was read and used.

Another option is to continue to leverage the TKG_CUSTOM_IMAGE_REPOSITORY handling like how it is done in the else clause in L466 to patch the image path to use the right repository.

vuil (commented):
Thanks for the changes, @joshrosso . Mostly nits on typos and names, but I think we need some modifications to make sure the airgap case is properly handled.

vuil:
cc @prkalle to also help take a look at the bom download logic.

pkg/v1/config/clientconfig.go
```diff
@@ -273,14 +285,24 @@ func GetClientConfig() (cfg *configv1alpha1.ClientConfig, err error) {
...
 		_ = StoreClientConfig(&c)
 	}
 
 	return &c, nil
 }
 
+// addDefaultEditionIfMissing returns true if the default edition was added to the configuration (because there was no edition)
+func addCompatibilityFileIfMissing(config *configv1alpha1.ClientConfig) bool {
+	if config.ClientOptions == nil || config.ClientOptions.CLI == nil || config.ClientOptions.CLI.CompatabilityFileLocation == "" {
+		addCompatabilityFile(config, resolveDefaultCompatabilityFile())
```
vuil, line 300: If I am reading correctly, the linker-flag provided values will only be used here. But once the edition is set, there will be no way to go back to using those values unless one manually updates the compatibilityFileLocation config value. Is that the intended behavior?

joshrosso: > But once the edition is set, there will be no way to go back to using those values unless one manually updates the compatibilityFileLocation config value. Is that the intended behavior?

I wasn't sure on this one. I do want to **bind** the edition change to a compatibility file change, but it seems like a valid use-case to change just the compatibility file.

Should I update this PR to offer a way to use `tanzu config` to change **just** the compatibility file?

pkg/v1/config/clientconfig.go
joshrosso: I did it to keep the structure consistent with:

https://github.com/vmware-tanzu/tanzu-framework/blob/caabeefb01e299e3b03dd8809925a595086eea53/pkg/v1/config/clientconfig.go#L182-L203

The compiler should be smart enough to inline it, so I think this is fine.

LMK if you or anyone feels passionately otherwise!

pkg/v1/tkg/tkgconfigbom/bom.go
joshrosso: > Another option is to continue to uniformly leverage the TKG_CUSTOM_IMAGE_REPOSITORY handling like how it is done in the else clause in L466 to patch the compatibilityFileLocation image path to use the right repository.

I'm open to this! Is this the approach that makes the most sense to framework?

If so, we could update the client config to be something like:

```yaml
baseRepoUrl: 
compatibilityPath:
```

pkg/v1/cli/command/core/config.go
joshrosso: yes,typo, fixed!

pkg/v1/config/clientconfig.go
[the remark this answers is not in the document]
joshrosso: fixed!

pkg/v1/config/clientconfig.go
```diff
@@ -273,14 +285,24 @@ func GetClientConfig() (cfg *configv1alpha1.ClientConfig, err error) {
...
 
-	if addedFeatureFlags || addedDefaultDiscovery || addedEdition {
+	if addedFeatureFlags || addedDefaultDiscovery || addedEdition || addedCompatabilityFile {
 		_ = StoreClientConfig(&c)
 	}
 
 	return &c, nil
 }
 
+// addDefaultEditionIfMissing returns true if the default edition was added to the configuration (because there was no edition)
```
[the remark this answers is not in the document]
joshrosso: fixed!

pkg/v1/config/clientconfig.go
```diff
@@ -618,3 +644,28 @@ func GetEdition() (string, error) {
...
+	if err != nil {
+		return "", err
+	}
+	if cfg != nil && cfg.ClientOptions != nil && cfg.ClientOptions.CLI != nil {
+		return cfg.ClientOptions.CLI.CompatabilityFileLocation, nil
+	}
+	return "", nil
+}
+
+func GetRepoFromCompatabilityFile() string {
```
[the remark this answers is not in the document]
joshrosso: good point. Update the signature. Thoughts?

joshrosso:
> cc @prkalle to also help take a look at the bom download logic.

Adding @prkalle as an assignee to review this PR.

Thanks!

apis/config/v1alpha1/clientconfig.go
```diff
@@ -6,8 +6,13 @@ package v1alpha1
 import (
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
+
+	"github.com/aunum/log"
```
stmcginnis, line 14: Need a line break between the third party import (aunum/log) and the local import.

apis/config/v1alpha1/clientconfig.go
```diff
@@ -143,3 +151,53 @@ func (c *ClientConfig) SetEditionSelector(edition EditionSelector) {
 	}
 	c.ClientOptions.CLI.UnstableVersionSelector = EditionStandard
 }
+
+// SetCompatibilityFile changes the compatibility file for the edition.
+func (c *ClientConfig) SetCompatibiltyFile(edition EditionSelector) error {
```
stmcginnis, line 156: s/SetCompatibiltyFile/SetCompatibiiltyFile/

pkg/v1/config/clientconfig.go
nrb: @joshrosso I think adding a `tanzu config` option is a good idea. Whether it should go in this PR or a different one I'll leave to Vui.

apis/config/v1alpha1/clientconfig.go
joshrosso: done!

pkg/v1/tkg/tkgconfigbom/bom.go
vuil: > > Another option is to continue to uniformly leverage the TKG_CUSTOM_IMAGE_REPOSITORY handling like how it is done in the else clause in L466 to patch the compatibilityFileLocation image path to use the right repository.
> 
> I'm open to this! Is this the approach that makes the most sense to framework?
For the purpose of making options/settings the user sets has consistent behavior, I'd say yes.
e.g. In the past situations where setting TKG_CUSTOM_IMAGE_REPOSITORY should have the desired effect but didn't were treated as bugs and fixed.

> 
> If so, we could update the client config to be something like:
> 
> ```yaml
> baseRepoUrl: 
> compatibilityPath:
> ```
I assume in this proposal, baseRepoURL is something like
projects.registry.vmware.com/tce
and compatibilityPath something like
tkg-compatibility
?

I was originally think just that the current state of TKG_CUSTOM_IMAGE_REPOSITORY be taken into account after reading the compatibility location cli option, but I like the idea of starting to use the config file option to record the equivalent as well. 
If we do the latter, however, we would need to do some refactoring to ensure all places (about 6-7 places) where TKG_CUSTOM_IMAGE_REPOSITORY is read is converted to prefer the client option value if it exists.
Otherwise there will be surprises when values of TKG_CUSTOM_IMAGE_REPOSITORY and the cli option differ for whatever reason.

pkg/v1/config/clientconfig.go
vuil: Much of how relevant the use case is depends on how we handle the issue in the previous comment. Say the implementation is done such that TKG_CUSTOM_IMAGE_REPOSITORY will continue to be respected regardless of what is set in the config file, then there will be little need to update the compat file location via `tanzu config`.

pkg/v1/tkg/tkgconfigbom/bom.go
joshrosso: I agree. I think this also makes things more explicit.

I'll try to make this change and evaluate how tough the plumbing is.

pkg/v1/config/clientconfig.go
joshrosso: totally agreed. I'll follow-up on https://github.com/vmware-tanzu/tanzu-framework/pull/1605#discussion_r799852572 and we can go from there.

pkg/v1/tkg/tkgconfigbom/bom.go
prkalle: I like the idea of breaking the `baseRepoUrl` and `compatibilityPath`.

joshrosso:
@vuil @prkalle I've reworked this PR and based on what we discussed and updated the original description. This model now respects overrides such as environment variables yet ensures defaults (set via LD flags) are reflected in the `config.yaml`. I think this gives us the best of both worlds.

Can you please re-review at your earliest?

Thank you!

apis/config/v1alpha1/clientconfig.go
```diff
@@ -19,6 +25,10 @@ const (
 	ExperimentalUnstableVersions VersionSelectorLevel = "experimental"
 	// NoUnstableVersions allows no unstable plugin versions, format major.minor.patch only
 	NoUnstableVersions VersionSelectorLevel = "none"
+	// DefaultTCECompatibilityFileLocation is the OCI location for TCE' compatibility file
```
prkalle, line 28: nit: should `DefaultTCECompatibilityFileLocation` be updated to `DefaultTCEBomRepo`

prkalle (approved):
LGTM. Thank you!

joshrosso:
> LGTM. Thank you!

:pray: , @vuil FYI if you could give this another review and possibly merge it, it'd be greatly appreciated.

vuil (approved):
Thanks for the updates to account for both the ld flags and environment variables.
LGTM!

Issue #1532, opened 2022-01-21
TKR infra-Machine webhooks (with packages)


imikushin:
Code to implement TKR infra-Machine webhooks AND package manifests to produce Carvel packages. We need such package for every infrastructure provider: aws, azure, docker, vsphere.

## Details
If the `run.tanzu.vmware.com/os-image-ref` annotation is present in the infra-Machine metadata, the webhook should populate the configured fields in the infra-Machine spec. 

For example, an AWSMachine object is being created and has the following in its metadata:
```yaml
  annotations:
    run.tanzu.vmware.com/os-image-ref: |
      id: ami-0f2e5eec7ae0a1986
``` 
The mutating webhook should read this annotation, and set 
`spec.ami.id := "ami-0f2e5eec7ae0a1986"` in the final AWSMachine object. 

The webhook should be installable as a Carvel package, so we need package manifests (package.yaml) for all infrastructure providers: aws, azure, docker, vsphere.

PR #1615, opened 2022-02-03
Relax the type constraint on OSImage.Spec.Image.Ref values

imikushin:
Allow values of the OSImage.Spec.Image.Ref map to be any type, not just
string.

We need to be able to pass values of any type in .spec.image.ref map, so that the infra-Machine webhook can inject them without typecasting directly into the infra-Machine object.

Helps implement #1532 

Added a unit test to make sure we can pass bool and string values.
Tested the generated CRD manually on a AWS based management cluster.

### Release note

```release-note
NONE
```

### PR Checklist


prkalle (approved):
LGTM. Thanks!

tenczar (approved):
LGTM

apis/run/v1alpha3/osimage_types.go
```diff
@@ -17,14 +19,38 @@ type OSInfo struct {
...
 	// Some of currently known types are: "ami", "azure", "docker", "ova", "vmop".
 	Type string `json:"type"`
 
 	// Ref is a key-value map identifying the image within the infrastructure provider. This is the data
 	// to be injected into the infra-Machine objects (like AWSMachine) on creation.
-	Ref map[string]string `json:"ref"`
+	// +kubebuilder:validation:Schemaless
+	// +kubebuilder:validation:Type=object
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Ref map[string]interface{} `json:"ref"`
```
vijaykatam, line 33: Question: If the target is just an annotations on a runtime.Object, I am confused why this interface{}, is it because CRD machinery not able to marshal properly if it sees anything other string?

imikushin: > If the target is just an annotations on a runtime.Object, ...

Not sure I understand the above premise.

> why interface{}

Because we need to be able to preserve the data type, e.g.  `spec.image.marketplace.thirdPartyImage` field in AzureMachine is bool, not string.

If we keep the map value type as `string`, the infra-Machine webhook will need to be smart, which means less simple and more prone to bugs.

> is it because CRD machinery not able to marshal properly if it sees anything other string?

CRD machinery on the kube-apiserver side is perfectly capable of handling "free-form" `object` type. The added `+kubebuilder:...` annotations simply drive the CRD manifest generator to insert appropriate statements into the CRD YAML. 

Also, controller-gen cannot generate DeepCopy implementation `interface{}` on its own, so it had to be handled, too.

vijaykatam: non-blocking, approving for now

vijaykatam (approved):

PR #1639, opened 2022-02-08
Add proxy configuration in tanzuClusterBootstrap

### LINKED ISSUE #1586: Proxy configuration in TanzuclusterBootstrapTemplate

Proxy config such as httpProxy, httpsProxy and noProxy is used by lot of addons. The proxy config needs to be aggregated to the template layer and cloned for clusters.

The cloned proxy information should be provided to all packages as annotations that the individual controllers can then pickup and use for generating secrets.

- [ ] APIs
- [x] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Additional context**

blc1996:

* Add proxy configuration in tanzuClusterBootstrap Spec
* tanzuClusterBootstrap controller propagates proxy config to cluster annotations, so that other config controllers can read cluster proxy config from it

Fixes #1586 

### Release note

```release-note
package-based-lcm: Consolidates cluster proxy configuration in TanzuClusterBootstrap
```

### PR Checklist


addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -584,3 +592,39 @@ func (r *TanzuClusterBootstrapReconciler) watchProvider(providerRef *corev1.Type
...
+	// use patchHelper to auto detect if there is diff in Cluster CR when performing update
+	patchHelper, err := clusterapipatchutil.NewHelper(cluster, r.Client)
+	if err != nil {
+		return err
+	}
+
+	HTTPProxy := ""
+	HTTPSProxy := ""
+	NoProxy := ""
+	if tcb.Spec.Proxy != nil {
```
maralavi, line 609: If tcb.Spec.Proxy be nil non of these changes are needed, could you add an if block at the very beginning of the function and immediately return on the following?
```
if tcb.Spec.Proxy != nil {
       return nil
}
```

blc1996: `tcb.Spec.Proxy == nil` means those proxy configs in annotations would be empty. We still want to try patch it, i.e. cluster annotations had some pre-existing proxy annotations and we want to rewrite to empty (User changes proxy to empty).

maralavi: Ok, I see. Thanks for the reply.

apis/run/v1alpha3/tanzuclusterbootstraptemplate_types.go
```diff
@@ -54,6 +56,15 @@ type ValuesFrom struct {
 	ProviderRef *corev1.TypedLocalObjectReference `json:"providerRef,omitempty"`
 }
 
+type ProxyConfig struct {
+	// +optional
+	HTTPProxy string `json:"http_proxy"`
+	// +optional
+	HTTPSProxy string `json:"https_proxy"`
+	// +optional
+	NoProxy string `json:"no_proxy"`
```
vijaykatam, line 65: Question: Should `NoProxy` be a list of strings?

vijaykatam (commented):
I think we need info on how proxy info will be provided from folks.

apis/run/v1alpha3/tanzuclusterbootstraptemplate_types.go
```diff
@@ -21,6 +21,8 @@ type TanzuClusterBootstrapTemplateSpec struct {
 	Kapp *TanzuClusterBootstrapPackage `json:"kapp,omitempty"`
 	// +optional
 	AdditionalPackages []*TanzuClusterBootstrapPackage `json:"additionalPackages,omitempty"`
+	// +optional
+	Proxy *ProxyConfig `json:"proxy,omitempty"`
```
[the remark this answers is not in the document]
blc1996: I was planning to add a config map to pass in proxy configs, but I saw this field in the design doc, so I switch to add this field. Let me know if we need to take other approaches.

apis/run/v1alpha3/tanzuclusterbootstraptemplate_types.go
blc1996: I think it's a comma delimited string
```
# A comma delimited list of domain names which kapp controller should
  # bypass the proxy for when making requests (optional)
  noProxy: "github.com,docker.io"
```

apis/run/v1alpha3/tanzuclusterbootstraptemplate_types.go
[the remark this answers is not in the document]
vijaykatam: I think this needs to be cluster variables since proxy is cross cutting. Lets wait for feedback before we decide on an approach.

apis/run/v1alpha3/tanzuclusterbootstraptemplate_types.go
blc1996: Saw the [link](https://gitlab.eng.vmware.com/core-build/guest-cluster-controller/-/blob/master/apis/run.tanzu/v1alpha2/tanzukubernetescluster_types.go#L267), will change to string array.

blc1996:
Will rebase and proceed after https://github.com/vmware-tanzu/tanzu-framework/pull/1669 is merged.

addons/pkg/constants/constants.go
```diff
@@ -126,6 +126,24 @@ const (
 
 	// InfrastructureRefDocker is the docker infrastructure
 	InfrastructureRefDocker = "DockerCluster"
+
+	// DefaultCNISelectionClusterVariableName is the default cluster variable name for cni selection
+	DefaultCNISelectionClusterVariableName = "tkg.tanzu.vmware.com/cni"
+
+	// DefaultHTTPProxyClusterClassVarName is the default cluster variable name for HTTP proxy setting
```
blc1996, line 133: Checked with Nick, they are using the var name they had directly. Not sure if we need to add `tkg.tanzu.vmware.com/` prefix here

pkg/v1/tkg/constants/defaults.go
```diff
@@ -9,8 +9,7 @@ import (
 
 // default value constants
 const (
-	DefaultCNIType                         = "antrea"
-	DefaultCNISelectionClusterVariableName = "tkg.tanzu.vmware.com/cni"
```
blc1996, line 13: I think this const should be put under addons/pkg/constants

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -584,3 +592,39 @@ func (r *TanzuClusterBootstrapReconciler) watchProvider(providerRef *corev1.Type
...
+		HTTPProxy = tcb.Spec.Proxy.HTTPProxy
+		HTTPSProxy = tcb.Spec.Proxy.HTTPSProxy
+		NoProxy = tcb.Spec.Proxy.NoProxy
+	}
+
+	if cluster.Annotations == nil {
+		cluster.Annotations = map[string]string{}
+	}
+	cluster.Annotations[types.TCBHTTPProxyConfigAnnotation] = HTTPProxy
```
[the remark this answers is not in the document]
blc1996: I think it makes sense to add even if they are empty. It will make it easier for downstream (kapp config controller) to read these annotations.

apis/run/v1alpha3/tanzuclusterbootstraptemplate_types.go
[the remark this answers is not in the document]
blc1996: Used cluster variable instead.

addons/main.go
```diff
@@ -57,44 +56,62 @@ func init() {
...
+	flag.StringVar(&addonFlags.proxyCACertClusterVarName, "proxy-ca-cert-cluster-var-name", constants.DefaultProxyCaCertClusterClassVarName, "Proxy CA certificate cluster variable name")
+	flag.StringVar(&addonFlags.ipFamilyClusterVarName, "ip-family-cluster-var-name", constants.DefaultIPFamilyClusterClassVarName, "IP family setting cluster variable name")
 
 	flag.Parse()
+}
+
+func main() {
+	// parse flags
+	flags := &addonFlags{}
+	parseAddonFlags(flags)
```
blc1996, line 114: This is to get rid of the funlen lint error (main function exceeds 50 lines)

addons/controllers/testdata/test-cluster-bootstrap.yaml
```diff
@@ -0,0 +1,61 @@
+---
+apiVersion: run.tanzu.vmware.com/v1alpha3
```
maralavi, line 2: The content of this file should be already covered in ```testdata/test-cluster-bootstrap-1.yaml```. can we drop this file then?

addons/pkg/types/clusterbootstrap.go
```diff
@@ -0,0 +1,22 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package types defines type constants.
+package types
+
+const (
+	// HTTPProxyConfigAnnotation is the cluster HTTP proxy annotation key
+	HTTPProxyConfigAnnotation = "tkg.tanzu.vmware.com/tkg_http_proxy"
```
blc1996, line 9: The reason for keeping both `HTTPProxyConfigAnnotation` and `DefaultIPFamilyClusterClassVarName` is that even though they keep the same value, they have different semantic meanings.

maralavi (approved):

addons/pkg/constants/constants.go
blc1996: Adding tkg.tanzu.vmware.com/ prefix for now. Need to sync with CLI team to finalize the naming convention.
cc @tenczar

addons/controllers/testdata/test-cluster-bootstrap.yaml
blc1996: deleted

addons/pkg/constants/constants.go
```diff
@@ -132,6 +132,24 @@ const (
 
 	// ClusterBootstrapManagedSecret is the name for the secrets that are managed by ClusterBootstrapController
 	ClusterBootstrapManagedSecret = "clusterbootstrap-secret"
+
+	// DefaultCNISelectionClusterVariableName is the default cluster variable name for cni selection
+	DefaultCNISelectionClusterVariableName = "tkg.tanzu.vmware.com/cni"
+
+	// DefaultHTTPProxyClusterClassVarName is the default cluster variable name for HTTP proxy setting
+	DefaultHTTPProxyClusterClassVarName = "tkg.tanzu.vmware.com/tkg_http_proxy"
```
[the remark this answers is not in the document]
blc1996: Changed to dashes

vijaykatam (approved):
LGTM

PR #1649, opened 2022-02-09
New cni selection using cluster class variable

### LINKED ISSUE #1617: CNI selection using cluster class variable
CNI selection should happen in tanzuclusterbootstrap_controller.go using the following

1. Update `TanzuClusterBootstrapTemplate` spec.cni to spec.cnis to indicate multiple supported cnis
2. Controller should pick the cni that matches a cluster class variable or default to the first one in the template spec.cnis

maralavi:
This PR makes cni selection using cluster class variable. If not cluster variable is defined, it uses the first entry defined in the template as the cnu.

Fixes https://github.com/vmware-tanzu/tanzu-framework/issues/1617

Added unit tests

### Release note

```release-note
package-based-lcm: make cni selection using cluster class variable
```

### PR Checklist


addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -584,3 +598,77 @@ func (r *TanzuClusterBootstrapReconciler) watchProvider(providerRef *corev1.Type
...
+		return "", nil
+	}
+	clusterVariables := cluster.Spec.Topology.Variables
+	for _, clusterVariable := range clusterVariables {
+		if clusterVariable.Name == r.CNISelectionClusterVariableName {
+			if err := json.Unmarshal(clusterVariable.Value.Raw, &clusterVariableValue); err != nil {
+				return "", errors.New("failed in json unmarshal")
+			}
+			switch t := clusterVariableValue.(type) {
+			case string:
```
vijaykatam, line 660: Should process only type `string`, for others return error.

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -348,11 +362,11 @@ func (r *TanzuClusterBootstrapReconciler) cloneSecretsAndProviders(cluster *clus
 	var createdSecrets []*corev1.Secret
 
 	packages := append([]*runtanzuv1alpha3.TanzuClusterBootstrapPackage{
-		bootstrap.Spec.CNI,
 		bootstrap.Spec.CPI,
 		bootstrap.Spec.CSI,
 		bootstrap.Spec.Kapp,
-	}, bootstrap.Spec.AdditionalPackages...)
+	}, bootstrap.Spec.CNIs...)
```
vijaykatam, line 368: cni selection should happen here

addons/controllers/tanzuclusterbootstrap_controller.go
```diff
@@ -584,3 +598,77 @@ func (r *TanzuClusterBootstrapReconciler) watchProvider(providerRef *corev1.Type
 		},
 	)
 }
+
+func (r *TanzuClusterBootstrapReconciler) reconcileCNISelection(
```
vijaykatam, line 602: I would suggest something below that gets called after `tanzuClusterBootstrap.Spec = tanzuClusterBootstrapTemplate.Spec.DeepCopy()` instead of patching the object after it is created.
```
func (r *TanzuClusterBootstrapReconciler) getCNI(tanzuClusterBootstrapTemplate *runtanzuv1alpha3.TanzuClusterBootstrapTemplate, cluster *clusterapiv1beta1.Cluster, log logr.Logger) (*runtanzuv1alpha3.TanzuClusterBootstrapPackage, error)
```
Once CNIs has been set and tanzuClusterBootstrap object created CNIs is immutable and will be enforced by a webhook.

addons/controllers/tanzuclusterbootstrap_controller.go
maralavi: I changed to do right before this function ```r.cloneSecretsAndProviders``` as ```createOrPatchTanzuClusterBootstrapFromTemplate``` seemed a better place to add and didn't need passing template object to ```cloneSecretsAndProviders```

addons/main.go
```diff
@@ -147,6 +152,17 @@ func main() {
 		os.Exit(1)
 	}
 
+	tanzuClusterBootstrapReconciler := controllers.NewtanzuClusterBootstrapReconciler(mgr.GetClient(),
```
blc1996, line 155: Do we want to add ClusterBootstrap controller to main.go now?

addons/main.go
```diff
@@ -147,6 +152,17 @@ func main() {
...
+		ctrl.Log.WithName("controllers").WithName("TanzuClusterBootstrap"),
+		mgr.GetScheme(),
+		cniSelectionClusterVarName,
```
blc1996, line 158: Can we create a config struct like `addonconfig.Config`? Some more configs will be added.

addons/main.go
maralavi: Not yet, just removed.

addons/main.go
maralavi: Sure, done

addons/controllers/testdata/test-cluster-bootstrap.yaml
```diff
@@ -20,6 +20,20 @@ metadata:
...
+  clusterNetwork:
+    pods:
+      cidrBlocks: [ "192.168.0.0/16","fd00:100:96::/48" ]
+    services:
+      cidrBlocks: [ "192.168.0.0/16","fd00:100:96::/48" ]
+  topology:
+    class: test-clusterclass-tcbt
+    version: v1.22.3
+    variables:
+      - name: tkg.tanzu.vmware.com/cni
```
blc1996, line 35: Look like the CLI team hasn't finalized the naming convention for clusterClass vars.
Saw something in Anuj's [PR](https://github.com/vmware-tanzu/tanzu-framework/pull/1668/files#diff-64eae10d16908dfcb6b093a12243425fbb7b16e7c8c0d8573a883b9ea39a371cR41)

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -607,3 +624,64 @@ func (r *ClusterBootstrapReconciler) watchProvider(providerRef *corev1.TypedLoca
...
+			clusterBootstrapPackage = clusterBootstrapTemplate.Spec.CNIs[0]
+		} else {
+			return nil, errors.New("no CNI was specified in the ClusterClass or in ClusterBootstrap.Spec.CNIs")
+		}
+	}
+
+	return clusterBootstrapPackage, nil
+}
+
+func (r *ClusterBootstrapReconciler) parseClusterVariableCNI(cluster *clusterapiv1beta1.Cluster) (string, error) {
```
blc1996, line 658: Can we make this a generic function that can be used to fetch different clusterClass vars?

maralavi: I don't think so. the reason is that each variable might have a different type in the schema (string, numeric, slice, ...) and this function would get too hacky if made generic to accommodate for all possible variable types. I'd rather having separate functions for different variables for now.

blc1996: Those we need will all be strings. Basically what's in https://github.com/vmware-tanzu/tanzu-framework/blob/main/pkg/v1/providers/config_default.yaml

blc1996: Maybe a generic function like `GetClusterVariableStringByName` for string would be enough.

maralavi: Yes I can do one only for string type variables

maralavi: Added ```parseClusterVariableString()``` for parsing string variables

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -349,11 +366,11 @@ func (r *ClusterBootstrapReconciler) cloneSecretsAndProviders(cluster *clusterap
 	var createdSecrets []*corev1.Secret
 
 	packages := append([]*runtanzuv1alpha3.ClusterBootstrapPackage{
-		bootstrap.Spec.CNI,
 		bootstrap.Spec.CPI,
 		bootstrap.Spec.CSI,
 		bootstrap.Spec.Kapp,
-	}, bootstrap.Spec.AdditionalPackages...)
+	}, bootstrap.Spec.CNIs...)
+	packages = append(packages, bootstrap.Spec.AdditionalPackages...)
```
vijaykatam, line 373: This line could be moved up into the previous append

addons/controllers/testdata/test-cluster-bootstrap.yaml
vijaykatam: Should be fine as long as the name is configurable. I think the var names will differ between TKGm and TKGS due to TKGm needing to support legacy configs.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -607,3 +624,64 @@ func (r *ClusterBootstrapReconciler) watchProvider(providerRef *corev1.TypedLoca
...
+			clusterBootstrapPackage = clusterBootstrapTemplate.Spec.CNIs[0]
+		} else {
+			return nil, errors.New("no CNI was specified in the ClusterClass or in ClusterBootstrap.Spec.CNIs")
+		}
+	}
+
+	return clusterBootstrapPackage, nil
+}
+
+func (r *ClusterBootstrapReconciler) parseClusterVariableString(cluster *clusterapiv1beta1.Cluster, variableName string) (string, error) {
```
vijaykatam, line 658: Could this be a standalone helper in pkg/util? Nothing specific to ClusterBootstrapReconciler in the method. 

Suggestion: Move to pkg/util and write a couple of unit tests.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -607,3 +624,64 @@ func (r *ClusterBootstrapReconciler) watchProvider(providerRef *corev1.TypedLoca
...
+	var (
+		clusterVariableValue interface{}
+		result               string
+	)
+
+	if cluster == nil {
+		return "", errors.New("cluster resource is not found")
```
vijaykatam, line 665: nitpick: error message should "cluster is nil"

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -607,3 +624,64 @@ func (r *ClusterBootstrapReconciler) watchProvider(providerRef *corev1.TypedLoca
...
+	}
+	if cluster.Spec.Topology == nil {
```
vijaykatam, line 667: or if `variableName` is empty

addons/controllers/testdata/test-cluster-bootstrap.yaml
```diff
@@ -20,6 +20,20 @@ metadata:
...
+        value: calico
 ---
```
vijaykatam, line 37: Could we add a test for the case when cluster variable does not exist?

addons/controllers/clusterbootstrap_controller.go
maralavi: No, cause now both ```CNIs``` and ```AdditionalPackages``` are slices, it doesn't let me to do that.

vijaykatam (commented):
Changes LGTM. Will approve once capi bump is merged and tests pass.

vijaykatam (approved):
LGTM. Good test coverage!

PR #1657, opened 2022-02-09
AWS filter out arm64 instance type options

### LINKED ISSUE #861: Kickstart UI should "recommend" some AWS instance types or be more selective in what it presents our users
[This issue originated in community-edition. It's largely feedback that the instance size list is very overwhelming.](https://github.com/vmware-tanzu/community-edition/issues/2134)

## Details

We should provide a selective list or file some way to have a better UX for choosing an instance size.

miclettej:
Adds filtering out of arm64-based instance types when retrieve lists of available instance types per region and availability zone. This filtering is behind a feature flag which is enabled by default.

Fixes #861 

Built and ran CLI locally, testing against my AWS account to confirm that instance types with CPU support architecture containing arm64 are not available in the UI.

### Release note
```release-note
ARM64 based instance types have been removed from lists of available instance types in the Kickstart UI
```

### PR Checklist


pkg/v1/tkg/aws/client.go
```diff
@@ -372,8 +374,9 @@ func (c *client) ListInstanceTypes(optionalAZName string) ([]string, error) {
 		if err != nil {
 			return nil, err
 		}
-		filteredInstances := filterInstanceType(candidates)
-		diffInstancesPerAz := getSetDifference(instances, filteredInstances)
+		// filter and build a list of unsupported instance types
+		unsupportedInstances := getUnsupportedInstanceTypes(candidates)
```
stmcginnis, line 378: Nothing blocking here, but just curious why we don't have one call here of `getSupportedInstanceTypes` rather than getting the unsupported and then filtering after that.

stmcginnis: Ah, nevermind. I've read through the following code a little closer and see there are multiple conditions where we need to filter out the unsupported instances from different sets of possible instance types.

stmcginnis (approved):
Looks good!

miclettej:
> Looks good to me, but appears there's now a merge conflict with one of the generated files.

Fixed, thank you!

PR #1480, opened 2022-01-12
Configure vSphere cpi to exclude kube-vip ip

adobley:
### What this PR does / why we need it
Prevents the assignment of the kube-vip IP to a node's IP.

Bumps the vsphere-cpi from 1.22.4 to 1.22.5 in `pkg/v1/providers`, a pre-requisite for this feature. The tkg-bom already includes 1.22.5, this PR brings the templates in sync.

### Which issue(s) this PR fixes

Fixes #

### Describe testing done for PR

We wrote the tests for the rendering of the provider templates. The functionality has been thoroughly tested in upstream.

Deployed cluster, ensured the VSPHERE_CONTROL_PLANE_ENDPOINT ip was not selected as the node ip on any nodes. Reviewed rendered configs and log output for expected information.  

### Release note

```release-note
Prevent the vsphere-cpi from assigning the VSPHERE_CONTROL_PLANE_ENDPOINT IP to node's ExternalIP when running kube-vip.
```

### PR Checklist

- [x] Squash the commits into one or a small number of logical commits
      
- [x] Use good commit [messages](https://github.com/vmware-tanzu/tanzu-framework/blob/main/CONTRIBUTING.md)
- [x] Ensure PR contains terms all contributors can understand and links all contributors can access

PR #1659, opened 2022-02-09
update ytt to 0.38.0

adobley:
this gives us access to the new IP module in ytt

This will be necessary for #1480 as the provider templates use the IP module

Ran all the tests

### Release note

```release-note
NONE
```

### PR Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -58,7 +58,7 @@ require (
 	github.com/k14s/imgpkg v0.17.0
 	github.com/k14s/kbld v0.32.0
 	github.com/k14s/semver/v4 v4.0.1-0.20210701191048-266d47ac6115
-	github.com/k14s/ytt v0.32.1-0.20210511155130-214258be2519
+	github.com/k14s/ytt v0.38.0
 	github.com/lithammer/dedent v1.1.0
 	github.com/logrusorgru/aurora v2.0.3+incompatible
 	github.com/mattn/go-isatty v0.0.14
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -64,7 +64,7 @@ require (
 	github.com/k14s/imgpkg v0.17.0
 	github.com/k14s/kbld v0.31.0
 	github.com/k14s/semver/v4 v4.0.1-0.20210701191048-266d47ac6115
-	github.com/k14s/ytt v0.32.1-0.20210511155130-214258be2519
+	github.com/k14s/ytt v0.37.0
```
yharish991, line 67: Question: is there a reason to upgrade ytt to 0.37.0? TKG BOM uses 0.38.0 and also we are using 0.38.0 in here https://github.com/vmware-tanzu/tanzu-framework/blob/main/hack/tools/Makefile#L29, should we upgrade it to 0.38.0 to keep it consistent across the project?

vuil: Seems reasonable, any concerns @adobley ?

adobley: I originally based it on the version of ytt that the kapp-controller we depend on in providers uses, but I'm happy to bump to 0.38.0.
I'll make those changes.

mcwumbly:
@adobley nit, but we should probably fixup that commit message to match the version number now too.

mcwumbly (approved):
lgtm.

mcwumbly:
need a review from @vmware-tanzu/tanzu-framework-reviewers

vuil (approved):
lgtm, thanks!

PR #1670, opened 2022-02-11
WIP: Adds webhook TLSCredentials datastructure.

adduarte:
webhook.TLSCredentials can hold the ca, and
server certificates, along with teh rsa keys that
can be used to stablish an TLS connection between
a webhook and the kube-apiserver

Fixes #

### Release note

```release-note

```

### PR Checklist

### DIFF
--- a/addons/main.go
+++ b/addons/main.go
@@ -29,14 +29,16 @@ import (
 	addonconfig "github.com/vmware-tanzu/tanzu-framework/addons/pkg/config"
 	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/constants"
 	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/crdwait"
+	"github.com/vmware-tanzu/tanzu-framework/addons/webhook"
 	runtanzuv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/run/v1alpha1"
 	runtanzuv1alpha3 "github.com/vmware-tanzu/tanzu-framework/apis/run/v1alpha3"
 	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/buildinfo"
 )
 
 var (
-	scheme   = runtime.NewScheme()
-	setupLog = ctrl.Log.WithName("setup")
+	scheme     = runtime.NewScheme()
+	setupLog   = ctrl.Log.WithName("setup")
+	webhooktls *webhook.TLSCredentials
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


addons/main.go
```diff
@@ -96,6 +98,16 @@ func main() {
...
+	webhooktls = &webhook.TLSCredentials{
+		CAName:       "self.signed.cert",
+		Organization: "tanzu.vmware",
+		Hostname:     "hostname",
+		DnsNames:     []string{"webhook-service", "webhook-service.default", "webhook-service.default.svc"},
+		ValidFrom:    time.Now(),
+		MaxAge:       time.Hour * 24 * 365,
+	}
+	webhooktls.GetServerRSAKeyPEM()
+
```
adduarte, line 110: This whole section does not belong here. It might be better to move it to the SetupWebhookWithManager function of each one of the webhook definitions.

addons/main.go
```diff
@@ -29,14 +29,16 @@ import (
...
 	runtanzuv1alpha3 "github.com/vmware-tanzu/tanzu-framework/apis/run/v1alpha3"
 	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/buildinfo"
 )
 
 var (
-	scheme   = runtime.NewScheme()
-	setupLog = ctrl.Log.WithName("setup")
+	scheme     = runtime.NewScheme()
+	setupLog   = ctrl.Log.WithName("setup")
+	webhooktls *webhook.TLSCredentials
```
adduarte, line 41: This might end up being a []*webhook.TLSCredentials if we have to define a serverCert for each package. Which might be the case because the hostname and dsnNames might need to match the service name.

addons/webhook/certificates.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+type TLSCredentials struct {
+	CAName       string
+	Organization string
+	Hostname     string
+	DnsNames     []string
+	ValidFrom    time.Time
+	MaxAge       time.Duration
+	CA_cert      *[]byte
```
adduarte, line 22: this should probably be private. (lowercase) best to create a function to provide the values if they are needed. Same as getPEM functions, something like getDER (the ca_cert, server_cert, and rsa keys are DER encoded). but it will be bes to force the caller to specify the encoding they want.

adduarte (commented):
needs some polishing

addons/main.go
adduarte: Take this out and instead add tests to addons/webhook/certificates_test.go

adduarte:
This works has been absorved into: https://github.com/vmware-tanzu/tanzu-framework/pull/1687

adduarte:
https://github.com/vmware-tanzu/tanzu-framework/pull/1687

Issue #1589, opened 2022-02-01
Package reconcile for additionalPackages


vijaykatam:
(This is used to request new product features)

**Describe the feature request**
All additionalPackages in TanzuClusterBootstrap CR need to be reconciled to a cluster. See placeholder TODOs in vmware-tanzu/tanzu-framework#1541

**Describe alternatives you've considered**

**Affected product area (please put an X in all that apply)**

- [ ] APIs
- [x] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Additional context**

danniel1205:
this issue has been addressed in https://github.com/vmware-tanzu/tanzu-framework/pull/1679.

Issue #1677, opened 2022-02-11
Clusterbootstrap controller might want to specify more fields on PackageInstall CR creation


danniel1205:
(This is used to request new product features)

**Describe the feature request**

`clusterbootstrap_controller` creates or patches PackageInstall CR for kapp-controller on the mgmt cluster and for other bootstrap packages on the workload cluster. As MVP, we only fill the minimum required fields. It is a good idea to do some research and see if any other fields need to be specified, i.e., `ext.packaging.carvel.dev/ytt-data-values-overlays` annotations, `pkgi.Spec.SyncPeriod`, etc.

**Describe alternatives you've considered**

NA

**Affected product area (please put an X in all that apply)**

- [ ] APIs
- [x] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Additional context**

danniel1205:
Related issue: https://github.com/vmware-tanzu/tanzu-framework/issues/1738

shivaani0505:
Change sync_period to 5 min

PR #1679, opened 2022-02-11
Add more reconciliation logic to clusterbootstrap controller for cluster creation

### LINKED ISSUE #1587: Package reconcile for Kapp,CNI, CPI, CSI

Package reconcile for Kapp, CNI, CPI and CSI packages in TanzuClusterBootstrap controller. See placeholder TODOs in vmware-tanzu/tanzu-framework#1541

- [ ] APIs
- [x] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Additional context**

danniel1205:

This PR adds more reconciliation logic to clusterbootstrap_controller.go:
- Create PackageInstall CR of kapp-controller on mgmt cluster, the created PackageInstall CR is for deploying kapp-controller resources on the workload cluster.
- Create service account, cluster role and cluster role binding for PackageInstall on workload cluster.
- Reconcile core bootstrap packages and create related resources on the workload cluster.
- Reconcile additional bootstrap packages and create related resources on workload cluster.
- Add positive tests to verify the reconcile logic.

This PR implements the logic to make cluster bootstrapping work as designed.

Fixes #1587, #1589

- Run `go test` on local env.

### Release note

```release-note
package-based-lcm: Add more reconciliation logic to clusterbootstrap controller for cluster creation
```

### PR Checklist


addons/controllers/clusterbootstrap_controller.go
```diff
@@ -203,7 +245,7 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
 	}
 
 	clusterBootstrapTemplate := &runtanzuv1alpha3.ClusterBootstrapTemplate{}
-	key := client.ObjectKey{Namespace: constants.TKGSystemNS, Name: tkrName}
+	key := client.ObjectKey{Namespace: r.Config.AddonNamespace, Name: tkrName}
```
blc1996, line 248: Nit: Not sure if `AddonNamespace` is a good name here.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +320,270 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+}
+
+// createOrPatchPackageOnRemote creates the Package CR on remote cluster. In order to install a package on remote cluster
+// the Package CR needs to be present.
+// createOrPatchPackageOnRemote returns a tuple: (<remote-package>, <error>)
+func (r *ClusterBootstrapReconciler) createOrPatchPackageOnRemote(cluster *clusterapiv1beta1.Cluster,
+	cbPkg *runtanzuv1alpha3.ClusterBootstrapPackage, clusterClient client.Client) (*kapppkgv1alpha1.Package, error) {
+
+	var err error
+	// Create or patch Package CR on remote cluster
```
blc1996, line 384: Question: Looks like we are not using remote App to deploy kapp controller any more. Instead, we are pushing package and pkgi to remote cluster, while not using pkgr?

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -313,16 +611,17 @@ func (r *ClusterBootstrapReconciler) createOrPatchPackageInstallSecret(cluster *
 			log.Info("provider status does not have secretRef", "provider", provider)
 			return nil, nil
 		}
-		key := client.ObjectKey{Namespace: namespace, Name: secretName}
+		key := client.ObjectKey{Namespace: cluster.Namespace, Name: secretName}
 		if err := r.Get(r.context, key, secret); err != nil {
 			log.Error(err, "unable to fetch secret", "objectkey", key)
 			return nil, err
 		}
 	}
```
blc1996, line 619: Looks like we are not handling inline case now?

blc1996 (commented):
Some questions and comments

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -7,11 +7,17 @@ import (
...
 	"errors"
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
+
+	kapppkgiv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/packaging/v1alpha1"
+	kapppkgv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apiserver/apis/datapackaging/v1alpha1"
+	versions "github.com/vmware-tanzu/carvel-vendir/pkg/vendir/versions/v1alpha1"
```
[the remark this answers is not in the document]
danniel1205: yes, that is what I thought, however golang-cli-lint complains that is not `goimports`ed if I combines. Let me see if I could do something to get rid of that linter complain.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -61,11 +68,12 @@ type ClusterBootstrapReconciler struct {
 }
 
 // NewClusterBootstrapReconciler returns a reconciler for ClusterBootstrap
-func NewClusterBootstrapReconciler(c client.Client, log logr.Logger, scheme *runtime.Scheme) *ClusterBootstrapReconciler {
+func NewClusterBootstrapReconciler(c client.Client, log logr.Logger, scheme *runtime.Scheme, config *addonconfig.Config) *ClusterBootstrapReconciler {
```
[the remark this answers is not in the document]
danniel1205: yea, ideally we should use a different config struct.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -164,21 +172,55 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
...
-	// handle additionalPackages
-	for _, additionalPkg := range clusterBootstrap.Spec.AdditionalPackages {
-		// TODO packageinstall (rbac, serviceaccount, package, packageinstall)
-		// https://github.com/vmware-tanzu/tanzu-framework/issues/1589
-		secret, err := r.createOrPatchPackageInstallSecret(cluster, additionalPkg, remoteClient, clusterBootstrap.Namespace, log)
-		if err != nil {
-			log.Error(err, "failed to createOrPatchPackageInstallSecret")
+	// Create a PackageInstall CR under the cluster namespace for deploying the kapp-controller on the remote cluster.
+	// We need kapp-controller to be deployed prior to CNI, CPI, CSI.
+	if err := r.createOrPatchKappPackageInstall(clusterBootstrap, cluster); err != nil {
```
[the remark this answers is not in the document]
danniel1205: the lcm of kapp-controller on mgmt is out of the scope of this PR. It should be handled separately as part of mgmt package.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -164,21 +172,55 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
...
+	// Create the ClusterRole on remote cluster, and bind it to the ServiceAccount created in above. kapp-controller
+	// reconciliation needs privileges.
+	if err := r.createOrPatchAddonRoleOnRemote(cluster, remoteClient); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Create or patch the resources for CNI, CPI, CSI to be running on the remote cluster.
+	// Those resources include Package CR, data value Secret, PackageInstall CR.
+	var corePackages []*runtanzuv1alpha3.ClusterBootstrapPackage
+	corePackages = append(corePackages, clusterBootstrap.Spec.CNI, clusterBootstrap.Spec.CPI, clusterBootstrap.Spec.CSI)
```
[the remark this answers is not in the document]
danniel1205: reflect works, but there are few reasons not doing that at this moment:
- Readability: Using reflect usually takes more time for other maintainers to understand the type conversion logic and hard for debugging.
- Complexity: We only have 3 types for core packages.

If in the future, we support more types which make us hard to enum, we probably could think of using reflect.

addons/controllers/clusterbootstrap_controller.go
danniel1205: no, it is not a good name. Switching to a different config struct will address this.

addons/controllers/clusterbootstrap_controller.go
danniel1205: correct.

addons/controllers/clusterbootstrap_controller.go
danniel1205: Inline is captured in a separate issue: https://github.com/vmware-tanzu/tanzu-framework/issues/1694. 

The reasons of addressing it separately are:
- Inline is not a small change including the tests. This PR combines 2 issues already.
- We try to make the MVP in first, so that we could start testing the MVP logic in real env earlier.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +319,270 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+	if err != nil {
+		return err
+	}
+	r.Log.Info(fmt.Sprintf("created the PackageInstall CR %s/%s on cluster %s/%s",
+		pkgi.Namespace, pkgi.Name, cluster.Namespace, cluster.Name))
+
+	return nil
+}
+
 // createOrPatchPackageInstallSecret creates or patches or the secret used for PackageInstall in a cluster
```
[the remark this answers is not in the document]
danniel1205: good catch, I guess it was a typo from previous code change. Fixed

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -164,21 +171,55 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
...
+		return ctrl.Result{}, err
+	}
+
+	// Create or patch the resources for CNI, CPI, CSI to be running on the remote cluster.
+	// Those resources include Package CR, data value Secret, PackageInstall CR.
+	var corePackages []*runtanzuv1alpha3.ClusterBootstrapPackage
+	corePackages = append(corePackages, clusterBootstrap.Spec.CNI, clusterBootstrap.Spec.CPI, clusterBootstrap.Spec.CSI)
+	for _, corePackage := range corePackages {
+		// The following nil check is redundant, we do not expect CNI, CPI, CSI to be nil and webhook should handle
+		// the validations against those fields. Having this nil check is mainly to allow local envtest could run when
```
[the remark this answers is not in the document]
danniel1205: fixed

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -164,21 +171,55 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
...
+		// any above component is missing.
+		if corePackage == nil {
+			continue
+		}
+		// There are different ways to have all the resources created or patched on remote cluster. Current solution is
+		// to handle packages in sequence order. I.e., Create all resources for CNI first, and then CPI, CSI. It is also
+		// possible to create all resources in a different order or in parallel. We will consider to use goroutines to create
+		// all resources in parallel on remote cluster if there is performance issue from sequential ordering.
+		if err := r.createOrPatchAddonResourcesOnRemote(cluster, corePackage, remoteClient); err != nil {
+			// For core packages, we require all their creation or patching to succeed, so if error happens against any of the
```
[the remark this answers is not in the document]
danniel1205: fixed

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +319,270 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
 	return nil, errors.New("should not happen")
 }
 
+// createOrPatchKappPackageInstall contains the logic that create/update PackageInstall CR for kapp-controller on
+// mgmt cluster. The kapp-controller runs on mgmt cluster reconciles the PackageInstall CR and creates kapp-controller resources
```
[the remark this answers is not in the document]
danniel1205: fixed

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +319,270 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+		remotePkgi.Spec.ServiceAccountName = r.Config.PkgiServiceAccount
+		remotePkgi.Spec.PackageRef = &kapppkgiv1alpha1.PackageRef{
+			RefName: remotePackageName,
+			VersionSelection: &versions.VersionSelectionSemver{
+				Prereleases: &versions.VersionSelectionSemverPrereleases{},
+			},
+		}
+		remotePkgi.Spec.Values = []kapppkgiv1alpha1.PackageInstallValues{
+			{SecretRef: &kapppkgiv1alpha1.PackageInstallValuesSecretRef{
+				Name: remotePackageName},
```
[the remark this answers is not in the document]
danniel1205: good catch, it should be `util.GenerateDataValueSecretName(cluster.Name, util.GetPackageShortName(cbPkg.RefName))`

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +319,270 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+	// addon config controller on local cluster.
+	remoteSecret, err := r.createOrPatchPackageInstallSecretOnRemote(cluster, cbPkg, clusterClient, r.Log)
+	// We expect there is NO error to create or patch the secret used for PackageInstall in a cluster.
+	if err != nil {
+		// Logging has been handled by createOrPatchPackageInstallSecret() already
+		return err
+	}
+	if remoteSecret == nil {
+		// The nil secret happens when tcbPkg.ValuesFrom is a ProviderRef and no data value secret for that provider is found
+		// on local cluster. This error occurs when handling additionalPackages. But it does not hurt to check nil here,
```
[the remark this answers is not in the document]
danniel1205: rephrased a little bit. Yes, it is the nil remoteSecret case.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +319,270 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+	r.Log.Info(fmt.Sprintf("created the Package CR %s on cluster %s/%s", remotePackage.Name, cluster.Namespace,
+		cluster.Name))
+
+	// Create or patch the data value secret on remote cluster. The data value secret has been generated by each
+	// addon config controller on local cluster.
+	remoteSecret, err := r.createOrPatchPackageInstallSecretOnRemote(cluster, cbPkg, clusterClient, r.Log)
+	// We expect there is NO error to create or patch the secret used for PackageInstall in a cluster.
+	if err != nil {
+		// Logging has been handled by createOrPatchPackageInstallSecret() already
+		return err
```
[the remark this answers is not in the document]
danniel1205: fixed

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +319,270 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+	}
+	if remoteSecret == nil {
+		// The nil secret happens when tcbPkg.ValuesFrom is a ProviderRef and no data value secret for that provider is found
+		// on local cluster. This error occurs when handling additionalPackages. But it does not hurt to check nil here,
+		// and return error which let the reconciler handles the retry.
+		// The detailed logging has been handled by createOrPatchPackageInstallSecret().
```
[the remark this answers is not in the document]
danniel1205: fixed

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +319,271 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+		// }
+		// ipkg.ObjectMeta.Annotations[addontypes.YttMarkerAnnotation] = ""
+		// ipkg.Spec.SyncPeriod = &metav1.Duration{Duration: r.Config.AppSyncPeriod}
+		// ipkg.Spec.ServiceAccountName = r.Config.AddonServiceAccount
+		pkgi.Spec.PackageRef = &kapppkgiv1alpha1.PackageRef{
+			RefName: clusterBootstrap.Spec.Kapp.RefName,
+			VersionSelection: &versions.VersionSelectionSemver{
+				Prereleases: &versions.VersionSelectionSemverPrereleases{},
+			},
+		}
```
shyaamsn, line 352: Question: Dont see the reference to Cluster and its kubeconfig here. Isnt it needed ?
https://github.com/vmware-tanzu/tanzu-framework/blob/main/addons/controllers/app_reconciler.go#L71

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +319,271 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+	}
+
+	pkgiMutateFn := func() error {
+		// TODO: Followup on the following fields, only add them if needed.
+		// https://github.com/vmware-tanzu/tanzu-framework/issues/1677
+		// if ipkg.ObjectMeta.Annotations == nil {
+		//	 ipkg.ObjectMeta.Annotations = make(map[string]string)
+		// }
+		// ipkg.ObjectMeta.Annotations[addontypes.YttMarkerAnnotation] = ""
+		// ipkg.Spec.SyncPeriod = &metav1.Duration{Duration: r.Config.AppSyncPeriod}
```
shyaamsn, line 345: Thought: We probably want to retain SyncPeriod to 5min similar to what was there in old addons controller.

addons/controllers/clusterbootstrap_controller.go
danniel1205: yes, for installing kapp-controller resources into remote workload cluster, we do need the reference to Cluster. Good catch.

addons/controllers/clusterbootstrap_controller.go
danniel1205: Yes, I left a TODO and filed an issue tracking this  https://github.com/vmware-tanzu/tanzu-framework/issues/1677. The reasons for leaving a TODO are because a) I guess it is not a must-to-have as MVP, I would follow up and add more fields like that if needed. b) after testing with real cluster, we would have better understanding on the user stories that requires the advanced configurations, and implement it all together.

addons/controllers/clusterbootstrap_controller_test.go
```diff
@@ -46,123 +53,219 @@ var _ = Describe("ClusterBootstrap Reconciler", func() {
...
+						if string(s.Data["values.yaml"]) != "foobar" {
+							return false
+						}
 
+						return true
+					}, waitTimeout, pollingInterval).Should(BeTrue())
+				})
+
+				By("verifying that kapp-controller PackageInstall CR is created under cluster namespace properly on the management cluster")
+				// Verify kapp-controller PackageInstall CR has been created under cluster namespace on management cluster
```
vijaykatam, line 176: kapp-controller is not a good example as it is preinstalled on mgmt cluster and we should not reconcile it, could we pick another package for the test and potentially assert that kapp-controller is skipped on mgmt cluster.

danniel1205: >kapp-controller is not a good example as it is preinstalled on mgmt cluster and we should not reconcile it

This test is actually asserting something different. Since kapp is one of the bootstrap packages, and when a cluster creation is initiated, a PackageInstall CR for kapp-controller will be created under a particular cluster namespace (for remote install).

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -35,6 +34,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/go-logr/logr"
```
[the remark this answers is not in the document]
danniel1205: `make go-lint` complains if putting it together with other github imports.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -164,21 +171,55 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
...
-	// handle additionalPackages
-	for _, additionalPkg := range clusterBootstrap.Spec.AdditionalPackages {
-		// TODO packageinstall (rbac, serviceaccount, package, packageinstall)
-		// https://github.com/vmware-tanzu/tanzu-framework/issues/1589
-		secret, err := r.createOrPatchPackageInstallSecret(cluster, additionalPkg, remoteClient, clusterBootstrap.Namespace, log)
-		if err != nil {
-			log.Error(err, "failed to createOrPatchPackageInstallSecret")
+	// Create a PackageInstall CR under the cluster namespace for deploying the kapp-controller on the remote cluster.
+	// We need kapp-controller to be deployed prior to CNI, CPI, CSI.
+	if err := r.createOrPatchKappPackageInstall(clusterBootstrap, cluster); err != nil {
```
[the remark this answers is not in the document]
danniel1205: I am assuming we want to skip ClusterBootstrap CR creation for mgmt cluster as well? And is the following [code](https://github.com/vmware-tanzu/tanzu-framework/blob/5ff74cc5c40b885027f5f6b8f9fc650300b503ca/addons/controllers/clusterbootstrap_controller.go#L199-L203) enforcing that ?

The whole bootstrap controller logic seems only applies to workload clusters, if so, I am thinking to add a predicate on watch.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +319,271 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+
+	pkgiMutateFn := func() error {
+		// TODO: Followup on the following fields, only add them if needed.
+		// https://github.com/vmware-tanzu/tanzu-framework/issues/1677
+		// if ipkg.ObjectMeta.Annotations == nil {
+		//	 ipkg.ObjectMeta.Annotations = make(map[string]string)
+		// }
+		// ipkg.ObjectMeta.Annotations[addontypes.YttMarkerAnnotation] = ""
+		// ipkg.Spec.SyncPeriod = &metav1.Duration{Duration: r.Config.AppSyncPeriod}
+		// ipkg.Spec.ServiceAccountName = r.Config.AddonServiceAccount
```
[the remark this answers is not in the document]
danniel1205: Cluster ref with kubeconfig should be sufficient. That is what we do today in app_reconciler. https://github.com/vmware-tanzu/tanzu-framework/blob/main/addons/controllers/app_reconciler.go#L68

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +319,271 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+		pkgi.Spec.PackageRef = &kapppkgiv1alpha1.PackageRef{
+			RefName: clusterBootstrap.Spec.Kapp.RefName,
```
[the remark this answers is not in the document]
danniel1205: fixed

addons/controllers/clusterbootstrap_controller.go
danniel1205: fixed

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +319,271 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+				Prereleases: &versions.VersionSelectionSemverPrereleases{},
+			},
+		}
+		pkgi.Spec.Values = []kapppkgiv1alpha1.PackageInstallValues{
+			{SecretRef: &kapppkgiv1alpha1.PackageInstallValuesSecretRef{
+				// The secret name could also be fetched from kappConfig.Spec.Status.SecretRef. However, the naming convention
+				// of that field is the same as the follow. To simplify the implementation and reduce the call to r.Client.Get()
+				// on kappConfig CR, use the util function to construct the secret name.
+				// TODO: Follow up with reviewer to see if this is feasible
+				Name: util.GenerateDataValueSecretName(cluster.Name, constants.KappControllerAddonName)},
```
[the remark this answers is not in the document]
danniel1205: If valuesFrom is a `Secret`, the actual secret obj name under cluster namespace has been mutated: https://github.com/vmware-tanzu/tanzu-framework/blob/5ff74cc5c40b885027f5f6b8f9fc650300b503ca/addons/controllers/clusterbootstrap_controller.go#L511. That means we could not simply fetch it from `valuesFrom` in this case.

If the valueFrom is a provider, the secret obj name is constructed by using above util function: https://github.com/vmware-tanzu/tanzu-framework/blob/5ff74cc5c4/addons/controllers/antrea/antreaconfig_controller.go#L180. In this case, we are able to fetch the secret name from `xxxConfig.Spec.Status.SecretRef`, but we have to be careful on error handling. Like, if a secret is not found, should we treat it as "no secret" or should we treat it as failure and blocks PKGI from creating.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +319,271 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+	if err = r.Client.Get(r.context, key, localPackage); err != nil {
+		// If there is an error to get the Carvel Package CR from local cluster, nothing needs to be created/cloned on remote.
+		// Let the reconciler try again.
+		r.Log.Error(err, fmt.Sprintf("unable to get package %s, nothing needs to be created on cluster %s/%s",
+			key.String(), cluster.Namespace, cluster.Name))
+		return nil, err
+	}
+	remotePackage := &kapppkgv1alpha1.Package{}
+	remotePackage.SetName(localPackage.Name)
+	// The Package CR on remote cluster needs to be under tkg-system namespace
```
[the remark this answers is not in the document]
danniel1205: fixed

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +320,280 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+	cbPkg *runtanzuv1alpha3.ClusterBootstrapPackage, clusterClient client.Client) (*kapppkgiv1alpha1.PackageInstall, error) {
+
+	// Create PackageInstall CRs on the remote workload cluster, kapp-controller will take care of reconciling them
+	remotePkgi := &kapppkgiv1alpha1.PackageInstall{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      util.GeneratePackageInstallName(cluster.Name, util.GetPackageShortName(cbPkg.RefName)),
+			Namespace: r.Config.BootstrapSystemNamespace,
+		},
+	}
+	remotePackageName := cbPkg.RefName
```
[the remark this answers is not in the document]
danniel1205: fixed

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +319,271 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+		return nil, err
+	}
+
+	return serviceAccount, nil
+}
+
+// createOrPatchAddonRoleOnRemote creates or patches the ClusterRole, ClusterRoleBinding on remote cluster.
+// The ClusterRole is bound to the ServiceAccount which is referenced by PackageInstall CR, so that kapp-controller on remote
+// cluster could have privileges to lifecycle manage package resources.
+func (r *ClusterBootstrapReconciler) createOrPatchAddonRoleOnRemote(cluster *clusterapiv1beta1.Cluster, clusterClient client.Client) error {
```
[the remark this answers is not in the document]
danniel1205: fixed

addons/controllers/clusterbootstrap_controller.go
danniel1205: well, I think I have made some changes to support the sync period. Above TODO and issue will track whether we should preserve additional annotations.

danniel1205: fixed

addons/controllers/clusterbootstrap_controller.go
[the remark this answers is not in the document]
danniel1205: added the logic to skip on management cluster within `createOrPatchKappPackageInstall`

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +319,271 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+func (r *ClusterBootstrapReconciler) createOrPatchKappPackageInstall(clusterBootstrap *runtanzuv1alpha3.ClusterBootstrap, cluster *clusterapiv1beta1.Cluster) error {
+	pkgi := &kapppkgiv1alpha1.PackageInstall{
+		ObjectMeta: metav1.ObjectMeta{
+			// The legacy addons controller uses <cluster name>-<addon name> convention for naming the PackageInstall CR.
+			// https://github.com/vmware-tanzu/tanzu-framework/blob/main/addons/controllers/package_reconciler.go#L195.
+			// util.GeneratePackageInstallName() follows the same pattern.
+			Name: util.GeneratePackageInstallName(cluster.Name, util.GetPackageShortName(clusterBootstrap.Spec.Kapp.RefName)),
+			// kapp-controller PackageInstall CR is installed under the same namespace as tanzuClusterBootstrap. The namespace
+			// is also the same as where the cluster belongs.
+			Namespace: clusterBootstrap.Namespace,
```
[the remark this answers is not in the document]
danniel1205: fixed

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +319,271 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+		// https://github.com/vmware-tanzu/tanzu-framework/issues/1677
+		// if ipkg.ObjectMeta.Annotations == nil {
+		//	 ipkg.ObjectMeta.Annotations = make(map[string]string)
+		// }
+		// ipkg.ObjectMeta.Annotations[addontypes.YttMarkerAnnotation] = ""
+		// ipkg.Spec.SyncPeriod = &metav1.Duration{Duration: r.Config.AppSyncPeriod}
+		// ipkg.Spec.ServiceAccountName = r.Config.AddonServiceAccount
+		pkgi.Spec.PackageRef = &kapppkgiv1alpha1.PackageRef{
+			RefName: clusterBootstrap.Spec.Kapp.RefName,
+			VersionSelection: &versions.VersionSelectionSemver{
```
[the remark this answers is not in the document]
danniel1205: fixed

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +320,280 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+			Namespace: r.Config.BootstrapSystemNamespace,
+		},
+	}
+	remotePackageName := cbPkg.RefName
+	_, err := controllerutil.CreateOrPatch(r.context, clusterClient, remotePkgi, func() error {
+		remotePkgi.Spec.ServiceAccountName = r.Config.PkgiServiceAccount
+		remotePkgi.Spec.PackageRef = &kapppkgiv1alpha1.PackageRef{
+			RefName: remotePackageName,
+			VersionSelection: &versions.VersionSelectionSemver{
+				Prereleases: &versions.VersionSelectionSemverPrereleases{},
```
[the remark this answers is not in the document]
danniel1205: fixed

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +320,280 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+			},
+		}
+		remotePkgi.Spec.Values = []kapppkgiv1alpha1.PackageInstallValues{
+			{SecretRef: &kapppkgiv1alpha1.PackageInstallValuesSecretRef{
```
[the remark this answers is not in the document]
danniel1205: fixed

addons/pkg/util/addon_util.go
```diff
@@ -8,16 +8,19 @@ import (
 	"fmt"
 	"net"
 	"strconv"
+	"strings"
 
-	"github.com/pkg/errors"
```
danniel1205, line 12: this is auto changed by `make fmt` from project root which makes `make go-lint` happy

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -242,15 +284,15 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(
 		clusterBootstrap.Namespace = cluster.Namespace
 		clusterBootstrap.Spec = clusterBootstrapTemplate.Spec.DeepCopy()
 		// get selected CNI and populate clusterBootstrap.Spec.CNIs with it
-		cniPackage, err := r.getCNI(clusterBootstrapTemplate, cluster, log)
+		cniPackage, err := r.getCNIClusterBootstrapPackage(clusterBootstrapTemplate, cluster, log)
```
maralavi, line 287: Would it be better to be called ```getCNIForClusterBootstrap```?

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -677,20 +1101,33 @@ func (r *ClusterBootstrapReconciler) getCNI(
 		return nil, err
 	}
 	if selectedCNI != "" {
+		var getPkgMetadataErrs []error
+		foundCNI := false
 		for _, cni := range clusterBootstrapTemplate.Spec.CNIs {
-			if selectedCNI == packageShortName(cni.RefName) {
+			// Package should be available in cluster namespace
+			pkgRefName, _, getPkgMetadataErr := util.GetPackageMetadata(r.context, r.Client, cni.RefName, cluster.Namespace)
```
maralavi, line 1108: Wouldn't using just ```err``` be easier to read than ```getPkgMetadataErr```?

addons/controllers/clusterbootstrap_controller_test.go
```diff
@@ -38,298 +46,393 @@ var _ = Describe("ClusterBootstrap Reconciler", func() {
...
-				var foundClusterOwnerRef bool
-				var foundClusterBootstrapOwnerRef bool
-				var foundLabels bool
-				var foundType bool
-				for _, ownerRef := range s.GetOwnerReferences() {
-					if ownerRef.UID == cluster.UID {
-						foundClusterOwnerRef = true
+				// TODO
+				By("CNI thing")
+				//// Verify CNI is populated in the cloned object with the value from the cluster variables definitions
```
maralavi, line 108: These commented tests will be fixed in a future PR?

addons/controllers/testdata/test-cluster-bootstrap-2.yaml
```diff
@@ -28,47 +20,195 @@ spec:
...
-          apiGroup: cni.tanzu.vmware.com
-          kind: CalicoConfig
-          name: test-cluster-tcbt-2
-    - refName: "antrea.tanzu.vmware.com.21.2.0--vmware.1-tkg.1-rc.1"
-      valuesFrom:
-        providerRef:
-          apiGroup: cni.tanzu.vmware.com
-          kind: AntreaConfig
-          name: test-cluster-tcbt-2
-
```
maralavi, line 54: Why removing all these? Isn't that why CNIs tests have been failing and now commented?

addons/controllers/testdata/test-cluster-bootstrap-2.yaml
```diff
@@ -28,47 +20,195 @@ spec:
...
+            rawOptions:
+              - --wait-timeout=5m
+              - --kube-api-qps=20
+              - --kube-api-burst=30
 ---
-apiVersion: cni.tanzu.vmware.com/v1alpha1
-kind: CalicoConfig
+apiVersion: data.packaging.carvel.dev/v1alpha1
+kind: Package
 metadata:
```
maralavi, line 68: Why these are removed? These were to test CNI selecltion

danniel1205: that is because when envtest get set up for clsuterbootstrap_controller_test.go, test-cluster-bootstrap-1.yaml has already contains those resources and get deployed. No need to duplicate them here.

addons/controllers/testdata/test-cluster-bootstrap-2.yaml
danniel1205: that is because when envtest get set up for clsuterbootstrap_controller_test.go, test-cluster-bootstrap-1.yaml has already contains those resources and get deployed. No need to duplicate them here.

addons/controllers/clusterbootstrap_controller_test.go
danniel1205: Those are the tests I am still fixing.

addons/controllers/clusterbootstrap_controller.go
danniel1205: there are few reasons: 
1. Shadowing: using `err` everywhere is simple, but it could shadow previous assigned `err`.
2. Self-Explanation: Using a self-explanation variable could help reader understand what kind of error it would hold, and simplifies the error type check. I.e., `if errors.IsNotFound(err)`
3. Scope: This scope of above error in particular is only for L1103-L1123, that error is not supposed to be a global accessible error.

addons/controllers/clusterbootstrap_controller.go
danniel1205: yes, nice catch.

addons/controllers/clusterbootstrap_controller_test.go
danniel1205: done now.

addons/controllers/kappcontrollerconfig_controller_test.go
```diff
@@ -9,11 +9,12 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
```
danniel1205, line 12: this new line is auto generated by `make fmt` from root make target, that make linter happy

addons/controllers/clusterbootstrap_controller.go
maralavi: If defined correctly it wouldn't shadow previous definitions. In my opinion using a single variable err is more readable:) But that is more programmer's choice I guess

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -300,79 +343,371 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(
...
+			},
 		}
-		provider, err := r.dynamicClient.Resource(*gvr).Namespace(namespace).Get(r.context, pkg.ValuesFrom.ProviderRef.Name, metav1.GetOptions{}, "status")
+		return nil
+	}
 
-		if err != nil {
-			log.Error(err, "unable to fetch provider", "provider", pkg.ValuesFrom.ProviderRef, "gvr", gvr)
-			return nil, err
+	_, err = controllerutil.CreateOrPatch(r.context, r.Client, pkgi, pkgiMutateFn)
```
maralavi, line 425: Question: As the installation is local in the management cluster (under workload cluster namespace), we use ```r.client``` as opposed to ```remoteClient```. Right?

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -300,79 +343,371 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(
...
+					UID:        cluster.UID,
+				},
+			},
+		},
 	}
 
-	if pkg.ValuesFrom.ProviderRef != nil {
-		gvr, err := r.getGVR(schema.GroupKind{Group: *pkg.ValuesFrom.ProviderRef.APIGroup, Kind: pkg.ValuesFrom.ProviderRef.Kind})
+	pkgiMutateFn := func() error {
+		// TODO: Followup on the following fields, only add them if needed.
```
maralavi, line 387: I believe we should also add a label containing the cluster-name. I can add that later

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -300,79 +343,371 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(
...
+	cbPkg *runtanzuv1alpha3.ClusterBootstrapPackage, clusterClient client.Client) (*kapppkgv1alpha1.Package, error) {
+
+	var err error
+	// Create or patch Package CR on remote cluster
+	localPackage := &kapppkgv1alpha1.Package{}
+	key := client.ObjectKey{Namespace: cluster.Namespace, Name: cbPkg.RefName}
+	if err = r.Client.Get(r.context, key, localPackage); err != nil {
+		// If there is an error to get the Carvel Package CR from local cluster, nothing needs to be created/cloned on remote.
+		// Let the reconciler try again.
+		r.Log.Error(err, fmt.Sprintf("unable to get package %s, nothing needs to be created on cluster %s/%s",
```
maralavi, line 448: Somehow the part for ```nothing needs to be created on cluster``` in the error message seems misleading to me

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -300,79 +343,371 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(
...
+		// Package.Spec.RefName and Package.Spec.Version are required fields for Package CR. We do not expect them to be
+		// empty and error should not happen when fetching them from a Package CR.
+		r.Log.Error(err, fmt.Sprintf("unable to fetch Package.Spec.RefName or Package.Spec.Version from Package %s/%s on cluster %s/%s",
+			r.Config.SystemNamespace, cbPkg.RefName, cluster.Namespace, cluster.Name))
+		return nil, err
+	}
+
+	// Create PackageInstall CRs on the remote workload cluster, kapp-controller will take care of reconciling them
+	remotePkgi := &kapppkgiv1alpha1.PackageInstall{
+		ObjectMeta: metav1.ObjectMeta{
```
maralavi, line 487: Same as above for kapp pkgi, I believe we should also add a label containing the cluster-name. I can add that later

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -300,79 +343,371 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(
...
+	// The secret will be created or patched under tkg-system namespace on remote cluster
+	remoteSecret.Namespace = r.Config.SystemNamespace
+	remoteSecret.Type = corev1.SecretTypeOpaque
 
 	dataValuesSecretMutateFn := func() error {
-		dataValuesSecret.Data = map[string][]byte{}
-		for k, v := range secret.Data {
-			dataValuesSecret.Data[k] = v
+		remoteSecret.Data = map[string][]byte{}
+		for k, v := range localSecret.Data {
```
maralavi, line 699: Shouldn't you use ```patchedSecret``` cause ```localSecret``` might have been patched?

addons/controllers/clusterbootstrap_controller.go
danniel1205: yes, that is correct.

addons/controllers/clusterbootstrap_controller.go
danniel1205: I filed an GH issue to add all required label/annotations in one shot. I did not blindly adding them at the moment is because I don't know what are the use cases for adding them. 

Ref: https://github.com/vmware-tanzu/tanzu-framework/issues/1677

addons/controllers/clusterbootstrap_controller.go
danniel1205: Re-phrased it a little bit: `unable to create or patch Package %s on cluster %s/%s. Error occurs when getting Package %s from the management cluster`

addons/controllers/clusterbootstrap_controller.go
danniel1205: yes, you are right.

danniel1205: fixed

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +320,280 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+// running on remote cluster will reconcile it and deploy resources.
+func (r *ClusterBootstrapReconciler) createOrPatchPackageInstallOnRemote(cluster *clusterapiv1beta1.Cluster,
+	cbPkg *runtanzuv1alpha3.ClusterBootstrapPackage, clusterClient client.Client) (*kapppkgiv1alpha1.PackageInstall, error) {
+
+	// Create PackageInstall CRs on the remote workload cluster, kapp-controller will take care of reconciling them
+	remotePkgi := &kapppkgiv1alpha1.PackageInstall{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      util.GeneratePackageInstallName(cluster.Name, util.GetPackageShortName(cbPkg.RefName)),
+			Namespace: r.Config.BootstrapSystemNamespace,
+		},
```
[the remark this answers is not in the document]
danniel1205: Filed GH issue as a followup to add required labels/annotations on PKGI as one shot. We might want to do some research to construct a list of required annotations.

https://github.com/vmware-tanzu/tanzu-framework/issues/1677

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +319,271 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+
+// createOrPatchAddonServiceAccountOnRemote creates or patches the addon ServiceAccount on remote cluster.
+// The ServiceAccount will be referenced by the PackageInstall CR, so that kapp-controller on remote cluster could consume
+// for PackageInstall reconciliation.
+func (r *ClusterBootstrapReconciler) createOrPatchAddonServiceAccountOnRemote(cluster *clusterapiv1beta1.Cluster, clusterClient client.Client) (*corev1.ServiceAccount, error) {
+	serviceAccount := &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      r.Config.PkgiServiceAccount,
+			Namespace: r.Config.BootstrapSystemNamespace,
+		},
```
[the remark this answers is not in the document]
danniel1205: Filed an good-first GH issue to check it: https://github.com/vmware-tanzu/tanzu-framework/issues/1738

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -278,14 +320,280 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(c
...
+// createOrPatchAddonResourcesOnRemote creates or patches the resources for a cluster bootstrap package on remote workload
+// cluster. The resources are [Package CR, Secret for PackageInstall, PackageInstall CR].
+func (r *ClusterBootstrapReconciler) createOrPatchAddonResourcesOnRemote(cluster *clusterapiv1beta1.Cluster,
+	cbPkg *runtanzuv1alpha3.ClusterBootstrapPackage, clusterClient client.Client) error {
+
+	remotePackage, err := r.createOrPatchPackageOnRemote(cluster, cbPkg, clusterClient)
+	if err != nil {
+		return err
+	}
+	r.Log.Info(fmt.Sprintf("created the Package CR %s on cluster %s/%s", remotePackage.Name, cluster.Namespace,
```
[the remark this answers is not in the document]
danniel1205: tracked as a followup: https://github.com/vmware-tanzu/tanzu-framework/issues/1739

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -664,7 +1031,64 @@ func (r *ClusterBootstrapReconciler) watchProvider(providerRef *corev1.TypedLoca
...
+			// In this case, we expect the secretRef to be present under status subresource and its value gets updated by
+			// the corresponding controller. However, the config controller might not create the secret in time.
+			r.Log.Info("provider status does not have secretRef", "GVR", gvr)
+			return "", nil
+		}
+		return secretName, nil
+	}
+
+	// When valuesFrom is not nil, but either valuesFrom.Inline, valuesFrom.SecretRef, or valuesFrom.providerRef is empty or nil,
+	// we interpret it as the data value secret for that package has not been available yet. One of those three fields needs
```
vijaykatam, line 1081: Should be a custom error if at all to allow skipping of the package. Lets log an issue to get this use case covered.

addons/controllers/clusterbootstrap_controller.go
vijaykatam: +1, err is not special here

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -606,26 +952,47 @@ func (r *ClusterBootstrapReconciler) updateValuesFromProvider(cluster *clusterap
...
 			providerLabels[types.ClusterNameLabel] = cluster.Name
 		}
 
-		newProvider.SetName(fmt.Sprintf("%s-%s-package", cluster.Name, packageShortName(pkg.RefName)))
-		log.Info("cloning provider", "provider", newProvider)
-		newProvider, err = r.dynamicClient.Resource(*gvr).Namespace(bootstrap.Namespace).Create(r.context, newProvider, metav1.CreateOptions{})
+		newProvider.SetName(fmt.Sprintf("%s-%s-package", cluster.Name, pkgRefName))
+		newProvider.SetNamespace(bootstrap.Namespace)
+		log.Info(fmt.Sprintf("cloning provider %s/%s to namespace %s", cbTemplateNamespace, newProvider.GetName(), bootstrap.Namespace), "gvr", gvr)
+		// newProvider and createdOrUpdatedProvider are different. The newProvider is the one we want apiserver to accept,
```
vijaykatam, line 970: this comment makes the creation seem a lot more complex than it actually is

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -460,32 +804,32 @@ func (r *ClusterBootstrapReconciler) ensureOwnerRef(ownerRef *metav1.OwnerRefere
...
-	return nil
-}
-
-func packageShortName(refName string) string {
-	if refName != "" {
-		refParts := strings.Split(refName, ".")
-		if len(refParts) > 0 {
-			return refParts[0]
+		newProvider = newProvider.DeepCopy()
+		newProvider.SetOwnerReferences(clusterapiutil.EnsureOwnerRef(provider.GetOwnerReferences(), *ownerRef))
```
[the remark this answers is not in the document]
blc1996: non-blocking

addons/controllers/clusterbootstrap_controller.go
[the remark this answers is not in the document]
danniel1205: made a change to put `"github.com/pkg/errors"` together with `sigs.xxx`. Ideally I would expect linter to be happy when grouping `"github.com/pkg/errors"` with other `github.xxx`.

addons/controllers/clusterbootstrap_controller.go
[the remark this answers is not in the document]
danniel1205: L699 is content-agnostic, it just copy whatever the secret.data contains. Logically the same as doing a copy.

addons/controllers/clusterbootstrap_controller.go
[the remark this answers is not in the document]
danniel1205: It is about the ordering. We are able to add ownerRef only when the referenced obj gets created. L312 explains it in comments.

addons/controllers/clusterbootstrap_controller.go
danniel1205: making this change and adding that comment is because there is a bug on the original code when there are errors on create.
```
newProvider, err = r.dynamicClient.Resource(*gvr).Namespace(bootstrap.Namespace).Create(r.context, newProvider, metav1.CreateOptions{})
```

`newProvider` is nil and the code returns it. `ensureOwnerRef` uses that nil object to update ownerRef which causes NPE.

We might need a follow up to refactor the code a little bit.

maralavi (approved):

blc1996 (approved):
LGTM

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -300,79 +342,371 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(
...
+	})
+	if err != nil {
+		r.Log.Error(err, fmt.Sprintf("unable to create or patch Package resource %s/%s on cluster: %s/%s",
+			remotePackage.Namespace, remotePackage.Name, cluster.Namespace, cluster.Name))
+		return nil, err
+	}
+	return remotePackage, nil
+}
+
+// createOrPatchPackageInstallOnRemote creates or patches PackageInstall CR on remote cluster. The kapp-controller
```
maralavi, line 469: Question: Are core/additional packages installed in another way (using management packages) for management cluster, and are they installed under the tkg-system?

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -300,79 +342,371 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(
...
+// running on remote cluster will reconcile it and deploy resources.
+func (r *ClusterBootstrapReconciler) createOrPatchPackageInstallOnRemote(cluster *clusterapiv1beta1.Cluster,
```
maralavi, line 471: Don't you need to check for ```constants.ManagementClusterRoleLabel``` to not end up installing the package in the management cluster here?

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -544,52 +887,54 @@ func (r *ClusterBootstrapReconciler) updateValuesFromSecret(cluster *clusterapiv
...
+			newSecret.Labels[types.PackageNameLabel] = util.ParseStringForLabel(pkg.RefName)
+			newSecret.Labels[types.ClusterNameLabel] = cluster.Name
+			// Set secret.Type to ClusterBootstrapManagedSecret to enable us to Watch these secrets
+			newSecret.Type = constants.ClusterBootstrapManagedSecret
+			return nil
+		})
+		if createOrPatchErr != nil {
+			return nil, createOrPatchErr
 		}
 		pkg.ValuesFrom.SecretRef = newSecret.Name
```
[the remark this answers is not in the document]
danniel1205: yea, pkg is a pointer so its filed will be updated.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -300,79 +342,371 @@ func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(
...
+func (r *ClusterBootstrapReconciler) createOrPatchKappPackageInstall(clusterBootstrap *runtanzuv1alpha3.ClusterBootstrap, cluster *clusterapiv1beta1.Cluster) error {
+	// Skip if the cluster object represents the management cluster
+	if _, exists := cluster.Labels[constants.ManagementClusterRoleLabel]; exists {
+		r.Log.Info(fmt.Sprintf("cluster %s/%s is management cluster, skip creating or patching the PackageInstall CR for kapp-controller", cluster.Namespace, cluster.Name))
+		return nil
+	}
 
-	secret := &corev1.Secret{}
+	// In order to create PackageInstall CR, we need to get the Package.Spec.RefName and Package.Spec.Version
+	packageRefName, packageVersion, err := util.GetPackageMetadata(r.context, r.Client, clusterBootstrap.Spec.Kapp.RefName, cluster.Namespace)
```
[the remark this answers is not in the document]
danniel1205: for core packages like kapp-controller, we do not allow nil and webhook should reject the request.

addons/controllers/clusterbootstrap_controller.go
danniel1205: only kapp-controller on utkg mgmt cluster is installed in a different way, others are following the same pattern

addons/controllers/clusterbootstrap_controller.go
danniel1205: we only skip kapp-controller installation on mgmt cluster. There is an check at L365. Other packages are following the same pattern to be installed on mgmt cluster.

vijaykatam (approved):
LGTM pending downstream tests

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -606,26 +994,47 @@ func (r *ClusterBootstrapReconciler) updateValuesFromProvider(cluster *clusterap
...
+		// created provider objects, there will errors. I.e., [invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update]
+		createdOrUpdatedProvider, err = r.dynamicClient.Resource(*gvr).Namespace(bootstrap.Namespace).Create(r.context, newProvider, metav1.CreateOptions{})
 		if err != nil {
-			log.Error(err, "unable to clone provider", "provider", newProvider, "gvr", gvr)
+			// There are possibilities that current reconciliation loop fails due to various reasons, and during next reconciliation
+			// loop, it is possible that the provider resource has been created. In this case, we want to run update/patch.
+			if apierrors.IsAlreadyExists(err) {
+				createdOrUpdatedProvider, err = r.dynamicClient.Resource(*gvr).Namespace(bootstrap.Namespace).Update(r.context, newProvider, metav1.UpdateOptions{})
+				if err != nil {
+					log.Info(fmt.Sprintf("updated provider %s/%s", createdOrUpdatedProvider.GetNamespace(), createdOrUpdatedProvider.GetName()), "gvr", gvr)
```
maralavi, line 1025: You should have nilness checked ```createdOrUpdatedProvider``` before accessing their metadata.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -606,26 +994,47 @@ func (r *ClusterBootstrapReconciler) updateValuesFromProvider(cluster *clusterap
...
+		// won't. The intent of this function is to return the actual object pointer that apiserver creates, so we should
+		// return createdOrUpdatedProvider instead of newProvider. Otherwise, when the caller wants to make changes to the
+		// created provider objects, there will errors. I.e., [invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update]
+		createdOrUpdatedProvider, err = r.dynamicClient.Resource(*gvr).Namespace(bootstrap.Namespace).Create(r.context, newProvider, metav1.CreateOptions{})
 		if err != nil {
-			log.Error(err, "unable to clone provider", "provider", newProvider, "gvr", gvr)
+			// There are possibilities that current reconciliation loop fails due to various reasons, and during next reconciliation
+			// loop, it is possible that the provider resource has been created. In this case, we want to run update/patch.
+			if apierrors.IsAlreadyExists(err) {
+				createdOrUpdatedProvider, err = r.dynamicClient.Resource(*gvr).Namespace(bootstrap.Namespace).Update(r.context, newProvider, metav1.UpdateOptions{})
```
maralavi, line 1023: I see issues with my own yaml files added around this line. When the resource already exists, the update fails with the following :
```
kappcontrollerconfigs.run.tanzu.vmware.com "test-wlc-kapp-controller.tanzu.vmware.com-package" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update
```

```resourceVersion``` is already set in ```createdOrUpdatedProvider``` after resource Get, but seems like ```newProvider``` lacks it.  Maybe usingPatch would have fixed this?

File: addons/controllers/clusterbootstrap_controller.go [hunk 55]
```diff
@@ -606,26 +994,47 @@ func (r *ClusterBootstrapReconciler) updateValuesFromProvider(cluster *clusterap
...
+				if err != nil {
+					log.Info(fmt.Sprintf("updated provider %s/%s", createdOrUpdatedProvider.GetNamespace(), createdOrUpdatedProvider.GetName()), "gvr", gvr)
+					return newProvider, nil
```
maralavi, line 1026: Also in that case when the update fails, ```createdOrUpdatedProvider``` would be nil and seems like you needed to record the name and the ```createdOrUpdatedProvider.GetName()``` & ```createdOrUpdatedProvider.GetName()``` into another variables after resource Get; as their access here can end up in nil pointer access for when the update fails

addons/controllers/clusterbootstrap_controller.go
maralavi: Non blocker for the merge. But just wanted to point this out as can cause problem later on.

addons/controllers/clusterbootstrap_controller.go
maralavi: And your log message here needs to get improved anyway, as apparently the update has failed here, not succeeded

addons/controllers/clusterbootstrap_controller.go
blc1996: Didn't look into it. The error message looks like it's trying to update a resource that haven't been created.

maralavi: No, the resource already exist. Few lines above Create fails with AlreadyExisting. But seems like some required fields like ```resourceVersion``` are missing in the ```newProvider``` which is passed to Update

alfredthenarwhal:
> @danniel1205:  /test install-vc7
Commit: dc0240c

Build failed! Build no: 1589

danniel1205:
> > @danniel1205:  /test install-vc7
> > Commit: [dc0240c](https://github.com/vmware-tanzu/tanzu-framework/commit/dc0240ce20c0fc3e5c9f244c9268ec253692cf83)
> 
> Build failed! Build no: 1589

Downstream pipeline machinery is broken, and the PR does not change anything on legacy code path. Skip downstream pipeline for now.

alfredthenarwhal:
> @danniel1205:  /test install-vc7
Commit: dc0240c

Tests failed! Build no: 1590

PR #1682, opened 2022-02-11
Use TKR label to decide whether addon controller or cluster bootstrap…

### LINKED ISSUE #1582: Addons controller behavior for TKRs v1alpha3
Existing addons-controller should reconcile only when there is a BOM ConfigMap associated with the TKR. TKRs upgraded from v1alpha1 to v1alpha3 will continue to reference BOM ConfigMap but new v1alpha3 TKR versions will not have BOM ConfigMap and should not be reconciled by addon_controller.go

- [ ] APIs
- [x] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Additional context**

### LINKED ISSUE #1674: Use TKR label on clusters to decide which controller processes addons
*  Legacy clusters will use the label `tanzuKubernetesRelease`
* Classy and new style clusters will use the label `run.tanzu.vmware.com/tkr`

Goals 
1. addons_controller.go should only process if it sees label `tanzuKubernetesRelease`
2. clusterbootstrap_controller.go should only process if it sees label `run.tanzu.vmware.com/tkr`
3. Info log when cluster is skipped.

Possible options
* Update `WithEventFilter` in  SetupWithManager

maralavi:

Use TKR label to decide whether addon controller or cluster bootstrap controller should reconcile cluster

Fixes https://github.com/vmware-tanzu/tanzu-framework/issues/1582
Fixes https://github.com/vmware-tanzu/tanzu-framework/issues/1674

Added unit tests

### Release note

```release-note
package-based-lcm: use TKR label to decide whether addon controller or cluster bootstrap controller should reconcile cluster
```

### PR Checklist

### DIFF
--- a/addons/controllers/addon_controller.go
+++ b/addons/controllers/addon_controller.go
@@ -88,6 +88,7 @@ func (r *AddonReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager
 		).
 		WithOptions(options).
 		WithEventFilter(clusterApiPredicates.ResourceNotPaused(r.Log)).
+		WithEventFilter(addonpredicates.ClusterHasLabel(constants.TKRLabel, r.Log)).
 		Build(r)
 	if err != nil {
 		r.Log.Error(err, "Error creating an addon controller")
@@ -115,10 +116,11 @@ func (r *AddonReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ct
 		return ctrl.Result{}, err
 	}
 
-	tkrName := util.GetTKRNameForCluster(ctx, r.Client, cluster)
-	if tkrName == "" {
-		log.Info("cluster does not have an associated TKR")
-		return ctrl.Result{}, nil
+	tkrName := cluster.Labels[constants.TKRLabel]
+	tkr, err := util.GetTKRByName(ctx, r.Client, tkrName)
+	if err != nil || tkr == nil {
+		log.Error(err, "unable to fetch TKR object", "name", tkrName)
+		return ctrl.Result{}, err
 	}
 
 	log.Info("Reconciling cluster")
--- a/addons/controllers/addon_controller.go
+++ b/addons/controllers/addon_controller.go
@@ -123,9 +123,9 @@ func (r *AddonReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ct
 		return ctrl.Result{}, err
 	}
 
-	// if tkr is not found, let the reconciler try again
+	// if tkr is not found, should not requeue for the reconciliation
 	if tkr == nil {
-		log.Info("TKR object not found")
+		log.Info("TKR object not found", "name", tkrName)
 		return ctrl.Result{}, nil
 	}
 
--- a/addons/controllers/addon_controller.go
+++ b/addons/controllers/addon_controller.go
@@ -118,9 +118,15 @@ func (r *AddonReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ct
 
 	tkrName := cluster.Labels[constants.TKRLabel]
 	tkr, err := util.GetTKRByName(ctx, r.Client, tkrName)
-	if err != nil || tkr == nil {
+	if err != nil {
 		log.Error(err, "unable to fetch TKR object", "name", tkrName)
 		return ctrl.Result{}, err
+	}
+
+	// if tkr is not found, let the reconciler try again
+	if tkr == nil {
+		log.Info("TKR object not found")
+		return ctrl.Result{}, nil
 	}
 
 	log.Info("Reconciling cluster")
--- a/addons/controllers/clusterbootstrap_controller.go
+++ b/addons/controllers/clusterbootstrap_controller.go
@@ -141,9 +141,9 @@ func (r *ClusterBootstrapReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return ctrl.Result{}, err
 	}
 
-	// if tkr is not found, let the reconciler try again
+	// if tkr is not found, should not requeue for the reconciliation
 	if tkr == nil {
-		log.Info("TKR object not found")
+		log.Info("TKR object not found", "name", tkrName)
 		return ctrl.Result{}, nil
 	}
 
--- a/addons/controllers/clusterbootstrap_controller.go
+++ b/addons/controllers/clusterbootstrap_controller.go
@@ -136,8 +136,14 @@ func (r *ClusterBootstrapReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	// make sure the TKR object exists
 	tkrName := cluster.Labels[constants.TKRLabelClassyClusters]
 	tkr, err := util.GetTKRByName(r.context, r.Client, tkrName)
-	if err != nil || tkr == nil {
+	if err != nil {
 		log.Error(err, "unable to fetch TKR object", "name", tkrName)
+		return ctrl.Result{}, err
+	}
+
+	// if tkr is not found, let the reconciler try again
+	if tkr == nil {
+		log.Info("TKR object not found")
 		return ctrl.Result{}, nil
 	}
 
--- a/addons/controllers/clusterbootstrap_controller.go
+++ b/addons/controllers/clusterbootstrap_controller.go
@@ -92,6 +92,7 @@ func (r *ClusterBootstrapReconciler) SetupWithManager(ctx context.Context, mgr c
 		).
 		WithOptions(options).
 		WithEventFilter(clusterApiPredicates.ResourceNotPaused(r.Log)).
+		WithEventFilter(predicates.ClusterHasLabel(constants.TKRLabelClassyClusters, r.Log)).
 		Build(r)
 	if err != nil {
 		r.Log.Error(err, "Error creating an addon controller")
@@ -132,11 +133,14 @@ func (r *ClusterBootstrapReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return ctrl.Result{}, err
 	}
 
-	tkrName := util.GetTKRNameForCluster(r.context, r.Client, cluster)
-	if tkrName == "" {
-		log.Info("cluster does not have an associated TKR")
+	// make sure the TKR object exists
+	tkrName := cluster.Labels[constants.TKRLabelClassyClusters]
+	tkr, err := util.GetTKRByName(r.context, r.Client, tkrName)
+	if err != nil || tkr == nil {
+		log.Error(err, "unable to fetch TKR object", "name", tkrName)
 		return ctrl.Result{}, nil
 	}
+
 	log.Info("Reconciling cluster")
 
 	// if deletion timestamp is set, handle cluster deletion
@@ -193,15 +197,11 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
 
 // createOrPatchclusterBootstrapFromTemplate will get, clone or update a ClusterBootstrap associated with a cluster
 // all linked secret refs and object refs are cloned into the same namespace as clusterBootstrap
-func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(cluster *clusterapiv1beta1.Cluster,
+func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(
+	cluster *clusterapiv1beta1.Cluster,
 	log logr.Logger) (*runtanzuv1alpha3.ClusterBootstrap, error) {
 
-	tkrName := util.GetTKRNameForCluster(r.context, r.Client, cluster)
-	if tkrName == "" {
-		log.Info("cluster does not have an associated TKR")
-		return nil, nil
-	}
-
+	tkrName := cluster.Labels[constants.TKRLabelClassyClusters]
 	clusterBootstrapTemplate := &runtanzuv1alpha3.ClusterBootstrapTemplate{}
 	key := client.ObjectKey{Namespace: constants.TKGSystemNS, Name: tkrName}
 	if err := r.Client.Get(r.context, key, clusterBootstrapTemplate); err != nil {
--- a/addons/pkg/util/cluster_util.go
+++ b/addons/pkg/util/cluster_util.go
@@ -107,40 +107,9 @@ func GetClusterClient(ctx context.Context, currentClusterClient client.Client, s
 	return c, nil
 }
 
-// GetTKRForCluster gets the TKR for cluster
-func GetTKRForCluster(ctx context.Context, c client.Client, cluster *clusterv1beta1.Cluster) (*runtanzuv1alpha1.TanzuKubernetesRelease, error) {
-	if c == nil || cluster == nil {
-		return nil, nil
-	}
-
-	tkrName := GetTKRNameForCluster(ctx, c, cluster)
-	if tkrName == "" {
-		return nil, nil
-	}
-
-	tkr, err := GetTKRByName(ctx, c, tkrName)
-	if err != nil {
-		return nil, err
-	}
-
-	return tkr, nil
-}
-
-// GetTKRNameForCluster get the TKR name for the cluster
-func GetTKRNameForCluster(ctx context.Context, c client.Client, cluster *clusterv1beta1.Cluster) string {
-	if c == nil || cluster == nil {
-		return ""
-	}
-
-	return cluster.Labels[constants.TKRLabel]
-}
-
-// GetBOMForCluster gets the bom associated with the cluster
+// GetBOMForCluster gets the bom associated with the legacy-style TKGm cluster
 func GetBOMForCluster(ctx context.Context, c client.Client, cluster *clusterv1beta1.Cluster) (*bomtypes.Bom, error) {
-	tkrName := GetTKRNameForCluster(ctx, c, cluster)
-	if tkrName == "" {
-		return nil, nil
-	}
+	tkrName := cluster.Labels[constants.TKRLabel]
 
 	bom, err := GetBOMByTKRName(ctx, c, tkrName)
 	if err != nil {
--- a/addons/predicates/tkr.go
+++ b/addons/predicates/tkr.go
@@ -4,6 +4,7 @@
 package predicates
 
 import (
+	"reflect"
 	"strings"
 
 	"github.com/go-logr/logr"
@@ -41,16 +42,21 @@ func ClusterHasLabel(label string, logger logr.Logger) predicate.Funcs {
 	}
 }
 
+// processIfClusterHasLabel determines if the input object is a cluster with a non-empty
+// value for the specified label. For other input object types, it returns true
 func processIfClusterHasLabel(label string, obj client.Object, logger logr.Logger) bool {
 	kind := obj.GetObjectKind().GroupVersionKind().Kind
-	cluster := &clusterapiv1beta1.Cluster{}
+	clusterKind := reflect.TypeOf(clusterapiv1beta1.Cluster{}).Name()
 
-	if kind == "" || kind != cluster.Kind {
+	if kind != clusterKind {
 		return true
 	}
 
-	if _, ok := obj.GetLabels()[label]; ok {
-		return true
+	labels := obj.GetLabels()
+	if labels != nil {
+		if l, ok := labels[label]; ok && l != "" {
+			return true
+		}
 	}
 
 	log := logger.WithValues("namespace", obj.GetNamespace(), strings.ToLower(kind), obj.GetName())
--- a/addons/predicates/tkr.go
+++ b/addons/predicates/tkr.go
@@ -4,7 +4,11 @@
 package predicates
 
 import (
+	"strings"
+
 	"github.com/go-logr/logr"
+	clusterapiv1beta1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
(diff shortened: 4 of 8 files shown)


addons/controllers/addon_controller.go
```diff
@@ -115,7 +115,19 @@ func (r *AddonReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ct
...
+	// These clusters need to get reconciled by the cluster bootstrap controller.
+	// only legacy-style clusters are labeled with "tanzuKubernetesRelease"
+	// and should be reconciled by the addons controller.
+	for k := range cluster.Labels {
+		if k == constants.TKRLabelClassyClusters {
+			return ctrl.Result{}, nil
+		}
+	}
+
+	tkrName := util.GetTKRNameForCluster(ctx, r.Client, cluster, constants.TKRLabel)
```
vijaykatam, line 130: Could we add the check to WithEventFilter? https://github.com/vmware-tanzu/tanzu-framework/blob/d5687c28ead83902696943eed40aeaf0efced0c5/addons/controllers/addon_controller.go#L90

vijaykatam (commented):
Will there be a case when both the tkr labels exist and which one should take precedence?

maralavi:
> Will there be a case when both the tkr labels exist and which one should take precedence?

@vijaykatam If I see any ```constants.TKRLabelClassyClusters``` among the labels, ```TanzuClusterBootstrapController``` will process the request (in line 125 of addon_controller.go)

addons/controllers/addon_controller.go
maralavi: @vijaykatam The label only exists in the cluster object and not other resources we watch. For that reason, can't add the check to ```WithEventFilter``` as the resources are not converted to cluster object yet. Please correct me otherwise.

vijaykatam: Please check how `clusterApiPredicates.ResourceNotPaused` works. It checks for annotations. You would be checking for labels.

maralavi: I've looked at ```clusterApiPredicates.ResourceNotPaused```. The pause annotation can exist for all resources we watch (cluster, secret, KubeadmControlPlane, configmap, tkr). But the tkr label only exists for the cluster object out of those. That's why I'm afraid adding the predicate for tkr labels in WithEventFilter would not address our concern.

vijaykatam: Could you check object meta for Cluster Kind so that it ignores other types? I am confused why the checks need to be duplicated if it is already in the event filter.

maralavi: Yes I'll check for the type. For the cases where we convert other kinds to cluster (https://github.com/vmware-tanzu/tanzu-framework/blob/main/addons/controllers/addon_controller.go#L63) for example ```AddonSecretToClusters```,  which they call ```EnqueueRequestsFromMapFunc``` directly after getting converted, will the same high level ```WithEventFilter```  be sufficient ? Please let me know if I was not clear

imikushin:
IMO, addon_controller should *only* reconcile clusters that have the legacy `tanzuKubernetesRelease` label and ignore any others.

clusterbootstrap_controller should only consider clusters that have the new `run.tanzu.vmware.com/tkr` label and ignore any others.

imikushin:
Therefore it's best to add these conditions to the `WithEventFilter()` predicates when building the controller. 

For example, the addon_controller should not be concerned if the cluster is "classy". Instead, it should ignore any clusters beside those with the `tanzuKubernetesRelease` label.

vijaykatam:
@imikushin - Just to confirm, a cluster will never have both tkr labels?

imikushin:
> @imikushin - Just to confirm, a cluster will never have both tkr labels?

Correct. Unless someone intentionally applies both :) 

Also, it's important, IMO, to not touch clusters that are _not_ associated with a TKR: that may be intentional, and we don't want to interfere in such cases.

addons/controllers/addon_controller.go
```diff
@@ -115,7 +119,19 @@ func (r *AddonReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ct
...
+	// are not labeled with any TKR related labels
+	// (in case of not having run.tanzu.vmware.com/resolve-tkr annotation).
+	// These clusters need to get reconciled by the cluster bootstrap controller.
+	// only legacy-style clusters are labeled with "tanzuKubernetesRelease"
+	// and should be reconciled by the addons controller.
+	for k := range cluster.Labels {
+		if k == constants.TKRLabelClassyClusters {
+			return ctrl.Result{}, nil
+		}
+	}
```
imikushin, line 132: We don't need this check if we've already filtered out anything else but clusters with `tanzuKubernetesRelease` label.

addons/controllers/addon_controller.go
```diff
@@ -115,7 +119,19 @@ func (r *AddonReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ct
...
+
+	tkrName := util.GetTKRNameForCluster(ctx, r.Client, cluster, constants.TKRLabel)
```
imikushin, line 134: We can get a label value with
```suggestion
	tkrName := cluster.Labels[constants.TKRLabel]
```

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -132,7 +132,8 @@ func (r *ClusterBootstrapReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return ctrl.Result{}, err
 	}
 
-	tkrName := util.GetTKRNameForCluster(r.context, r.Client, cluster)
+	// TODO: What about those classy clusters that aren't labeled with any TKR related labels?
```
imikushin, line 135: We should leave such clusters alone :)

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -132,7 +132,8 @@ func (r *ClusterBootstrapReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+	tkrName := util.GetTKRNameForCluster(r.context, r.Client, cluster, constants.TKRLabelClassyClusters)
```
imikushin, line 136: Please get the label directly. By this point we should have filtered out any clusters that are not carrying the `run.tanzu.vmware.com/tkr` label.
 
```suggestion
	tkrName := cluster.Labels[constants.TKRLabelClassyClusters]
```

maralavi:
> > @imikushin - Just to confirm, a cluster will never have both tkr labels?
> 
> Correct. Unless someone intentionally applies both :)
> 
> Also, it's important, IMO, to not touch clusters that are _not_ associated with a TKR: that may be intentional, and we don't want to interfere in such cases.

@imikushin  What about “classy” clusters that don’t have any TKR labels? None of these controllers should process those?

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -196,7 +197,8 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
 func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(cluster *clusterapiv1beta1.Cluster,
 	log logr.Logger) (*runtanzuv1alpha3.ClusterBootstrap, error) {
 
-	tkrName := util.GetTKRNameForCluster(r.context, r.Client, cluster)
+	// TODO: What about those classy clusters that aren't labeled with any TKR related labels?
```
imikushin, line 200: We should leave them alone :)

imikushin:
Exactly!

vijaykatam:
@maralavi - Looks like all we need is a predicate function, we should not be changing any of GetTKRNameForCluster. Lets leave those as is and just add an event filter. Something like below should probably work.

addons_controller.go
```
...
		WithEventFilter(clusterApiPredicates.ResourceNotPaused(r.Log)).
		WithEventFilter(addonpredicates.ClusterHasLabel("tanzuKubernetesRelease", r.Log)).
...
```

clusterbootstrap_controller.go
```
		WithEventFilter(clusterApiPredicates.ResourceNotPaused(r.Log)).
		WithEventFilter(predicates.ClusterHasLabel("run.tanzu.vmware.com/tkr", r.Log)).
```

tkr.go
```
func ClusterHasLabel(label string, logger logr.Logger) predicate.Funcs {
	return predicate.Funcs{
		UpdateFunc: func(e event.UpdateEvent) bool {
			return processIfClusterHasLabel(label, e.ObjectNew, logger.WithValues("predicate", "updateEvent"))
		},
		CreateFunc: func(e event.CreateEvent) bool {
			return processIfClusterHasLabel(label, e.Object, logger.WithValues("predicate", "createEvent"))
		},
		DeleteFunc: func(e event.DeleteEvent) bool {
			return processIfClusterHasLabel(label, e.Object, logger.WithValues("predicate", "deleteEvent"))
		},
		GenericFunc: func(e event.GenericEvent) bool {
			return processIfClusterHasLabel(label, e.Object, logger.WithValues("predicate", "genericEvent"))
		},
	}
}

func processIfClusterHasLabel(label string, obj client.Object, logger logr.Logger) bool {
	kind := strings.ToLower(obj.GetObjectKind().GroupVersionKind().Kind)
	cluster := &clusterapiv1beta1.Cluster{}
	// only cluster objects, ignore others
	if kind != cluster.Kind {
		return true
	}

	_, ok := obj.GetLabels()[label]
	if ok {
		return true
	}
	log := logger.WithValues("namespace", obj.GetNamespace(), kind, obj.GetName())
	log.V(6).Info("Cluster resource does not have label", "label", label)
	return false
}
```

imikushin:
Imagine someone who created their own cluster class, perhaps by following a tutorial for upstream Cluster API that doesn't have any idea about TKRs or BootstrapTemplates. We should not create surprises for them.

vijaykatam:
That's right, we already skip processing if there is no associated TKR object. This check will add a prereq for even getting into reconcile loop.

maralavi:
@vijaykatam I've added the EvenFilter in the my last commit. I can refine that a bit, and remove the other checks. My concern was about the order in which ```WithEventFilter``` gets applied relative to the ```EventHandler``` for the watches. If this get applied to all the cluster objects obtained after applying the ```EventHandler```, then we are good. Asked in controller-runtime public channel to make sure.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -196,7 +197,8 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
...
+	tkrName := util.GetTKRNameForCluster(r.context, r.Client, cluster, constants.TKRLabelClassyClusters)
```
imikushin, line 201: We should already know the TKR name here. So, no need to get and check. We could just pass it as an argument to createOrPatchclusterBootstrapFromTemplate().

maralavi:
> WithEventFilter

addons/pkg/util/cluster_util.go
```diff
@@ -107,37 +107,26 @@ func GetClusterClient(ctx context.Context, currentClusterClient client.Client, s
...
 
-	return cluster.Labels[constants.TKRLabel]
+	return ""
 }
 
-// GetBOMForCluster gets the bom associated with the cluster
+// GetBOMForCluster gets the bom associated with the legacy-style TKGm cluster
 func GetBOMForCluster(ctx context.Context, c client.Client, cluster *clusterv1beta1.Cluster) (*bomtypes.Bom, error) {
-	tkrName := GetTKRNameForCluster(ctx, c, cluster)
+	tkrName := GetTKRNameForCluster(ctx, c, cluster, constants.TKRLabel)
```
imikushin, line 129: There's only one usage of this func: from `(r *AddonReconciler) reconcileNormal()`. Again, at the time we're in reconcileNormal we already know the TKR name, so no need to obtain and check for it here.

imikushin: In fact, this function should now become useless. So, it's probably a good idea to deprecate it.

maralavi: Well, TKR versions v1alpha1 still need it?

maralavi:
Addressed all the comments

addons/controllers/addon_controller.go
```diff
@@ -115,12 +116,6 @@ func (r *AddonReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ct
 		return ctrl.Result{}, err
 	}
 
-	tkrName := util.GetTKRNameForCluster(ctx, r.Client, cluster)
```
vijaykatam, line 118: I am wondering if we should keep this because the predicate only checks for the existence of a label key and not the actual existence of a TKR object and corresponding BOM configmap. Note that BOM configmap is only needed for addon_controller and not clusterbootstrap_controller.go

Without the check reconcileNormal of addon_controller will return an error and keep re-queuing if BOM is not found.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -193,15 +190,11 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
 
 // createOrPatchclusterBootstrapFromTemplate will get, clone or update a ClusterBootstrap associated with a cluster
 // all linked secret refs and object refs are cloned into the same namespace as clusterBootstrap
-func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(cluster *clusterapiv1beta1.Cluster,
+func (r *ClusterBootstrapReconciler) createOrPatchclusterBootstrapFromTemplate(
+	cluster *clusterapiv1beta1.Cluster,
+	tkrName string,
```
vijaykatam, line 195: tkrName doesn't need to be passed as it can be obtained from cluster object

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -193,15 +190,11 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
...
 	log logr.Logger) (*runtanzuv1alpha3.ClusterBootstrap, error) {
 
-	tkrName := util.GetTKRNameForCluster(r.context, r.Client, cluster)
```
vijaykatam, line 199: Here we should check if the tkr object actually exists.

addons/controllers/addon_controller.go
maralavi: This function ```GetTKRNameForCluster()``` was solely checking the label previously, it wasn't checking if the tkr object existed. Please note that ```GetTKRForCluster()``` was not used at all.

addons/predicates/tkr.go
```diff
@@ -18,3 +22,38 @@ func TKR(log logr.Logger) predicate.Funcs {
...
+			return processIfClusterHasLabel(label, e.Object, logger.WithValues("predicate", "genericEvent"))
+		},
+	}
+}
+
+func processIfClusterHasLabel(label string, obj client.Object, logger logr.Logger) bool {
+	kind := obj.GetObjectKind().GroupVersionKind().Kind
+	cluster := &clusterapiv1beta1.Cluster{}
+
+	if kind == "" || kind != cluster.Kind {
```
imikushin, line 48: `cluster.Kind` is always ` == ""` here (based on `cluster := &clusterapiv1beta1.Cluster{}` in the above line).

We don't need to check for `kind == ""`: we simply need to get Cluster kind properly. We can do that, e.g. like this:
```go
var clusterKind = reflect.TypeOf(clusterapiv1beta1.Cluster{}).Name()
```
Which is best to do as a standalone var (not every time when you need it).

BTW, do we have a unit test for this? This function looks to be perfectly unit-testable.

addons/predicates/tkr.go
```diff
@@ -18,3 +22,38 @@ func TKR(log logr.Logger) predicate.Funcs {
...
+		return true
+	}
+
+	if _, ok := obj.GetLabels()[label]; ok {
```
imikushin, line 52: If `obj.Labels == nil`, this is going to panic. So, need to check for that first.

addons/predicates/tkr.go
```diff
@@ -18,3 +22,38 @@ func TKR(log logr.Logger) predicate.Funcs {
...
+		return true
```
imikushin, line 53: Perhaps also need to make sure the value is not empty.

imikushin (commented):
This change needs unit tests

addons/predicates/tkr.go
maralavi: @imikushin I had to add this check for ```kind == "" ``` as the the tests for addons controller are using dynamic client to create resources from unstructured map created by parsing the yaml. https://github.com/vmware-tanzu/tanzu-framework/blob/main/addons/testutil/test_helpers.go#L60
After debugging, I found out that the kind field of the resource does not get populated there (and doesn't get updated if I try to update it later). So these check for the emptiness of the kind field was to cover those. Tried a lot but didn't find any way to fix that seemingly a bug in test utils

maralavi: You are right. I'll add unit test for this function.

maralavi: @imikushin You were right! !!```reflect.TypeOf(clusterapiv1beta1.Cluster{}).Name()``` is the correct way to go. I wonder there are so many usages of ```cluster := &clusterapiv1beta1.Cluster{}``` followed by ```cluster.Kind``` in our code base, but I just checked and the output would be an empty string. Not sure how those are working!

addons/predicates/tkr.go
```diff
@@ -18,3 +23,43 @@ func TKR(log logr.Logger) predicate.Funcs {
...
+			return processIfClusterHasLabel(label, e.Object, logger.WithValues("predicate", "genericEvent"))
+		},
+	}
+}
+
+// processIfClusterHasLabel determines if the input object is a cluster with a non-empty
+// value for the specified label. For other input object types, it returns true
+func processIfClusterHasLabel(label string, obj client.Object, logger logr.Logger) bool {
+	kind := obj.GetObjectKind().GroupVersionKind().Kind
+	clusterKind := reflect.TypeOf(clusterapiv1beta1.Cluster{}).Name()
```
imikushin, line 49: nit: There's no need to calculate `clusterKind` this every time `processIfClusterHasLabel()` is called. It's basically a constant, so we can take it out of the func body into a standalone var:
```go
var clusterKind = reflect.TypeOf(clusterapiv1beta1.Cluster{}).Name()
```

addons/predicates/tkr.go
```diff
@@ -18,3 +23,43 @@ func TKR(log logr.Logger) predicate.Funcs {
...
+			return processIfClusterHasLabel(label, e.Object, logger.WithValues("predicate", "deleteEvent"))
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return processIfClusterHasLabel(label, e.Object, logger.WithValues("predicate", "genericEvent"))
+		},
+	}
+}
+
+// processIfClusterHasLabel determines if the input object is a cluster with a non-empty
+// value for the specified label. For other input object types, it returns true
```
imikushin, line 46: Thanks for adding the comment!

addons/controllers/addon_controller.go
```diff
@@ -115,10 +116,11 @@ func (r *AddonReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ct
...
-	if tkrName == "" {
-		log.Info("cluster does not have an associated TKR")
-		return ctrl.Result{}, nil
+	tkrName := cluster.Labels[constants.TKRLabel]
+	tkr, err := util.GetTKRByName(ctx, r.Client, tkrName)
```
imikushin, line 120: We need to add a check to `GetTKRByName()` that the returned error is not `IsNotFound(err)`:
```go
	if err := c.Get(ctx, tkrNamespaceName, tkr); err != nil && !apierrors.IsNotFound(err) {
		return nil, err
	}
```

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -132,11 +133,14 @@ func (r *ClusterBootstrapReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-	if tkrName == "" {
-		log.Info("cluster does not have an associated TKR")
+	// make sure the TKR object exists
+	tkrName := cluster.Labels[constants.TKRLabelClassyClusters]
+	tkr, err := util.GetTKRByName(r.context, r.Client, tkrName)
+	if err != nil || tkr == nil {
+		log.Error(err, "unable to fetch TKR object", "name", tkrName)
 		return ctrl.Result{}, nil
```
imikushin, line 141: We don't want to swallow `err`.
```suggestion
		return ctrl.Result{}, err
```

The only case when we want that, (looks like it) should be handled by `util.GetTKRByName()` and it's when `apierrors.IsNotFound(err) == true`

addons/predicates/tkr_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+		When("cluster label does not match the input label", func() {
+			BeforeEach(func() {
+				clusterObj.Labels = map[string]string{"otherLabel": testTKRLabel}
+				result = processIfClusterHasLabel(constants.TKRLabel, clusterObj, log)
+			})
+			It("should return false", func() {
+				Expect(result).To(BeFalse())
+			})
+		})
```
imikushin, line 76: nit: This looks like a duplicate of the previous test case

imikushin (commented):
Left the last (I hope) round of comments.

We're almost there!

addons/controllers/addon_controller.go
maralavi: I just changed with a minor difference that I return nil in ```GetTKRByName()``` for the object on ```apierrors.IsNotFound(err)``` to easier check this case in the reconcile functions

addons/predicates/tkr_test.go
maralavi: Aah yes, thanks for catching it

maralavi:
> Left the last (I hope) round of comments.
> 
> We're almost there!

Thanks @imikushin . Just addressed them now

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -132,11 +133,20 @@ func (r *ClusterBootstrapReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-		log.Info("cluster does not have an associated TKR")
+	// make sure the TKR object exists
+	tkrName := cluster.Labels[constants.TKRLabelClassyClusters]
+	tkr, err := util.GetTKRByName(r.context, r.Client, tkrName)
+	if err != nil {
+		log.Error(err, "unable to fetch TKR object", "name", tkrName)
+		return ctrl.Result{}, err
+	}
+
+	// if tkr is not found, let the reconciler try again
```
imikushin, line 144: Same comment as https://github.com/vmware-tanzu/tanzu-framework/pull/1682/files#r808287608

imikushin (approved):
Good work @maralavi!

addons/controllers/addon_controller.go
```diff
@@ -115,9 +116,16 @@ func (r *AddonReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ct
...
-	if tkrName == "" {
-		log.Info("cluster does not have an associated TKR")
+	tkrName := cluster.Labels[constants.TKRLabel]
+	tkr, err := util.GetTKRByName(ctx, r.Client, tkrName)
+	if err != nil {
+		log.Error(err, "unable to fetch TKR object", "name", tkrName)
+		return ctrl.Result{}, err
+	}
+
+	// if tkr is not found, let the reconciler try again
```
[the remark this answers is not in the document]
maralavi: Oh I knew it that nil error would not requeue anymore. Guess I added a wrong comment as I wrote this pretty late. Thanks for catching it. I'll fix in a minute.

addons/controllers/addon_controller.go
vijaykatam: You are right, I got confused by the similarities in the function names. Checking for BOM would be great in the addons_controller.

addons/predicates/tkr.go
vijaykatam: yup, the Kind is indeed empty. Could you log an issue for some of the other controllers that are doing this?

vijaykatam (approved):
LGTM, minor non-blocking comment.

addons/predicates/tkr.go
maralavi: Sure @vijaykatam. Filed [this issue](https://github.com/vmware-tanzu/tanzu-framework/issues/1711)

addons/controllers/addon_controller.go
maralavi: @vijaykatam BOM is already checked in the ```ReconcileNormal()``` logic```bom, err := util.GetBOMForCluster(ctx,r.Client, cluster)```.
And as far I can tell, ```ReconcileDelete()``` wouldn't need the BOM to exist.

PR #1686, opened 2022-02-12
fix docker builds

### LINKED ISSUE #1685: Fix docker builds for controller images
**Bug description**

The Docker client sends the entire "build context" to the Docker daemon. That build context is the entire project root directory including the generated CLI plugin artifacts, which we don't need to build the controller images in framework repo. By fixing this we can significantly save docker build time

- [ ] APIs
- [ ] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [x] Test and Release
- [ ] User Experience

**Expected behavior**

yharish991:
This PR reduces docker build time by using .dockerignore file to exclude files(cli plugin artifacts) that are not needed for building the images

Fixes https://github.com/vmware-tanzu/tanzu-framework/issues/1685

ran `make docker-all` and verified that the images are built

### Release note

```release-note
NONE
```

### PR Checklist

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+artifacts/
+artifacts-admin/
+hack/
+packages/
+docs/
+test/


addons/Dockerfile.dockerignore
```diff
@@ -0,0 +1,2 @@
+artifacts
```
rajathagasthya, line 1: Just a single `.dockerignore` in repo root is enough (note the file needs to be named just `.dockerignore`). It's not necessary to have it in all these directories since all the `docker-build` targets `cd` to repo root for passing context to Docker.

nit: add a trailing `/` to these entries indicate they are directories.

rajathagasthya (changes requested):

.dockerignore
```diff
@@ -0,0 +1,2 @@
+artifacts/
```
yharish991, line 1: there are other directories that can be added here, but this is good starting to reduce the build time significantly

addons/Dockerfile.dockerignore
rajathagasthya: I didn't know about `Dockerfile.dockerignore`. Thanks for the explanation!

rajathagasthya (approved):
LGTM. This is a really useful change!

stmcginnis (approved):
Looks good - thanks for cleaning that up!

PR #1687, opened 2022-02-14
Adds support for webhooks using self generated certificates

adduarte:
    Adds support for webhooks using self generated certificates
    
    Allow webhooks to use certificate pairs
    created by the addon-manager itself instead of relying on an
    external service like cert-manager.
    
    - Server certificate and key are written to
      CertDir for webhook sever access.
    
    - Validating and Defaulting Webhook configurations
      are updated with the CA certificate

Fixes 1583

ginkgo testing with provided _test.go:
    - verified certificates and keys are generated and written to managers.webhook.CertDir
    - verified webhook validating and defaulting configurations are updated with generated CA cert
    - verified certificates can be validated.
    - all testing done through controller-runtime envtest. 
    - unit tests and integration tests submitted with pr. They are all passing.

### Release note

``` 
package-based-lcm: Adds support for webhooks using self generated certificates
```

### PR Checklist


addons/main.go
```diff
@@ -29,14 +29,16 @@ import (
...
 	runtanzuv1alpha3 "github.com/vmware-tanzu/tanzu-framework/apis/run/v1alpha3"
 	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/buildinfo"
 )
 
 var (
-	scheme   = runtime.NewScheme()
-	setupLog = ctrl.Log.WithName("setup")
+	scheme     = runtime.NewScheme()
+	setupLog   = ctrl.Log.WithName("setup")
+	webhooktls *webhook.TLSCredentials
```
adduarte, line 41: probably don't need this.

addons/main.go
```diff
@@ -96,6 +98,18 @@ func main() {
...
+		"hostname",
+		[]string{"webhook-service", "webhook-service.default", "webhook-service.default.svc"},
+		time.Now(),
+		time.Hour*24*365)
+	if err != nil {
+		setupLog.Error(err, "Unable to generate tls certificates for webhooks")
+	} else {
+		webhooktls.WriteFiles()
+	}
+
```
adduarte, line 112: This whole section needs to be rework. 
We still need to get the list of webhooks which will need to be updated with the correct ca certificates.

addons/webhook/certificates.go
```diff
@@ -0,0 +1,203 @@
...
+	for idx, webhook := range mWebhook.Webhooks {
+		if bytes.Equal(webhook.ClientConfig.CABundle, caCert) {
+			continue
+		}
+		updated = true
+		webhook.ClientConfig.CABundle = caCert
+		mWebhook.Webhooks[idx] = webhook
+	}
+	return updated
+}
```
adduarte, line 203: This two functions area almost idential, but for the different types in their signature. There is probably a better way to do this. Maybe using interfaces.

addons/webhook/certificates.go
```diff
@@ -0,0 +1,203 @@
...
+// must be named tls.key and tls.crt, respectively.
+
+//We need to come up with the functions below but for now we can use the assumed defaults
+func certfile_path() string {
+	return "/tmp/k8s-webhook-server/serving-certs/tsl.crt"
+}
+
+func keyfile_path() string {
+	return "/tmp/k8s-webhook-server/serving-certs/tsl.key"
+}
```
adduarte, line 177: This are using the default from the documentation but the paths can actually be pulled from the manager. 
Probably better to add the paths as an argument to NewTLSCredentials.

adduarte (commented):
needs more work

addons/webhook/certificates.go
```diff
@@ -0,0 +1,203 @@
+package webhook
+
+import (
```
vijaykatam, line 3: knative has a well defined and fairly isolated cert creation utility. Could we leverage https://github.com/knative/pkg/blob/main/webhook/certificates/resources/certs.go#L144 instead of rolling our own?

adduarte: that sounds like a good idea. let me see if it fits all the pieces. It looks like it unless we need the CAkey but I don't think we need it.

addons/webhook/certificates.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	if _, err := client.AdmissionregistrationV1().MutatingWebhookConfigurations().Update(context.TODO(), mWhCfg, metav1.UpdateOptions{}); err != nil {
+		return fmt.Errorf("error updating addons CA cert of MutatingWebhookConfiguration %s: %v", name, err)
+	}
+
+	return nil
+}
+
+//func UpdateTLSSecret(ctx context.Context, name, namespace string, maxAge time.Duration, certDir, tlsCertName, tlsKeyName string) (*corev1.Secret, error) {
+func UpdateTLSSecret(name string, wh *webhook.Server, namespace string) (*corev1.Secret, error) {
```
adduarte, line 100: It would be great to be able to use the create secret from https://github.com/knative/pkg/blob/3c00bb0157b9242359df6868e720861b6807773f/webhook/certificates/resources/secret.go#L45, but that one has a timelimit of one week hard coded, also we would need to create exctra functions to write the cert to the correct files. Might as well do it here below and be done.

addons/webhook/certificates.go
adduarte: irrelevant now

addons/webhook/certificates.go
```diff
@@ -0,0 +1,125 @@
...
+
+	//update the webhooks with the new CA from the secret created
+	k8sConfig, err := ctrl.GetConfig()
+	if err != nil {
+		return err
+	}
+	client, err := kubernetes.NewForConfig(k8sConfig)
+	if err != nil {
+		return err
+	}
```
adduarte, line 45: It would be nice to have obtain the "client" (kubernetes.interface) from the manager itself, but I am not sure if that is possible.

addons/main.go
```diff
@@ -147,6 +148,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	//name of all the validating webhook configurations. the following need to be populatec with the names of the
+	//validating and webhooks configurations.
+	var vWhCfgNames []string
+	var mWhCfgNames []string
```
adduarte, line 154: This section needs to be done after all the webhooks are registered with the manager and we have all the names of the webhooks. we definetily don't want to hardcode the names.

addons/webhook/certificates.go
```diff
@@ -0,0 +1,125 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func UpdateValidatingWebhook(client kubernetes.Interface, name string, secret *corev1.Secret) error {
+
+	vWhCfg, err := client.AdmissionregistrationV1().ValidatingWebhookConfigurations().Get(context.TODO(), name, metav1.GetOptions{})
```
danniel1205, line 66: Question: Do we want to use `context.TODO()` or using https://github.com/vmware-tanzu/tanzu-framework/blob/main/addons/main.go#L97 ?

addons/main.go
```diff
@@ -147,6 +148,12 @@ func main() {
...
+	webhook.SetupWebhooks(&mgr, addonNamespace, vWhCfgNames, mWhCfgNames)
```
danniel1205, line 155: Question: Do we want to handle the error here? Or we want to proceed when it fails to set up webhook.

addons/main.go
maralavi: type ```populatec```

addons/webhook/certificates.go
```diff
@@ -0,0 +1,125 @@
...
+	"path"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"time"
+)
+
+const (
+	ServerKey = "server-key.pem"
+	// ServerCert is the name of the key associated with the secret's public key.
```
maralavi, line 21: This description here seems to belong to ```ServerKey```. Doesn't it?

addons/webhook/certificates.go
```diff
@@ -0,0 +1,125 @@
...
+	"k8s.io/client-go/kubernetes"
+	certutil "k8s.io/client-go/util/cert"
+	"k8s.io/client-go/util/keyutil"
+	rsrcs "knative.dev/pkg/webhook/certificates/resources"
+	"path"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"time"
+)
```
maralavi, line 17: Could you please reorder the imports into 3 categories separated by new line: 1) go internal 2) external imports 3) internal tanzu imports

addons/webhook/certificates.go
```diff
@@ -0,0 +1,125 @@
...
+func UpdateValidatingWebhook(client kubernetes.Interface, name string, secret *corev1.Secret) error {
+
+	vWhCfg, err := client.AdmissionregistrationV1().ValidatingWebhookConfigurations().Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	for idx, webhook := range vWhCfg.Webhooks {
+		webhook.ClientConfig.CABundle = secret.Data[CACert]
+		vWhCfg.Webhooks[idx] = webhook
```
maralavi, line 73: Have you tested that this line work without append and doesn't it index out of range?

addons/webhook/certificates.go
vijaykatam: Seems like it was due to [recent security change](https://github.com/knative/pkg/pull/1998) where expiry is 7 days and rotation is 1 day before expiry. Seems like something we should adopt as well.

addons/webhook/certificates.go
vijaykatam: manager has the controller-runtime client and not the more generic kubernetes.interface client

addons/webhook/certificates.go
```diff
@@ -0,0 +1,125 @@
...
+const (
+	ServerKey = "server-key.pem"
+	// ServerCert is the name of the key associated with the secret's public key.
+	ServerCert = "server-cert.pem"
+	// CACert is the name of the key associated with the certificate of the CA for
+	// the keypair.
+	CACert = "ca-cert.pem"
+)
+
+func SetupWebhooks(mgr *manager.Manager, namespace string, vWhCfgNames, mWhCfgNames []string) error {
```
vijaykatam, line 28: more like SetupWebhookCerts

addons/main.go
```diff
@@ -147,6 +148,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	//name of all the validating webhook configurations. the following need to be populatec with the names of the
+	//validating and webhooks configurations.
+	var vWhCfgNames []string
```
vijaykatam, line 153: suggestion: var validatingWebhooks []string

addons/webhook/certificates.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return err
+	}
+
+	for idx, webhook := range vWhCfg.Webhooks {
+		webhook.ClientConfig.CABundle = secret.Data[CACert]
+		vWhCfg.Webhooks[idx] = webhook
+	}
+	if _, err := client.AdmissionregistrationV1().ValidatingWebhookConfigurations().Update(context.TODO(), vWhCfg, metav1.UpdateOptions{}); err != nil {
+		return fmt.Errorf("error updating addons CA cert of ValidatingWebhookConfiguration %s: %v", name, err)
```
vijaykatam, line 76: I would make the error more generic so that other webhooks outside addons can use it.

vijaykatam (commented):
I would consider moving these generic helpers to pkg/v1/sdk/webhooks or something similar cc @rajathagasthya @jmoroski

addons/webhook/certificates.go
```diff
@@ -0,0 +1,125 @@
...
+
+func SetupWebhooks(mgr *manager.Manager, namespace string, vWhCfgNames, mWhCfgNames []string) error {
+
+	//create the secret and write the certificate files based on the existing webhook server config
+	wh := (*mgr).GetWebhookServer()
+	secret, err := UpdateTLSSecret("addons-webhook-tls", wh, namespace)
+	if err != nil {
+		return err
+	}
+
```
shyaamsn, line 36: Rotation could be setup here similar to antrea: https://github.com/antrea-io/antrea/blob/main/pkg/apiserver/certificate/certificate.go#L100

vijaykatam: Timer based with select is probably better(this is sleep). Also needs to reload webhook http server with new certs.

addons/main.go
adduarte: done

addons/main.go
adduarte: Yes. we probably should. But the question is what we should do? 
Is not catastrophic failure. Addon manager can still do its job. The user won't be able to use the api to configure the packages.

addons/webhook/certificates.go
adduarte: done

addons/webhook/certificates.go
adduarte: correct. Removing comment. is redundant.

addons/webhook/certificates.go
adduarte: done

addons/webhook/certificates.go
adduarte: Its starting to look like we need a certificate controller.. Certificate roration coudl be taken care of *and* we need to no only update existing webhook Configurations but also configurations which are created *after* the add on manger is started.

addons/webhook/certificates.go
adduarte: yea. Perhaps is a good idea to propagae the context from the addon main.

addons/webhook/certificates.go
adduarte: hmm. I am not sure it would index out of range. 
idx is obtained from vWhCfg.Webhooks. so it does have the index. 
It is also not and appen but a replacement.  but maybe I am missing something. let me check.

addons/webhook/certificates.go
adduarte: I think is best to use the one provided by knative then. I'll make the changes.

adduarte: replaced the with the the mentioned "create secret" from knative/pkg

addons/webhook/certificates.go
adduarte: tested by removing "webhook.ClientConfig.CABundle = secret.Data[CACert]"
It did not index out of range. 
i.e. I tested with:
``` 
	for idx, whook := range mWhCfg.Webhooks {
		// whook.ClientConfig.CABundle = secret.Data[CACert]
		mWhCfg.Webhooks[idx] = whook
	}
``` 
And it did not error out. It behaved as expected. It did not update  with the new CACert but it did not error out.

addons/webhook/certificates.go
adduarte: done

addons/webhooks/suite_test.go
```diff
@@ -0,0 +1,167 @@
...
+	RegisterFailHandler(Fail)
+
+	RunSpecsWithDefaultAndCustomReporters(t,
+		"Addon Webhooks Suite",
+		[]Reporter{printer.NewlineReporter{}})
+}
+
+var _ = BeforeSuite(func(done Done) {
+
+	By("bootstrapping test environment")
```
danniel1205, line 56: question: I think `--admission-control=` needs to be added to `envtest.DefaultKubeAPIServerFlags` for test setup.

addons/webhook/certificates.go
adduarte: I think we will submit rotation of certificates as a follow up pr to keep this pr small and not break current functioanlity

addons/webhooks/suite_test.go
adduarte: fixed this thanks!

addons/controllers/certificates_test.go
```diff
@@ -0,0 +1,107 @@
...
+			key, err := keyutil.PrivateKeyFromFile(keyFilePath)
+			Expect(err).ToNot(HaveOccurred())
+			orgKey, err := keyutil.ParsePrivateKeyPEM(secret.Data[certificates.ServerKey])
+			Expect(err).ToNot(HaveOccurred())
+			Expect(key).To(Equal(orgKey))
+		})
+	})
+
+	Context("Defautling and Validating Configurations", func() {
+		It("should be updated with CA bundle ", func() {
```
blc1996, line 41: Should we do resource clean up after the test case is executed?

adduarte: Do you mean delete the created configurations?

blc1996: Yes, I assume each test will create it's own configuration? Or you thing other tests could come and add rules here?

addons/controllers/certificates_test.go
```diff
@@ -0,0 +1,142 @@
...
+			defaultingWebhookConfig.Webhooks = []adminregv1.MutatingWebhook{
+				{
+					Name: "mcalicoconfig.kb.io",
+					ClientConfig: adminregv1.WebhookClientConfig{
+						Service:  &srvc,
+						CABundle: []byte{'b', 'a', 'd'}},
+					SideEffects:             &se,
+					AdmissionReviewVersions: []string{"v1"},
+				},
+			}
```
adduarte, line 90: This whole section creating webhook configurations should be done through testuilts.CreateResoruces(), that way later on we can just point it to the yaml files we will use to deploy the real pacakge.

addons/controllers/certificates/certificates.go
```diff
@@ -0,0 +1,159 @@
+package certificates
```
rajathagasthya, line 1: I agree with @vijaykatam's suggestion that this should be in a package outside `addons`. I don't know if that's necessarily `pkg/v1/sdk/controllers`, but I guess it's better than `pkg/v1/utils`?

cc @jmoroski @yharish991

addons/controllers/certificates_test.go
adduarte: Lets leave it here for now.

addons/controllers/certificates/certificates.go
```diff
@@ -0,0 +1,159 @@
...
+			if err != nil {
+				certificateIsValid = false
+			} else {
+				certificateIsValid = true
+			}
+		}
+	}
+}
+
+func SyncValidatingConfigs(ctx context.Context, client kubernetes.Interface, labelSelector string, secret *corev1.Secret) error {
```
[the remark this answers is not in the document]
adduarte: will do. Manage function is being removed to another pr.

addons/controllers/certificates/certificates.go
adduarte: done. Moved to pkg/v1/webhooks/certificates.go

addons/controllers/certificates_test.go
```diff
@@ -0,0 +1,118 @@
...
+			Eventually(func() string { return webhookServer.CertDir }, "10s").ShouldNot(BeEmpty()) //Until the webhook server has comleted startup CertDir returns ""
+			certFilePath := path.Join(webhookServer.CertDir, webhookServer.CertName)
+			Expect(certFilePath).ShouldNot(BeEmpty())
+			keyFilePath := path.Join(webhookServer.CertDir, webhookServer.KeyName)
+			Expect(keyFilePath).ShouldNot(BeEmpty())
+			Expect(webhookServer).ToNot(BeNil())
+			secret, err := webhooks.UpdateTLSSecret(ctx, webhookScrtName, webhookServiceName, webhookServer, addonNamespace)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(secret).NotTo(BeNil())
+			cert, err := cert2.CertsFromFile(certFilePath)
```
shivaani0505, line 30: QQ - In this PR, do we have to create the certificates and keys manually for running tests ? I was trying to automate that as part of the tests so that when different components use webhooks they don't have to generate/rotate/manage certificates for integration tests.

adduarte: nope. 
In your test after the envtest has been created started and the manager has been created you call 
webhookserver = manager.getwebhook(), and then somewhere in your test you make the call: 

``` secret, err := webhooks.UpdateTLSSecret(ctx, webhookScrtName, webhookServiceName, webhookServer, addonNamespace)

you can see it here. 
https://github.com/vmware-tanzu/tanzu-framework/pull/1687/files#diff-03f91d8a3c07135ca5f572b41fb7e213ada82e2fc17fdac963a16388ee24af57R159

in this case the variable "webhookServer" is global and has a reference to the manager webhook server. See here:
https://github.com/vmware-tanzu/tanzu-framework/pull/1687/files#diff-03f91d8a3c07135ca5f572b41fb7e213ada82e2fc17fdac963a16388ee24af57R76

**Comment by alfredthenarwhal** [comment]
> @adduarte: /retest
Commit: 2a0bd32

Tests failed! Build no: 1574

**Comment by alfredthenarwhal** [comment]
> @navidshaikh:  /test install-vc7
Commit: 2a0bd32

Tests failed! Build no: 1575

**Comment by adduarte** [comment]
@alfredthenarwhal looking into this failurs. It seems someting in the test_suite.go file is causing it.

**Comment by alfredthenarwhal** [comment]
> @adduarte: /retest

Tests can be triggered only by the trusted reviewers (users mentioned in CODEOWNERS file).

File: pkg/v1/webhooks/certificates.go [hunk 24]
```diff
@@ -0,0 +1,131 @@
...
+		return secret, err
+	}
+	if err := keyutil.WriteKey(keyPath, secret.Data[resources.ServerKey]); err != nil {
+		return secret, err
+	}
+
+	return secret, nil
+}
+
+func UpdateCertificates(ctx context.Context, k8sConfig *rest.Config, certPath, keyPath, secretName, namespace, serviceName, labelSelector string) (*corev1.Secret, error) {
```
blc1996, line 84: Should other webhooks test files call this function to update the certificate before running the test cases?

addons/controllers/suite_test.go
```diff
@@ -142,13 +150,20 @@ var _ = BeforeSuite(func(done Done) {
 	Expect(err).ToNot(HaveOccurred())
 	Expect(dynamicClient).ToNot(BeNil())
 
+	tmpDir, err := os.MkdirTemp("/tmp", "webhooktest")
+	Expect(err).ToNot(HaveOccurred())
+	certPath = path.Join(tmpDir, "tls.cert")
+	keyPath = path.Join(tmpDir, "tls.key")
```
blc1996, line 156: Question: I assume the certificates specified here will not be valid in the first place?

addons/controllers/certificates_test.go
```diff
@@ -0,0 +1,90 @@
...
+			err = k8sClient.List(ctx, orgvwcfgs)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(orgvwcfgs.Items).ToNot(BeEmpty())
+
+			orgmwcfgs := &adminregv1.MutatingWebhookConfigurationList{}
+			err = k8sClient.List(ctx, orgmwcfgs)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(orgmwcfgs.Items).ToNot(BeEmpty())
+
+			secret, err := webhooks.UpdateCertificates(ctx, k8sConfig, certPath, keyPath, webhookScrtName, addonNamespace, webhookServiceName, "")
```
blc1996, line 63: Question: Do we assume we will set certain label for the webhookConfigurations that we create?

addons/controllers/certificates_test.go
```diff
@@ -0,0 +1,90 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
```
blc1996, line 1: Why do we have both `webhooks/certificates_test.go` and `controllers/certificates_test.go`?

blc1996 (approved):
Some non-blocking comments. LGTM if all the testcases pass.

addons/controllers/certificates_test.go
adduarte: yes

addons/controllers/suite_test.go
adduarte: this are not certificates. this are just the fulll file path where the certificates will be written to.

addons/controllers/certificates_test.go
adduarte: controllers/certificates_test.go are the integration tests. they required envtest. The ones under webhooks/certificates.test.go are the unit test. They do not requried envtest. 
The rest of the work (next patches will probably have their code live under controllers/ as they only apply to the addons package.

pkg/v1/webhooks/certificates.go
adduarte: yes.   This will write the certificates to the provided paths. In reallity this exists independent of the webhook server. 
What it basically does is write the tls.cert and tls.keys to the given directories, and inject  the CA bundle into the webhook configs. 
 so the webhook server does not need to even be running. It just hast to be configured with the matching certDir

you basically want certPath = certDir + "tls.cert" and keypath = CertDir + "tls.key"
Then when this gets run the cerfiticates are "installed" and the webhook will automatically pick it up (if it is running already) or will pick it up on start up. 

The CertDir needs to be created already. 

So it should be something like

install webhook configuration (manfiests.yml)
create  certdir ( to prevent racing condition when webhook takes too long to start)
run this function (changed name from UpdateCertificates to "InstallNewCertificates" 
configure webhookserver with certdir and start it
then ready to run test. 

The order could be change a bit, just need to make sure the CertDir is available before you run this function.

vijaykatam (approved):
LGTM. Please update release note to empty from TBD prior to merging.

shivaani0505 (approved):
LGTM

PR #1707, opened 2022-02-16
Automate release note generation process for Tanzu Framework Release

### LINKED ISSUE #1706: Update release Github Action workflow to auto generate release notes

Whenever a new tag is created, currently a new draft release is created using github actions. Update the github action to include the release note generation as well.

Alternate is to generate the release notes manually and that is what is currently followed

- [ ] APIs
- [ ] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [X] Test and Release
- [ ] User Experience

**Additional context**

sriramandev:
Auto generate release note for every new release. When tag is pushed, automate the release note creation for the new release.

This PR updates the existing release workflow in github actions to include release notes for the current tag release to be added when creating the draft release.

Fixes # [1706](https://github.com/vmware-tanzu/tanzu-framework/issues/1706)

Ran workflow in fork to generate the release notes for dummy release. First created and pushed a dummy tag based on the working branch. This triggered the release workflow github action which created the draft release with the release notes. Attached below snapshots of the same
image
image
image
image
image
image

### Release note

```release-note

```

### PR Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -61,12 +61,14 @@ jobs:
       shell: bash
       run: |
         set -o xtrace
+        RELEASE_BRANCH="main"
         RELEASE_VERSION=${GITHUB_REF#refs/*/}
         major=$(echo "$RELEASE_VERSION"  | sed 's/^v\(.*\)/\1/' | cut -d. -f1)
         minor=$(echo "$RELEASE_VERSION" | cut -d. -f2)
         revision=$(echo "$RELEASE_VERSION" | cut -d. -f3)
         if [ "$revision" -gt 0 ];then
           revision=$(($revision-1))
+          RELEASE_BRANCH="release-${major}.${minor}"
         elif [ "$minor" -gt 0 ]; then
           minor=$(($minor-1))
         elif [ "$major" -gt 0 ]; then
@@ -88,6 +90,7 @@ jobs:
         echo "::set-output name=prev-release-version::$LAST_TAG"
         echo "::set-output name=repo-name::$(echo '${{ github.repository }}' | awk -F '/' '{print $2}')"
         echo "::set-output name=repo-org::$(echo '${{ github.repository }}' | awk -F '/' '{print $1}')"
+        echo "::set-output name=release-branch::$RELEASE_BRANCH"
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
@@ -110,7 +113,7 @@ jobs:
             --github-base-url https://github.com \
             --org ${{ steps.generate-release-details.outputs.repo-org }} \
             --repo ${{ steps.generate-release-details.outputs.repo-name }} \
-            --branch main \
+            --branch ${{ steps.generate-release-details.outputs.release-branch }} \
             --required-author "" \
             --start-sha ${{ steps.generate-release-details.outputs.start-sha }} \
             --end-sha ${{ steps.generate-release-details.outputs.end-sha }} \
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -29,23 +29,23 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v1
 
-    - name: Get dependencies
-      run: |
-        go get -v -t -d ./...
-        if [ -f Gopkg.toml ]; then
-            curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
-            dep ensure
-        fi
-
-    - name: Build
-      run: make all
-
-    - name: Verify
-      run: |
-        make verify
-
-    - name: Test
-      run: make test
+#    - name: Get dependencies
+#      run: |
+#        go get -v -t -d ./...
+#        if [ -f Gopkg.toml ]; then
+#            curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
+#            dep ensure
+#        fi
+#
+#    - name: Build
+#      run: make all
+#
+#    - name: Verify
+#      run: |
+#        make verify
+#
+#    - name: Test
+#      run: make test
 
     - id: upload-cli-admin-artifacts
       # disable unsigned/untested artifacts upload to bucket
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -56,10 +56,140 @@ jobs:
...
+      shell: bash
+      run: |
+        echo "::set-output name=repo-name::$(echo '${{ github.repository }}' | awk -F '/' '{print $2}')"
+        echo "::set-output name=repo-org::$(echo '${{ github.repository }}' | awk -F '/' '{print $1}')"
+
+    - name: Get last release info
+      uses: octokit/request-action@v2.x
+      id: get-last-release
+      with:
+        route: GET /repos/{owner}/{repo}/releases/latest
```
rajaskakodkar, line 70: For last release: Getting the latest release will work only for minor version change. This will break for a patch release. I think the last release should be calculated based on the current release. For e.g if RELEASE_VERSION = X.Y.Z, LAST_RELEASE = X.Y.Z-1 if Z>0 X.Y-1.Z if Y>0 X-1.Y.Z if X>0. I think this is inline with semver versioning which we follow - https://semver.org/

.github/workflows/release.yaml
```diff
@@ -56,10 +56,140 @@ jobs:
...
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Generate details for creating release notes
+      id: generate-release-details
+      shell: bash
+      run: |
+        LAST_TAG=${{ fromJson(steps.get-last-release.outputs.data).tag_name }}
+        echo "last tag - $LAST_TAG"
+        LAST_TAG_SHA=$(git rev-list -n 1 ${{ fromJson(steps.get-last-release.outputs.data).tag_name }})
```
rajaskakodkar, line 82: Shouldn't this be START_SHA?

.github/workflows/release.yaml
```diff
@@ -56,10 +56,140 @@ jobs:
...
+        echo "Last tag SHA - $LAST_TAG_SHA"
+        END_SHA=$(git rev-parse HEAD)
```
rajaskakodkar, line 84: Do we need to iterate over sha array or can we get END_SHA from $(git rev-parse -n 1 RELEASE_VERSION) because the release tag has already been pushed before this job gets triggered?

rajaskakodkar (changes requested):

rajaskakodkar:
This is awesome 🚀 I have left some comments otherwise lgtm!

.github/workflows/release.yaml
sriramandev: Updated to use semver based logic instead of api based retrieval. Only downside of this I believe is, we cannot skip any release version  eg- after 0.18.0 some reason we do not want 0.19.0 but 0.20.0 - not possible. However, using the latest release api has downside as well that would not work with patches. 
As a result, have modified to use semver based as suggested.

.github/workflows/release.yaml
sriramandev: Originally was planning to take the commit after last release tag commit. Hence the variable name. Based on discussion, making it simpler by using last release tag sha as the start sha.

.github/workflows/release.yaml
sriramandev: Originally was planning to take the commit after last release tag commit. Hence loop to find the correct sha. Based on discussion, making it simpler by using last release tag sha as the start sha.

.github/workflows/release.yaml
```diff
@@ -56,10 +56,110 @@ jobs:
...
+        echo "::set-output name=release-version::$RELEASE_VERSION"
+        echo "::set-output name=prev-release-version::$LAST_TAG"
+        echo "::set-output name=repo-name::$(echo '${{ github.repository }}' | awk -F '/' '{print $2}')"
+        echo "::set-output name=repo-org::$(echo '${{ github.repository }}' | awk -F '/' '{print $1}')"
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Install release-note dependency
+      run: |
+        go get k8s.io/release/cmd/release-notes@latest
```
rajaskakodkar, line 96: I am wondering if we should pin the sha which works now instead of `latest` for the sake of reproducibility.

rajaskakodkar:
@sriramandev with new changes, can you please run this on your fork for some test cases like minor version change, patch version change to make sure it works? 

I think the only bit remaining after testing is reverting the commented make targets.

sriramandev:
> @sriramandev with new changes, can you please run this on your fork for some test cases like minor version change, patch version change to make sure it works?
> 
> I think the only bit remaining after testing is reverting the commented make targets.

Have run now both the minor bump and patch bump. Below are snapshots of the same.
Minor release snapshots
image
image
image

Patch release snapshots
image
image

.github/workflows/release.yaml
sriramandev: The reason I am not pinning to SHA is simply because other wise we will need to regularly update the workflow every time a new release notes generator version is released.

rajaskakodkar: ack! The only issue with release-notes@latest is that if release-notes introduces a different ux in the tool then our workflow will break. I think we go ahead with latest for now because ux changing is a less likely scenario

.github/workflows/release.yaml
```diff
@@ -61,12 +61,14 @@ jobs:
...
       run: |
         set -o xtrace
+        RELEASE_BRANCH="main"
         RELEASE_VERSION=${GITHUB_REF#refs/*/}
         major=$(echo "$RELEASE_VERSION"  | sed 's/^v\(.*\)/\1/' | cut -d. -f1)
         minor=$(echo "$RELEASE_VERSION" | cut -d. -f2)
         revision=$(echo "$RELEASE_VERSION" | cut -d. -f3)
         if [ "$revision" -gt 0 ];then
           revision=$(($revision-1))
+          RELEASE_BRANCH="release-${major}.${minor}"
```
rajaskakodkar, line 71: Good catch!

rajaskakodkar (approved):
lgtm!

.github/workflows/release.yaml
```diff
@@ -56,10 +56,113 @@ jobs:
...
+      with:
+        route: POST /repos/{owner}/{repo}/releases/generate-notes
+        owner: ${{ steps.generate-release-details.outputs.repo-org }}
+        repo: ${{ steps.generate-release-details.outputs.repo-name }}
+        tag_name: ${{ steps.generate-release-details.outputs.release-version }}
+        previous_tag_name: ${{ steps.generate-release-details.outputs.prev-release-version }}
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: 'Identify New Contributors section'
```
navidshaikh, line 134:
```suggestion
    - name: Identify New Contributors section
```

.github/workflows/release.yaml
```diff
@@ -56,10 +56,113 @@ jobs:
...
+      id: get-new-contributors
+      run: |
+        githubOutput="/tmp/${{ steps.time.outputs.time }}-github-output"
+        echo "${{ fromJson(steps.get-github-release-notes.outputs.data).body }}" > "$githubOutput"
+
+
```
navidshaikh, line 140:
```suggestion

```

.github/workflows/release.yaml
```diff
@@ -56,10 +56,113 @@ jobs:
...
+    - name: Install release-note dependency
+      run: |
+        go get k8s.io/release/cmd/release-notes@latest
+
+    - name: Get Time
+      id: time
+      uses: nanzm/get-time-action@v1.1
+      with:
+        format: 'YYYYMMDDHHmmss'
```
navidshaikh, line 105: why not use the release version to generate the tmp files but use time ?

navidshaikh (approved):
lgtm!

PR #1722, opened 2022-02-22
Adding webhook for AntreaConfig and CalicoConfig

### LINKED ISSUE #1690: Webhook for AntreaConfig and CalicoConfig
- AntreaConfig immutable fields are TLSCipherSuites, DefaultMTU, TrafficEncapMode, NoSNAT, DisableUDPTunnelOffload
- CalicoConfig immutable fields are namespace and VethMTU

shivaani0505:
Signed-off-by: Shivaani Gupta <gshivaani@vmware.com>

- Adding webhook for AntreaConfig and CalicoConfig
- Removed manual steps for testing webhooks. Modified controllers/suite_test.go to include setting up of webhooks. 

This includes the following:
- Generate self-signed certificate for CA and signed certificate and key for webhook server
- Code to inject caBundle to the web hooks

The overall flow (and changes in the PR therein) is thus - 
1. The scaffolding for Antrea and Calico webhooks are autogenerated using kubebuilder
2. Logic is implemented in Validating and Mutating/Defaulting functions for Antrea and Calico
3. The scaffolded files include autogenerated markers for kubebuilder. The webhook configuration file (manifests.yaml) is also autogenerated using kubebuilder (controller-gen) tool.  
4. Test suite changes are made to include setting up of webhooks. This includes the following steps:
    1. Apply the auto-generated manifests.yaml file to create webhook configurations
    2. Generate self-signed CA certificate and certificate and keys for webhooks in code (no manual setup required)
    3. Patch the caBundle (CA certificate) in the webhooks
    4. Setup the webhook with the manager calling the scaffolded function (SetupWebhookWithManager)

Fixes #

- The webhook server is provided the server certificate and key in the default directory “/tmp/k8s-webhook-server/serving-certs” 
- The webhook configurations load the caBundle into the API server by directly providing the CA cert as bundle

### Release note

```release-note
Adding webhook for AntreaConfig and CalicoConfig
```

### PR Checklist


addons/controllers/webhooks/manifests.yaml
```diff
@@ -0,0 +1,95 @@
+---
```
blc1996, line 1: Could we define and create these configurations with GO code instead of defining them in yaml file? I feel in this way it would be easier to change and maintain.

shivaani0505: These configurations are auto-generated by using kubebuilder. They are generated from the markers that are in the scaffolded code. If we make all these in GO code, then I think it is harder to maintain. Thinking aloud here - say we need to add webhooks for another component… we will not have to do anything at all except adding 2 lines to call the function SetupWebhookWithManager for that new component, and of course implement the logic in validating and mutating webhooks. Everything else will be autogenerated. It felt much easier doing this way, both in terms of how much changes we need to do and also to maintain. But if the team decides to go with GO code instead, I can change accordingly.

adduarte: We probably want to move to using the yml files and just using a CrateResources(). 
in other words. Generate the yaml files which we will eventually pacakge along with the addons package, and apply them. that way we get to test all at once.

config/crd/patches/cainjection_in_calicoconfigs.yaml
```diff
@@ -0,0 +1,8 @@
+# The following patch adds a directive for certmanager to inject CA into the CRD
+# CRD conversion requires k8s 1.13 or later.
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    cert-manager.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
```
adduarte, line 7: This should not be here. we won't be using cert-manager to inject CAbundle

addons/controllers/suite_test.go
```diff
@@ -240,3 +257,126 @@ var _ = AfterSuite(func() {
...
+		setupLog.Error(err, "unable to create webhook for Calico")
+		os.Exit(1)
+	}
+}
+
+// setupCertsAndKeysForWebhooks sets up the certificate for CA and the cert and keys for webhook server
+func setupCertsAndKeysForWebhooks() {
+
+	// Use the default directory and file names used by webhook server for TLS certs and keys
+	certDir = "/tmp/k8s-webhook-server/serving-certs"
```
adduarte, line 322: don't hard code the directoryy.  you should use mgr.certDir. or webhook.certDir, or one of the other ways to pull the certDir value from the existing manager

addons/controllers/suite_test.go
```diff
@@ -240,3 +257,126 @@ var _ = AfterSuite(func() {
...
+	f, err = os.Open(webhooksFile)
+	Expect(err).ToNot(HaveOccurred())
+	defer f.Close()
+	err = testutil.CreateResources(f, cfg, dynamicClient)
+	Expect(err).ToNot(HaveOccurred())
+
+	// Update the caBundle in validating and mutating webhooks
+	k8sConfig := mgr.GetConfig()
+	client, err := kubernetes.NewForConfig(k8sConfig)
+	validatingWebhookConf, err := client.AdmissionregistrationV1().ValidatingWebhookConfigurations().
```
adduarte, line 281: Could't no you just call  
``` 
addons/testutils/test_helpers.go.CreateResources(addons/controllers/webhooks/manifests.yaml)) 
```

That way this test not only validates the webhooks but also the format and functionality of the manifest.yml files. 

then if you really want to, you  cabundle by doing something like:
```
	allValidatingWebhookConfigurations, err := client.AdmissionregistrationV1().ValidatingWebhookConfigurations().List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
	if err != nil {
		return err
	}
	for idx := range allValidatingWebhookConfigurations.Items {
		configName := allValidatingWebhookConfigurations.Items[idx].Name
		validatingWebhookConfiguration, err := client.AdmissionregistrationV1().ValidatingWebhookConfigurations().Get(ctx, configName, metav1.GetOptions{})
		if err != nil {
			return err
		}

		for idx := range validatingWebhookConfiguration.Webhooks {
			validatingWebhookConfiguration.Webhooks[idx].ClientConfig.CABundle = secret.Data[resources.CACert]
		}
		if _, err := client.AdmissionregistrationV1().ValidatingWebhookConfigurations().Update(ctx, validatingWebhookConfiguration, metav1.UpdateOptions{}); err != nil {
			return fmt.Errorf("error updating CA cert of ValidatingWebhookConfiguration %s: %v", configName, err)
		}
	}
```

addons/controllers/suite_test.go
```diff
@@ -240,3 +257,126 @@ var _ = AfterSuite(func() {
...
+	for idx := range mutatingWebhookConf.Webhooks {
+		mutatingWebhookConf.Webhooks[idx].ClientConfig.CABundle = caPEM.Bytes()
+	}
+	_, err = client.AdmissionregistrationV1().MutatingWebhookConfigurations().
+		Update(ctx, mutatingWebhookConf, metav1.UpdateOptions{})
+	Expect(err).ToNot(HaveOccurred())
+
+	// Configure and start the webhook server
+	wh := mgr.GetWebhookServer()
+	wh.CertDir = "/tmp/k8s-webhook-server/serving-certs"
```
adduarte, line 303: no need. this is the default. btw the webhook server gets started the first time you run 
wh := mgr.GetWebhookServer() and the correct way to set the certDir is to do it in test_suite.go when you start the manager with the mgr.options { certDir: "whatever path you want"} in the test_suite.go file.

config/crd/patches/cainjection_in_antreaconfigs.yaml
```diff
@@ -0,0 +1,8 @@
+# The following patch adds a directive for certmanager to inject CA into the CRD
```
adduarte, line 1: what are these? I thought we were not going to use cert-manager

adduarte (changes requested):
the manifests that are created by kubebuilder will not work with our setup without modifications. They have to be edited to fit our case.  specially the use of cert-manager. 

also there are not tests here that verify the validity of the manifest.yml files. 

all the variables which apply to the webhook server need to be derived from the running manager.webhookserver. and set on manager start here https://github.com/vmware-tanzu/tanzu-framework/blob/0bbc3a4fc9abef66d49d62cac55625cc24e20527/addons/controllers/suite_test.go#L149
in the "options" field

addons/controllers/webhooks/manifests.yaml
adduarte: I think part of this pr's purpose IS to create the yaml files.

adduarte: ^^ I take that back. I guess this pr is not about manifests.yamls. ?

addons/controllers/webhooks/manifests.yaml
```diff
@@ -0,0 +1,90 @@
...
+kind: MutatingWebhookConfiguration
+metadata:
+  creationTimestamp: null
+  name: mutating-webhook-configuration
+webhooks:
+- admissionReviewVersions:
+  - v1
+  - v1beta1
+  clientConfig:
+    url: "https://127.0.0.1:9443/mutate-cni-tanzu-vmware-com-v1alpha1-antreaconfig"
```
adduarte, line 12: is there any reason to use a service instead here? see https://github.com/kubernetes/api/blob/b8c40e080bc5e830097df540d4ef804034cb5bdb/admissionregistration/v1/types.go#L498
 
hardconding the ip seems could bring down some problems say if a firewall or some sort of access is on (like selinux). using a service would allow for the actual cluster itself to correctly setup the service (host:port part of the url).

addons/controllers/suite_test.go
adduarte: take a look at https://github.com/vmware-tanzu/tanzu-framework/pull/1740/files

addons/controllers/webhooks/manifests.yaml
shivaani0505: As mentioned above, the manifests.yaml is automatically generated using the markers. So, it is just a part of this change.

shivaani0505:
Closing this PR as new [PR](https://github.com/vmware-tanzu/tanzu-framework/pull/1796) has been created after refactoring [PR](https://github.com/vmware-tanzu/tanzu-framework/pull/1687)

PR #1724, opened 2022-02-23
Not using WithVersion no longer causes error and allows searching for any version

### LINKED ISSUE #1294: Allow capability API discover resources when version is unknown

As a capability API user, I would like to have an API allows me to discover a resource without specifying the version, because I do not know what version of that resource is currently running in the cluster.

Example: Package plugin would like to use `capdiscovery.Group("packageInstallAPIQuery", "packaging.carvel.dev").WithVersions("xxxx").WithResource("packageinstalls")` to check if `packageInstall` CRD has been deployed. However, the cli code does not know what version of that CRD is current deployed/served. We could put `v1alpha1` as of today and fully rely on the conversion webhook from carvel in the future once they have newer version, but it does not scale if `v1alpha1` gets deprecated (We have to fix it passively and ask user to upgrade their CLI plugin).

A related discussion could be found here: https://github.com/vmware-tanzu/tanzu-framework/pull/1242/files#r761528548

Using other APIs(client) to get `packageInstall` CRD, if the result is non-empty then we know `packgeInstall` is available.

- [x] APIs
- [ ] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Additional context**

codegold79:

This PR,

- allows for `WithVersions(...)` method to be omitted, even if `WithResource(...)` was used
- fixes a bug where using empty string version, `WithVersions("")` will return `Found: true` result even though a specified resource does not exist. Now, using an empty string in the `WithVersions` method will return an error
- using an empty string in the `WithResource` method, in other words, `WithResource("")` will return an error
- adds unit tests to test combinations of using WithVersions and WithResource methods

More details are in [this comment](https://github.com/vmware-tanzu/tanzu-framework/issues/1294#issuecomment-1042383302).

Fixes #1294

When using capability discovery, allow searching for any version of an API resource by ~~both~~ the following way~~s~~:

(1) omitting WithVersions method
~~(2) using `WithVersions("")`~~

### Release note

```release-note
GVR Capability Discovery query now supports searching for any version of a resource by omitting the `WithVersions` method. In addition, `WithVersions` or `WithResource` methods return an error when an empty string is passed as an argument.
```

### PR Checklist

### Additional information

~~I have made it such that using multiple empty strings as the only arguments in the WithVersions method acts just like using `WithVersion("")` and will return an error.~~

Using `WithVersions("")` or having an empty string in a list of versions like `WithVersions("v1", "v2beta", "")` will return an error.

I hope these changes will allow for not having to hard-code versions when looking for capabilities. More information in [this comment](https://github.com/vmware-tanzu/tanzu-framework/pull/1242#discussion_r762207075).

### DIFF
--- a/pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
+++ b/pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
@@ -4,15 +4,25 @@
 package discovery
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 )
 
-// Group represents any API group that may exist on a cluster, with ability to specify versions and resource to check for GVR.
+var (
+	// errInvalidWithResourceMethodArgument occurs when an empty string argument is passed into the WithResource method.
+	errInvalidWithResourceMethodArgument = errors.New("WithResource method must have non-empty string argument; omit method to indicate any resource")
+
+	// errInvalidWithVersionsMethodArgument occurs when empty string argument(s) are passed into the WithVersions method.
+	errInvalidWithVersionsMethodArgument = errors.New("WithVersions method must be comprised of non-empty string argument(s); omit method to indicate any version")
+)
+
+// Group represents any API group that may exist on a cluster.
 func Group(queryName, group string) *QueryGVR {
 	return &QueryGVR{
 		name:  queryName,
@@ -20,11 +30,24 @@ func Group(queryName, group string) *QueryGVR {
 	}
 }
 
+// nullString represents a string that may be null or not explicitly set.
+type nullString struct {
+	String string
+	IsSet  bool
+}
+
+func (s *nullString) set(value string) {
+	if s != nil {
+		s.String = value
+		s.IsSet = true
+	}
+}
+
 // QueryGVR provides insight to the clusters GVRs
 type QueryGVR struct {
 	name          string
 	group         string
-	resource      string
+	resource      nullString
 	versions      []string
 	unmatchedGVRs []string
 }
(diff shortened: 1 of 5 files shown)


codegold79:
I marked this PR as `draft` because of recent discussions with @jmoroski and @rajathagasthya.

This PR allows (as it was allowed before) in some cases for using empty string arguments for `WithResource` and `WithVersions` methods. I will change that to not allowing empty string arguments.

As it is in the current PR, when the `WithResource` and `WithVersions`, it is like saying "any resource", or "any versions", respectively.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
```diff
@@ -35,94 +58,236 @@ func (q *QueryGVR) Name() string {
...
+	for _, ver := range q.versions {
+		ver = strings.TrimSpace(ver)
+		if ver != "" {
+			has = true
 		}
-	} else {
-		gvrs = append(gvrs, schema.GroupVersionResource{Group: q.group})
+	}
+	return has
+}
```
jmoroski, line 134: I don't think this function is complete. Specifying `WithVersion("v1", "")` is an error in my opinion. But this will return true because it only sets true, never false.

If we've decided that the empty string is an invalid value for version, then that should always be true regardless of the position of the empty string.

codegold79: After some thought, I agree with you, @jmoroski.

What the code currently does is if someone really wanted to sneak in an empty version, as long as it had one other version, I'd let the query go through. In the end, the query would not error, but would return `Found: false` because it could not find the empty version. 

However, allowing this loophole is inconsistent with our design decision and makes understanding our API more complicated than it needs to be.

I'll update the logic to not allow any empty string versions, even if there are valid versions listed.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
```diff
@@ -35,94 +58,239 @@ func (q *QueryGVR) Name() string {
...
 	return q
 }
 
-// Run discovery
+// Run discovery.
 func (q *QueryGVR) Run(config *clusterQueryClientConfig) (bool, error) {
-	if q.group != "" && q.resource != "" && len(q.versions) == 0 {
-		return false, fmt.Errorf("cannot check for group and resource existence without version info; use WithVersion method")
+	if err := q.validate(config); err != nil {
+		return false, fmt.Errorf("validating query: %w", err)
```
rajathagasthya, line 77: nit: error message needs to add sufficient context when you are wrapping. You could change this to `query validation failed: %w` (still not additional context IMO) or remove the error wrapping entirely.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
```diff
@@ -35,94 +58,239 @@ func (q *QueryGVR) Name() string {
...
-	q.unmatchedGVRs = unmatched
-	return len(unmatched) == 0, nil
+	return unmatched, nil
 }
 
-func (q *QueryGVR) groupExists(group string, groupList *metav1.APIGroupList) bool {
-	for i := range groupList.Groups {
-		if strings.EqualFold(groupList.Groups[i].Name, group) {
+// groupsContains looks for a group in a slice of APIGroup.
+func (q *QueryGVR) groupsContains(groups []*metav1.APIGroup, group string) bool {
```
rajathagasthya, line 236: nit: `s/groupsContains/containsGroup`

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
```diff
@@ -35,94 +58,239 @@ func (q *QueryGVR) Name() string {
...
 }
 
-// Reason surfaces what didnt match
+// extractGroup extracts a group from the GroupVersion field In APIResourceList.
+// Examples are,
+// - GroupVersion:"autoscaling/v1"
+// - GroupVersion:"v1"
+// In the first example, the group returned will be "autoscaling".
+// In the second example, the group returned will be an empty string.
+func extractGroup(groupVersion string) string {
```
rajathagasthya, line 285: optional: we could use [this function](https://github.com/kubernetes/apimachinery/blob/57893b822e174aa3bdaeee434bc42d22b37fc2dd/pkg/runtime/schema/group_version.go#L209-L211) to extract group to be future-proof. But if you think it's an overkill, I'm okay with this code.

pkg/v1/sdk/capabilities/discovery/cluster_test.go
```diff
@@ -1,9 +1,10 @@
-// Copyright 2021 VMware, Inc. All Rights Reserved.
+// Copyright 2022 VMware, Inc. All Rights Reserved.
```
rajathagasthya, line 1: Same as above.

pkg/v1/sdk/capabilities/discovery/fake.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2021 VMware, Inc. All Rights Reserved.
+// Copyright 2022 VMware, Inc. All Rights Reserved.
```
rajathagasthya, line 1: Ditto.

pkg/v1/sdk/capabilities/discovery/fake.go
```diff
@@ -14,8 +14,9 @@ import (
 // NewFakeClusterQueryClient returns a fake ClusterQueryClient for use in tests.
 func NewFakeClusterQueryClient(resources []*metav1.APIResourceList, scheme *runtime.Scheme, objs []runtime.Object) (*ClusterQueryClient, error) {
 	fakeDynamicClient := dynamicFake.NewSimpleDynamicClient(scheme, objs...)
-	fakeDiscoveryClient := &fakediscovery.FakeDiscovery{Fake: &k8stesting.Fake{
-		Resources: resources,
-	}}
+	fakeDiscoveryClient := &fakediscovery.FakeDiscovery{
```
rajathagasthya, line 17: This looks like an unrelated change?

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
```diff
@@ -35,94 +58,239 @@ func (q *QueryGVR) Name() string {
...
+			return true
+		}
+	}
+	return false
+}
+
+func (q *QueryGVR) unmatchedGroupResource(cfg *clusterQueryClientConfig) (string, error) {
+	groups, resources, err := cfg.discoveryClientset.ServerGroupsAndResources()
+	if err != nil {
+		return "", fmt.Errorf("discovering server group and resource: %w", err)
```
rajathagasthya, line 142: Same here. This needs to say `failed to discover server group and resource: %w` or remove the wrapping.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
```diff
@@ -35,94 +58,239 @@ func (q *QueryGVR) Name() string {
...
+	}
+
+	return "", nil
+}
+
+func (q *QueryGVR) unmatchedGroupVersions(cfg *clusterQueryClientConfig) ([]string, error) {
+	groupList, err := cfg.discoveryClientset.ServerGroups()
 	if err != nil {
-		return false, err
+		return nil, fmt.Errorf("discovering server groups: %w", err)
```
rajathagasthya, line 170: Ditto.

rajathagasthya (changes requested):
Looks pretty good on first pass! Just some minor comments to improve readability.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
```diff
@@ -12,19 +13,41 @@ import (
...
+// specify versions and resource to check for GVR.
 func Group(queryName, group string) *QueryGVR {
 	return &QueryGVR{
 		name:  queryName,
 		group: group,
 	}
 }
 
+// nullString represents a string that may be null or not explicitly set.
+type nullString struct {
```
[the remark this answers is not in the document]
codegold79: I was following the NullString pattern that a standard library uses. You can see it at https://pkg.go.dev/database/sql#NullString

I know what you're proposing, using `*string` is very popular, but I think creating a new type, `NullString` is just as valid. I like the `NullString` type because it's so easy to forget to do a nil check and end up with a nil dereference panic. The NullString, not being a pointer, is less prone to panic.

However, again, your suggestion is very popular. @rajathagasthya , would you still recommend I change to using `*string`? Even though I slightly prefer `NullString` more, I also realize people will be more familiar with `*string`, so I'm fine going either way.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
codegold79: I'm against putting the word "failed" or "error" in an error message. It is redundant as it is clear in both the output and in the code that the message is due to an error. The Thanos Style Guide [expresses this sentiment](https://github.com/thanos-io/thanos/blob/main/docs/contributing/coding-style-guide.md#wrap-errors-for-more-context-dont-repeat-failed--there) more strongly than I do.

However, I agree more context can be added. I'll change it to something like,

```go
fmt.Errorf("GroupVersionResource API query validation: %w", err)
```

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
codegold79: I chose `groupsContains` because it reads similar to the standard library method `strings.Contains`. I even ordered the arguments to be first the haystack and the needle (groups first, then specific group second).

Since it's a nit, I'll leave it how it is. However, if someone has a strong opinion, I don't mind changing it to `containsGroup`.

pkg/v1/sdk/capabilities/discovery/fake.go
codegold79: Unrelated, but it's easier to read with the change (and more standard). Should I change it back?

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
codegold79: I would argue that having the word "failed" is redundant. My reasoning is stated in [a previous comment.](https://github.com/vmware-tanzu/tanzu-framework/pull/1724#discussion_r815194795)

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
codegold79: I would argue that having the word "failed" is redundant. My reasoning is stated in https://github.com/vmware-tanzu/tanzu-framework/pull/1724#discussion_r815194795

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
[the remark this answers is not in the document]
rajathagasthya: It's just as valid, but perhaps more verbose than necessary? Kubernetes uses [pointers for optional things](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required) extensively, hence the suggestion. But not a blocker.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
```diff
@@ -12,19 +13,41 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
-// Group represents any API group that may exist on a cluster, with ability to specify versions and resource to check for GVR.
+var (
+	// ErrInvalidWithResourceMethodArgument occurs when an empty string argument is passed into the WithResource method.
+	ErrInvalidWithResourceMethodArgument = errors.New("WithResource method must have non-empty string argument; omit method to indicate any resource")
```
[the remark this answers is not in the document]
codegold79: If we make these errors private, users of our API will no longer be able to use the standard library method `errors.Is(...)` to determine if a certain error happened. They would be forced to match on error string using something like `strings.Contains(...)`, which I think is not as clean and has more room for mistakes.

I am thinking maybe we can have the errors public/exported like the  io package does: https://pkg.go.dev/io#pkg-variables

On the other hand, I'm not sure if users of our API will ever want to be able to match on the the  `ErrInvalidWithResourceMethodArgument` or `ErrInvalidWithVersionsMethodArgument` errors. If we can't see them ever wanting to be able to look for a particular error, then I agree that I should make these errors private. What do you think?

I'm thinking even if our API users were to write tests for themselves, they might want to look out for these two invalid method arguments errors.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
codegold79: Oh, good find. I like using existing functions provided by Kubernetes.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
rajathagasthya: Not sure I agree with the Thanos style guide when multiple errors are wrapped, but this looks good:

```
fmt.Errorf("GroupVersionResource API query validation: %w", err)
```

pkg/v1/sdk/capabilities/discovery/fake.go
rajathagasthya: Okay to keep since you already have this. But in general, if `gofmt` doesn't complain it just adds noise to the review.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
```diff
@@ -1,9 +1,10 @@
-// Copyright 2021 VMware, Inc. All Rights Reserved.
+// Copyright 2022 VMware, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
codegold79: I changed the copyright back to 2021.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
codegold79: Error message has been updated with more context.

pkg/v1/sdk/capabilities/discovery/cluster_test.go
codegold79: I changed the copyright back to 2021.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
codegold79: I got rid of the `extractGroup` function and used Kubernetes' `schema.ParseGroupVersion(...)` instead.

pkg/v1/sdk/capabilities/discovery/fake.go
codegold79: I changed the copyright back to 2021.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
```diff
@@ -35,94 +58,233 @@ func (q *QueryGVR) Name() string {
...
 
-	var gvrs []schema.GroupVersionResource
-	if len(q.versions) > 0 {
-		for _, v := range q.versions {
-			gvrs = append(gvrs, schema.GroupVersionResource{Group: q.group, Version: v, Resource: q.resource})
+	if q.resource.IsSet && q.resource.String == "" {
+		return ErrInvalidWithResourceMethodArgument
+	}
+
+	if q.versions != nil && containsEmpty(q.versions) {
```
yharish991, line 120: should you be validating the versions first and then the resource later?
this `Group("test", "").WithVersions("").WithResource("")` returns `ErrInvalidWithResourceMethodArgument`, it should return `ErrInvalidWithVersionsMethodArgument` if we are immediately returning the error

This also makes me think, should we append the errors, because in the above case we have invalid versions and resource?

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
yharish991: I vote for `containsGroup` as well

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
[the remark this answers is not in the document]
yharish991: I'm fine either way

pkg/v1/sdk/capabilities/discovery/cluster_test.go
```diff
@@ -141,3 +148,239 @@ func TestClusterQueries(t *testing.T) {
...
+				WithVersions("", "", "", "").
+				WithResource("pods"),
+			err: ErrInvalidWithVersionsMethodArgument,
+		},
+		{
+			description: "having empty string version among many returns an error",
+			query:       Group("test", "").WithVersions("", "v1", "v2", "v3"),
+			err:         ErrInvalidWithVersionsMethodArgument,
+		},
+	}
```
yharish991, line 345: I would also like to see test cases for below scenarios as well 
`Group("test", "").WithVersions("", "", "", "").WithResource("")`
`Group("test", "autoscaling").WithVersions("", "", "", "").WithResource("")`

yharish991 (commented):
great job, added a few minor comments

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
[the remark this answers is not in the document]
codegold79: > It's just as valid, but perhaps more verbose than necessary? 

I agree that it's more verbose, but I'm willing to take a slight increase in verbosity to be more safe and not have to deal with nil checks every time I want to dereference the string pointer.

Since both are valid and no strong opinions either way, I'll keep the `nullString` custom type.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
codegold79: Changed `groupsContains` -> `containsGroup`.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
```diff
@@ -12,19 +13,41 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
-// Group represents any API group that may exist on a cluster, with ability to specify versions and resource to check for GVR.
+var (
+	// ErrInvalidWithResourceMethodArgument occurs when an empty string argument is passed into the WithResource method.
+	ErrInvalidWithResourceMethodArgument = errors.New("WithResource method must have non-empty string argument; omit method to indicate any resource")
+
+	// ErrInvalidWithVersionsMethodArgument occurs when only empty string argument(s) are passed into the WithVersions method.
```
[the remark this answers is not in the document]
codegold79: Good catch. I forgot to update this comment when I made the change to disallow any empty string(s) in the `WithVersions` method. The comments and error message have been updated.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
codegold79: > should you be validating the versions first and then the resource later?

I suppose given the example you gave, validating versions should come first. However, if the developer used a different order, i.e. `Group("test", "").WithResource("").WithVersions("")`, then I think the resource error should come first. 

But, I don't think it's worth the effort to ensure a "correct" order of error messages.

> should we append the errors, because in the above case we have invalid versions and resource?

I don't mind doing that; I went ahead and updated the error wrapping to include both errors if they both occur.

pkg/v1/sdk/capabilities/discovery/cluster_test.go
codegold79: Tests added.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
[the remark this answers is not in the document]
rajathagasthya: > On the other hand, I'm not sure if users of our API will ever want to be able to match on the the ErrInvalidWithResourceMethodArgument or ErrInvalidWithVersionsMethodArgument errors.

This is really the intention behind my earlier comment. I don't see these errors being too useful for consumers. I also don't think anyone writing unit tests will be testing validation behavior of our underlying APIs (or they shouldn't at least). A good proxy for consumer usage is to look at `pkg/v1/sdk/capabilities/discovery/tkg` and see how the tests are written.

I don't feel there is a strong enough reason, so I'm inclined to avoid another API contract with exported errors that makes it harder to change later.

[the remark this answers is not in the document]
codegold79: > I don't see these errors being too useful for consumers. I also don't think anyone writing unit tests will be testing validation behavior of our underlying APIs (or they shouldn't at least)...I don't feel there is a strong enough reason, so I'm inclined to avoid another API contract 

Yup, if consumers have no use for these errors, I should make them private.

Errors are now not exported.

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
yharish991: I'm kind of used to using `failed`, `error occurred` etc when adding the error messages, so, not sure I agree with the thanos style guide(I understand their intentions).. also, to keep it consistent across the project can we continue to use the prefixes in error messages?

jmoroski: I did a check for just `Errorf` uses and we seem to be all over the map across the teams and packages within this repo:
https://github.com/vmware-tanzu/tanzu-framework/search?q=errorf

Some that say "error" or "failed", some that say "couldn't do 'x'", some that acknowledge an unexpected state, etc.

While we don't have a cross-repo standard in place we _should_ value consistency in a case like this. I'd recommend finding a few other error messages in the query/discovery/capabilities area of the code and making this look like that.

codegold79: I updated the error message to be prefaced with "failed...".

pkg/v1/sdk/capabilities/discovery/cluster_gvr.go
codegold79: Per team guidance, I updated the error message to be prefaced with "failed...".

rajathagasthya (approved):
Thanks for a useful feature addition!

jmoroski (approved):

PR #1730, opened 2022-02-23
Refactor labels

### LINKED ISSUE #1620: Kickstart UI: refactor labels
**Problem description**
We currently have two different implementation of labels for load-balancer and metadata steps. It would be nice to create a component and re-use it between the two steps.
In addition, we can use a simple array of KeyValuePairs for the backing data structure. This would permit the user to edit both the key and value fields for existing labels.
We can allow a trashcan icon when there is only one key-value pair and have it simply clear the values out of the fields.
(That allows consistency: there is always a trashcan icon.)

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**

2022amanda:
Kickstart UI: refactor labels

Fixes #1620

### Release note

```release-note

```

### PR Checklist


raymondz1:
There is one css issue. The error message is wrapped. 
Screen Shot 2022-02-24 at 3 30 35 PM

PR #1742, opened 2022-02-26
fix local docker registry port

yharish991:
This PR updates the local docker registry port to 5001 as 5000 port used by the control center in macOS Monterey.

Fixes https://github.com/vmware-tanzu/tanzu-framework/issues/1743

make local-registry
PACKAGE_REGISTRY=management make package-bundles
### Release note

```release-note
NONE
```

### PR Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -709,7 +709,7 @@ package-bundles: management-package-bundles ## Build tar bundles for multiple pa
 
 .PHONY: management-package-bundles
 management-package-bundles: tools management-imgpkg-lock-output ## Build tar bundles for packages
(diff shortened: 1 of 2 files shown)


stmcginnis (approved):
Looks good.

Outside the scope of this change, but there are additional changes needed to actually use a local registry that I don't even see in the original PR adding this make target. It looks like we are only using as a convenient way to have something local to push a bundle and query its sha256. Might be good at some point to at least add some comments to the Makefile explaining its presence there.

codegold79 (approved):

PR #1759, opened 2022-03-02
Remove double compile of core from builder

stmcginnis:

The `builder cli compile` command currently builds the "core" tanzu
binary twice - once with the passed in version number, and once with the
version set to "latest".

It looks like this "latest" version may have been for a purpose that was
never fully realized, and there are no current uses of this extra
binary. The signing process we have explicitly ignores this artifact as
well.

This is a minor savings, but since we do not use or need this "latest"
binary, this updates the builder command to only compile once for the
version requested.

Performed local builds, verified there were no errors uncovered by not having a "latest" version of `tanzu`.

### Release note

```release-note
Build process will no longer create a `tanzu` binary with the version set to "latest".
```

### PR Checklist

### DIFF
--- a/pkg/v1/builder/command/cli.go
+++ b/pkg/v1/builder/command/cli.go
@@ -177,13 +177,6 @@ func compileCore(corePath string, arch cli.Arch) cli.Plugin {
 		os.Exit(1)
 	}
 
-	// TODO (pbarker): should copy.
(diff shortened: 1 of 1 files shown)


rajathagasthya (approved):

anujc25 (approved):
LGTM

PR #1786, opened 2022-03-05
Detect go modules for linting

### LINKED ISSUE #1785: Linting of new go modules should be automatic
**Bug description**

As it is today, when new go modules are added to the repo, that module path needs to be added to the `go-lint` target in the Makefile. Since `golangci-lint`  only runs within a single go module, it needs to be explicitly added so it gets run against the new module's code.

The make target should be smart enough to detect all go modules and automatically make sure linting is done.

stmcginnis:

We've had the issue that any time we need to add a new go module to the
repo, the person adding it needs to remember to include it in the
Makefile as an additional location for the `go-lint` and `modules`
target since those operate on a per-module level.

This updates those targets so they automatically iterate through all
detected directories containing a `go.mod` file. That way when someone
adds new modules in the future, they don't need to be as aware of this
linting limitation.

Fixes #1785 

Ran `make go-lint` and `make modules` locally.

### Release note

```release-note

```

### PR Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,8 @@
 
 include ./common.mk
 
+SHELL := /usr/bin/env bash
+
 # Image URL to use all building/pushing image targets
 IMG ?= controller:latest
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
@@ -24,6 +26,7 @@ ADDONS_DIR := addons
 YTT_TESTS_DIR := pkg/v1/providers/tests
 PACKAGES_SCRIPTS_DIR := $(abspath hack/packages/scripts)
 UI_DIR := pkg/v1/tkg/web
+GO_MODULES=$(shell find . -path "*/go.mod" | grep -v "^./pinniped" | xargs -I _ dirname _)
 
 # Add tooling binaries here and in hack/tools/Makefile
 GOLANGCI_LINT      := $(TOOLS_BIN_DIR)/golangci-lint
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -24,6 +26,7 @@ ADDONS_DIR := addons
 YTT_TESTS_DIR := pkg/v1/providers/tests
 PACKAGES_SCRIPTS_DIR := $(abspath hack/packages/scripts)
 UI_DIR := pkg/v1/tkg/web
+GO_MODULES=$(shell find . -path "*/go.mod" | xargs -I _ dirname _)
```
rajathagasthya, line 29:
```
$ find . -path "*/go.mod" | xargs -I _ dirname _
.
./addons
./addons/pinniped/post-deploy
./hack/tools
./hack/packages/kbld-image-replace
./pinniped/generated/1.18/apis
./pinniped/generated/1.18/client
./pinniped/generated/1.20/apis
./pinniped/generated/1.20/client
./pinniped/generated/1.19/apis
./pinniped/generated/1.19/client
./pinniped/generated/1.17/apis
./pinniped/generated/1.17/client
./pinniped
./pkg/v1/providers/tests
```

This seems to included all the `generated/*` modules from the cloned pinniped repo. Should we filter these out?

stmcginnis: Ah, good point. Probably not. Let me filter those out quick...

rajathagasthya (approved):
Thanks for the changes!

vuil (approved):
lgtm, this is a nice bit of generalization!

PR #1794, opened 2022-03-06
Check out current source for linting

### LINKED ISSUE #1793: Lint job does not run against proposed PR changes
**Bug description**

It was noticed in https://github.com/vmware-tanzu/tanzu-framework/pull/1512 that there were changes to the `lint` target, but those changes were not reflected in the job execution. It looks like this is because the Lint job does not perform the checkout action, so it doesn't actually run against the proposed changes.

- [ ] APIs
- [ ] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [X] Test and Release
- [ ] User Experience

stmcginnis:

It was noticed in another PR that changes to the linter target were not
picked up, missing some issues. On investigation, it was found that this
was a configuration issue in the Lint job. GitHub Actions need to
explicitly check out the code of the PR to pick up the proposed changes,
so even though the PR made changes, they were not reflected in the
results of the job.

This adds the checkout action to the Lint job to make sure it always
runs against the proposed changes.

Fixes #1793

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -89,6 +89,9 @@ jobs:
         go-version: '1.17'
       id: go
 
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v1
+
     - name: Run lint checks
       run: |
         make lint


stmcginnis:
Looks like we may need the increased golangci-lint timeout being done in https://github.com/vmware-tanzu/tanzu-framework/pull/1786 first.

.github/workflows/main.yaml
```diff
@@ -89,6 +89,9 @@ jobs:
         go-version: '1.17'
       id: go
 
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v1
+
```
navidshaikh, line 94: we checkout at line no: 84 as well, its about ordering? 🤔

stmcginnis: Wow, for some reason when I did this I did not see that on line 84 at all! I'll close this PR.

Though that still doesn't explain why the other jobs failed to see the changes in the code when running linting. I guess that will require further investigation.

PR #1795, opened 2022-03-07
Update checkout action to v3 release

stmcginnis:

There is a newer v3 release of the `actions/checkout` GitHub action.
This updates all usage of this action to use the latest version to pick
up fixes and improvements.

Related #1793

Difficult to test locally, so testing is going to need to be done by getting results of the actions that run.


stmcginnis:
Looks like the behavior of the cloned repo changes significantly with the v3 action. This isn't necessarily required, so just going to close for now.

PR #1802, opened 2022-03-07
UI: remove node server mock certificate and key

miclettej:
Removes mock certificate and key that were used for UI's node.js server only when
running the server in HTTPS mode. This is no longer used, so removing the mock cert and
key to avoid these creds from showing up in an audit.

Confirmed that node.js server still serves mock REST APIs for UI in local dev mode

### Release note

```release-note

```

### PR Checklist


raymondz1 (approved):
LGTM

anujc25 (approved):

PR #1803, opened 2022-03-08
Adds function to continuously manage webhook TLS certificates

### LINKED ISSUE #1583: Admission Webhook for AddonConfig APIs
Per uTKG addon mgmt design, each addon will have its own config API. We would like to have an admission webhook to handle the CRUD of Addon Config API based on the design(We would have some objects/fields immutable).

Details:

The admission webhook should be validating all the newly introduced APIs

- [ ] #1689
- [x] #1690

adduarte:
- New self signed server certificates, key, and ca cert are created
  and placed in a k8s opaque secret.
- The server certificate and key are written to the specified path
- Webhook Mutating and Validating configurations are augmented with new CA bundle
- The new certificates are monitored for validity and rotated when appropriate.

An example of integrationg in addons/main.go

```
#define WebhookTLS struct will all relevant information: 

	webhookTLS := webhooks.WebhookTLS{
		Ctx:           ctx,
		K8sConfig:     k8sConfig,
		CertPath:      "/tmp/somedirectory/tls.cert",    // or use default value ofr mgr.Option{CertDir: }
		KeyPath:       "/tmp/somedirectory/tls.key",
		Name:         "tls-secret",
		ServiceName:   "webhook-tls",
		LabelSelector: "webhook-cert=self-managed",  // this is the label that will be used to select the webhook configs
		Logger:        setupLog,
		Namespace:     addonNamespace,
		RotationTime: time.Hour * 24,        // time.Duration in which certificates will be rotated. Max one week.
	}

     # go function: 
     checkFrequency = time.Hour // Check certificates every hour
     err := webhookTLS.ManageCertificates(managementFrequency)
     if err != nil {
       #do something about/with err because certificate management failed to start
     }


    # Start manager with ctrl.Options { ... CertDir: "/tmp/somedirectory" }
```

Fixes #1583

### Release note

```release-note
package-based-lcm: Adds function to continuously manage webhook TLS certificates
```

### PR Checklist

### DIFF
--- a/addons/controllers/suite_test.go
+++ b/addons/controllers/suite_test.go
@@ -79,7 +79,7 @@ var (
 	cancel        context.CancelFunc
 	certPath      string
 	keyPath       string
-	// clientset     *kubernetes.Clientset
+	tmpDir        string
 )
 
 func TestAddonController(t *testing.T) {
@@ -164,7 +164,7 @@ var _ = BeforeSuite(func(done Done) {
 	Expect(err).ToNot(HaveOccurred())
 	Expect(dynamicClient).ToNot(BeNil())
 
-	tmpDir, err := os.MkdirTemp("/tmp", "webhooktest")
+	tmpDir, err = os.MkdirTemp("/tmp", "webhooktest")
 	Expect(err).ToNot(HaveOccurred())
 	certPath = path.Join(tmpDir, "tls.cert")
 	keyPath = path.Join(tmpDir, "tls.key")
--- a/addons/controllers/webhooks_test.go
+++ b/addons/controllers/webhooks_test.go
@@ -1,0 +1,153 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package controllers
+
+import (
+	"context"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	cert2 "k8s.io/client-go/util/cert"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/webhooks"
+)
+
+const (
+	oneDay   = time.Hour * 24
+	oneWeek  = time.Hour * 24 * 7
+	zeroTime = time.Second * 0
+)
+
+var setupLog = ctrl.Log.WithName("controllers").WithName("Addon")
+
+var _ = Describe("when webhook TLS is being continuously managed", func() {
+	Context("if check frequency time is reached", func() {
+		It("tls certificates should be validated and rotated if invalid", func() {
+			privCtx := context.Background()
+			cacelCtx, cancelFn := context.WithCancel(privCtx)
+			rotationTime := time.Second
+			managementFrequency := rotationTime / 2
+			webhookTLS := webhooks.WebhookTLS{
+				Ctx:           cacelCtx,
+				K8sConfig:     k8sConfig,
+				CertPath:      certPath,
+				KeyPath:       keyPath,
+				Name:          webhookScrtName,
+				ServiceName:   webhookServiceName,
+				LabelSelector: "webhook-cert=self-managed",
+				Logger:        setupLog,
+				Namespace:     addonNamespace,
+				RotationTime:  rotationTime,
+			}
+
+			err := webhookTLS.ManageCertificates(managementFrequency)
+			Expect(err).ToNot(HaveOccurred())
+			time.Sleep(managementFrequency + rotationTime) // we wait longer than one rotation to allow for certificate installation
+			firstCert, err := cert2.CertsFromFile(webhookTLS.CertPath)
+			Expect(err).ToNot(HaveOccurred())
+			firstCertPEM, err := cert2.EncodeCertificates(firstCert[0])
+			Expect(err).ToNot(HaveOccurred())
+
+			time.Sleep(2 * rotationTime) // wait enough time to ensure the certificate has been rotated.
+
+			secondCert, err := cert2.CertsFromFile(webhookTLS.CertPath)
+			Expect(err).ToNot(HaveOccurred())
+			secondCertPEM, err := cert2.EncodeCertificates(secondCert[0])
+			Expect(err).ToNot(HaveOccurred())
+
+			cancelFn()
+
+			//The certificates should be different due to rotation
+			Expect(secondCertPEM).ToNot(Equal(firstCertPEM))
+
+		})
+	})
+	Context("if rotation time is longer than one week", func() {
+		It("should set rotation to one week", func() {
+			rotationTime := oneWeek + oneDay
+			managementFrequency := rotationTime / 2
+			webhookTLS := webhooks.WebhookTLS{
+				Ctx:           context.Background(),
+				K8sConfig:     k8sConfig,
+				CertPath:      certPath,
+				KeyPath:       keyPath,
+				Name:          webhookScrtName,
+				ServiceName:   webhookServiceName,
+				LabelSelector: "webhook-cert=self-managed",
+				Logger:        setupLog,
+				Namespace:     addonNamespace,
+				RotationTime:  rotationTime,
+			}
+			err := webhookTLS.ManageCertificates(managementFrequency)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(webhookTLS.RotationTime).To(Equal(oneWeek))
+
+		})
+	})
+	Context("if rotation time is 0 or less than 0", func() {
+		It("should set rotation to one day", func() {
+			rotationTime := zeroTime
+			managementFrequency := oneDay //any value larger than 0 will work here
+			webhookTLS := webhooks.WebhookTLS{
+				Ctx:           context.Background(),
+				K8sConfig:     k8sConfig,
+				CertPath:      certPath,
+				KeyPath:       keyPath,
+				Name:          webhookScrtName,
+				ServiceName:   webhookServiceName,
+				LabelSelector: "webhook-cert=self-managed",
+				Logger:        setupLog,
+				Namespace:     addonNamespace,
+				RotationTime:  rotationTime,
+			}
+			err := webhookTLS.ManageCertificates(managementFrequency)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(webhookTLS.RotationTime).To(Equal(oneDay))
+
+		})
+	})
+	Context("if rotation time is one week", func() {
+		It("tls certificates should be validated and not rotated until one week", func() {
+			privCtx := context.Background()
+			cacelCtx, cancelFn := context.WithCancel(privCtx)
+			rotationTime := oneWeek
+			managementFrequency := time.Second / 2 //manage the certificates every 1/2 second for testing purposes
+			webhookTLS := webhooks.WebhookTLS{
+				Ctx:           cacelCtx,
+				K8sConfig:     k8sConfig,
+				CertPath:      certPath,
+				KeyPath:       keyPath,
+				Name:          webhookScrtName,
+				ServiceName:   webhookServiceName,
+				LabelSelector: "webhook-cert=self-managed",
+				Logger:        setupLog,
+				Namespace:     addonNamespace,
+				RotationTime:  rotationTime,
+			}
+
+			err := webhookTLS.ManageCertificates(managementFrequency)
+			Expect(err).ToNot(HaveOccurred())
+			time.Sleep(managementFrequency * 3) // manage the certificates three times
+			firstCert, err := cert2.CertsFromFile(webhookTLS.CertPath)
+			Expect(err).ToNot(HaveOccurred())
+			firstCertPEM, err := cert2.EncodeCertificates(firstCert[0])
+			Expect(err).ToNot(HaveOccurred())
+
+			time.Sleep(managementFrequency * 3) // manage the certificate three times
+			secondCert, err := cert2.CertsFromFile(webhookTLS.CertPath)
+			Expect(err).ToNot(HaveOccurred())
+			secondCertPEM, err := cert2.EncodeCertificates(secondCert[0])
+			Expect(err).ToNot(HaveOccurred())
+
+			cancelFn()
+
+			//The certificates should be the same since we did not hit one week rotation
+			Expect(secondCertPEM).To(Equal(firstCertPEM))
+
+		})
+	})
+
+})
--- a/pkg/v1/webhooks/certificates.go
+++ b/pkg/v1/webhooks/certificates.go
@@ -11,25 +11,28 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	certutil "k8s.io/client-go/util/cert"
 	"k8s.io/client-go/util/keyutil"
+	"k8s.io/client-go/util/retry"
 	"knative.dev/pkg/webhook/certificates/resources"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func addCertsToWebhookConfigs(ctx context.Context, client kubernetes.Interface, labelSelector string, secret *corev1.Secret) error {
+func addCertsToWebhookConfigs(ctx context.Context, k8sclient kubernetes.Interface, labelSelector string, secret *corev1.Secret) error {
 	if labelSelector == "" {
 		return fmt.Errorf("label selector not provided for webhook configurations udpate")
 	}
-	allValidatingWebhookConfigurations, err := client.AdmissionregistrationV1().ValidatingWebhookConfigurations().List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
+	allValidatingWebhookConfigurations, err := k8sclient.AdmissionregistrationV1().ValidatingWebhookConfigurations().List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
 	if err != nil {
 		return err
 	}
 	for idx := range allValidatingWebhookConfigurations.Items {
 		configName := allValidatingWebhookConfigurations.Items[idx].Name
-		validatingWebhookConfiguration, err := client.AdmissionregistrationV1().ValidatingWebhookConfigurations().Get(ctx, configName, metav1.GetOptions{})
+		validatingWebhookConfiguration, err := k8sclient.AdmissionregistrationV1().ValidatingWebhookConfigurations().Get(ctx, configName, metav1.GetOptions{})
 		if err != nil {
 			return err
 		}
@@ -37,18 +40,18 @@ func addCertsToWebhookConfigs(ctx context.Context, client kubernetes.Interface, 
 		for idx := range validatingWebhookConfiguration.Webhooks {
 			validatingWebhookConfiguration.Webhooks[idx].ClientConfig.CABundle = secret.Data[resources.CACert]
 		}
-		if _, err := client.AdmissionregistrationV1().ValidatingWebhookConfigurations().Update(ctx, validatingWebhookConfiguration, metav1.UpdateOptions{}); err != nil {
+		if _, err := k8sclient.AdmissionregistrationV1().ValidatingWebhookConfigurations().Update(ctx, validatingWebhookConfiguration, metav1.UpdateOptions{}); err != nil {
 			return fmt.Errorf("error updating CA cert of ValidatingWebhookConfiguration %s: %v", configName, err)
 		}
 	}
 
-	allMutatingWebhookConfigurations, err := client.AdmissionregistrationV1().MutatingWebhookConfigurations().List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
+	allMutatingWebhookConfigurations, err := k8sclient.AdmissionregistrationV1().MutatingWebhookConfigurations().List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
 	if err != nil {
 		return err
 	}
 	for idx := range allMutatingWebhookConfigurations.Items {
 		configName := allMutatingWebhookConfigurations.Items[idx].Name
-		mutatingWebhookConfiguration, err := client.AdmissionregistrationV1().MutatingWebhookConfigurations().Get(ctx, configName, metav1.GetOptions{})
+		mutatingWebhookConfiguration, err := k8sclient.AdmissionregistrationV1().MutatingWebhookConfigurations().Get(ctx, configName, metav1.GetOptions{})
 		if err != nil {
 			return err
 		}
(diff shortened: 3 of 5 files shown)


addons/controllers/webhooks/webhooks.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	return nil
+}
+
+func ManageTLSCertificates(whtls *WebhookTLS, checkFrequency time.Duration) {
+	for {
+		select {
+		case <-whtls.Ctx.Done():
+			return
+		default:
```
vijaykatam, line 50: lets update to `time.After` and remove `time.Sleep`
```
 case <-time.After(checkFrequency):
```

addons/controllers/webhooks/webhooks.go
```diff
@@ -0,0 +1,58 @@
...
+			err := whtls.Manage()
+			if err == nil {
```
vijaykatam, line 52: err != nil and include namespace and labelselector in the log

addons/controllers/webhooks_test.go
```diff
@@ -0,0 +1,155 @@
...
+		"Webhooks TLS",
+		[]Reporter{printer.NewlineReporter{}})
+}
+
+var (
+	setupLog = ctrl.Log.WithName("controllers").WithName("Addon")
+)
+var _ = Describe("when TLS secret is created", func() {
+
+	Context("If the certificates do not exists", func() {
```
danniel1205, line 35: nit: usually lower case. e.g., `if xxxx`

addons/controllers/suite_test.go
```diff
@@ -79,6 +79,7 @@ var (
 	cancel        context.CancelFunc
 	certPath      string
 	keyPath       string
+	tmpDir        string
 	// clientset     *kubernetes.Clientset
```
danniel1205, line 83: question: this might not be from this PR, but could we remove the commented line in above ?

pkg/v1/webhooks/certificates.go
```diff
@@ -122,3 +135,30 @@ func ValidateTLSSecret(tlsSecret *corev1.Secret, certGracePeriod time.Duration)
 	}
 	return nil
 }
+
+// Write secret to persistent storage
+func updatePersistentTLSSecret(ctx context.Context, clusterClient client.Client, tlsSecret *corev1.Secret) error {
```
vijaykatam, line 140: Rename to updateTLSsecret. Nothing special about saving to k8s.

addons/controllers/webhooks/webhooks.go
```diff
@@ -0,0 +1,58 @@
...
+	Name          string
+	ServiceName   string
+	LabelSelector string
+	Logger        logr.Logger
+	TLSSecret     *corev1.Secret
+	Namespace     string
+	InvalidWithIn time.Duration
+}
+
+func (w *WebhookTLS) Manage() error {
```
vijaykatam, line 32: Do we expect this to be used by callers directly or is an internal API?

addons/controllers/webhooks_test.go
```diff
@@ -0,0 +1,155 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			orgKey, err = keyutil.ParsePrivateKeyPEM(testTLSSecret.TLSSecret.Data[resources.ServerKey])
+			Expect(err).ToNot(HaveOccurred())
+			Expect(key).To(Equal(orgKey))
+			Expect(certPEM1).To(Equal(certPEM2))
+		})
+	})
+})
+
+var _ = Describe("when TLS certificates are being managed", func() {
```
vijaykatam, line 115: Does this test verify cert rotation? If so, lets update the test description.

addons/controllers/webhooks_test.go
```diff
@@ -0,0 +1,155 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(certPEM).To(Equal(testTLSSecret.TLSSecret.Data[resources.ServerCert]))
+			key, err := keyutil.PrivateKeyFromFile(keyPath)
+			Expect(err).ToNot(HaveOccurred())
+			orgKey, err := keyutil.ParsePrivateKeyPEM(testTLSSecret.TLSSecret.Data[resources.ServerKey])
+			Expect(err).ToNot(HaveOccurred())
+			Expect(key).To(Equal(orgKey))
+		})
+	})
+	Context("If the certificates does exists", func() {
```
danniel1205, line 66: nit: typo `do exist`

addons/controllers/webhooks_test.go
```diff
@@ -0,0 +1,155 @@
...
+		It("should install new certificates", func() {
```
danniel1205, line 67: question: why we should install new certificates if the certificates exist ?

addons/controllers/webhooks_test.go
```diff
@@ -0,0 +1,155 @@
...
+				ServiceName:   webhookServiceName,
+				LabelSelector: "webhook-cert=self-managed",
+				Logger:        setupLog,
+				TLSSecret:     nil,
+				Namespace:     addonNamespace,
+				InvalidWithIn: time.Hour * 24,
+			}
+			err := testTLSSecret.Manage()
+			Expect(err).ToNot(HaveOccurred())
+			Expect(err).ToNot(HaveOccurred())
```
danniel1205, line 83: blocking: seems L82, L83 are duplicates.

addons/controllers/webhooks_test.go
```diff
@@ -0,0 +1,155 @@
...
```
danniel1205, line 52: blocking: dup assertion in above.

addons/controllers/webhooks_test.go
```diff
@@ -0,0 +1,155 @@
...
+			Expect(certPEM1).To(Equal(testTLSSecret.TLSSecret.Data[resources.ServerCert]))
+			key, err := keyutil.PrivateKeyFromFile(keyPath)
+			Expect(err).ToNot(HaveOccurred())
+			orgKey, err := keyutil.ParsePrivateKeyPEM(testTLSSecret.TLSSecret.Data[resources.ServerKey])
+			Expect(err).ToNot(HaveOccurred())
+			Expect(key).To(Equal(orgKey))
+
+			err = testTLSSecret.Manage()
+			Expect(err).ToNot(HaveOccurred())
+			Expect(err).ToNot(HaveOccurred())
```
danniel1205, line 98: blocking: dup assertion in above.

addons/controllers/webhooks_test.go
```diff
@@ -0,0 +1,155 @@
...
+		})
+	})
+})
+
+var _ = Describe("when TLS certificates are being managed", func() {
+	Context("TLS certificate Management", func() {
+		It("should continuously update and validate certificates ", func() {
+			privCtx := context.Background()
+			cacelCtx, cancelFn := context.WithCancel(privCtx)
+			invalidTime := (time.Hour * 24 * 7) - (time.Second * 10)
```
danniel1205, line 120: question: why is `(time.Hour * 24 * 7) - (time.Second * 10)`, would it be a good idea to add some comments ?

addons/controllers/webhooks_test.go
```diff
@@ -0,0 +1,155 @@
...
+				ServiceName:   webhookServiceName,
+				LabelSelector: "webhook-cert=self-managed",
+				Logger:        setupLog,
+				TLSSecret:     nil,
+				Namespace:     addonNamespace,
+				InvalidWithIn: invalidTime,
+			}
+
+			go webhooks.ManageTLSCertificates(&webhookTLS, time.Second*10)
+			time.Sleep(time.Second * 10)
```
danniel1205, line 136: question: why 10 sec ? Would it be possible to add some comments ? I am assuming it is related to the `time.Second*10` passed in on L135, but just a guess. And would it be possible to reduce the time so that tests could run faster ?

addons/controllers/webhooks_test.go
```diff
@@ -0,0 +1,155 @@
...
+			firstCert, err := cert2.CertsFromFile(webhookTLS.CertPath)
+			Expect(err).ToNot(HaveOccurred())
+			firstCertPEM, err := cert2.EncodeCertificates(firstCert[0])
+			Expect(err).ToNot(HaveOccurred())
+
+			time.Sleep(time.Second * 20)
```
danniel1205, line 142: question: trying to understand why 20sec here, would it be a good idea to have some comments ?

addons/controllers/webhooks/webhooks.go
```diff
@@ -0,0 +1,58 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package webhooks provides functions to manage webhook TLS certificates
+package webhooks
```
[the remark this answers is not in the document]
vijaykatam: +1, I think these are pretty reusable but Adolfo mentioned issues with tests

addons/controllers/webhooks/webhooks.go
```diff
@@ -0,0 +1,58 @@
...
+type WebhookTLS struct {
+	Ctx           context.Context
+	K8sConfig     *rest.Config
+	CertPath      string
+	KeyPath       string
+	Name          string
+	ServiceName   string
+	LabelSelector string
+	Logger        logr.Logger
+	TLSSecret     *corev1.Secret
```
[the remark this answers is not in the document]
vijaykatam: agreed, blocker

addons/controllers/webhooks/webhooks.go
```diff
@@ -0,0 +1,58 @@
...
+		w.Logger.Info("Installing new certificates")
+		w.TLSSecret, err = webhooks.InstallNewCertificates(w.Ctx, w.K8sConfig, w.CertPath, w.KeyPath, w.Name, w.Namespace, w.ServiceName, w.LabelSelector)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func ManageTLSCertificates(whtls *WebhookTLS, checkFrequency time.Duration) {
```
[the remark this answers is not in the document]
vijaykatam: +1, make it simpler for the caller

addons/controllers/webhooks/webhooks.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func (w *WebhookTLS) Manage() error {
+	err := webhooks.ValidateTLSSecret(w.TLSSecret, w.InvalidWithIn)
+	if err != nil {
+		w.Logger.Info("Invalid webhook tls secret: " + err.Error())
+		w.Logger.Info("Installing new certificates")
+		w.TLSSecret, err = webhooks.InstallNewCertificates(w.Ctx, w.K8sConfig, w.CertPath, w.KeyPath, w.Name, w.Namespace, w.ServiceName, w.LabelSelector)
+		if err != nil {
+			return err
```
danniel1205, line 39: question: do we want to put some logging before returning error ? Or it has ben handled by `InstallNewCertificates` already ?

pkg/v1/webhooks/certificates.go
danniel1205: I guess you mean `updateTLSSecret` or maybe `createOrUpdateTLSSecret` (which is more consistent with the naming convention we used in other controllers). But this is just my personal opinion, whatever name makes sense would be feasible.

pkg/v1/webhooks/certificates.go
```diff
@@ -122,3 +135,30 @@ func ValidateTLSSecret(tlsSecret *corev1.Secret, certGracePeriod time.Duration)
...
+		err = clusterClient.Create(ctx, tlsSecret)
+		if err != nil {
+			return err
+		}
+	}
+	if err != nil {
+		return err
+	}
+
+	err = clusterClient.Update(ctx, tlsSecret)
```
danniel1205, line 159: question: do we want to do `retry.RetryOnConflict()` Or let caller to handle the errors such as "the object has been modified, please get the latest and try again"

addons/controllers/webhooks/webhooks.go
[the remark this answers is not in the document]
adduarte: yeap. Agreed. we don't need the TLSSecret, will hide it from interface. (lowercase it)

addons/controllers/webhooks/webhooks.go
adduarte: yeap. will add. InstallNewCertificates will not log, it returns error with info, so we should probably log it as you point out.

addons/controllers/webhooks/webhooks.go
```diff
@@ -0,0 +1,58 @@
...
+		}
+	}
+	return nil
+}
+
+func ManageTLSCertificates(whtls *WebhookTLS, checkFrequency time.Duration) {
+	for {
+		select {
```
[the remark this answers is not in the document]
adduarte: +1

addons/controllers/webhooks/webhooks.go
adduarte: +1

addons/controllers/webhooks/webhooks.go
adduarte: +1

addons/controllers/webhooks_test.go
adduarte: not nit. You are correct. it should be lower

addons/controllers/webhooks_test.go
adduarte: It should read "replace with new certificate".

addons/controllers/webhooks_test.go
adduarte: they are

addons/controllers/webhooks_test.go
adduarte: +1

addons/controllers/webhooks_test.go
adduarte: Here I am jus trying to make sure we don't pickup the same certificate. 
The MageneTLSCertificate is configured to rotate the certificate every 10 seconds. 
I guess we could make this 15 seconds, in theory any number larger than 10 would work, but 20 is a good buffer I think

pkg/v1/webhooks/certificates.go
adduarte: concistency is good

addons/controllers/webhooks/webhooks.go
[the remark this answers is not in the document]
adduarte: hid tlsSecret from user. added ServerCert(), ServerKey(), and CACert() instead to obtain the respective []byte. 
This decouples the structure of the *coreve1.Secret being used to store the values.

addons/controllers/webhooks/webhooks.go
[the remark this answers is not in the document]
adduarte: switched to ticker instead of sleep.

addons/controllers/webhooks/webhooks.go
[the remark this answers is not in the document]
adduarte: done

addons/controllers/webhooks_test.go
adduarte: done

addons/controllers/suite_test.go
adduarte: done

addons/controllers/webhooks/webhooks.go
adduarte: impleented ticker instead.

pkg/v1/webhooks/certificates.go
adduarte: done

addons/controllers/webhooks/webhooks.go
adduarte: Callers will use it directly. For example, ManageTLSCertificates, is a caller and will only have access to it if we make expose it. 
(I think that is what you mean).

addons/controllers/webhooks/webhooks.go
adduarte: done

addons/controllers/webhooks_test.go
adduarte: done

addons/controllers/webhooks_test.go
adduarte: done

addons/controllers/webhooks_test.go
adduarte: it should reach "should replace with new certificate after rotation time"

adduarte: done

adduarte: Removed these tests since they are just a subset of the "continoulsy manage test"
no need to double test.

addons/controllers/webhooks_test.go
adduarte: Rename "InvalidWithIn" to "RotationTime" and instead derived the invalidwithin time from it. Makes more sense from user perspective if they provide the time they want the certificate to be rotated vs the time the certificates is invalid before 7 days.... (see confusing to say :) )

addons/controllers/webhooks_test.go
adduarte: Reworked the tests to instead of use a explicit time, to use a time derived from the rotation time provided by user. 
Makes test more flexible, and clear.

addons/controllers/webhooks_test.go
adduarte: Made tests derive their "wait" times from the rotation time provided for certificates.

addons/controllers/webhooks_test.go
adduarte: Removed test as it is redundant. Other tests cover these same steps.

addons/controllers/webhooks_test.go
adduarte: removed tests as redundant.

addons/controllers/webhooks_test.go
adduarte: removed redundant test

addons/controllers/webhooks_test.go
adduarte: removed redundant test

addons/controllers/webhooks/webhooks.go
adduarte: Managed() is now meant to be called.  "err" will contain the error itself. 
It is up to the user (caller) to decide if the error should be logged or not. Otherwise we end up with multiple logs of the same error.  and since the what the error will be is not predictable we can just return it and let the calle handle the logging, as is done in ManageTLSCertificates

addons/controllers/webhooks/webhooks.go
adduarte: Renamed Mange -> UpdateAndCreate to remove ambiguity. It really does not manage. It updates or creates. 
The caller to this function will be the manager.

addons/controllers/webhooks/webhooks.go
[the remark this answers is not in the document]
adduarte: We can move this to the pkg directory, and look at weather or not is worth moving the tests to that directory as well. (in another PR).

addons/controllers/webhooks/webhooks.go
```diff
@@ -0,0 +1,58 @@
...
+	ServiceName   string
+	LabelSelector string
+	Logger        logr.Logger
+	TLSSecret     *corev1.Secret
+	Namespace     string
+	InvalidWithIn time.Duration
+}
+
+func (w *WebhookTLS) Manage() error {
+	err := webhooks.ValidateTLSSecret(w.TLSSecret, w.InvalidWithIn)
```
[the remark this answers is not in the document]
adduarte: Added logic to read secret from cluster and update WebhooTLS.secret if exists, or create if not found

pkg/v1/webhooks/certificates.go
adduarte: Done.

pkg/v1/webhooks/webhooks.go
```diff
@@ -0,0 +1,107 @@
...
+		Namespace: w.Namespace,
+		Name:      w.Name}, currentSecret)
+	if err == nil {
+		w.secret = currentSecret
+	} else if !apierrors.IsNotFound(err) { // secret not found = "Create" case.
+		return err
+	}
+
+	err = ValidateTLSSecret(w.secret, gracePeriod)
+	if err != nil {
```
vijaykatam, line 57: question: so any error results is new cert?

pkg/v1/webhooks/certificates.go
```diff
@@ -122,3 +136,32 @@ func ValidateTLSSecret(tlsSecret *corev1.Secret, certGracePeriod time.Duration)
 	}
 	return nil
 }
+
+// write secret to cluster
+func updateOrCreateTLSSecret(ctx context.Context, clusterClient client.Client, tlsSecret *corev1.Secret) error {
+	// if secret is nil nothign to do
```
vijaykatam, line 142: nit: typo

pkg/v1/webhooks/certificates.go
```diff
@@ -122,3 +136,32 @@ func ValidateTLSSecret(tlsSecret *corev1.Secret, certGracePeriod time.Duration)
 	}
 	return nil
 }
+
+// write secret to cluster
+func updateOrCreateTLSSecret(ctx context.Context, clusterClient client.Client, tlsSecret *corev1.Secret) error {
```
vijaykatam, line 141: I would rename clusterClient to k8sclient, otherwise it can be confused with a remote clusterClient

vijaykatam (commented):
Minor comments

pkg/v1/webhooks/webhooks.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+const oneWeek = time.Hour * 24 * 7
+
+func (w *WebhookTLS) UpdateOrCreate() error {
+	if w.RotationTime > oneWeek {
+		w.Logger.Info("rotation time will be set to maximum allowed value of one Week")
+		w.RotationTime = oneWeek
+	}
+	gracePeriod := oneWeek - w.RotationTime // Time left to end of one week when certificates need to be rotated.
```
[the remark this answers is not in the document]
adduarte: yes. Probably, should add a default if no graceperiod is given

pkg/v1/webhooks/webhooks.go
adduarte: yes.  We could add more specific tests. Also perhaps I will make it return a boolean instead and we can log if an error.

pkg/v1/webhooks/webhooks.go
```diff
@@ -0,0 +1,107 @@
...
+func (w *WebhookTLS) CACert() []byte {
+	return w.secret.Data[resources.CACert]
+}
+
+func (w *WebhookTLS) ManageCertificates(frequency time.Duration) error {
+	err := w.UpdateOrCreate()
+	if err != nil {
+		return err
+	}
+	w.Logger.Error(err, w.Name)
```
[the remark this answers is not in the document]
adduarte: no, will remove.

pkg/v1/webhooks/certificates.go
adduarte: done

pkg/v1/webhooks/webhooks.go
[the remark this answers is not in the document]
adduarte: done

pkg/v1/webhooks/webhooks.go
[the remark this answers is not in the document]
adduarte: done

pkg/v1/webhooks/webhooks.go
adduarte: expanded ValidateTLSSecret to return a boolean and an error. currently any error will result on new certificate, but we can augment if required based on the retruned logic and error.

pkg/v1/webhooks/webhooks.go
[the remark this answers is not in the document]
adduarte: added code so the graceperiod is never less than zero (by enforcing a max rotation of oneweek which results in  0 graceperiod(which is valid, the certificate is valid for the whole week) , and minimum rotation of  > 0 which results in graceperiod < oneWeek. For example if rotation is 1 ms, then graceperiod is oneWeek - 1ms.

vijaykatam (approved):

PR #1809, opened 2022-03-09
Feature gate ClusterBootstrap and AddonConfig controllers

### LINKED ISSUE #1774: Feature gate all new addon controllers
All new controllers for unified TKG should be feature gated by https://github.com/vmware-tanzu/tanzu-framework/blob/main/pkg/v1/config/clientconfig.go#L56 so that 1.6 release does not activate them.

1. Feature gate controllers in https://github.com/vmware-tanzu/tanzu-framework/blob/main/addons/main.go with FeatureFlagPackageBasedLCM
2. Downstream build has https://docs.google.com/document/d/1Sp4oEf6J1R9Vu-DISdtpRrcyjR_IvMkqzdaZc7q0Ebw/ `gobuild.uTKGEnabled=True/false` which should enable or disable feature gate as input to package install for addons-controller

vijaykatam:
* Addons-manager binary feature gated with --feature-gate-cluster-bootstrap
* Addons-manager-v2 package feature gated to toggle CRD install

Feature ClusterBootstrap and AddonConfig controllers to disable by default

Fixes #1774

Generate manifests with feature off and on and diff

### Release note

```release-note
package-based-lcm: Feature gate ClusterBootstrap and AddonConfig controllers
```

### PR Checklist


packages/management/addons-manager-v2/bundle/config/overlays/feature-gate-clusterbootstrap.yaml
```diff
@@ -0,0 +1,12 @@
+#@ load("@ytt:data", "data")
+#@ load("@ytt:overlay", "overlay")
+
+#@ antreaconfigscrd = overlay.subset({"kind": "CustomResourceDefinition", "metadata": {"name": "antreaconfigs.cni.tanzu.vmware.com"}})
+#@ calicoconfigscrd = overlay.subset({"kind": "CustomResourceDefinition", "metadata": {"name": "calicoconfigs.cni.tanzu.vmware.com"}})
```
shyaamsn, line 5: **question**: We havent added CPI to the package ?

shyaamsn (approved):
LGTM

packages/management/addons-manager-v2/bundle/config/overlays/feature-gate-clusterbootstrap.yaml
vijaykatam: Don't see it. I'll add it and CPI controller as well.

blc1996 (approved):
LGTM. We can also add CPI controller as the first MR for it has been merged.

addons/main.go
```diff
@@ -188,3 +198,62 @@ func setupChecks(mgr ctrl.Manager) {
...
+	}).SetupWithManager(ctx, mgr, controller.Options{MaxConcurrentReconciles: 1}); err != nil {
+		setupLog.Error(err, "unable to create KappControllerConfig", "controller", "kapp")
+		os.Exit(1)
+	}
+	if err := (&cpicontroller.CPIConfigReconciler{
+		Client: mgr.GetClient(),
+		Log:    ctrl.Log.WithName("CPIConfig"),
+		Scheme: mgr.GetScheme(),
+	}).SetupWithManager(ctx, mgr, controller.Options{MaxConcurrentReconciles: 1}); err != nil {
+		setupLog.Error(err, "unable to create KappControllerConfig", "controller", "kapp")
```
danniel1205, line 233: nit: I guess above msg means `CPIConfigReconciler`

addons/main.go
```diff
@@ -188,3 +198,62 @@ func setupChecks(mgr ctrl.Manager) {
...
+		ctrl.Log.WithName("ClusterBootstrapController"),
+		mgr.GetScheme(),
+		&addonconfig.ClusterBootstrapControllerConfig{
+			CNISelectionClusterVariableName: constants.DefaultCNISelectionClusterVariableName,
+			HTTPProxyClusterClassVarName:    constants.DefaultHTTPProxyClusterClassVarName,
+			HTTPSProxyClusterClassVarName:   constants.DefaultHTTPSProxyClusterClassVarName,
+			NoProxyClusterClassVarName:      constants.DefaultNoProxyClusterClassVarName,
+			ProxyCACertClusterClassVarName:  constants.DefaultProxyCaCertClusterClassVarName,
+			IPFamilyClusterClassVarName:     constants.DefaultIPFamilyClusterClassVarName,
+			SystemNamespace:                 constants.TKGSystemNS,
```
danniel1205, line 248: question: do we want to make this configurable via a flag ? Remember TKGs uses a different system namespace.

addons/main.go
```diff
@@ -188,3 +198,62 @@ func setupChecks(mgr ctrl.Manager) {
...
+			PkgiServiceAccount:              constants.PackageInstallServiceAccount,
```
danniel1205, line 249: question: do we want to use the value from a flag var (`addonServiceAccount`) ? The same might apply to the following few lines.

packages/management/addons-manager-v2/bundle/config/overlays/overlay-addons-manager.yaml
```diff
@@ -1,57 +0,0 @@
...
-subjects:
-#@overlay/match by=overlay.subset({"kind":"ServiceAccount", "name": "tanzu-addons-manager-sa"})
-- kind: ServiceAccount
-  name: tanzu-addons-manager-sa
-  namespace: #@ data.values.tanzuAddonsManager.namespace
-
-#@overlay/match by=overlay.subset({"kind":"Deployment", "metadata": {"name": "tanzu-addons-controller-manager"}})
----
-apiVersion: apps/v1
-kind: Deployment
```
danniel1205, line 32: question: is the reason of removing this file because all overlay logic/template has been handled by `upstrea/addons-manager.yaml` ?

packages/management/addons-manager-v2/bundle/config/upstream/cpi.tanzu.vmware.com_cpiconfigs.yaml
```diff
@@ -0,0 +1,237 @@
+
```
danniel1205, line 1: question: will `make manifest` automatically update this file ? Or we have to manually sync the updates whenever there is a API change.

addons/main.go
vijaykatam: yup will fix

addons/main.go
vijaykatam: Agreed, the system namespace needs to be configurable. I will use existing flag and update package to pass the param as well.

addons/main.go
vijaykatam: We made service account, etc configurable for exisint addons-manager and never really saw use case. I think the idea is needs to be unique within the namespace and I prefixed with `tanzu-cluster-bootstrap-` which should be pretty unique. I suggest we revisit making configurable if we really run into it during testing.

packages/management/addons-manager-v2/bundle/config/upstream/cpi.tanzu.vmware.com_cpiconfigs.yaml
vijaykatam: No, vendir sync needs to be done after picking which commit to target.

danniel1205 (approved):

PR #1816, opened 2022-03-09
Load all credential helpers

scothis:

Any place we load one of the k8s credential helpers, load them all.
Since we don't know what clusters users will use Tanzu with, we need to
support everything.

A user reported being unable to use the `tanzu package` cli plugin with TKGi

```
Error: Unable to set up rest mapper: no Auth Provider found for name "oidc"
Error: exit status 1
```

### Release note

```release-note
Load all kubernetes credential helpers
```

### PR Checklist


maralavi:
Question: Does the following file needs change as well?
https://github.com/vmware-tanzu/tanzu-framework/blob/main/pkg/v2/tkr/webhook/infra-machine/main.go#L13

maralavi:
@scothis Thanks for the PR. You need to resolve the conflict for the ```addons/go.mod``` and the above question I left. Otherwise, the PR looks good to me.

shivaani0505:
PR looks good to me. Please resolve conflicts for go.mod.

scothis:
rebased

maralavi (approved):

danniel1205:
Thanks for the PR. @scothis could you help add `Describe testing done for PR` in PR description ? Thanks

scothis:
The short answer is "none", other than my current usage of the cli didn't break. I don't have the bandwidth to fully test oidc. If someone else who has access to a cluster can test it out, that would be appreciated.

vijaykatam:
Unless I am missing something this change removes gcp auth since root import only does oidc https://github.com/kubernetes/client-go/blob/master/plugin/pkg/client/auth/plugins.go.

Is that expected since gcp k8s 1.25 is [dropping support](https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke) for this plugin? https://github.com/kubernetes/client-go/blob/master/plugin/pkg/client/auth/gcp/gcp.go#L120

vijaykatam (approved):
Approved with the caveat that only tested on OIDC. Not tested on GCP.

Issue #1709, opened 2022-02-16
Antrea routable pod customization


vijaykatam:
AntreaConfig will be customized and pre-created when using routable pod feature. ClusterBootstrapController should be able to handle if a config already exists and only set ownerReferences.

Done criteria:
Env test demonstrating a pre-created AntreaConfig is updated with ownerReferences

shivaani0505:
@vijaykatam , My Analysis:
1. I see that NSXT_POD_ROUTING_ENABLED variable is available as a part of the Tanzu Cluster Configuration Variables
excel sheet

2. It is a configuration variable which can also be set in Tanzu Kubernetes cluster yaml 
( Section: NSX-T Pod Routing in https://docs.vmware.com/en/VMware-Tanzu-Kubernetes-Grid/1.4/vmware-tanzu-kubernetes-grid-14/GUID-tanzu-config-reference.html#common-variables-for-all-infrastructure-providers-0 )

3. Further, if NSXT_POD_ROUTING_ENABLED is true, ANTREA_PROXY should be set to true (https://github.com/vmware-tanzu/tanzu-framework/blob/e8a083109169b54f6d1ca1c5df16019dfcff1d8f/pkg/v1/providers/ytt/02_addons/cni/antrea/antrea_addon_data.lib.yaml#L58) ( for vsphere only). Also, set the following 2 variables trafficEncapMode: "noEncap"
      noSNAT: true

4. NSXT_POD_ROUTING_ENABLED can be fetched from cluster.Spec.Topology.Variables and should be able to use as “[tkg.tanzu.vmware.com/nsxt_pod_routing_enabled](http://tkg.tanzu.vmware.com/nsxt_pod_routing_enabled)”.  I saw some variables in excel sheet for cluster class are using this pattern => change to lower case have prefix as “[tkg.tanzu.vmware.com](http://tkg.tanzu.vmware.com/)[/](http://tkg.tanzu.vmware.com/)“ I have pinged @lzhan for confirmation.

shivaani0505:
@vijaykatam, some updates: 
- We can parse  NSXT_POD_ROUTING_ENABLED cluster variable by adding a utility function 
- NSXT_POD_ROUTING_ENABLED can be fetched from cluster.Spec.Topology.Variables
[...]

PR #1881, opened 2022-03-22
Support clone embedded local object reference into cluster namespace

### LINKED ISSUE #1773: Support cloning of embedded secrets in Config resources
Cloning of providers in ClusterBootstrapTemplate currently does not support cloning of references to secrets. Clone should inspect for secrets by potentially using a naming convention to clone any referenced secrets.

danniel1205:
Signed-off-by: Daniel Guo <danniel1205@gmail.com>

When clusterbootstrap_controller clones cluster bootstrap objects from system namespace into cluster namespace, it is possible that the object has embedded local object reference. Take `VSphereCPIConfig` as an example:

```
apiVersion: cpi.tanzu.vmware.com/v1alpha1
kind: VSphereCPIConfig
metadata:
  name: test-cluster-cpi
  namespace: tkg-system
spec:
  vsphereCPI:
    mode: vsphereCPI
    region: test-region
    zone: test-zone
    insecureFlag: true
    vmInternalNetwork: internal-net
    vmExternalNetwork: external-net
    vmExcludeInternalNetworkSubnetCidr: 192.168.3.0/24
    vmExcludeExternalNetworkSubnetCidr: 22.22.3.0/24
    vSphereCredentialLocalObjRef:
      kind: Secret
      name: cpi-vsphere-credential
```

When cloning `VSphereCPIConfig` CR, we need to clone the `vSphereCredentialLocalObjRef` resource as well.

This PR adds the support of cloning embedded local object reference into cluster namespace within `clusterbootstrap_controller.go` when handles bootstrap package. Detailed changes are listed as below:

- Change the secret reference within VSphereCPIConfig to be typed local object reference. Current approach with SecretReference allows the referenced Secret object to be under a different namespace. Allowing cross-namespace brings security concerns and make implementation complicated.
- Add util function to extract the GroupKind and resource names from unstructured object.
- Make changes to clusterbootstrap_controller.go to clone the embedded local object reference as part of updateValuesFrom logic.
- Make enhancements on local envtest.

Fixes #1773 

- Added tests into local envtest and verified the change.

### Release note

```release-note
package-based-lcm: 
The following fields of VSphereCPIConfig have been renamed:
- spec.vsphereCPI.vSphereCredentialRef -> spec.vsphereCPI.vSphereCredentialLocalObjRef
- spec.nsxt.nsxtCredentialsRef -> spec.nsxt.credentialLocalObjRef

And the type of above two fields are changed from SecretReference to TypedLocalObjectReference
```

### PR Checklist


HanFa:
Since the PR, it looks like the VSphereCPIConfig CR cannot cross-ref credentials from other namespace. I am good with this as long as we can ensure the invariant that they always exists within the same namespace.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1179,10 +1238,11 @@ func (r *ClusterBootstrapReconciler) updateValuesFromProvider(cluster *clusterap
 			// There are possibilities that current reconciliation loop fails due to various reasons, and during next reconciliation
 			// loop, it is possible that the provider resource has been created. In this case, we want to run update/patch.
 			if apierrors.IsAlreadyExists(err) {
-				// Setting the resource version is because it's been removed at L984 with 'unstructured.RemoveNestedField(newProvider.Object, "metadata")'
```
vijaykatam, line 1182: If resource already exists, lets fetch it from the server and only patch ownerReferences. This should cover the use case for Antrea routable pod customization https://github.com/vmware-tanzu/tanzu-framework/issues/1709. 

Patching ownerreferences should be something like https://gist.github.com/vijaykatam/a091168e747872a2176de16f720a6370

Please also add an envtest that pre-creates AntreaConfig.

vijaykatam: Both ownerrefs and labels.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1122,6 +1125,62 @@ func (r *ClusterBootstrapReconciler) updateValuesFromSecret(cluster *clusterapiv
...
+				return err
+			}
+
+			copiedObj := fetchedObj.DeepCopy()
+			unstructured.RemoveNestedField(copiedObj.Object, "metadata", "resourceVersion")
+			copiedObj.SetNamespace(cluster.Namespace)
+			_, err = r.dynamicClient.Resource(*gvr).Namespace(cluster.Namespace).Create(r.context, copiedObj, metav1.CreateOptions{})
+			if err != nil {
+				if apierrors.IsAlreadyExists(err) {
+					var jsonData []byte
```
lubronzhan, line 1161: Could we add log here to explicitly state that?
```
Resource %s already exist in target namespace %s, only patch the ownerRef and labels
```

addons/controllers/clusterbootstrap_controller.go
lubronzhan: Does that mean during the cluster/tkr upgrade process, GCM should also be responsible for updating the AntreaConfig to new version? So the GCM should watch the TKR associated with the cluster.

addons/controllers/clusterbootstrap_controller.go
danniel1205: yes, it is a good idea.

addons/controllers/clusterbootstrap_controller.go
danniel1205: yes, this also applies to ProviderRef. Above comments have been addressed with the latest commit.

addons/pkg/util/provider_util_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+		It("should return non-empty result", func() {
+			Expect(yaml.Unmarshal([]byte(validInput1), testData)).Should(Succeed())
+			result := util.ExtractTypedLocalObjectRef(testData, "LocalObjRef")
+			Expect(len(result)).To(Equal(2))
+			Expect(len(result[schema.GroupKind{Group: "", Kind: "Secret"}])).To(Equal(2))
+			Expect(len(result[schema.GroupKind{Group: "", Kind: "ConfigMap"}])).To(Equal(1))
+		})
+	})
+
```
vijaykatam, line 95: Could we also add a test for nested TypedLocalObjectRef?

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1122,6 +1125,62 @@ func (r *ClusterBootstrapReconciler) updateValuesFromSecret(cluster *clusterapiv
...
+				gvr.String(), resourceName, provider.GetNamespace(), cluster.Namespace))
+			fetchedObj, err := r.dynamicClient.Resource(*gvr).Namespace(provider.GetNamespace()).Get(r.context, resourceName, metav1.GetOptions{})
+			if err != nil {
+				r.Log.Error(err, fmt.Sprintf("unable to get provider: %s with name: %s under namespace: %s",
+					providerGVK.String(), provider.GetName(), provider.GetNamespace()))
+				return err
+			}
+
+			copiedObj := fetchedObj.DeepCopy()
+			unstructured.RemoveNestedField(copiedObj.Object, "metadata", "resourceVersion")
```
vijaykatam, line 1156: Why are we removing resourceVersion?

vijaykatam, line 1156: Why are we removing resourceVersion?

addons/controllers/clusterbootstrap_controller.go
vijaykatam: I guess I am missing where ownerRef and labels are being patched.

vijaykatam: I guess I am missing where ownerRef and labels are being patched.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1122,6 +1125,62 @@ func (r *ClusterBootstrapReconciler) updateValuesFromSecret(cluster *clusterapiv
...
+			copiedObj.SetNamespace(cluster.Namespace)
+			_, err = r.dynamicClient.Resource(*gvr).Namespace(cluster.Namespace).Create(r.context, copiedObj, metav1.CreateOptions{})
+			if err != nil {
+				if apierrors.IsAlreadyExists(err) {
+					var jsonData []byte
+					if jsonData, err = copiedObj.MarshalJSON(); err != nil {
+						return err
+					}
+					_, err = r.dynamicClient.Resource(*gvr).Namespace(cluster.Namespace).Patch(r.context, copiedObj.GetName(), types.MergePatchType, jsonData, metav1.PatchOptions{})
```
vijaykatam, line 1165: Is this safe? Seems like we are merging the whole object and with MergePatchType we can overwrite array elements when all we want is updated labels and ownerRef's

vijaykatam, line 1165: Is this safe? Seems like we are merging the whole object when all we want is updated labels and ownerRef's

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1179,12 +1239,19 @@ func (r *ClusterBootstrapReconciler) updateValuesFromProvider(cluster *clusterap
...
+				// Instantiate an empty unstructured and only set ownerReferences and Labels for patching
+				patchObj := unstructured.Unstructured{}
+				patchObj.SetLabels(newProvider.GetLabels())
+				patchObj.SetOwnerReferences(newProvider.GetOwnerReferences())
+				patchData, err := patchObj.MarshalJSON()
+				if err != nil {
+					log.Error(err, fmt.Sprintf("unable to patch provider %s/%s", newProvider.GetNamespace(), newProvider.GetName()), "gvr", gvr)
+					return nil, err
+				}
+				createdOrUpdatedProvider, err = r.dynamicClient.Resource(*gvr).Namespace(cluster.Namespace).Patch(r.context, newProvider.GetName(), types.MergePatchType, patchData, metav1.PatchOptions{})
```
vijaykatam, line 1252: Same comment, is this safe? We should be only patching ownerRefs and labels.

vijaykatam, line 1252: Same comment, is this safe? We should be only patching ownerRefs and labels.

addons/controllers/testdata/test-cluster-bootstrap-2.yaml
```diff
@@ -212,3 +237,39 @@ spec:
...
+      deploy:
+        - kapp:
+            rawOptions:
+              - --wait-timeout=30s
+              - --kube-api-qps=20
+              - --kube-api-burst=30
+  releasedAt: "2021-12-20T10:59:32Z"
+  valuesSchema:
+    openAPIv3:
+      title: vsphere-cpi.tanzu.vmware.com.1.22.3--vmware.1-tkg.1 values schema
```
vijaykatam, line 275: Could we also add a test case for pre-created Antrea routable config? This would cover the use case for customizing an addon config prior to creating a cluster.

vijaykatam, line 275: Could we also add a test case for pre-created Antrea routable config? This would cover the use case for customizing an addon config prior to creating a cluster.

vijaykatam (commented):
Nice work on embedded cloning and looks pretty close.

addons/controllers/clusterbootstrap_controller.go
danniel1205: yes, we should only patch ownerRefs and labels. That is why I added L1243-L1247. Those few lines are doing the following few things:
- create an empty map
- set only metadata.ownerRefs and metadata.labels field
- Marshal to json bytes. I.e., `{"metadata":{"labels":{"foo":"bar"},"ownerReferences":[{"apiVersion":"Foo","kind":"Bar","name":"Zoo","uid":""}]}}`
- Call dynamic client to patch

An alternative is to use `fmt.Sprintf()` to construct the json string manually. Using `unstructured.Unstructured{}` is just to create the patch data in a more "structured", readable and maintainable way.

addons/controllers/clusterbootstrap_controller.go
danniel1205: because `resoruceVersion` should not be set on object to be created. It is only used for concurrent control on updates. K8S errors out if you create an object with `resourceVersion` set.

addons/controllers/clusterbootstrap_controller.go
danniel1205: It would be nice if I could learn a little bit more on the requirements of `ownerRef and labels` for embedded object.

Take the following as an example, and have few questions:
```
apiVersion: cpi.tanzu.vmware.com/v1alpha1
kind: VSphereCPIConfig
metadata:
  name: test-cluster-cpi
  namespace: tkg-system
spec:
  vsphereCPI:
    mode: vsphereCPI
    region: test-region
    zone: test-zone
    insecureFlag: true
    vmInternalNetwork: internal-net
    vmExternalNetwork: external-net
    vmExcludeInternalNetworkSubnetCidr: 192.168.3.0/24
    vmExcludeExternalNetworkSubnetCidr: 22.22.3.0/24
    vSphereCredentialLocalObjRef:
      kind: Secret
      name: cpi-vsphere-credential
```
- Who is supposed to be the owner of secret `cpi-vsphere-credential` ?
  - VSphereCPIConfig ?
  - CAPI Cluster ?
- What labels do we have to add ?
  - `tkg.tanzu.vmware.com/package-name` ? I guess we do not want to add this label since this secret is not the data value secret for a package. It is only meaningful within the scope of VSphereCPIConfig.
  - `tkg.tanzu.vmware.com/cluster-name` ?  I do not think we want to add this, it will be enqueued for reconciliation by the event handler we set for secrets.

addons/controllers/clusterbootstrap_controller.go
danniel1205: I think the logic for embedded object is to createOrUpdate ? If so, `patch whole object` == `update`. One of the benefits of using patch is to get rid of `object has been modified` error from `update` and do not have to handle retry logic.

addons/pkg/util/provider_util_test.go
danniel1205: Is this what you mean ?
https://github.com/vmware-tanzu/tanzu-framework/pull/1881/files#diff-9154d6ec46b64e61c9c1aebd17c054bc3d4d6df4e272fbfa44bd484258d9c113R43

addons/controllers/clusterbootstrap_controller.go
danniel1205: added ownerrefs.

danniel1205: added ownerrefs.

addons/controllers/clusterbootstrap_controller.go
danniel1205: changed to patch owner ref only

danniel1205: changed to patch owner ref only

vijaykatam (approved):
LGTM

PR #1889, opened 2022-03-23
package/pinniped: generate RBAC using kubebuilder controller-gen

benjaminapetersen:
Signed-off-by: Benjamin A. Petersen <ben@benjaminapetersen.me>

Adds `controller-gen` to the tanzu-framework `hack/tools` directory.  
Ensures that `controller-gen` can be installed on a users machine.
Generates the RBAC for pinniped-config-controller using `controller-gen` 

Fixes #

### Release note

```release-note

```

### PR Checklist


addons/pinniped/config-controller/controllers/controllers.go
```diff
@@ -73,6 +73,10 @@ func (c *PinnipedController) SetupWithManager(manager ctrl.Manager) error {
 	return nil
 }
 
+// +kubebuilder:rbac:groups="",resources=secrets,verbs=list;watch;get;patch;update;delete
+// +kubebuilder:rbac:groups="",resources=configmaps,verbs=watch;get
```
sabbey37, line 77: can we add the `list` verb here, I think you were right when you called out that we might need it.

benjaminapetersen: aha yes!

hack/tools/Makefile
```diff
@@ -117,6 +121,11 @@ $(CONVERSION_GEN): go.mod
 	mkdir -p $(BIN_DIR)
 	go build -tags=tools -o $(BIN_DIR) k8s.io/code-generator/cmd/conversion-gen
 
+controller-gen: $(CONTROLLER_GEN) $(SRCS) ## Build controller-gen
+$(CONTROLLER_GEN): go.mod
```
vijaykatam, line 125: This is different from controller-gen in root Makefile. Could we make both consistent?

benjaminapetersen:
Linking up a previous unmerged PR #1105 with some similar intent (more generic, mine is to support my own use case, doesn't include as much general cleanup).

benjaminapetersen:
Holding this atm:
- Given a prev PR #1105 that installs `controller-gen` and `kustomize` (unmerged, some conflicts)
- Opened PR #1918 with `controller-gen` and `kustomize` in `/hack/tools` (subset of above)

If #1918 is preferrable, will go with that, then I will update this PR (#1889) to build on top of it.

benjaminapetersen:
Reopening based on #1918

PR #1908, opened 2022-03-24
addons/pinniped/config-controller: Watch clusterbootstrap secrets

sabbey37:
watch v1alpha3 secrets created by clusterbootstrap controller and add information from pinniped-info configmap

Fixes #

- added unit tests, will verify with clusterbootstrap controller

### PR Checklist

- [ x] Squash the commits into one or a small number of logical commits
      
- [x ] Use good commit [messages](https://github.com/vmware-tanzu/tanzu-framework/blob/main/CONTRIBUTING.md)
- [ x] Ensure PR contains terms all contributors can understand and links all contributors can access


addons/pinniped/config-controller/constants/constants.go
```diff
@@ -16,30 +16,42 @@ const (
...
 	TKGDataValueFieldName = "values.yaml"
 
-	// NamespaceLogKey is the log key for "namespace"
-	NamespaceLogKey = "namespace"
+	// SecretNamespaceLogKey is the log key for "secret namespace"
+	SecretNamespaceLogKey = "secret namespace"
 
-	// NameLogKey is the log key for "name"
-	NameLogKey = "name"
+	// SecretNameLogKey is the log key for "secret name"
```
benjaminapetersen, line 22: +1 for the specificity here!

addons/pinniped/config-controller/controllers/controllers.go
```diff
@@ -92,49 +79,50 @@ func (c *PinnipedController) Reconcile(ctx context.Context, req ctrl.Request) (r
 		log.V(1).Info("pinniped-info configmap not found, setting value to nil")
 		pinnipedInfoCM.Data = nil
 	}
-	// if req is empty, CM changed, let's loop through all clusters and create/update/delete secrets
+	// if req is empty, CM changed, let's loop through all secrets and create/update/delete as needed
```
benjaminapetersen, line 82: How about calling a separate function rather than a comment & all of this inline?

addons/pinniped/config-controller/controllers/controllers.go
```diff
@@ -92,49 +79,50 @@ func (c *PinnipedController) Reconcile(ctx context.Context, req ctrl.Request) (r
...
 				return reconcile.Result{}, err
 			}
 		}
 		return reconcile.Result{}, nil
 	}
 
-	// Get cluster from rec
-	cluster := clusterapiv1beta1.Cluster{}
-	if err := c.client.Get(ctx, req.NamespacedName, &cluster); err != nil {
+	// Get secret from rec
```
benjaminapetersen, line 112: I'm tempted to think this could be a separate function as well, keep `Reconcile()` fairly small & call out to other funcs that do the actual work.

addons/pinniped/config-controller/controllers/controllers.go
```diff
@@ -92,49 +79,50 @@ func (c *PinnipedController) Reconcile(ctx context.Context, req ctrl.Request) (r
...
 		if k8serror.IsNotFound(err) {
-			if err := c.reconcileClusterDelete(ctx, req.NamespacedName); err != nil {
-				return reconcile.Result{}, fmt.Errorf("failed to reconcile cluster delete: %w", err)
-			}
+			// If secret not found, assume cluster was deleted, secret will be deleted
+			log.V(1).Info("could not find secret, assuming it has been deleted")
 			return reconcile.Result{}, nil
 		}
-		log.Error(err, "error getting cluster")
+		log.Error(err, "error getting secret")
```
benjaminapetersen, line 120: Shall we be verbose and log the secret name also?

addons/pinniped/config-controller/controllers/controllers.go
```diff
@@ -92,49 +79,50 @@ func (c *PinnipedController) Reconcile(ctx context.Context, req ctrl.Request) (r
...
 
-	if utils.IsManagementCluster(&cluster) {
-		log.V(1).Info("skipping reconciliation of management cluster")
-		return reconcile.Result{}, nil
-	}
-
-	if err := c.reconcileAddonSecret(ctx, &cluster, pinnipedInfoCM); err != nil {
-		log.Error(err, "Error reconciling addon secret")
+	if err := c.reconcileSecret(ctx, &secret, pinnipedInfoCM); err != nil {
+		log.Error(err, "Error reconciling secret")
```
benjaminapetersen, line 125: Same as above?

addons/pinniped/config-controller/controllers/controllers.go
```diff
@@ -158,17 +146,41 @@ type concierge struct {
...
 
-// nolint:funlen,nolintlint // Eh, we can live with a function of this length
-func (c *PinnipedController) reconcileAddonSecret(ctx context.Context, cluster *clusterapiv1beta1.Cluster, pinnipedInfoCM *corev1.ConfigMap) error {
-	log := c.Log.WithValues(constants.NamespaceLogKey, cluster.Namespace, constants.NameLogKey, cluster.Name)
-	// check if cluster is scheduled for deletion, if so, delete addon secret on mgmt cluster
-	if !cluster.GetDeletionTimestamp().IsZero() {
-		log.V(1).Info("cluster is getting deleted, deleting addon secret")
-		if err := c.reconcileClusterDelete(ctx, client.ObjectKeyFromObject(cluster)); err != nil {
-			return fmt.Errorf("failed to reconcile cluster delete: %w", err)
+// TODO: simplify this function
```
benjaminapetersen, line 149: +1 for simplify & smaller helpers but cool with a todo for later also!

addons/pinniped/config-controller/controllers/controllers.go
```diff
@@ -158,17 +146,41 @@ type concierge struct {
...
+	secretNamespace := secret.Namespace
+	log := c.Log.WithValues(constants.SecretNamespaceLogKey, secretNamespace, constants.SecretNameLogKey, secretName)
+	// check if secret is scheduled for deletion, if so, skip reconcile
+	if !secret.GetDeletionTimestamp().IsZero() {
+		log.V(1).Info("secret is getting deleted, skipping reconcile")
+		return nil
+	}
+	clusterName, err := utils.GetClusterNameFromSecret(secret)
+	if err != nil {
+		log.Error(err, "could not get cluster name from secret")
```
benjaminapetersen, line 162: shall we specify the secret in logs again?

addons/pinniped/config-controller/controllers/controllers.go
sabbey37: the request object (which includes the name/namespace) has been added to the logs above in this line:
```
log := c.Log.WithName("reconcile").WithValues("request object", req)
```

addons/pinniped/config-controller/controllers/controllers.go
sabbey37: the request object (which includes the name/namespace) has been added to the logs above in this line:
```
log := c.Log.WithName("reconcile").WithValues("request object", req)
```

addons/pinniped/config-controller/controllers/controllers.go
```diff
@@ -197,35 +209,39 @@ func (c *PinnipedController) reconcileAddonSecret(ctx context.Context, cluster *
...
+			Namespace: secretNamespace,
+			Name:      secretName,
 		},
 	}
-	// TODO: remove this
-	pinnipedAddonSecret.Type = constants.TKGAddonType
-	pinnipedAddonSecret.Data = make(map[string][]byte)
 
-	pinnipedAddonSecretMutateFn := func() error {
+	pinnipedSecretMutateFn := func() error {
```
benjaminapetersen, line 219: Kind of inclined to extract this inline func out into a separate named func since it is big enough to be "significant".

benjaminapetersen: (also ok to leave as a refactor TODO)

addons/pinniped/config-controller/controllers/controllers.go
sabbey37: This line above adds that information to all the logs:
```
log := c.Log.WithValues(constants.SecretNamespaceLogKey, secretNamespace, constants.SecretNameLogKey, secretName)
```

here is more information about this convention: https://github.com/kubernetes-sigs/controller-runtime/blob/9ee63fc65a9720568f74df9901b16883621636b4/TMP-LOGGING.md

addons/pinniped/config-controller/controllers/controllers.go
```diff
@@ -234,17 +250,17 @@ func (c *PinnipedController) reconcileAddonSecret(ctx context.Context, cluster *
...
-		dataValueYamlBytes = append([]byte(valuesYAMLPrefix), dataValueYamlBytes...)
-		pinnipedAddonSecret.Data[constants.TKGDataValueFieldName] = dataValueYamlBytes
+		pinnipedSecret.Data[constants.TKGDataValueFieldName] = dataValueYamlBytes
 
 		return nil
 	}
 
-	log.V(1).Info("creating or patching addon secret")
-	result, err := controllerutil.CreateOrPatch(ctx, c.client, pinnipedAddonSecret, pinnipedAddonSecretMutateFn)
+	log.V(1).Info("creating or patching secret")
```
benjaminapetersen, line 261: More specificity in this log?

addons/pinniped/config-controller/controllers/controllers.go
sabbey37: yeah, I think I might do the refactor now since I hate the TODO :/

addons/pinniped/config-controller/controllers/controller_test.go
```diff
@@ -42,215 +44,166 @@ var _ = Describe("Controller", func() {
...
 
 		AfterEach(func() {
-			delete(ctx, cluster)
-		})
-
-		When("cluster is created", func() {
-			It("creates a secret with identity_management_type set to none", func() {
-				Eventually(addonSecretFunc(ctx, cluster, nil)).Should(Succeed())
-			})
+			deleteObject(ctx, cluster)
```
ankeesler, line 68: since we did `delete()` -> `deleteObject()`, should we also do `create()` -> `createObject()`?

addons/pinniped/config-controller/controllers/controller_test.go
```diff
@@ -378,47 +389,46 @@ func delete(ctx context.Context, o client.Object) {
 	}).Should(Succeed())
 }
 
-func addonSecretFunc(ctx context.Context, cluster *clusterapiv1beta1.Cluster, configMap *corev1.ConfigMap) func(Gomega) {
+func cbSecretFunc(ctx context.Context, cluster *clusterapiv1beta1.Cluster, configMap *corev1.ConfigMap) func(Gomega) {
 	return func(g Gomega) {
+		clusterCopy := cluster.DeepCopy()
```
ankeesler, line 394: how come we need to make a live `GET` for the cluster again? shouldn't the `cluster` argument already contain the correct information?

addons/pinniped/config-controller/controllers/controller_test.go
```diff
@@ -42,215 +44,166 @@ var _ = Describe("Controller", func() {
...
-			delete(ctx, cluster)
+			It("updates the secret with the proper data values", func() {
+				Eventually(cbSecretFunc(ctx, cluster, nil)).Should(Succeed())
+			})
 		})
 
-		When("the secret gets deleted", func() {
-			var gotSecret *corev1.Secret
-
+		When("random values are added to the secret", func() {
```
ankeesler, line 97: praise: i really like this test

ankeesler: in the future, we should add a test that makes sure that if someone has set an `overlay.yaml` in the `Secret`, that is preserved

ankeesler (approved):
looking really good sarah. a comment here or there, but none should block the merge

addons/pinniped/config-controller/utils/utils.go
```diff
@@ -5,27 +5,50 @@
...
-func HasAddonLabel(secret *corev1.Secret, label string) bool {
-	return secret.Labels[constants.TKGAddonLabel] == label
+// ContainsPackageName returns true if the `tkg.tanzu.vmware.com/package-name` label contains the package name we pass in
+func ContainsPackageName(secret *corev1.Secret, expectedName string) bool {
+	name, labelExists := secret.Labels[constants.PackageNameLabel]
+	return labelExists && strings.Contains(name, expectedName)
+}
+
+func GetClusterNameFromSecret(secret *corev1.Secret) (string, error) {
+	for _, ownerRef := range secret.OwnerReferences {
```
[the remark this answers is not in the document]
sabbey37: Both OwnerRef and the `ClusterNameLabel` should be there since they get added by the [clusterbootstrap controller](https://github.com/vmware-tanzu/tanzu-framework/blob/main/addons/controllers/clusterbootstrap_controller.go#L1096-L1111).  They also utilize both [when mapping secrets to clusters](https://github.com/vmware-tanzu/tanzu-framework/blob/main/addons/controllers/clusterbootstrap_handlers.go#L97-L116).  Looking more at their description, it seems `OwnerRef` would be preferred in our case since the `Type` can be matched, but at this point in the code (we're in the reconcile phase), we have already filtered down to the secrets we care about and are just trying to find the cluster that matches that secret... so I figured why not try to get the cluster name both ways?  But I'm open.

addons/pinniped/config-controller/utils/utils.go
```diff
@@ -59,43 +82,58 @@ func GetInfraProvider(cluster *clusterapiv1beta1.Cluster) (string, error) {
...
+	}
+
+	return pinnipedInfoCM, nil
+}
+
+func GetClusterFromSecret(ctx context.Context, c client.Client, secret *corev1.Secret) (*clusterapiv1beta1.Cluster, error) {
+	secretNamespace := secret.Namespace
+	clusterName, err := GetClusterNameFromSecret(secret)
+	if err != nil {
+		err = fmt.Errorf("could not get cluster name from secret: %w", err)
```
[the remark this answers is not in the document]
sabbey37: 1000%, thank you for seeing this!  I'll fix it.

addons/pinniped/config-controller/controllers/handlers.go
```diff
@@ -70,12 +51,30 @@ func (c *PinnipedController) withAddonLabel(addonLabel string) predicate.Funcs {
...
+	}
+	// Predicate func will get called for all events (create, update, deleteObject, generic)
+	return builder.WithPredicates(
+		predicate.Funcs{
+			CreateFunc: func(e event.CreateEvent) bool { return containsPackageName(e.Object, packageName) },
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				return containsPackageName(e.ObjectOld, packageName) || containsPackageName(e.ObjectNew, packageName)
+			},
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				c.Log.V(1).Info(
```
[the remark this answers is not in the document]
sabbey37: totes!

addons/pinniped/config-controller/controllers/controllers.go
```diff
@@ -76,156 +62,150 @@ func (c *PinnipedController) SetupWithManager(manager ctrl.Manager) error {
...
+		return err
 	}
 
-	log.V(1).Info("retrieved data from pinniped-info configmap", "supervisorAddress", supervisorAddress, "supervisorCABundle", supervisorCABundle)
+	return nil
+}
 
-	pinnipedAddonSecret := &corev1.Secret{
+func (c *PinnipedController) reconcileDataValues(ctx context.Context, secret *corev1.Secret, cluster *clusterapiv1beta1.Cluster, pinnipedInfoCM *corev1.ConfigMap) error {
+	log := c.Log.WithValues(constants.SecretNamespaceLogKey, secret.Namespace, constants.SecretNameLogKey, secret.Name)
```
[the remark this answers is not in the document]
sabbey37: ^I think it actually adds to it, but let me double check.

addons/pinniped/config-controller/controllers/controller_test.go
sabbey37: sure, I had to change `delete` to `deleteObject` because I was using the actual `delete` (for removing a key from a map), but I can change `create` as well.

addons/pinniped/config-controller/controllers/controller_test.go
sabbey37: we do it to get the cluster's `UID`

addons/pinniped/config-controller/controllers/controller_test.go
sabbey37: oooh good thought.  I'll add that now.

addons/pinniped/config-controller/utils/utils.go
```diff
@@ -5,27 +5,50 @@
 package utils
 
 import (
+	"context"
```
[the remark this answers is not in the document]
sabbey37: ooh, I'll fix it though

addons/pinniped/config-controller/go.sum
```diff
@@ -1,11 +1,16 @@
+bazil.org/fuse v0.0.0-20160811212531-371fbbdaa898/go.mod h1:Xbm+BRKSBEpa4q4hTSxohYNQpsxXPbPry4JJWOB3LB8=
```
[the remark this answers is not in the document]
sabbey37: thank you for calling this out, i'll look into it

addons/pinniped/config-controller/controllers/controllers.go
```diff
@@ -76,156 +62,150 @@ func (c *PinnipedController) SetupWithManager(manager ctrl.Manager) error {
...
+			var (
+				issuerExists bool
+				bundleExists bool
+			)
+			supervisorAddress, issuerExists = pinnipedInfoCM.Data[constants.IssuerKey]
+			supervisorCABundle, bundleExists = pinnipedInfoCM.Data[constants.IssuerCABundleKey]
+			//TODO: do we only want to set idmgmttype to oidc if both issuer/cabundle exist?
+			if issuerExists && bundleExists {
+				identityManagementType = constants.OIDC
+			}
```
sabbey37, line 182: Also wanted to know what y'all thought of this (leaving `identity_management_type: none` if neither the `issuer` nor the `caBundle` exist in the pinniped info configmap.

addons/pinniped/config-controller/go.sum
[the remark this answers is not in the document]
benjaminapetersen: `go mod tidy` ensures go.mod and go.sum are correct?

addons/pinniped/config-controller/controllers/controllers.go
benjaminapetersen: ok cool. ill have to run this again and look at it. I'm not super familiar with the parameterized logging yet 👍

addons/pinniped/config-controller/controllers/controllers.go
benjaminapetersen: That seems to make sense.  Also, if one exists but not the other, ought we log/error?  That seems to be a weird state.

addons/pinniped/config-controller/go.sum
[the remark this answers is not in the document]
ankeesler: that is my understanding. from `go help mod tidy`:
```Tidy makes sure go.mod matches the source code in the module.
It adds any missing modules necessary to build the current module's
packages and dependencies, and it removes unused modules that
don't provide any relevant packages. It also adds any missing entries
to go.sum and removes any unnecessary ones.
```

addons/pinniped/config-controller/controllers/controller_test.go
ankeesler: shouldn't a `cluster` contain a `UID` after the `create()` call? i thought `create()` mutated the object that was passed to it for creation

sabbey37: In our `createObject` we make a copy of the object that is passed, so the original object won't be mutated, which makes sense so that we are able easily reuse those vars (cluster, secret, etc.) for create/delete ops.  But since we've changed our code to not watch clusters, I can likely just create the cluster once and let the var get mutated so we don't have to do the live get.  Another option is returning the mutated object in `createObject`.  wdyt?

addons/pinniped/config-controller/controllers/suite_test.go
```diff
@@ -95,7 +95,6 @@ var _ = BeforeSuite(func() {
 
 var _ = AfterSuite(func() {
 	By("tearing down the test environment")
-	// TODO: should we delete namespace here?
```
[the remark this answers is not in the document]
sabbey37: It does not seem like a typical practice (deleting the namespace)... and it hasn't seemed necessary, but looking around controller-runtime docs I haven't found anything specific about it yet.  I can leave the TODO for now.

addons/pinniped/config-controller/controllers/controllers.go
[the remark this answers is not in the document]
sabbey37: oh geez, sorry, I just reread your comment.  Yeah, I debated on this one... (using the `c.Log` and setting those values again vs. passing in the `log` var as an arg to the function (which I did in some places when the func wasn't associated w/the `PinnipedController`).  I didn't feel like we needed the cluster namespace and name at this point since we'd gotten past getting the cluster and errors that might fall out of that, so I only set the secret name/namespace.  I'm open to passing in the log as an argument since it preserves the values we've set though.  Let me know what you think.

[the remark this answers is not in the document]
sabbey37: adding it is no problem, you think it's better to pass the logger in as an arg to these functions?

PR #1781, opened 2022-03-04
Add PackageInstallStatus controller & remote watch

maralavi:
Signed-off-by: Marjan Alavi <malavi@vmware.com>

### What this PR does / why we need it
Adds PackageInstallStatus controller & remote watch

Upon reconciliation, the pkgi conditions are set in the ClusterBootstrapStatus.Conditions, one entry per core/additional packages; where Type and Status are derived by summarization of same fields in pkgi status.conditions for the same pkgi.

### Which issue(s) this PR fixes

Fixes https://github.com/vmware-tanzu/tanzu-framework/issues/1633

### Describe testing done for PR

Added unit tests and tested in tkgm workload cluster

### Release note

```release-note
package-based-lcm: Add PackageInstallStatus controller & remote watch managed packages
```

### PR Checklist

- [X] Squash the commits into one or a small number of logical commits
      
- [X] Use good commit [messages](https://github.com/vmware-tanzu/tanzu-framework/blob/main/CONTRIBUTING.md)
- [X] Ensure PR contains terms all contributors can understand and links all contributors can access

PR #1910, opened 2022-03-24
Logic to handle workload cluster deletion

### LINKED ISSUE #1591: Handle cluster delete for additionalPackages

Cluster deletion should trigger delete of all additionalPackages installed in a cluster so that external resources such as LoadBalancers and PersistentVolumes are deleted. Should not delete Kapp, CNI, CPI and CSI packages on the target cluster.

- [ ] APIs
- [x] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Additional context**

### LINKED ISSUE #1913: Workload cluster delete leaves kapp-controller pkgi
**Bug description**
Deleting a workload cluster leaves pkgi related to workload cluster kapp-controller.

```
k get pkgi workload-kapp-controller -o yaml      
apiVersion: packaging.carvel.dev/v1alpha1
kind: PackageInstall
metadata:
  creationTimestamp: "2022-03-23T20:02:26Z"
  deletionGracePeriodSeconds: 0
  deletionTimestamp: "2022-03-23T22:03:01Z"
  finalizers:
  - finalizers.packageinstall.packaging.carvel.dev/delete
  generation: 2
  name: workload-kapp-controller
  namespace: default
  ownerReferences:
  - apiVersion: cluster.x-k8s.io/v1beta1
    kind: Cluster
    name: workload
    uid: cf5d32a6-cd36-43ea-97ec-a2ba80549eb7
  resourceVersion: "2072344"
  uid: 512fd451-2866-4b8e-b57c-6e8852408d62
spec:
  cluster:
    kubeconfigSecretRef:
      key: value
      name: workload-kubeconfig
  packageRef:
    refName: kapp-controller.tanzu.vmware.com
    versionSelection:
      constraints: 0.30.1+vmware.1-tkg.1-zshippable
      prereleases: {}
  syncPeriod: 10m0s
  values:
  - secretRef:
      name: workload-kapp-controller.tanzu.vmware.com-package-kapp-controller-data-values
status:
  conditions:
  - message: Error (see .status.usefulErrorMessage for details)
    status: "True"
    type: DeleteFailed
  friendlyDescription: 'Delete failed: Error (see .status.usefulErrorMessage for details)'
  lastAttemptedVersion: 0.30.1+vmware.1-tkg.1-zshippable
  observedGeneration: 2
  usefulErrorMessage: 'Preparing kapp: Getting kubeconfig secret: secrets "workload-kubeconfig"
    not found'
  version: 0.30.1+vmware.1-tkg.1-zshippable
```

- [ ] APIs
- [x] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**
workload cluster kapp pkgi should be deleted

1. Delete workload cluster
2. Wait for it to finish
3. Check for pkgi

adduarte:
Logic to handle workload cluster deletion
    
     - adds logic to add/remove finalizers to cluster resources
       based on installpackages
    
     - adds machine controller
    
      - anotates machines based on finalizers
    
     - adds NoopDelete=true to kapp-controller to bypass deletion of remote app
    
     - adds owner refernces to cluster resources to control the flow of garbage collection
    
     - makes cluster delete timeout configurable
    
     - adds machine controller to featureGateClusterBootstrap
    
     - finalizers are removed and added to clusters, clusterbootstrap, and kubeconfigsecrets
    
     - new tests added for machine controller, and clusterboostrap controller.
     
   

    

Fixes #
Possibly Fixes #1591
Possible Fixes #1913

Created/deleted aws workload cluster to verify functionality 

### Release note

```release-note

```

### PR Checklist

### DIFF
--- a/addons/controllers/clusterbootstrap_controller.go
+++ b/addons/controllers/clusterbootstrap_controller.go
@@ -29,6 +29,7 @@ import (
 	clusterapiutil "sigs.k8s.io/cluster-api/util"
 	clusterapipatchutil "sigs.k8s.io/cluster-api/util/patch"
 	clusterApiPredicates "sigs.k8s.io/cluster-api/util/predicates"
+	secretutil "sigs.k8s.io/cluster-api/util/secret"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -172,9 +173,7 @@ func (r *ClusterBootstrapReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	// if deletion timestamp is set, handle cluster deletion
 	if !cluster.GetDeletionTimestamp().IsZero() {
-		// TODO handle delete
-		// https://github.com/vmware-tanzu/tanzu-framework/issues/1591
-		return ctrl.Result{}, nil
+		return r.reconcileDelete(cluster, log)
 	}
 	return r.reconcileNormal(cluster, log)
 }
@@ -199,11 +198,6 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
 		r.Log.Info(fmt.Sprintf("cluster %s/%s does not have status phase %s", cluster.Namespace, cluster.Name, clusterapiv1beta1.ClusterPhaseProvisioned))
 		return ctrl.Result{}, nil
 	}
-	remoteClient, err := util.GetClusterClient(r.context, r.Client, r.Scheme, clusterapiutil.ObjectKey(cluster))
-	if err != nil {
-		log.Error(err, "Error getting remote cluster client")
-		return ctrl.Result{Requeue: true}, err
-	}
 
 	// Create a PackageInstall CR under the cluster namespace for deploying the kapp-controller on the remote cluster.
 	// We need kapp-controller to be deployed prior to CNI, CPI, CSI. This will be a no-op if the cluster object is mgmt
@@ -213,9 +207,32 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
 		return ctrl.Result{}, err
 	}
 
+	remoteClient, err := util.GetClusterClient(r.context, r.Client, r.Scheme, clusterapiutil.ObjectKey(cluster))
+	if err != nil {
+		return ctrl.Result{Requeue: true}, err
+	}
+
 	if err := r.prepareRemoteCluster(cluster, remoteClient); err != nil {
 		return ctrl.Result{}, err
 	}
+
+	_, err = r.createOrPatchResourcesForCorePackages(cluster, clusterBootstrap, remoteClient, log)
+	if err != nil {
+		return ctrl.Result{RequeueAfter: constants.RequeueAfterDuration}, err
+	}
+
+	_, err = r.createOrPatchResourcesForAdditionalPackages(cluster, clusterBootstrap, remoteClient, log)
+	if err != nil {
+		return ctrl.Result{RequeueAfter: constants.RequeueAfterDuration}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *ClusterBootstrapReconciler) createOrPatchResourcesForCorePackages(cluster *clusterapiv1beta1.Cluster,
+	clusterBootstrap *runtanzuv1alpha3.ClusterBootstrap,
+	remoteClient client.Client,
+	log logr.Logger) (ctrl.Result, error) {
 
 	// Create or patch the resources for CNI, CPI, CSI to be running on the remote cluster.
 	// Those resources include Package CR, data value Secret, PackageInstall CR.
@@ -238,15 +255,21 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
 			// packages, we return error and let the reconciler retry again.
 			log.Error(err, fmt.Sprintf("unable to create or patch all the required resources for %s on cluster: %s/%s",
 				corePackage.RefName, cluster.Namespace, cluster.Name))
-			return ctrl.Result{RequeueAfter: constants.RequeueAfterDuration}, err
+			return ctrl.Result{}, err
 		}
 	}
+	return ctrl.Result{}, nil
+}
 
-	// Create or patch the resources for additionalPackages
+func (r *ClusterBootstrapReconciler) createOrPatchResourcesForAdditionalPackages(cluster *clusterapiv1beta1.Cluster,
+	clusterBootstrap *runtanzuv1alpha3.ClusterBootstrap,
+	remoteClient client.Client,
+	log logr.Logger) (ctrl.Result, error) {
+
 	for _, additionalPkg := range clusterBootstrap.Spec.AdditionalPackages {
 		if err := r.createOrPatchAddonResourcesOnRemote(cluster, additionalPkg, remoteClient); err != nil {
 			// Logging has been handled in createOrPatchAddonResourcesOnRemote()
-			return ctrl.Result{Requeue: true}, err
+			return ctrl.Result{}, err
 		}
 		// set watches on provider objects in additional packages if not already set
 		if additionalPkg.ValuesFrom != nil && additionalPkg.ValuesFrom.ProviderRef != nil {
@@ -255,8 +278,53 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
 			}
 		}
 	}
-
+	if len(clusterBootstrap.Spec.AdditionalPackages) > 0 { // If we reach this and there are at least one additional package, we need to add finalizer
+		err := r.addFinalizersToClusterResources(cluster, log)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
 	return ctrl.Result{}, nil
+}
+
+func (r *ClusterBootstrapReconciler) addFinalizersToClusterResources(cluster *clusterapiv1beta1.Cluster, log logr.Logger) error {
+	err := r.addFinalizer(cluster)
+	if err != nil {
+		log.Error(err, "failed to add finalizer to cluster ")
+		return err
+	}
+
+	clusterKubeConfigSecret := &corev1.Secret{}
+	key := client.ObjectKey{Namespace: cluster.Namespace, Name: secretutil.Name(cluster.Name, secretutil.Kubeconfig)}
+	err = r.Client.Get(r.context, key, clusterKubeConfigSecret)
+	if err != nil {
+		return err
+	}
+	err = r.addFinalizer(clusterKubeConfigSecret)
+	if err != nil {
+		log.Error(err, "failed to add finalizer to cluster kubeconfig secret")
+		return err
+	}
+
+	clusterBootstrap := &runtanzuv1alpha3.ClusterBootstrap{}
+	err = r.Client.Get(r.context, client.ObjectKeyFromObject(cluster), clusterBootstrap)
+	if err != nil {
+		return err
+	}
+	err = r.addFinalizer(clusterBootstrap)
+	if err != nil {
+		log.Error(err, "failed to add finalizer to clusterboostrap")
+		return err
+	}
+
+	return nil
+}
+
+func (r *ClusterBootstrapReconciler) addFinalizer(o client.Object) error {
+	controllerutil.AddFinalizer(o, addontypes.AddonFinalizer)
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		return r.Client.Update(r.context, o)
+	})
 }
 
 // createOrPatchClusterBootstrapFromTemplate will get, clone or update a ClusterBootstrap associated with a cluster
@@ -543,6 +611,7 @@ func (r *ClusterBootstrapReconciler) createOrPatchKappPackageInstall(clusterBoot
 		//	 ipkg.ObjectMeta.Annotations = make(map[string]string)
 		// }
 		// ipkg.ObjectMeta.Annotations[addontypes.YttMarkerAnnotation] = ""
+		pkgi.Spec.NoopDelete = true
 		pkgi.Spec.SyncPeriod = &metav1.Duration{Duration: r.Config.PkgiSyncPeriod}
 		pkgi.Spec.PackageRef = &kapppkgiv1alpha1.PackageRef{
 			// clusterBootstrap.Spec.Kapp.RefName is Package.Name. I.e., kapp-controller.tanzu.vmware.com.0.28.0+vmware.1-tkg.1-rc.1
@@ -643,6 +712,12 @@ func (r *ClusterBootstrapReconciler) createOrPatchPackageInstallOnRemote(cluster
 			Name:        util.GeneratePackageInstallName(cluster.Name, remotePackageRefName),
 			Namespace:   r.Config.SystemNamespace,
 			Annotations: map[string]string{addontypes.ClusterNameAnnotation: cluster.Name, addontypes.ClusterNamespaceAnnotation: cluster.Namespace},
+			OwnerReferences: []metav1.OwnerReference{{
+				APIVersion: clusterapiv1beta1.GroupVersion.String(),
+				Kind:       cluster.Kind,
+				Name:       cluster.Name,
+				UID:        cluster.UID,
+			}},
 		},
 	}
 
@@ -1533,5 +1608,157 @@ func (r *ClusterBootstrapReconciler) reconcileClusterProxyAndNetworkSettings(clu
 		return err
 	}
 
+	return nil
+}
+
+func (r *ClusterBootstrapReconciler) makeClusterIsReadyForDeletion(cluster *clusterapiv1beta1.Cluster, log logr.Logger) (bool, error) {
+	clusterBootstrap := &runtanzuv1alpha3.ClusterBootstrap{}
+	err := r.Client.Get(r.context, client.ObjectKeyFromObject(cluster), clusterBootstrap)
+	if err != nil && !apierrors.IsNotFound(err) {
+		log.Error(err, "failed to lookup clusterbootstrap")
+		return false, err
+	}
+	if apierrors.IsNotFound(err) {
+		log.Info("ClusterBootstrap not found")
+	} else {
+		remoteClient, err := util.GetClusterClient(r.context, r.Client, r.Scheme, clusterapiutil.ObjectKey(cluster))
+		if err != nil {
+			return false, err
+		}
+		if hasPackageInstalls(r.context, remoteClient, cluster, r.Config.SystemNamespace, clusterBootstrap.Spec.AdditionalPackages, log) {
+			err = r.removeAdditionalPackageInstalls(remoteClient, cluster, clusterBootstrap, log)
+			if err != nil {
+				return false, err
+			}
+			return false, nil
+		}
+	}
+	log.Info("cluster is ready for deletion")
+	return true, nil
+}
+
+func (r *ClusterBootstrapReconciler) removeFinalizersFromClusterResources(cluster *clusterapiv1beta1.Cluster, log logr.Logger) error {
+	log.Info("removing finalizer from clusterbootstrap if present")
+	clusterBootstrap := &runtanzuv1alpha3.ClusterBootstrap{}
+	err := r.Client.Get(r.context, client.ObjectKeyFromObject(cluster), clusterBootstrap)
+	if err != nil && !apierrors.IsNotFound(err) {
+		log.Error(err, "failed to lookup clusterbootstrap")
+		return err
+	}
+	if !apierrors.IsNotFound(err) {
+		err = r.removeFinalizer(clusterBootstrap)
+		if err != nil {
+			return err
+		}
+	}
+
+	log.Info("removing finalizer from cluster kubeconfig secret if present")
+	clusterKubeConfigSecret := &corev1.Secret{}
+	key := client.ObjectKey{Namespace: cluster.Namespace, Name: secretutil.Name(cluster.Name, secretutil.Kubeconfig)}
+	err = r.Client.Get(r.context, key, clusterKubeConfigSecret)
+	if err == nil {
+		err = r.removeFinalizer(clusterKubeConfigSecret)
+		if err != nil {
+			return err
+		}
+	} else if !apierrors.IsNotFound(err) {
+		log.Error(err, "unable to fetch cluster kubeconfig secret")
+		return err
+	}
+
+	log.Info("removing finalizer from cluster if present")
+	err = r.removeFinalizer(cluster)
+	if err != nil {
+		return err
+	}
+
+	log.Info("all finalizers have been removed from cluster resources")
+	return nil
+}
+
+func (r *ClusterBootstrapReconciler) reconcileDelete(cluster *clusterapiv1beta1.Cluster, log logr.Logger) (ctrl.Result, error) {
+	okToRemoveFinalizers := false
+	var err error
+
+	timeOutReached := time.Now().After(cluster.GetDeletionTimestamp().Add(r.Config.ClusterDeleteTimeout))
+	if timeOutReached {
+		okToRemoveFinalizers = true
+	} else {
+		okToRemoveFinalizers, err = r.makeClusterIsReadyForDeletion(cluster, log)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	if !okToRemoveFinalizers {
+		return ctrl.Result{RequeueAfter: constants.RequeueAfterDuration}, nil
+	}
+
+	if err = r.removeFinalizersFromClusterResources(cluster, log); err != nil {
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *ClusterBootstrapReconciler) removeFinalizer(o client.Object) error {
+	controllerutil.RemoveFinalizer(o, addontypes.AddonFinalizer)
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		return r.Client.Update(r.context, o)
+	})
+}
+
+func hasPackageInstalls(ctx context.Context, remoteClient client.Client,
+	cluster *clusterapiv1beta1.Cluster, namespace string, packages []*runtanzuv1alpha3.ClusterBootstrapPackage,
+	log logr.Logger) bool {
+
+	for _, pkg := range packages {
+		pkgInstallName := util.GeneratePackageInstallName(cluster.Name, pkg.RefName)
+		if packageInstallExists(ctx, pkgInstallName, namespace, remoteClient, log) {
+			return true
+		}
+	}
+	return false
+}
+
+func packageInstallExists(ctx context.Context, pkgInstallName, pkgInstallNamespace string,
+	remoteClient client.Client, log logr.Logger) bool {
+
+	pkgInstall := &kapppkgiv1alpha1.PackageInstall{}
+	if err := remoteClient.Get(ctx, client.ObjectKey{Name: pkgInstallName, Namespace: pkgInstallNamespace}, pkgInstall); err != nil {
+		if apierrors.IsNotFound(err) {
+			return false
+		}
+		log.Error(err, "could not verify if package install is present")
+		return true
+	}
+	return true
+}
+
+func (r *ClusterBootstrapReconciler) removeAdditionalPackageInstalls(remoteClient client.Client, cluster *clusterapiv1beta1.Cluster, clusterBootstrap *runtanzuv1alpha3.ClusterBootstrap, log logr.Logger) error {
+	// Removes all additional package install CRs from the cluster
+
+	err := r.Client.Get(r.context, client.ObjectKeyFromObject(cluster), clusterBootstrap)
+	if err != nil {
+		return err
+	}
+	if clusterBootstrap == nil {
+		return nil
+	}
+	for _, additionalPkg := range clusterBootstrap.Spec.AdditionalPackages {
+		additionalPkgInstall := &kapppkgiv1alpha1.PackageInstall{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      util.GeneratePackageInstallName(cluster.Name, additionalPkg.RefName),
+				Namespace: r.Config.SystemNamespace,
+			},
+		}
+		log.Info(fmt.Sprintf("deleteting package install %s/%s", additionalPkgInstall.Namespace, additionalPkgInstall.Name))
+		err = remoteClient.Delete(r.context, additionalPkgInstall)
+		if err != nil && !apierrors.IsNotFound(err) {
+			log.Error(err, fmt.Sprintf("unable to delete package install for %s/%s",
+				additionalPkgInstall.Namespace, additionalPkg.RefName))
+			return err
+		}
+	}
 	return nil
 }
--- a/addons/controllers/clusterbootstrap_controller_test.go
+++ b/addons/controllers/clusterbootstrap_controller_test.go
@@ -25,6 +25,7 @@ import (
 	vspherecsiv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/csi/v1alpha1"
 	runtanzuv1alpha3 "github.com/vmware-tanzu/tanzu-framework/apis/run/v1alpha3"
 
+	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -34,6 +35,7 @@ import (
 	clusterapiutil "sigs.k8s.io/cluster-api/util"
 	"sigs.k8s.io/cluster-api/util/secret"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 var _ = Describe("ClusterBootstrap Reconciler", func() {
@@ -41,7 +43,6 @@ var _ = Describe("ClusterBootstrap Reconciler", func() {
 		clusterName             string
 		clusterNamespace        string
 		clusterResourceFilePath string
-		enableWebhook           bool
 	)
 
 	// Constants defined in testdata manifests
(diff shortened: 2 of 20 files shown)


addons/controllers/antrea/antreaconfig_controller.go
```diff
@@ -179,11 +179,18 @@ func (r *AntreaConfigReconciler) ReconcileAntreaConfigDataValue(
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      util.GenerateDataValueSecretName(cluster.Name, constants.AntreaAddonName),
 			Namespace: antreaConfig.Namespace,
+			OwnerReferences: []metav1.OwnerReference{{
```
blc1996, line 182: Nice catch

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -167,7 +173,7 @@ func (r *ClusterBootstrapReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	if !cluster.GetDeletionTimestamp().IsZero() {
```
blc1996, line 173: Question: Do we need to add finalizer to the Cluster CR?

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1361,3 +1370,80 @@ func (r *ClusterBootstrapReconciler) reconcileClusterProxyAndNetworkSettings(clu
...
+	clusterBootstrap := &runtanzuv1alpha3.ClusterBootstrap{}
+	err = remoteClient.Get(r.context, client.ObjectKeyFromObject(cluster), clusterBootstrap)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if clusterBootstrap == nil {
+		return ctrl.Result{}, nil
+	}
+
+	err = r.removeAdditionalPackageInstalls(cluster, log)
```
blc1996, line 1386: If the remote cluster is being deleted, meaning all the resources on it will be gone. I assume we probably don't need to clean up the resources in the remote cluster. We just need to clean up all the resources we created in the mgmt cluster.

blc1996: Question resolved: Cluster deletion should trigger delete of all additionalPackages installed in a cluster so that external resources such as LoadBalancers and PersistentVolumes are deleted. Should not delete Kapp, CNI, CPI and CSI packages on the target cluster.

addons/controllers/clusterbootstrap_controller.go
adduarte: since the clusterboostrap CR has a  reference pointing to the cluster, the line of thought is that we just need to add it to the clusterboostrap. Right now, the only thing we use the finalizer is to signal the machine-controller that the machines should be marked with the not delete pretermination hook. so we can get away with just adding it to the clusterboostrap CR. The only reason I can think of adding it to the cluster (at this time) would be if the clusterboostrap is deleted before the cluster and the machines, but I think that would be a "bug" in our logic.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1361,3 +1388,107 @@ func (r *ClusterBootstrapReconciler) reconcileClusterProxyAndNetworkSettings(clu
...
+	if err := remoteClient.Get(ctx, client.ObjectKey{Name: pkgInstallName, Namespace: pkgInstallNamespace}, pkgInstall); err != nil {
+		if apierrors.IsNotFound(err) {
+			return false
+		}
+		log.Error(err, "Could not verify if package install is present")
+		return true
+	}
+	return true
+}
+
```
adduarte, line 1461: perhaps there is something in https://github.com/vmware-tanzu/tanzu-framework/pull/1781 we can use instead of packagesInstallExists and hasPackageInstalls

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1361,3 +1388,107 @@ func (r *ClusterBootstrapReconciler) reconcileClusterProxyAndNetworkSettings(clu
...
+	}
+	for _, additionalPkg := range clusterBootstrap.Spec.AdditionalPackages {
+		additionalPkgInstall := &kapppkgiv1alpha1.PackageInstall{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      util.GeneratePackageInstallName(cluster.Name, additionalPkg.RefName),
+				Namespace: r.Config.SystemNamespace,
+			},
+		}
+		log.Info(fmt.Sprintf("Deleteting package install %s", additionalPkgInstall.Name))
+		err = remoteClient.Delete(r.context, additionalPkgInstall)
```
adduarte, line 1486: again, perhaps we can use something from https://github.com/vmware-tanzu/tanzu-framework/pull/1781 to remove the package install CRs.

adduarte:
Perhaps we should not be merged until after https://github.com/vmware-tanzu/tanzu-framework/pull/1781

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -210,6 +213,24 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
...
+	}
+
+	_, err = r.createOrPatchResourcesForAdditionalPackages(cluster, clusterBootstrap, remoteClient, log)
+	if err != nil {
+		return ctrl.Result{Requeue: true}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
```
adduarte, line 228: createOrPatchResourcesForCorePackages and createOrPatchResourcesForAdditionalPackages had to be introduced to reduce the cyclomatic complexity or go lint complains: 
```clusterbootstrap_controller.go:179:1: cyclomatic complexity 17 of func `(*ClusterBootstrapReconciler).reconcileNormal` is high (> 15) (gocyclo)
func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.Cluster, log logr.Logger) (ctrl.Result, error) {
^

File: addons/controllers/clusterbootstrap_controller.go [hunk 7]
```diff
@@ -210,6 +213,24 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
...
+func (r *ClusterBootstrapReconciler) createOrPatchResourcesForCorePackages(cluster *clusterapiv1beta1.Cluster,
```
adduarte, line 229: added to reduce cyclomatic complexity of reconcileNormal

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -231,27 +252,50 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
...
 			log.Error(err, fmt.Sprintf("unable to create or patch all the required resources for %s on cluster: %s/%s",
 				corePackage.RefName, cluster.Namespace, cluster.Name))
-			return ctrl.Result{RequeueAfter: time.Second * 10}, err
+			return ctrl.Result{}, err
 		}
 	}
+	return ctrl.Result{}, nil
+}
+
+func (r *ClusterBootstrapReconciler) createOrPatchResourcesForAdditionalPackages(cluster *clusterapiv1beta1.Cluster,
```
adduarte, line 261: added to reduce cyclomatic complexity of reconcileNormal

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -210,6 +213,24 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
 		return ctrl.Result{}, err
 	}
 
+	_, err = r.createOrPatchResourcesForCorePackages(cluster, clusterBootstrap, remoteClient, log)
```
blc1996, line 216: Kapp controller is also a core package, could you try include `createOrPatchKappPackageInstall ` in this function as well?

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1361,3 +1412,108 @@ func (r *ClusterBootstrapReconciler) reconcileClusterProxyAndNetworkSettings(clu
 
 	return nil
 }
+
+func (r *ClusterBootstrapReconciler) reconcileDelete(cluster *clusterapiv1beta1.Cluster, log logr.Logger) (ctrl.Result, error) {
+	remoteClient, err := util.GetClusterClient(r.context, r.Client, r.Scheme, clusterapiutil.ObjectKey(cluster))
```
blc1996, line 1417: Suggestion: Can we move this statement up into `Reconcile` and pass remoteClient into `reconcileNormal` and `reconcileDelete` as parameter, so that we wouldn't need to call `util.GetClusterClient` twice in the code?

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1361,3 +1412,108 @@ func (r *ClusterBootstrapReconciler) reconcileClusterProxyAndNetworkSettings(clu
...
+	if err != nil {
+		return ctrl.Result{}, nil
+	}
+	clusterBootstrap := &runtanzuv1alpha3.ClusterBootstrap{}
+	err = remoteClient.Get(r.context, client.ObjectKeyFromObject(cluster), clusterBootstrap)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if clusterBootstrap == nil {
```
blc1996, line 1426: I think this nil check has already been covered by the error check in line 1423.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1361,3 +1412,108 @@ func (r *ClusterBootstrapReconciler) reconcileClusterProxyAndNetworkSettings(clu
...
+		}
+		log.Error(err, "Could not verify if package install is present")
+		return true
+	}
+	return true
+}
+
+func (r *ClusterBootstrapReconciler) removeAdditionalPackageInstalls(cluster *clusterapiv1beta1.Cluster, log logr.Logger) error {
+	// Removes all additional package install CRs from the cluster
+	remoteClient, err := util.GetClusterClient(r.context, r.Client, r.Scheme, clusterapiutil.ObjectKey(cluster))
```
blc1996, line 1489: Same here, remote client can be passed in as parameter.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1361,3 +1412,108 @@ func (r *ClusterBootstrapReconciler) reconcileClusterProxyAndNetworkSettings(clu
...
+	if err != nil {
+		log.Error(err, "error getting remote cluster client")
+		return err
+	}
+	clusterBootstrap := &runtanzuv1alpha3.ClusterBootstrap{}
```
blc1996, line 1494: clusterBootstrap can be passed in as parameter, no need to fetch it again.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1361,3 +1412,108 @@ func (r *ClusterBootstrapReconciler) reconcileClusterProxyAndNetworkSettings(clu
...
+	if clusterBootstrap == nil {
+		return ctrl.Result{}, nil
+	}
+
+	err = r.removeAdditionalPackageInstalls(cluster, log)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	packageInstallsArePresent := hasPackageInstalls(r.context, remoteClient, cluster, r.Config.SystemNamespace, clusterBootstrap.Spec.AdditionalPackages, log)
```
blc1996, line 1435: Is this check meant to be wrapped in a retry loop?

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1361,3 +1412,108 @@ func (r *ClusterBootstrapReconciler) reconcileClusterProxyAndNetworkSettings(clu
...
+	timeOutReached := time.Now().After(cluster.GetDeletionTimestamp().Add(clusterDeleteTimeout))
+
+	if packageInstallsArePresent || timeOutReached {
```
blc1996, line 1438: If `hasPackageInstalls`  is wrapped in a retry look, we probably don't need timeOutReached.

vuil:
Thanks for the changes. Please add a short release note about the change and describe the testing done for it.

addons/controllers/clusterbootstrap_controller.go
adduarte: investigated and did not find anything we could use here

addons/controllers/clusterbootstrap_controller.go
adduarte: done - not applicable

addons/controllers/clusterbootstrap_controller.go
adduarte: Fixed  to RequeueAfter: constants.RequeueAfterDuration in both cases. The functions only return err != nil. Is up to the caller to modifie the reque  request

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -231,27 +252,50 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
...
 			}
 		}
 	}
-
 	return ctrl.Result{}, nil
 }
 
+func (r *ClusterBootstrapReconciler) addFinalizer(o client.Object) error {
+	controllerutil.AddFinalizer(o, types.AddonFinalizer)
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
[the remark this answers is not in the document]
adduarte: The object (in this case a of this pr aruntanzuv1alpha3.ClusterBootstrap) could be being updated by another process. Sometimes while running the envTest, the error: 
cannot be updated: the object has been modified; please apply your changes to the latest version and try again
was seen. 
Will rerun tests to verify that is the case.

addons/controllers/clusterbootstrap_controller.go
adduarte: The requeAfter values are different for kappcontroller than for CorePackages. 
Not sure we can lump them together, but if we can make the same requeAdfter for kappcontroller it can be done. 
I'll submit on its own commit so we can roll out if we don't like it

addons/controllers/clusterbootstrap_controller.go
adduarte: Done

addons/controllers/clusterbootstrap_controller.go
adduarte: agreed. I don't see a way for err == nil and clusterBoostrap == nil

adduarte: removed

addons/controllers/clusterbootstrap_controller.go
adduarte: done

addons/controllers/clusterbootstrap_controller.go
blc1996: It's non-blocking, I guess what I meant was to put `createOrPatchKappPackageInstall` inside `createOrPatchResourcesForCorePackages`, as the scope of latter seems include the first function.

addons/controllers/clusterbootstrap_controller.go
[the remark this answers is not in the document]
blc1996: Thanks for the clarification

addons/controllers/clusterbootstrap_controller.go
blc1996: typo: retry look -> retry loop

addons/controllers/clusterbootstrap_controller.go
adduarte: timeOutReached is the max time we want to spend pausing the deletion of the cluster. 
Once that time is reached we go ahead and continue deleting regardless of weather or not packageInstall crs are present. 
Deleting the packageInstall CRs could take quite a bit of time. We could sit here and continuously retry, but perhaps is best to just reque with a requeAfter,  let the system do something els, and come back again and check the timeout before we make any more calls to the api, since regardless of what the api says, we will delete cluster after timeOutReached = true.

addons/controllers/clusterbootstrap_controller.go
adduarte: Reworked the logic so as to only make calls to api itimeoutReached = false.

addons/controllers/clusterbootstrap_controller.go
adduarte: done

addons/controllers/clusterbootstrap_controller.go
adduarte: yes. but in the original logic if createOrPatchKappPackageInstall fails reconcileNormal returns ctrl.Result{}, err, whilst if the failure is when installing a corepackage (createOrPatchResroucesForCorePackages) then reconcileNormal returns  ctrl.Result{requeueAfter: constants.RequeueAfterDuration}, err 

If we merge the functions, then reconcileNormal will have to return ctrl.Result{requeueAfter: constants.RequeueAfterDuration}, err 
for both. (maybe that is what we want, but probably best done in a different pr then). 
Not sure what the effects might be. I think we best not change it, we would only be moving the code around for neatness, I would rather not, just in case it could introduce a bug.

blc1996: It's non-blocking. Let's defer this one.

addons/controllers/suite_test.go
```diff
@@ -331,6 +341,14 @@ var _ = BeforeSuite(func(done Done) {
 	err = clusterbootstrapTemplateWebhook.SetupWebhookWithManager(mgr)
 	Expect(err).ToNot(HaveOccurred())
 
+	// Setup the tkg-system namespace resources.
+	// We do it here because  specs may be executed in parallel and
+	// this is the only way to assure the tkg-system resources are ready before all specs start
+	f, err := os.Open("testdata/test-tkg-system-ns-resources.yaml")
```
blc1996, line 347: This will void the ClusterBootstrap and ClusterBootstrapTemplate webhook tests. The webhooks are enabled in `clusterbootstrap_controller_test.go` before creating these resources. Now we have moved these resources out, they will be created before the webhooks are in place, so won't be validated and tested.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1464,3 +1539,157 @@ func (r *ClusterBootstrapReconciler) reconcileClusterProxyAndNetworkSettings(clu
...
+
+func (r *ClusterBootstrapReconciler) clusterIsReadyForDeletion(cluster *clusterapiv1beta1.Cluster, log logr.Logger) (bool, error) {
+	clusterBootstrap := &runtanzuv1alpha3.ClusterBootstrap{}
+	err := r.Client.Get(r.context, client.ObjectKeyFromObject(cluster), clusterBootstrap)
+	if err != nil && !apierrors.IsNotFound(err) {
+		log.Error(err, "failed to lookup clusterbootstrap")
+		return false, err
+	}
+	if apierrors.IsNotFound(err) {
+		log.Info("cluster bootstrap not found")
```
blc1996, line 1551: nit: cluster bootstrap -> ClusterBootstrap

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1464,3 +1539,157 @@ func (r *ClusterBootstrapReconciler) reconcileClusterProxyAndNetworkSettings(clu
...
+
+func (r *ClusterBootstrapReconciler) reconcileDelete(cluster *clusterapiv1beta1.Cluster, log logr.Logger) (ctrl.Result, error) {
+	okToRemoveFinalizers := false
+	var err error
+
+	timeOutReached := time.Now().After(cluster.GetDeletionTimestamp().Add(r.Config.ClusterDeleteTimeout))
+	if timeOutReached {
+		okToRemoveFinalizers = true
+	} else {
+		okToRemoveFinalizers, err = r.clusterIsReadyForDeletion(cluster, log)
```
blc1996, line 1618: Nit: Looks like we are checking for deletion status, and doing the actual deletion in this function as well. However the function name suggests it's only checking for deletion status, could use a better name.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1464,3 +1539,157 @@ func (r *ClusterBootstrapReconciler) reconcileClusterProxyAndNetworkSettings(clu
...
+}
+
+func (r *ClusterBootstrapReconciler) removeAdditionalPackageInstalls(remoteClient client.Client, cluster *clusterapiv1beta1.Cluster, clusterBootstrap *runtanzuv1alpha3.ClusterBootstrap, log logr.Logger) error {
+	// Removes all additional package install CRs from the cluster
+
+	err := r.Client.Get(r.context, client.ObjectKeyFromObject(cluster), clusterBootstrap)
+	if err != nil {
+		return err
+	}
+	if clusterBootstrap == nil {
```
blc1996, line 1676: Do we need this check? If `clusterBootstrap == nil`, I guess it will return directly in line 1674.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1464,3 +1539,157 @@ func (r *ClusterBootstrapReconciler) reconcileClusterProxyAndNetworkSettings(clu
...
+		return false, err
+	}
+	if apierrors.IsNotFound(err) {
+		log.Info("cluster bootstrap not found")
+	} else {
+		remoteClient, err := util.GetClusterClient(r.context, r.Client, r.Scheme, clusterapiutil.ObjectKey(cluster))
+		if err != nil {
+			return false, err
+		}
+		if hasPackageInstalls(r.context, remoteClient, cluster, r.Config.SystemNamespace, clusterBootstrap.Spec.AdditionalPackages, log) {
```
blc1996, line 1557: I'm not sure whether we need this `hasPackageInstalls` check before running `removeAdditionalPackageInstalls`. As we will just continue is all the errors are `apierrors.IsNotFound`.

addons/controllers/clusterbootstrap_controller.go
adduarte: done

addons/controllers/clusterbootstrap_controller.go
adduarte: renamed to makeClusterReadyForDeletion

addons/controllers/clusterbootstrap_controller.go
adduarte: hasPackageInstall is wha actually returns 'false' for clusterIsReadyForDeletion (makeClusterReadyForDeletion). 
So we do need to keep the check, and I think we probably want to place the call to removeAdditionaPackageInstalls inside the if statement so we can return a false, nil right away. 
i.e. 
if we find packages we ask to remove them and signal that cluster is not ready for deletion.

addons/controllers/suite_test.go
adduarte: I ran all tests and they pass but if I understand correctly, this might have invalidated the webhook tests regardless of their result. 
We need to figure out a way to setup the tkg-system resources needed by all tests before we call any tests. 
in ginkgo v3 there is a before all tests option, but we don't have that here. 
The reason I had to do it, was because since the machine_controller tests are in a different spec, they sometimes get executed before the tkg-system resources are completely setup and they fail. and we end up with a flaky test. 

Also without this change, running things like "ginkgo -v -f "ClusterBootstrap is paused" will fail because the tkg-system resources in test-cluster-bootstrap-1.yaml are not created.  It might seem not a big deal, but it means that specs can't be safely run in parallel. 

also can't source test-cluster-boostrap-1.yaml on every spec (when statement) because it complains about defining the resources more than once.

addons/controllers/clusterbootstrap_controller_test.go
```diff
@@ -158,6 +160,59 @@ var _ = Describe("ClusterBootstrap Reconciler", func() {
...
+				key := client.ObjectKey{Namespace: cluster.Namespace, Name: secret.Name(clusterName, secret.Kubeconfig)}
+				Eventually(func() bool {
+					err := k8sClient.Get(ctx, key, clusterKubeConfigSecret)
+					if err != nil {
+						return false
+					}
+					return controllerutil.ContainsFinalizer(clusterKubeConfigSecret, addontypes.AddonFinalizer)
+				}, waitTimeout, pollingInterval).Should(BeTrue())
+
+				By("installpackage should have been created for each additional package in the clusterBoostrap")
```
vijaykatam, line 192: nitpick: installpackage -> packageinstall, same comment for line 196

addons/controllers/machine_controller.go
```diff
@@ -0,0 +1,183 @@
...
+	var requests []ctrl.Request
+	for i := range machines.Items {
+		requests = append(requests, ctrl.Request{
+			NamespacedName: types.NamespacedName{
+				Namespace: machines.Items[i].Namespace,
+				Name:      machines.Items[i].Name,
+			},
+		})
+	}
+	r.Log.Info("generated", "requests", requests)
```
vijaykatam, line 180: I would remove this log

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1464,3 +1539,157 @@ func (r *ClusterBootstrapReconciler) reconcileClusterProxyAndNetworkSettings(clu
...
+
+	log.Info("removing finalizer from cluster kubeconfig secret if present")
+	clusterKubeConfigSecret := &corev1.Secret{} // TODO (adduarte): test without deleting clusterKubeConfig
+	key := client.ObjectKey{Namespace: cluster.Namespace, Name: secretutil.Name(cluster.Name, secretutil.Kubeconfig)}
+	err = r.Client.Get(r.context, key, clusterKubeConfigSecret)
+	if err != nil && !apierrors.IsNotFound(err) {
+		log.Error(err, "unable to fetch cluster kubeconfig secret")
+		return err
+	}
+	if !apierrors.IsNotFound(err) {
```
vijaykatam, line 1592: Check if err != nil or move into block above

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1464,3 +1539,157 @@ func (r *ClusterBootstrapReconciler) reconcileClusterProxyAndNetworkSettings(clu
...
+		err = r.removeFinalizer(clusterBootstrap)
+		if err != nil {
+			return err
+		}
+	}
+
+	log.Info("removing finalizer from cluster kubeconfig secret if present")
+	clusterKubeConfigSecret := &corev1.Secret{} // TODO (adduarte): test without deleting clusterKubeConfig
```
vijaykatam, line 1585: I am confused what this TODO means

vijaykatam (commented):
Minor comments, mostly looks good otherwise.

addons/controllers/suite_test.go
blc1996: Had a discussion with Adolfo offline. We are going to enable the CB and CBT webhooks in suite_test.go before the resources are applied.
With this, it is enforced that all the testcases that creates CB and CBT need to have all the resources required, otherwise will be rejected by the webhook.

addons/controllers/clusterbootstrap_controller.go
adduarte: already done. need to remove thanks.  we need to keep clusterKubeConfigSecret around until we are done because we need it. Originally it was not included, and test where failing. the TODO was a reminder to verify this. its done.

addons/main.go
```diff
@@ -105,6 +106,8 @@ func parseAddonFlags(addonFlags *addonFlags) {
 		"The minimum interval at which watched resources are reconciled (e.g. 10m)")
 	flag.DurationVar(&addonFlags.appSyncPeriod, "app-sync-period", 5*time.Minute, "Frequency of app reconciliation (e.g. 5m)")
 	flag.DurationVar(&addonFlags.appWaitTimeout, "app-wait-timeout", 30*time.Second, "Maximum time to wait for app to be ready (e.g. 30s)")
+	flag.DurationVar(&addonFlags.clusterDeleteTimeout, "cluster-delete-timeout", 30*time.Second, "Maximum time to wait for addon resources to be deleted before allowing cluster deletion to proceed")
```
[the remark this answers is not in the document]
adduarte: yes. it is low. Thanks for pointing it out. testing shows this should probably be ~ 5 minute mark.

addons/controllers/machine_controller.go
```diff
@@ -0,0 +1,183 @@
...
+
+type MachineReconciler struct {
+	Client     client.Client
+	Log        logr.Logger
+	Scheme     *runtime.Scheme
+	controller controller.Controller
+	ctx        context.Context
+}
+
+const PreTerminateAddonsAnnotationPrefix = clusterapiv1beta1.PreTerminateDeleteHookAnnotationPrefix + "/addons"
```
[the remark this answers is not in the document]
adduarte: will do. - also will make this a constant just to be cleaner

addons/controllers/machine_controller.go
```diff
@@ -0,0 +1,183 @@
...
+const PreTerminateAddonsAnnotationValue = "true"
```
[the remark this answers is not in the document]
adduarte: will change.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -254,10 +277,55 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
...
+			return ctrl.Result{}, err
+		}
+	}
 	return ctrl.Result{}, nil
 }
 
+func (r *ClusterBootstrapReconciler) addFinalizersToClusterResources(cluster *clusterapiv1beta1.Cluster, log logr.Logger) error {
+	err := r.addFinalizer(cluster)
+	if err != nil {
+		log.Info("failed to add finalizer to cluster ")
```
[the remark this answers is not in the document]
adduarte: ack

addons/controllers/clusterbootstrap_controller_test.go
adduarte: ack

addons/controllers/machine_controller.go
adduarte: ack

addons/controllers/antrea/antreaconfig_controller.go
```diff
@@ -179,11 +179,17 @@ func (r *AntreaConfigReconciler) ReconcileAntreaConfigDataValue(
...
+				Kind:       cluster.Kind,
+				Name:       cluster.Name,
+				UID:        cluster.UID,
+			}},
 		},
+		Type: corev1.SecretTypeOpaque,
 	}
 
 	antreaDataValuesSecretMutateFn := func() error {
-		antreaDataValuesSecret.Type = corev1.SecretTypeOpaque
```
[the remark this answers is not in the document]
adduarte: it should not be.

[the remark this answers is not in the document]
adduarte: done

addons/controllers/antrea/antreaconfig_controller.go
```diff
@@ -179,11 +179,17 @@ func (r *AntreaConfigReconciler) ReconcileAntreaConfigDataValue(
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      util.GenerateDataValueSecretName(cluster.Name, constants.AntreaAddonName),
 			Namespace: antreaConfig.Namespace,
+			OwnerReferences: []metav1.OwnerReference{{
+				APIVersion: clusterapiv1beta1.GroupVersion.String(),
```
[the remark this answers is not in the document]
adduarte: moved owner reference setting to mutating function instead. 
Only if there is no owner references set do we add one.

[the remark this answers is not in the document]
adduarte: done
removed all code for setting owner references as it is not necessary.

addons/controllers/clusterbootstrap_controller.go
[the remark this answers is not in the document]
adduarte: done

addons/controllers/clusterbootstrap_controller.go
adduarte: done

addons/controllers/clusterbootstrap_controller.go
adduarte: done
removed TODO comment

addons/controllers/clusterbootstrap_controller.go
adduarte: will refactor to make it easier to read. 
Basically at this point if we have no errors (err=nil) we remove the finalizer, but if have err=IsNotFound we do nothing, because there is no kubeconfig, which means it was already deleted, somehwere else. 
so we can proceed with removing the rest of the finalizers.

adduarte: done
refactored code as suggested

addons/controllers/clusterbootstrap_controller_test.go
adduarte: done
fixed in both places

addons/controllers/machine_controller.go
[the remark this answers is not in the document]
adduarte: done

addons/controllers/machine_controller.go
```diff
@@ -0,0 +1,183 @@
...
+	}()
+	// Get the cluster to which the current machine belongs
+	cluster, err := util.GetClusterByName(ctx, r.Client, machine.ObjectMeta.Namespace, machine.Spec.ClusterName)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// case when machine is being deleted but cluster is not
+	if !machine.GetDeletionTimestamp().IsZero() && cluster.GetDeletionTimestamp().IsZero() {
+		log.Info(fmt.Sprintf("machine  is being deleted but its parent cluster is not, removing %s if present", PreTerminateAddonsAnnotationPrefix))
```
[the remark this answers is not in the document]
adduarte: done
fixed.

addons/controllers/machine_controller.go
adduarte: done

addons/main.go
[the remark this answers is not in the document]
adduarte: changed to 10 minutes (macthing syncPeriod).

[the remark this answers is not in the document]
adduarte: done

addons/controllers/machine_controller.go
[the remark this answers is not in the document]
adduarte: changed to tkg.tanzu.vmware.com/addons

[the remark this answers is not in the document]
adduarte: done

vijaykatam (approved):
LGTM

PR #1920, opened 2022-03-25
Add TKR nodeselector and handle cluster upgrade pause in ClusterBootstrap controller

### LINKED ISSUE #1850: Package upgrade on cluster version change
This is a follow up to https://github.com/vmware-tanzu/tanzu-framework/pull/1713. Packages managed by ClusterBootstrap must be updated with new versions and configs when status.resolvedTKR is different from TKR label on a cluster. 

1. CLI/other clients will set paused to true when setting a new version in KCP
2. TKR controller will label the cluster with version corresponding to KCP.
3. ClusterBootstrap controller must roll out package upgrades with the following requirements

   a. Ensure packages have node selector with new TKR. (Assume TKR controller has already labeled     nodes with TKR version
   b. Ensure deployment strategy is RollingUpdate
   c. Ensure daemonset strategy is OnDelete. In prev investigation we found kapp-controller gets stuck on daemonset. Investigate and log upstream issue if needed.

blc1996:

Add TKR nodeselector and handle cluster upgrade unpause in ClusterBootstrap controller

Fixes #1850 

Added test cases into env test

### Release note

```release-note
package-based-lcm: Add TKR nodeselector to all packageInstalls and handle cluster upgrade unpause in ClusterBootstrap controller
```

### PR Checklist


addons/controllers/clusterbootstrap_controller.go
```diff
@@ -170,17 +171,22 @@ func (r *ClusterBootstrapReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	log.Info("Reconciling cluster")
 
+	remoteClient, err := util.GetClusterClient(r.context, r.Client, r.Scheme, clusterapiutil.ObjectKey(cluster))
```
blc1996, line 174: The reason for moving this `GetClusterClient` up is because the remoteClient can be shared by reconcileDelete, so we'd better call it once and pass the remoteClient to individual functions as param.

addons/controllers/testdata/dependency/crd/vmoperator.vmware.com_virtualmachines.yaml
```diff
@@ -0,0 +1,535 @@
+
```
blc1996, line 1: `vm-operator-api` doesn't include CRD definition yaml files, so we need to manually import.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -902,6 +927,41 @@ func (r *ClusterBootstrapReconciler) createOrPatchPackageInstallSecretOnRemote(c
...
+					},
+					Daemonset: addontypes.DaemonsetUpdateInfo{
+						UpdateStrategy: constants.TKGSDaemonsetUpdateStrategy,
+					},
+				}
+				TKRDataValueYamlBytes, err := yaml.Marshal(upgradeDataValues)
+				if err != nil {
+					return err
+				}
+				// TODO: Make sure either YTT schema don't care about unspecified field, or all packages have `nodeSelector` field by default in schema.yaml
```
vijaykatam, line 960: I guess we need to check if packages have a schema and if they contain the data values being added here.

addons/pkg/types/addons.go
```diff
@@ -63,3 +63,32 @@ type Image struct {
 	ImagePath string `yaml:"imagePath"`
 	Tag       string `yaml:"tag"`
 }
+
+// UpgradeDataValues contains the package nodeSelector and update strategy information
+type UpgradeDataValues struct {
```
vijaykatam, line 68: This is also used for install. I would call it TKGSDataValues.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -902,6 +927,41 @@ func (r *ClusterBootstrapReconciler) createOrPatchPackageInstallSecretOnRemote(c
...
+		if err != nil {
+			return err
+		}
+		if infraRef == tkgconstants.InfrastructureProviderVSphere {
+			ok, err := util.IsTKGSCluster(r.context, r.Client, cluster)
+			if err != nil {
+				return err
+			}
+			if ok {
+				upgradeDataValues := addontypes.UpgradeDataValues{
```
vijaykatam, line 941: I would rename the var and struct to TKGSDataValues since these are used for initial install as well.

addons/controllers/suite_test.go
```diff
@@ -121,7 +122,7 @@ var _ = BeforeSuite(func(done Done) {
 	Expect(externalCRDPaths).ToNot(BeEmpty())
 	testEnv.CRDDirectoryPaths = externalCRDPaths
 	testEnv.CRDDirectoryPaths = append(testEnv.CRDDirectoryPaths,
-		filepath.Join("..", "..", "config", "crd", "bases"), filepath.Join("testdata"))
+		filepath.Join("..", "..", "config", "crd", "bases"), filepath.Join("testdata"), filepath.Join("testdata", "dependency", "crd"))
```
vijaykatam, line 125: Should we get the crd from the [repo itself](https://github.com/vmware-tanzu/vm-operator/tree/main/config/crd/bases) like we are doing for other external crds for consistency?

vijaykatam (commented):
Minor comments. I think @shyaamsn wants a temporary feature flag to gate in addition to checking TKGS cluster so that we can merge this week.

addons/controllers/suite_test.go
blc1996: Including vm-operator will introduce many dependency complications in go.mod. This is the reason they introduced `vm-operator-api`, but unfortunately, it doesn't include CRD files.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -256,9 +257,32 @@ func (r *ClusterBootstrapReconciler) reconcileNormal(cluster *clusterapiv1beta1.
...
+		return ctrl.Result{Requeue: true}, err
+	}
+
 	return ctrl.Result{}, nil
 }
 
+// handleClusterUnpause unpauses the cluster if the cluster pause annotation is set by cluster pause webhook (cluster has "tkg.tanzu.vmware.com/paused" annotation)
+func (r *ClusterBootstrapReconciler) handleClusterUnpause(cluster *clusterapiv1beta1.Cluster, clusterBootstrap *runtanzuv1alpha3.ClusterBootstrap, log logr.Logger) error {
+	if cluster.Spec.Paused && cluster.Annotations != nil {
+		if value, ok := cluster.Annotations[tkgconstants.ClusterPauseLabel]; ok && value == clusterBootstrap.Status.ResolvedTKR {
```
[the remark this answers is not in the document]
blc1996: If successive TKR's are rolled out, the value of `cluster.Annotations[tkgconstants.ClusterPauseLabel]` will also be updated by the clusterPause webhook to use the new TKR version, so I think it won't get stuck. The cluster will be unpaused when the new version is reconciled and marked as resolvedTKR.

addons/controllers/clusterbootstrap_controller.go
blc1996: Since we are adding an additional feature flag for it, removing this comment

addons/pkg/types/addons.go
blc1996: renamed

addons/controllers/clusterbootstrap_controller.go
blc1996: renamed

packages/management/addons-manager-v2/bundle/config/upstream/webhook-manifests.lib.yaml
```diff
@@ -18,7 +18,7 @@ apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
   labels:
-    tkg.tanzu.vmware.com/addon-webhooks: ""
+    tkg.tanzu.vmware.com/addon-webhooks: "addon-webhooks"
```
vijaykatam, line 21: I don't think you need a value, unless we use the value to different different label values with the same key

blc1996: If you noticed in main.go, we are calling
```
webhooks.InstallNewCertificates(ctx, mgr.GetConfig(), certPath, keyPath, constants.WebhookScrtName, flags.addonNamespace, constants.WebhookServiceName, constants.AddonWebhookLabelKey+"="+constants.AddonWebhookLabelValue)
```
and `constants.AddonWebhookLabelValue` has been changed to `addon-webhooks` in yesterday's Adolfo's cluster deletion PR.

addons/pkg/util/cluster_util.go
```diff
@@ -163,3 +165,42 @@ func GetClusterLabel(clusterLabels map[string]string, labelKey string) string {
...
+// IsTKGSCluster checks if the cluster is a TKGS cluster
+func IsTKGSCluster(ctx context.Context, c client.Client, cluster *clusterv1beta1.Cluster) (bool, error) {
+	// Verify if operating on a TKGS cluster by checking if virtualmachine objects exist for the cluster label
+	virtualMachineList := &vmoperatorv1alpha1.VirtualMachineList{}
+	listOptions := client.MatchingLabels{
+		tkgconstants.CAPVClusterSelectorKey: cluster.Name,
+	}
+
+	if err := c.List(ctx, virtualMachineList, listOptions); err != nil {
+		// If CRD resource doesn't exist on cluster, it will throw an unKnownError with the error message that contains `no matches for kind "VirtualMachine"`
```
vijaykatam, line 199: I probably didn't catch this earlier but this is brittle. I would use the capability client to check if resource is available, see https://github.com/vmware-tanzu/tanzu-framework/blob/main/cmd/cli/plugin/package/main.go#L72 for usage in package client.

addons/controllers/suite_test.go
vijaykatam: Ah, more trouble than actually worth it. Could you add a code comment with this last bit of explanation that vm-operator-api causes depedency conflicts.

packages/management/addons-manager-v2/bundle/config/upstream/webhook-manifests.lib.yaml
vijaykatam: Yeah, I didn't realize that. This is fine

addons/pkg/util/cluster_util.go
blc1996: I added some custom logic using discovery client. 

The reason why I didn't use capability library is because we are using `memCache` discovery client. It will simply return a plain text error `not found`, if Group doesn't exist in the cluster. https://github.com/kubernetes/client-go/blob/master/discovery/cached/memory/memcache.go#L98

While capability library library is trying to catch `apierrors.IsNotFound` https://github.com/vmware-tanzu/tanzu-framework/blob/main/pkg/v1/sdk/capabilities/discovery/cluster_gvr.go#L221-L223.  Since the error we got is not an IsNotFound api error, it will be returned as an error. However, we want the error to be nil if the resource is not found.

blc1996: A possible workaround to use capability library is by adding `strings.Contains(err.Error(), "not found")` into https://github.com/vmware-tanzu/tanzu-framework/blob/main/pkg/v1/sdk/capabilities/discovery/cluster_gvr.go#L221-L223. But I think we should be fine with adding some custom logic instead of using capability library.

vijaykatam (approved):
LGTM.

addons/controllers/testdata/dependency/crd/vmoperator.vmware.com_virtualmachines.yaml
Roangift: What do you need

PR #1938, opened 2022-03-28
Clusterbootstrap controller needs to handle inline data value from ClusterBootstrapPackage

### LINKED ISSUE #1694: Clusterbootstrap controller needs to handle inline data value from ClusterBootstrapPackage

There are three types of data value could be provided in ClusterBootstrapPackage: SecretRef, ProviderRef, Inline. Current implementation covers only SecretRef and ProviderRef. We also need to handle the inline case.

- [ ] APIs
- [x] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Additional context**

shivaani0505:
Signed-off-by: Shivaani Gupta <gshivaani@vmware.com>

This change handles packages with inline config. It changes the reconcile logic in createOrPatchPackageInstallSecretOnRemote by creating a new data value secret with the given inline config that is synced to the remote cluster. First time creation and upgrades are both handled as stated above. 

Fixes #1694 

- Added test case for a package with inline config

### Release note

```release-note

```

### PR Checklist


apis/run/v1alpha3/clusterbootstraptemplate_types.go
```diff
@@ -50,16 +52,46 @@ type ClusterBootstrapPackage struct {
...
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Inline map[string]interface{} `json:"inline,omitempty"`
 	// +optional
 	SecretRef string `json:"secretRef,omitempty"`
 	// +optional
 	ProviderRef *corev1.TypedLocalObjectReference `json:"providerRef,omitempty"`
 }
 
-//+kubebuilder:object:root=true
+// DeepCopyInto is a deepcopy function, copying the receiver, writing into out. in must be non-nil.
```
danniel1205, line 69: question: trying to understand why the deep copy func gets generated here instead of `zz_generated.deepcopy.go` ?

danniel1205:
question: do we want to handle the inline valueFrom at https://github.com/vmware-tanzu/tanzu-framework/blob/main/addons/controllers/clusterbootstrap_controller.go#L1254 ?

apis/run/v1alpha3/clusterbootstraptemplate_types.go
shivaani0505: These are custom definitions added similar to OSImage type
(Refer to comment by @vijaykatam => https://github.com/vmware-tanzu/tanzu-framework/issues/1694#issuecomment-1059422094)

shivaani0505:
> https://github.com/vmware-tanzu/tanzu-framework/blob/main/addons/controllers/clusterbootstrap_controller.go#L1254

After discussing with Vijay, we decided to go with Option 1 from the below two options: 
**Option 1:** 
We modify the reconcile logic in createOrPatchPackageInstallSecretOnRemote to handle inline config directly. For the packages that have inline ref, we will create a new data value secret that will contain the config provided inline. First time creation and upgrades are handled the same way - as stated above, the reconciler will simply create a new data value secret for the workload cluster in createOrPatchPackageInstallSecretOnRemote

**Option 2:** 
Instead of handling inline config directly in reconciliation, we can create a secret in the management cluster that will contain the inline config (i.e. the map) provided for the package. The life-cycle of this secret will be similar to the cloned secrets that we do for packages with SecretRef in BootstrapController 
For the first time, Reconciler will create data value secret for workload cluster corresponding to this BootstrapController-managed secret we created on the management cluster and pass it to the workload cluster
For upgrades, Reconciler will update the data value secret corresponding to the secret we create for inline config


So, we are not going to create a secret in the management cluster for packages with inline config. Looking at the code, I see one place where it might be assumed that a secret must be always present on the management cluster, which is for Kapp package install ( https://github.com/shivaani0505/tanzu-framework/blob/57fae012638a07f6db5ffdf85d1a6cc999342dff/addons/controllers/clusterbootstrap_controller.go#L559). We will need to discuss this.

PR #1942, opened 2022-03-29
adding vspherecsi api and controller

### LINKED ISSUE #1579: vSphereCSIConfig API and Controller

Create Core Addons Config API CRD - Create vSphereCSI CRD and controller to create data values secret

peterochodo:
- so that addon manager can watch and process CSIConfig objects --> which are necessary for configuring vsphere-csi and vsphere-pv-csi addons

- created a Cluster object, created a CSIConfig object with the same name as Cluster, verified that expected secret was created with the expected data
- deleted Cluster object and verified that CSIConfig object and corresponding secret were also deleted
- created CSIConfig object without corresponding Cluster object and verified that it was ignored

Fixes #1579 

### Fix
- https://github.com/vmware-tanzu/tanzu-framework/issues/1579

### Release note
- enables configuration of vsphere-pv-csi and vsphere-csi addons via CSIConfig api


addons/controllers/csi/vspherecsiconfig_controller.go
```diff
@@ -0,0 +1,173 @@
...
+			return ctrl.Result{}, nil
+		}
+
+		logger.Error(err, "Unable to fetch VSphereCSIConfig resource")
+		return ctrl.Result{}, err
+	}
+
+	cluster, err := r.getOwnerCluster(ctx, vcsiConfig)
+	if cluster == nil {
+		return ctrl.Result{RequeueAfter: 20 * time.Second}, err // retry until corresponding cluster is found
```
danniel1205, line 60: suggestion: will it be a good idea to make the 20 seconds as a constant so that other `xxxConfig` controller could reuse ?

addons/controllers/csi/vspherecsiconfig_controller.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	return result, nil
+}
+
+func (r *VSphereCSIConfigReconciler) reconcileVSphereCSIConfigNormal(ctx context.Context,
+	csiCfg *csiv1alpha1.VSphereCSIConfig,
+	cluster *clusterapiv1beta1.Cluster) (ctrl.Result, error) {
+
+	logger := log.FromContext(ctx)
```
danniel1205, line 114: suggestion: would it a good idea to pass in the logger defined at https://github.com/vmware-tanzu/tanzu-framework/pull/1942/files#diff-3410610ce615475fb7620f1c681cff9349758e47a817f2c178706e9507648413R79 ?

addons/controllers/csi/vspherecsiconfig_controller.go
```diff
@@ -0,0 +1,173 @@
...
+
+	ownerRef := metav1.OwnerReference{
+		APIVersion: clusterapiv1beta1.GroupVersion.String(),
+		Kind:       cluster.Kind,
+		Name:       cluster.Name,
+		UID:        cluster.UID,
+	}
+
+	if !clusterapiutil.HasOwnerRef(csiCfg.OwnerReferences, ownerRef) {
+		csiCfg.OwnerReferences = clusterapiutil.EnsureOwnerRef(csiCfg.OwnerReferences, ownerRef)
```
danniel1205, line 124: question: looks like we are mutating csiConfig pointer directly here. Do we want to use `client.Patch()` instead of the directly modification?

addons/controllers/csi/vspherecsiconfig_controller.go
```diff
@@ -0,0 +1,173 @@
...
+
+	opResult, err := controllerutil.CreateOrPatch(ctx, r.Client, secret, mutateFn)
+
+	if err != nil {
+		logger.Error(err, "Error creating or patching VSphereCSIConfig data values secret")
+		return ctrl.Result{}, err
+	}
+
+	logger.Info(fmt.Sprintf("'%s' the secret '%s'", opResult,
+		fmt.Sprintf("%s/%s", secret.Namespace, secret.Name)))
```
danniel1205, line 168: suggestion: I guess we could simplify above logging message construction by using one `fmt.Sprintf()` ? And seems no need of having single quote ?

addons/controllers/csi/vspherecsiconfig_controller.go
```diff
@@ -0,0 +1,173 @@
...
+		For(&csiv1alpha1.VSphereCSIConfig{}).
+		WithOptions(options).
+		Complete(r)
+}
+
+func (r *VSphereCSIConfigReconciler) reconcileVSphereCSIConfig(ctx context.Context,
+	csiCfg *csiv1alpha1.VSphereCSIConfig,
+	cluster *clusterapiv1beta1.Cluster) (result ctrl.Result, retErr error) {
+
+	logger := log.FromContext(ctx)
```
danniel1205, line 79: question: would it be a good idea to reuse `logger := log.FromContext(ctx).WithValues("VSphereCSIConfig", req.NamespacedName)` from L45 ?

danniel1205, line 79: question: would it be a good idea to reuse `logger := log.FromContext(ctx).WithValues("VSphereCSIConfig", req.NamespacedName)` from L45 ?

addons/controllers/csi/vspherecsiconfig_controller.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	defer func() {
+		if retErr != nil {
+			// don't modify VSphereCSIConfig if there is an error
+			return
+		}
+
+		if err := patchHelper.Patch(ctx, csiCfg); err != nil {
+			logger.Error(err, "Failed to patch VSphereCSIConfig")
```
danniel1205, line 93: nit: let's make the logging message consistent with `r.Log.Error(err, "Error patching CalicoConfig")` ?

danniel1205, line 93: nit: let's make the logging message consistent with `r.Log.Error(err, "Error patching CalicoConfig")` ?

addons/controllers/csi/vspherecsiconfig_controller.go
```diff
@@ -0,0 +1,173 @@
...
+		csiCfg.OwnerReferences = clusterapiutil.EnsureOwnerRef(csiCfg.OwnerReferences, ownerRef)
+	}
+
+	addonName := ""
+	switch csiCfg.Spec.VSphereCSI.Mode {
+	case VSphereCSINonParavirtualMode:
+		addonName = constants.CSIAddonName
+	case VSphereCSIParavirtualMode:
+		addonName = constants.PVCSIAddonName
+	}
```
danniel1205, line 133: question: do we have a validation against `csiCfg.Spec.VSphereCSI.Mode` ? Trying to understand what addonName we want to assign if it is a no-match against above two cases.

danniel1205, line 133: question: do we have a validation against `csiCfg.Spec.VSphereCSI.Mode` ? Trying to understand what addonName we want to assign if it is a no-match against above two cases.

addons/controllers/csi/vspherecsiconfig_controller.go
```diff
@@ -0,0 +1,173 @@
...
+
+	if err != nil {
+		logger.Error(err, "Error creating or patching VSphereCSIConfig data values secret")
+		return ctrl.Result{}, err
+	}
+
+	logger.Info(fmt.Sprintf("'%s' the secret '%s'", opResult,
+		fmt.Sprintf("%s/%s", secret.Namespace, secret.Name)))
+
+	csiCfg.Status.SecretRef = &secret.Name
```
danniel1205, line 170: suggestion: it is risky to mutate the csiCfg pointer directly. I assume you want to do `client.Update()` or `client.Patch` on csiCfg ?

danniel1205, line 170: suggestion: it is risky to mutate the csiCfg pointer directly. I assume you want to do `client.Update()` or `client.Patch` on csiCfg ?

addons/controllers/csi/vspherecsiconfig_datavalues.go
```diff
@@ -0,0 +1,42 @@
...
+
+package controllers
+
+// DataValues is the data values type generated by the VSphereCSIConfig CR
+type DataValues struct {
+	VSphereCSI   *DataValuesVSphereCSI   `yaml:"vsphereCSI,omitempty"`
+	VSpherePVCSI *DataValuesVSpherePVCSI `yaml:"vspherePVCSI,omitempty"`
+}
+
+type DataValuesVSpherePVCSI struct {
```
danniel1205, line 12: question: do we want `omitempty` in this struct ?

addons/controllers/csi/vspherecsiconfig_datavalues.go
```diff
@@ -0,0 +1,42 @@
...
+	ClusterName                      string `yaml:"clusterName"`
+	ClusterUID                       string `yaml:"clusterUID"`
+	Namespace                        string `yaml:"namespace"`
+	SupervisorMasterEndpointHostname string `yaml:"supervisorMasterEndpointHostname"`
+	SupervisorMasterPort             int32  `yaml:"supervisorMasterPort"`
+}
+
+type DataValuesVSphereCSI struct {
```
danniel1205, line 20: question: do we want `omitempty` in this struct ?

danniel1205, line 20: question: do we want `omitempty` in this struct ?

addons/controllers/csi/vspherecsiconfig_utils.go
```diff
@@ -0,0 +1,259 @@
...
+	config := vcsiConfig.Spec.VSphereCSI.ParavirtualConfig
+
+	if config == nil {
+		return nil, errors.Errorf("ParavirtualConfig missing in Spec.VSphereCSI")
+	}
+
+	// first populate derived values
+	dvs.VSpherePVCSI = &DataValuesVSpherePVCSI{}
+
+	dvs.VSpherePVCSI.Namespace = tryParseClusterVariableString(ctx, cluster, NamespaceVarName)
```
danniel1205, line 53: question: could we use `res, err := util.ParseClusterVariableString(cluster, variableName)` directly ? Trying to educate myself on what benefits there are to have a wrapper `tryParseClusterVariableString()`

addons/controllers/csi/vspherecsiconfig_utils.go
```diff
@@ -0,0 +1,259 @@
...
+	dvs.VSpherePVCSI.ClusterName = tryParseClusterVariableString(ctx, cluster, ClusterNameVarName)
+
+	dvs.VSpherePVCSI.ClusterUID = string(cluster.UID) // only place to get UID
+
+	// default values from https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/vsphere-pv-csi/2.4.1/bundle/config/values.yaml
+	dvs.VSpherePVCSI.SupervisorMasterEndpointHostname = "supervisor.default.svc"
```
danniel1205, line 59: question: fetch from `ParavirtualConfig.SupervisorMasterEndpointHostname` ?

addons/controllers/csi/vspherecsiconfig_utils.go
```diff
@@ -0,0 +1,259 @@
...
+	dvs.VSpherePVCSI.SupervisorMasterPort = 6443
```
danniel1205, line 60: question: fetch from `PravirtualConfig.SupervisorMasterPort` ?

danniel1205, line 60: question: trying to understand why we want to set the default here ? Shouldn't we always read from VSphereCSIConfig and assign whatever value it has ?

addons/controllers/csi/vspherecsiconfig_utils.go
```diff
@@ -0,0 +1,259 @@
...
+	dvs.VSphereCSI.Server = tryParseClusterVariableString(ctx, cluster, VSphereServerVarName)
+	dvs.VSphereCSI.Datacenter = tryParseClusterVariableString(ctx, cluster, VSphereDatacenterVarName)
+	dvs.VSphereCSI.PublicNetwork = tryParseClusterVariableString(ctx, cluster, VSphereNetworkVarName) //TODO: IS 'VSPHERE_NETWORK' same as 'public network'?
+	dvs.VSphereCSI.Username = tryParseClusterVariableString(ctx, cluster, VSphereUsernameVarName)
+	dvs.VSphereCSI.Password = tryParseClusterVariableString(ctx, cluster, VSpherePasswordVarName)
+	dvs.VSphereCSI.Region = tryParseClusterVariableString(ctx, cluster, VSphereRegionVarName)
+	dvs.VSphereCSI.Zone = tryParseClusterVariableString(ctx, cluster, VSphereZoneVarName)
+	dvs.VSphereCSI.VSphereVersion = tryParseClusterVariableString(ctx, cluster, VSphereVersionVarName)
+	dvs.VSphereCSI.WindowsSupport = tryParseClusterVariableBool(ctx, cluster, IsWindowsWorkloadClusterVarName)
+	dvs.VSphereCSI.TLSThumbprint = tryParseClusterVariableString(ctx, cluster, VSphereTLSThumbprintVarName)
```
danniel1205, line 110: question: the same question on the wrapper func in above. Would it be a good idea to use `res, err := util.ParseClusterVariableString(cluster, variableName)` directly ?

addons/controllers/csi/vspherecsiconfig_utils.go
```diff
@@ -0,0 +1,259 @@
...
+
+	if cluster.Annotations != nil {
+		dvs.VSphereCSI.HttpProxy = cluster.Annotations[pkgtypes.HTTPProxyConfigAnnotation]
+		dvs.VSphereCSI.HttpsProxy = cluster.Annotations[pkgtypes.HTTPSProxyConfigAnnotation]
+		dvs.VSphereCSI.NoProxy = cluster.Annotations[pkgtypes.NoProxyConfigAnnotation]
```
danniel1205, line 115: question: do we want to check if `cluster.Annotations[pkgtypes.HTTPProxyConfigAnnotation]` exist ? The case I could think of is when `cluster.Annotations != nil`, but there are no proxy related annotations, would it safe to set the proxy for vSphereCSI ?

addons/controllers/csi/vspherecsiconfig_utils.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	// populated from default value in https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/vsphere-csi/2.4.1/bundle/config/values.yaml
+	dvs.VSphereCSI.UseTopologyCategories = false
+	dvs.VSphereCSI.ProvisionTimeout = "300s"
+	dvs.VSphereCSI.AttachTimeout = "300s"
+	dvs.VSphereCSI.ResizerTimeout = "300s"
+	dvs.VSphereCSI.DeploymentReplicas = 3
+	dvs.VSphereCSI.InsecureFlag = true
```
danniel1205, line 124: question:  should we always assign whatever value vsphereCSIConfig ? IIUC, the defaulting thing should be done within `vsphereCSIConfig` (by a webhook who sets all the defaults if not provided)

addons/controllers/csi/vspherecsiconfig_utils.go
```diff
@@ -0,0 +1,259 @@
...
+	return dvs, nil
+}
+
+// getOwnerCluster verifies that the VSphereCSIConfig has a cluster as its owner reference,
+// and returns the cluster. It tries to read the cluster name from the VSphereCSIConfig's owner reference objects.
+// If not there, we assume the owner cluster and VSphereCSIConfig always has the same name.
+func (r *VSphereCSIConfigReconciler) getOwnerCluster(ctx context.Context,
+	vcsiConfig *csiv1alpha1.VSphereCSIConfig) (*clusterapiv1beta1.Cluster, error) {
+
+	logger := log.FromContext(ctx)
```
danniel1205, line 200: suggestion: reuse the logger ?

addons/controllers/csi/vspherecsiconfig_utils.go
```diff
@@ -0,0 +1,259 @@
...
+	cluster := &clusterapiv1beta1.Cluster{}
+	clusterName := vcsiConfig.Name
```
danniel1205, line 202: nit: could you help add some comments on why vcsiConfig.Name is the clusterName ?

config/rbac/vspherecsiconfig_editor_role.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - csi.tanzu.vmware.com
+  resources:
+  - vspherecsiconfigs/status
+  verbs:
+  - get
```
danniel1205, line 24: question: we update the status as well, right ?

packages/management/addons-manager-v2/vendir.yml
```diff
@@ -48,6 +48,13 @@ directories:
         newRootPath: config/crd/bases/cpi.tanzu.vmware.com_vspherecpiconfigs.yaml
         includePaths:
           - config/crd/bases/cpi.tanzu.vmware.com_vspherecpiconfigs.yaml
+      - path: csi.tanzu.vmware.com_vspherecsiconfigs.yaml
+        git:
+          url: git@github.com:vmware-tanzu/tanzu-framework.git
+          ref: ffffffffffffffffffffffffffffffffffffffff  #to be set after yaml is committed to main branch
```
danniel1205, line 54: suggestion: could we file an GH issue as a TODO to track ?

addons/controllers/csi/vspherecsiconfig_constants.go
```diff
@@ -0,0 +1,25 @@
...
+package controllers
+
+// supported modes for VSphereCSIConfig
+const (
+	VSphereCSINonParavirtualMode = "vsphereCSI"
+	VSphereCSIParavirtualMode    = "vsphereParavirtualCSI"
+)
+
+// cluster variable names required by VSphereCSIConfig to derive values
+const (
```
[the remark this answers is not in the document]
peterochodo: updated

addons/controllers/csi/vspherecsiconfig_controller.go
peterochodo: updated

addons/controllers/csi/vspherecsiconfig_controller.go
peterochodo: logging reworked

addons/controllers/csi/vspherecsiconfig_controller.go
peterochodo: done

addons/controllers/csi/vspherecsiconfig_controller.go
peterochodo: logging reworked

peterochodo: logging re-worked

addons/controllers/csi/vspherecsiconfig_controller.go
peterochodo: done

peterochodo: done

addons/controllers/csi/vspherecsiconfig_controller.go
peterochodo: yes. Kubebuilder should autogenerate validation code for us. See api definition:

```
type VSphereCSI struct {
	// The vSphere mode. Either `vsphereCSI` or `vsphereParavirtualCSI`.
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:Enum=vsphereCSI;vsphereParavirtualCSI
	Mode string `json:"mode"`

	*NonParavirtualConfig `json:"config,omitempty"`
}
```

addons/controllers/csi/vspherecsiconfig_controller.go
peterochodo: yes it's ok. If reconciliation is successful, a patch function is invoked to update the csiconfig object in kubeapiserver just before the reconciliation iteration comes to an end.

addons/controllers/csi/vspherecsiconfig_datavalues.go
peterochodo: yes. values.yaml should only have 'vsphereCSI' or 'vspherePVCSI'. Whichever is defined.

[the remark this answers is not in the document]
peterochodo: yes. values.yaml should only have 'vsphereCSI' or 'vspherePVCSI'. Whichever is defined.

addons/controllers/csi/vspherecsiconfig_datavalues.go
peterochodo: yes. values.yaml should only have 'vsphereCSI' or 'vspherePVCSI'. Whichever is defined.

peterochodo: no. All configuration values are resolved by the controller and populated accordingly in values.yaml

addons/controllers/csi/vspherecsiconfig_utils.go
peterochodo: this is better because we don't have to test for 'err' multiple times then set an empty string over and over again.

addons/controllers/csi/vspherecsiconfig_utils.go
peterochodo: - the VSphereCSIConfig can be configured to have a few configurations. The controller is required to populate defaults. So no, we don't totally rely on VSphereCSIConfig object.
- true. Using a defaulting webhook is best practice.

addons/controllers/csi/vspherecsiconfig_utils.go
peterochodo: this is better because we don't have to test for 'err' multiple times then set an empty string over and over again.

addons/controllers/csi/vspherecsiconfig_utils.go
peterochodo: Not necessary. Golang will return an empty string from the map if the given key does not exist in it.

addons/controllers/csi/vspherecsiconfig_utils.go
peterochodo: comment added

config/rbac/vspherecsiconfig_editor_role.yaml
peterochodo: - These permissions are meant for the user (who creates a vspherecsiconfig object), not the controller code. Status sub-resources should not be modified by a user, just the controller code.

```
Status subresource
The /status subresource is used to split the user-provided specification of a CR instance from the controller-provided status. The main motivation for this is privilege separation:

The user usually should not write status fields.

The controller should not write specification fields.
```

peterochodo:
replaced by https://github.com/vmware-tanzu/tanzu-framework/pull/2046

addons/controllers/csi/vspherecsiconfig_utils.go
peterochodo: - these fields have been removed from vspherecsiconfig api

PR #1985, opened 2022-04-02
Add signal library, simpler context handling in plugins, and orderly shutdown of Tanzu CLI

### LINKED ISSUE #201: Allow cli plugins to gracefully exit on interrupt signals

The tanzu cli currently exits immediately when it receives an interrupt or sigterm signal. When invoking a plugin, this has the effect of immediately killing the plugin process. Commands should have a way to gracefully exit before being fully killed.

Especially command that initiate network requests should be allowed to finish and cleanup state before being killed. Otherwise, users could be left in an ambiguous state.

When execution is passed to a plugin, either all signals should be passed to the plugin; or, the first signal should be passed to the plugin, and a second signal should result in the command being terminated.

For controllers, I've used [SetupSignalHandler](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/manager/signals#SetupSignalHandler) which returns a context that is canceled after the first signal and exits on the second signal. The context's done channel can be watched to start the graceful termination on the remove process.

There are certainly other valid approaches depending the the desired higher level semantics.

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] Installation
- [x] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Additional context**

codegold79:

This PR provides a signal handling library within Tanzu Framework that plugin developers may and are encouraged to use (but not required to) to capture SIGINT and SIGTERM system signals. It updates [Plugin Developer documentation](https://github.com/vmware-tanzu/tanzu-framework/blob/main/docs/cli/plugin_implementation_guide.md) explaining what the library does. This PR provides an example of how to use the library by updating the [featuregate plugin](https://github.com/vmware-tanzu/tanzu-framework/tree/main/cmd/cli/plugin/feature).

This PR also provides an easier way to pass context to all the Cobra commands a plugin executes. Context facilitates graceful shutdown of plugins by providing information as to when a program should begin to exit.

Tanzu CLI will use the new signals library to update the context that was passed while executing plugins (child processes). When a SIGTERM or SIGINT is detected, Tanzu CLI will wait a predetermined amount of time to give plugins a chance to gracefully exit. The CLI will shut down when all plugins exit or when the time runs out, whichever happens sooner.

There are a few reasons why this PR is beneficial:

- By providing a signals library, documentation, and an example, it will be easier for plugin developers to implement graceful shutdown of plugins when SIGTERM or SIGINT are sent. Currently, there is no consistency or guidance in signal handling and may lead to unnecessarily confusing user experience.
- By allowing for context to be sent during plugin execution, it will be easier for plugin developers to use a single context for all the commands in a plugin. Currently, separate contexts must be created for each command in a plugin.
- Tanzu CLI will now wait for plugins to shutdown when a SIGINT or SIGTERM is sent instead of shutting down immediately (default and current behavior). This behavior is more orderly and allows for final logging (and potentially other actions in the future) before plugins exit.

Fixes #201

Created unit tests to ensure signals library intercepts system SIGINT and SIGTERM signals and responds in the expected amount of time.

Created a very simple plugin in my [personal repository](https://github.com/codegold79/tanzu-playground/tree/main/plugin-play) to demonstrate the signal handling and graceful shutdown works as intended when installed and run via Tanzu CLI.

### Release note

```release-note
Provide signals library for signal handling as well as the ability for each plugin to pass context to all its commonds upon execution. Tanzu CLI now waits for all plugins to exit or a timeout (whichever happens first) when SIGTERM or SIGINT is captured instead of attempting to shutdown immediately.
```

### PR Checklist

### DIFF
--- a/cmd/cli/plugin/feature/main.go
+++ b/cmd/cli/plugin/feature/main.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"context"
+	"os"
 	"time"
 
 	"github.com/aunum/log"
@@ -38,11 +39,11 @@ func main() {
 
 	ctx := signals.SetupSignalHandler()
 	ctx, cancel := context.WithTimeout(ctx, contextTimeout)
-	defer cancel()
 	p.AddContext(ctx)
 
 	if err := p.Execute(); err != nil {
-		log.Errorf("executing plugin: %w", err)
-		return
+		cancel()
+		os.Exit(1)
 	}
+	cancel()
 }
--- a/cmd/cli/plugin/feature/main.go
+++ b/cmd/cli/plugin/feature/main.go
@@ -4,7 +4,7 @@
 package main
 
 import (
-	"os"
+	"context"
 	"time"
 
 	"github.com/aunum/log"
@@ -12,6 +12,7 @@ import (
 	cliv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/cli/v1alpha1"
 	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/buildinfo"
 	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli/command/plugin"
+	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli/signals"
 )
 
 var descriptor = cliv1alpha1.PluginDescriptor{
@@ -35,7 +36,13 @@ func main() {
 		FeatureDeactivateCmd,
 	)
 
+	ctx := signals.SetupSignalHandler()
+	ctx, cancel := context.WithTimeout(ctx, contextTimeout)
+	defer cancel()
+	p.AddContext(ctx)
+
 	if err := p.Execute(); err != nil {
-		os.Exit(1)
+		log.Errorf("executing plugin: %w", err)
+		return
 	}
 }
--- a/pkg/v1/cli/signals/signals.go
+++ b/pkg/v1/cli/signals/signals.go
@@ -1,0 +1,57 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package signals intercepts system signals and returns a context.
+package signals
+
+import (
+	"context"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/aunum/log"
+)
+
+// SetupSignalHandler will return a context that will be canceled if the system
+// sends either a SIGINT or SIGTERM signal.
+func SetupSignalHandler() context.Context {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		sig, ok := <-sigCh
+		if !ok {
+			return
+		}
+
+		log.Infof("%s: shutting down", sig)
+		cancel()
+	}()
+
+	return ctx
+}
+
+// SetupSignalHandlerWithTimeout will return a context that will be canceled
+// after a specified amount of time after the system sends either a SIGINT or
+// SIGTERM signal.
+func SetupSignalHandlerWithTimeout(timeout time.Duration) context.Context {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		sig, ok := <-sigCh
+		if !ok {
+			return
+		}
+
+		log.Infof("%s signal: waiting %v before forcing shut down", sig, timeout)
+		<-time.After(timeout)
+		cancel()
+	}()
+
+	return ctx
+}
--- a/pkg/v1/cli/signals/signals_test.go
+++ b/pkg/v1/cli/signals/signals_test.go
@@ -31,32 +31,34 @@ func TestSetupSignalHandler(t *testing.T) {
 	}
 
 	for _, tc := range tests {
-		ctx := SetupSignalHandler()
+		t.Run(tc.desc, func(t *testing.T) {
+			ctx := SetupSignalHandler()
 
-		startTime := time.Now()
-		var wg sync.WaitGroup
+			startTime := time.Now()
+			var wg sync.WaitGroup
 
-		wg.Add(1)
-		go func() {
-			defer wg.Done()
-			for {
-				select {
-				case <-ctx.Done():
-					return
-				default:
-					if time.Since(startTime) > tc.tolerance {
-						t.Errorf("%s: want: immediate interrupt signal, got: no signal after %v", tc.desc, tc.tolerance)
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				for {
+					select {
+					case <-ctx.Done():
 						return
+					default:
+						if time.Since(startTime) > tc.tolerance {
+							t.Errorf("want: immediate interrupt signal, got: no signal after %v", tc.tolerance)
+							return
+						}
 					}
 				}
-			}
-		}()
+			}()
 
-		// Send system signal.
-		if err := syscall.Kill(syscall.Getpid(), tc.signal); err != nil {
-			t.Errorf("want: no error, got: %v", err)
-		}
-		wg.Wait()
+			// Send system signal.
+			if err := syscall.Kill(syscall.Getpid(), tc.signal); err != nil {
+				t.Errorf("want: no error, got: %v", err)
+			}
+			wg.Wait()
+		})
 	}
 }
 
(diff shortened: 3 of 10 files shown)


cmd/cli/plugin/feature/main.go
```diff
@@ -35,7 +36,13 @@ func main() {
...
 	)
 
+	ctx := signals.SetupSignalHandler()
+	ctx, cancel := context.WithTimeout(ctx, contextTimeout)
+	defer cancel()
+	p.AddContext(ctx)
+
 	if err := p.Execute(); err != nil {
-		os.Exit(1)
+		log.Errorf("executing plugin: %w", err)
```
rajathagasthya, line 45: Does this add another line to CLI output? We might not want that depending on how errors are printed out inside `p.Execute()`.

pkg/v1/cli/signals/signals.go
```diff
@@ -0,0 +1,57 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
```
rajathagasthya, line 1: This package looks like generally useful outside of the CLI. Should this be in `pkg/v1` instead of `pkg/v1/cli`?

rajathagasthya, line 1: This package looks like generally useful outside of the CLI. Should this be in `pkg/v1` instead of `pkg/v1/cli`?

pkg/v1/cli/signals/signals_test.go
```diff
@@ -0,0 +1,115 @@
...
+import (
+	"sync"
+	"syscall"
+	"testing"
+	"time"
+)
+
+// TestSetupSignalHandler ensures that when a signal is sent, context channel
+// done is sent right away.
+func TestSetupSignalHandler(t *testing.T) {
```
rajathagasthya, line 15: Great set of tests!

rajathagasthya, line 15: Great set of tests!

pkg/v1/cli/signals/signals_test.go
```diff
@@ -0,0 +1,115 @@
...
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for {
+				select {
+				case <-ctx.Done():
+					return
+				default:
+					if time.Since(startTime) > tc.tolerance {
+						t.Errorf("%s: want: immediate interrupt signal, got: no signal after %v", tc.desc, tc.tolerance)
```
rajathagasthya, line 48: Just use "subtests" with `t.Run` and you don't need to prefix `tc.desc` everywhere. See https://go.dev/blog/subtests or [one of the examples](https://github.com/vmware-tanzu/tanzu-framework/blob/e3dbe729ace6d139ede1118916e3024d0d20747c/pkg/v1/sdk/features/client/client_test.go#L53).

rajathagasthya, line 48: Just use "subtests" with `t.Run` and you don't need to prefix `tc.desc` everywhere. See https://go.dev/blog/subtests or [one of the examples](https://github.com/vmware-tanzu/tanzu-framework/blob/e3dbe729ace6d139ede1118916e3024d0d20747c/pkg/v1/sdk/features/client/client_test.go#L53).

rajathagasthya (commented):
Code LGTM! Just a few minor questions/suggestions.

cmd/cli/plugin/feature/main.go
```diff
@@ -35,7 +36,13 @@ func main() {
 		FeatureDeactivateCmd,
 	)
 
+	ctx := signals.SetupSignalHandler()
```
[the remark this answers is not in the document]
codegold79: I'm not sure what sorts of plugins are being built, but certainly, some don't need context at all, or if a context is being passed, it's just a background context that doesn't do anything.

For example, a plugin can operate so quickly that there isn't even enough time for an interrupt or terminate signal to be sent before the program exits.

Even, as you say, "It's highly likely most plugins will be using some sort of context," should we make the default based on our opinion of what signal handler to use? If a plugin developer doesn't like our signals library (and, say, for example want to use the Kubernetes controller-runtime signals library) or doesn't even want to pass around a functioning context, should we make it so they have to update and change what we provided as default?

Based on your feedback in that if you think a default is the right way to go, even if it may have to be removed or changed depending on the plugin developer, I'll go ahead and add the line you highlighted, `ctx := signals.SetupHandler()` to the builder template.

cmd/cli/plugin/feature/main.go
codegold79: > Does this add another line to CLI output? 
Yes, if there is an error that occurs while executing a command, the error and its wrapped messages will be logged here.

> We might not want that depending on how errors are printed out inside p.Execute()

Unless I missed it, I don't see anywhere that errors are printed/logged in [p.Execute ](https://github.com/vmware-tanzu/tanzu-framework/blob/e0d0bde9c480aa8fd8bf09c48071a4994e94816b/pkg/v1/cli/command/plugin/plugin.go#L53)or the cobra Execute() method it calls.

I think without my log message, if an error happens during p.Execute(), no one would know why the plugin exited. We should log the error message here in `main.go`. Are you  seeing something different?

pkg/v1/cli/signals/signals.go
codegold79: Do you think there will be a need for signal handling outside of Tanzu CLI or its plugins? I couldn't think of a need, but maybe you can. If so, where would you recommend I put this package?

pkg/v1/cli/signals/signals_test.go
codegold79: Good call--I'll get those tests updated with subtests.

cmd/cli/plugin/feature/main.go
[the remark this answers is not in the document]
rajathagasthya: Let's keep it as is for now and evaluate as we see how plugins use context. But a couple of thoughts:

1. I can't think of a plugin that doesn't need to use context. Most, if not all, reach out to k8s API. A background context isn't good practice. So it's okay to mandate graceful shutdown.
2. We can be opinionated on an out-of-the-box signal handler that satisfies most use cases. If authors want to change it, they can as they have full control once the plugin has been scaffolded.

cmd/cli/plugin/feature/main.go
rajathagasthya: I can see plugins print error messages. I wonder where/how that's happening though?

pkg/v1/cli/signals/signals.go
rajathagasthya: I can envision needing signal handling if we build a "construction kit" for controllers, but that's far into the future, so let's just keep it here.

cmd/cli/plugin/feature/main.go
codegold79: Any chance you remember what the error message was about or if I can reproduce it to see it?

rajathagasthya: Here's one:
```
$ tanzu feature list
Error: couldn't get featureGateRunner: current server "" not found in tanzu config
Usage:
  tanzu feature list [flags]

Examples:

	# List a clusters Features
	tanzu feature list --activated
	tanzu feature list --unavailable
	tanzu feature list --deactivated

Flags:
  -a, --activated            List only activated Features
  -d, --deactivated          List only deactivated Features
  -e, --extended             Include extended output. Higher latency as it requires more API calls.
  -f, --featuregate string   List Features gated by a particular FeatureGate (default "tkg-system")
  -h, --help                 help for list
  -o, --output string        Output format (yaml|json|table)
  -u, --unavailable          List only Features specified in the gate but missing from cluster

Error: exit status 1

✖  exit status 1
```

codegold79: Thanks--I see what you mean, now. I'll look into it.

codegold79: I see where the error is printed; it's in [Cobra's code](https://github.com/spf13/cobra/blob/5271cf43427f43a0815f460bd80ab86e5aebc708/command.go#L968).

And I realize it's probably good to have `exit status 1`, which I had gotten rid of. I put it back with the latest commit. I purposely made that a separate commit so you can see the change clearly, @rajathagasthya. Don't worry--I'll be sure to squash before merging.

Thanks for your careful attention to this PR, @rajathagasthya. I think it was a bad decision on my part to log errors in main in this case and not exit with an error code.

rajathagasthya: > I purposely made that a separate commit so you can see the change clearly

Thanks for not force pushing!

> I'll be sure to squash before merging

And again for excellent git hygiene :)

rajathagasthya (approved):
Very good code, docs and tests. Thanks!

vuil:
Apologies for the delay, I plan to take a look by EOW. 
Question in the meantime: @codegold79 were you able to verify if the library works as expected on windows (say on gitbash?)

codegold79:
Question by @vuil:
> were you able to verify if the library works as expected on windows (say on gitbash?)

No, I'm afraid not. I'll have to find someone with a Windows environment who can test it.

PR #2083, opened 2022-04-08
Clusterbootstrap controller needs to handle inline data value from ClusterBootstrapPackage

### LINKED ISSUE #1694: Clusterbootstrap controller needs to handle inline data value from ClusterBootstrapPackage

shivaani0505:

Clusterbootstrap controller needs to handle inline data value from ClusterBootstrapPackage

Fixes #1694 

### Release note

```release-note
package-based-lcm: Clusterbootstrap controller needs to handle inline data value from ClusterBootstrapPackage
```

### PR Checklist


shivaani0505:
Implementation based on:
**Option 2:** 
Instead of handling inline config directly in reconciliation, we can create a secret in the management cluster that will contain the inline config (i.e. the map) provided for the package. The life-cycle of this secret will be similar to the cloned secrets that we do for packages with SecretRef in BootstrapController 
For the first time, Reconciler will create data value secret for workload cluster corresponding to this BootstrapController-managed secret we created on the management cluster and pass it to the workload cluster
For upgrades, Reconciler will update the data value secret corresponding to the secret we create for inline config

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1342,14 +1396,31 @@ func (r *ClusterBootstrapReconciler) watchProvider(providerRef *corev1.TypedLoca
...
+func (r *ClusterBootstrapReconciler) GetDataValueSecretNameFromBootstrapPackage(cbPkg *runtanzuv1alpha3.ClusterBootstrapPackage, cluster *clusterapiv1beta1.Cluster) (string, error) {
 	// When valuesFrom is nil, we interpret it as no data values are needed for the package installation.
 	if cbPkg.ValuesFrom == nil {
 		r.Log.Info(fmt.Sprintf("no data values are provided to the ClusterBootstrapPackage.ValuesFrom field. ClusterBootstrapPackage.RefName: %s", cbPkg.RefName))
 		return "", nil
 	}
 
-	// TODO: Handle inline valueFrom. https://github.com/vmware-tanzu/tanzu-framework/issues/1694
+	if cbPkg.ValuesFrom.Inline != nil {
+		packageRefName, _, err := util.GetPackageMetadata(r.context, r.aggregatedAPIResourcesClient, cbPkg.RefName, cluster.Namespace)
```
danniel1205, line 1407: suggestion(non-blocking): since we are using the live client `aggregatedAPIResourcesClient`, that means each call will talk to apiserver directly. And there are 6 places in clusterbootstrap_controller invokes it. We might want to file a GH issue to have it improved: reuse the result of util.GetPackageMetadata() as much as we could.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1087,6 +1097,50 @@ func (r *ClusterBootstrapReconciler) periodicGVRCachesClean() {
...
+			{
+				APIVersion: clusterapiv1beta1.GroupVersion.String(),
+				Kind:       cluster.Kind,
+				Name:       cluster.Name,
+				UID:        cluster.UID,
+			},
+		}
+
+		inlineSecret.Data = map[string][]byte{}
+		inlineConfigYamlBytes, err := yaml.Marshal(pkg.ValuesFrom.Inline)
```
danniel1205, line 1122: question: Since we are changing the pkg.ValuesFrom.Inline to be a map[string]interface{}, wondering how does it handle the YTT headers like:
```
#@data/values
#@overlay/match-child-defaults missing_ok=True
```
If we need to consider the YTT header case, do we want to prepend it to `inlineConfigYamlBytes` ?

addons/controllers/testdata/test-cluster-bootstrap-2.yaml
```diff
@@ -132,6 +132,48 @@ spec:
 ---
 apiVersion: data.packaging.carvel.dev/v1alpha1
 kind: Package
+metadata:
+  name: foobar2.example.com.1.18.2
+  namespace: cluster-namespace-2
```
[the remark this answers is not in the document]
shivaani0505: The tests were breaking , so I put this in

addons/controllers/clusterbootstrap_controller.go
shivaani0505: this can be done in follow up PR

danniel1205 (approved):
lgtm

PR #2092, opened 2022-04-11
Private registry for Windows node

hxietkg:
Certificate file for private registry will be added to containerd's config
file: '/etc/containerd/config.toml', Windows nodes have a different path
from Linux nodes, it is 'C:\Program Files\containerd\config.toml', so the
current implemention doesn't work for Windows containerd.

Another problem is that containerd 1.6.0(and versions < 1.6.0) is built
with go-lang 1.16/1.17, but go-lang 1.16i/1.17 doesn't support system cert
pool on Windows, it was fixed from 1.18, see:
  https://github.com/golang/go/commit/3544082f75fd3d2df7af237ed9aef3ddd499ab9c
so we will hit the error with containerd 1.6.0 and old versions:
  "failed to get system cert pool: crypto/x509: system root pool is not
   available on Windows"
Main branch of containerd is using go-lang 1.18, it works.

1. Current implementation add the cert to the wrong config.toml on Windows node, this PR will correct it.
2. Current implementation does not support insecure registry, just add it.

Fixes #

build containerd.exe from main branch of github.com/containerd/containerd, then run following tests:
1. test the secure registry with cert.
2. test insecure registry

### Release note

```release-note

```

### PR Checklist

### DIFF
--- a/pkg/v1/providers/tests/unit/windows_test.go
+++ b/pkg/v1/providers/tests/unit/windows_test.go
@@ -31,6 +31,8 @@ var _ = Describe("Windows Ytt Templating", func() {
 			filepath.Join(yamlRoot, "ytt", "03_customizations", "02_avi", "ako-deployment.lib.yaml"),
 			//filepath.Join(YAML_ROOT, "provider-bundle", "providers", "ytt", "02_addons", "cpi", "cpi_addon_data.lib.yaml"),
 			filepath.Join(yamlRoot, "ytt", "03_customizations", "03_windows"),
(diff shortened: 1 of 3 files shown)


knabben:
LGTM

I would way for the 1.18 on containerd in the TKR before bringing this change though, adding this before can cause more confusion and new issues.

knabben:
LGTM

I would way for the 1.18 on containerd in the TKR before bringing this change though, adding this before can cause more confusion and new issues.

randomvariable:
Unfortunately needs to be redone for ClusterClass. One for your radar @knabben

PR #2164, opened 2022-04-18
Remove namespace in CalicoConfig and change the output format of vethMTU to string

12345lcr:
* In converted data values yaml, `vethMTU` should be string instead of int64
* Remove `namespace` in CalicoConfig because it is an invalid field in calico packages which never took effect before and about to be deprecated

Fixes #

### Release note

```release-note
package-based-lcm: Update CalicoConfig `vethMTU` to string and remove `namespace`
```

### PR Checklist

Seems the `make test` in addon is broken.

### DIFF
--- a/addons/controllers/calico/calicoconfig_utils.go
+++ b/addons/controllers/calico/calicoconfig_utils.go
@@ -6,6 +6,7 @@ package controllers
 import (
 	"fmt"
 	"net"
+	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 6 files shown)


vijaykatam:
Looks good. Added release note. Lets follow up with a vendir update to sync the CRD changes into the package in https://github.com/vmware-tanzu/tanzu-framework/blob/main/packages/management/addons-manager/vendir.yml

vijaykatam (approved):

PR #2179, opened 2022-04-19
Update cc vars to organize/simplify data values

### LINKED ISSUE #2336: Update clusterclass variable defintions

Update the variables defined in our clusterclass definitions to logically group related properties and fields. Effort should be made, where possible, to homogenize variables across providers to align the definitions with one another.

- [ ] APIs
- [ ] Addons
- [x] CLI
- [x] Docs
- [ ] IAM
- [x] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [x] User Experience

**Additional context**

tenczar:
This change introduces new variable structures to simplify the variables
that are applied to clusterclass based clusters.

When moving to clusterclass based clusters, our initial implementation directly transcribed old TKG cluster properties to clusterclass variables one-to-one. This maintain the old flat key/value mapping relationship that existed in that old properties file. This change groups these variables, renames them, organizes them for clarity and simplicity. Several broad categories exist for properties that are shared across providers, these include audit logging, proxy,  and custom image repo. Each provider now also has a controlPlane and worker variable that contain provider specific customizations for each provider.

Fixes #2336 

Went through the normalized clustergen diffs between the old legacy cluster config and the new clusterclass based config. Care was taken to minimize as many diffs as possible to maintain functionality between old and new. Some diff are unavoidable. These largely come in the naming/annotating of machine deployments and machine health check values.

### Release note

```release-note

```

### PR Checklist

### DIFF
--- a/pkg/v1/providers/infrastructure-aws/v1.2.0/yttcc/overlay.yaml
+++ b/pkg/v1/providers/infrastructure-aws/v1.2.0/yttcc/overlay.yaml
@@ -2,7 +2,7 @@
 #@ load("@ytt:data", "data")
 #@ load("lib/helpers.star", "get_bom_data_for_tkr_name", "kubeadm_image_repo", "get_az_from_region", "get_default_tkg_bom_data")
 #@ load("lib/validate.star", "validate_configuration")
-#@ load("/lib/config_variable_association.star", "config_variable_association", "get_cluster_variables")
+#@ load("/lib/config_variable_association.star", "config_variable_association", "get_cluster_variables", "get_aws_vars")
 
 #@ validate_configuration("aws")
 
@@ -60,7 +60,7 @@ spec:
       #@overlay/match by=overlay.index(0)
       - class: tkg-worker
         name: md-0
-        replicas: #@ data.values.WORKER_MACHINE_COUNT
+        replicas: #@ data.values.WORKER_MACHINE_COUNT_0
         #@overlay/match missing_ok=True
         failureDomain: #@ default_az_0
         metadata:
@@ -76,8 +76,9 @@ spec:
         #@ if data.values.NODE_MACHINE_TYPE_1:
         variables:
           overrides:
-          - name: NODE_MACHINE_TYPE
-            value: #@ data.values.NODE_MACHINE_TYPE_1
+          - name: worker
+            value:
+              instanceType: #@ data.values.NODE_MACHINE_TYPE_1
         #@ end
       #@overlay/append
       - class: tkg-worker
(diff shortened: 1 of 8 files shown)


saji-pivotal (approved):
LGTM. But I'd definitely want another person to review this as well

vuil:
@tenczar please review https://storage.googleapis.com/tkg-clustergen/2179/20220510200221/clustergen_cc.diff.txt
and ack if the diffs looks good.

tenczar:
I have checked the diffs and confirmed that the diffs are expected. This includes updating the azure-json secret to properly use the cluster name, fixing the sshAuthorizedKey, and removing updating the pause image path when not necessary.

pkg/v1/providers/infrastructure-aws/v1.2.0/yttcc/overlay.yaml
```diff
@@ -87,15 +88,16 @@ spec:
...
     variables:
-    #@ vars = get_cluster_variables()
+    #@ vars = get_aws_vars()
     #@ for configVariable in vars:
-    #@  if vars[configVariable] != None and configVariable in ["AWS_REGION", "AWS_SSH_KEY_NAME", "BASTION_HOST_ENABLED", "AWS_VPC_CIDR", "AWS_VPC_ID", "AWS_NODE_AZ", "AWS_PRIVATE_NODE_CIDR", "AWS_PRIVATE_SUBNET_ID", "AWS_PUBLIC_NODE_CIDR", "AWS_PUBLIC_SUBNET_ID", "CONTROL_PLANE_MACHINE_TYPE", "NODE_MACHINE_TYPE", "CLUSTER_NAME", "AWS_CONTROL_PLANE_OS_DISK_SIZE_GIB", "AWS_NODE_OS_DISK_SIZE_GIB", "AWS_SECURITY_GROUP_BASTION", "AWS_SECURITY_GROUP_CONTROLPLANE", "AWS_SECURITY_GROUP_APISERVER_LB", "AWS_SECURITY_GROUP_NODE", "AWS_SECURITY_GROUP_LB", "AWS_LOAD_BALANCER_SCHEME_INTERNAL", "AWS_IDENTITY_REF_KIND", "AWS_IDENTITY_REF_NAME", "CLUSTER_PLAN", "proxy", "TKG_PROXY_CA_CERT", "TKG_CUSTOM_IMAGE_REPOSITORY", "TKG_CUSTOM_IMAGE_REPOSITORY_SKIP_TLS_VERIFY", "TKG_CUSTOM_IMAGE_REPOSITORY_CA_CERTIFICATE", "ENABLE_AUDIT_LOGGING"]:
+    #@  if vars[configVariable] != None and configVariable in ["region", "sshKeyName", "bastion", "network", "controlPlane", "worker", "loadBalancerSchemeInternal", "identityRef", "imageRepository", "trust", "auditLogging"]:
```
vuil, line 100: question: if the purpose of get_aws_vars is to construct the appropriate set of variables specific to aws, do we still need this check? (same for other infras as well)

tenczar:
Are there specific examples of things you are still seeing. The major one that is new in this change is additional proxy configuration that was added to our legacy configs by Lubron about a month ago. This covers all of those `which` commands, and the extra cert file that was added. The `ipv6: false` change is just reverting to what the legacy config had. Everything else from your original comment should be addressed. If you can point me to specific instances where this isn't the case that would be helpful.

pkg/v1/providers/infrastructure-aws/v1.2.0/yttcc/overlay.yaml
tenczar: It is impractical to remove this at the moment. The optional_params in our tests will create test files that mix and match aws and vsphere properties. When those get passed through as clusterclass variables there are errors when a variable is added which doesn't have a corresponding variable definition in the clusterclass. The current implementation allows us to filter these variables. I'm very open to setting up a new issue to address problems with our clustergen tests generation.

vuil:
> Are there specific examples of things you are still seeing. The major one that is new in this change is additional proxy configuration that was added to our legacy configs by Lubron about a month ago. This covers all of those which commands, and the extra cert file that was added. The ipv6: false change is just reverting to what the legacy config had. Everything else from your original comment should be addressed. If you can point me to specific instances where this isn't the case that would be helpful

Thanks for the explanation on the ipv6, proxy and the cert diffs. There were another set of diffs (second to last hunk from bottom, just above the `which` stuff). If that is also intended then we are good.

vuil (approved):
Lgtm. Thanks for the updates!

Issue #2193, opened 2022-04-20
vsphere-cpi package has schema inconsistencies and API mismatch


vijaykatam:
**Bug description**
Schema in https://github.com/vmware-tanzu/community-edition/tree/main/addons/packages/vsphere-cpi/1.23.0 has inconsistent 
usage of boolean and string. They do not match the API type defined in https://github.com/vmware-tanzu/tanzu-framework/blob/main/apis/cpi/v1alpha1/vspherecpiconfig_types.go

Package install fails, example
```
    \    |\n  26 |         insecureFlag: false\n     |\n\n     = found: boolean\n
    \    = expected: string (by schema.yaml:63)\n\n/etc/kappctrl-mem-tmp/kapp-controller-template-values3133008665/values.yaml:\n
    \    |\n  27 |         remoteAuth: false\n     |\n\n     = found: boolean\n     =
```

**Affected product area (please put an X in all that apply)**

- [x] APIs
- [x] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**
Consistent usage of schema for boolean parameters and matching types in the API

**Steps to reproduce the bug**
Create a cluster with vsphere cpi in supervisor environment

PR #2196, opened 2022-04-20
Fix VSphereCPIConfig data values type inconsistency and directly use endpoint IP.

HanFa:

1. Currently the data values type conflicts with the TCE data value schema. This can be fixed by using boolean fields defined in the schema, instead of deprecated one of type string.

2. We cannot directly use the FQDN supervisor.local.svc for the value of SUPERVISOR_CONTROL_PLANE_IP, because of the fact that CPI boots up before core-dns is ready. The IP lookup will fail, causing error when initializing the cloud provider.

This PR, instead of using FQDN, it looks for the supervisor control plane IP with following steps.

    Check if a k8s service kube-system/kube-apiserver-lb-svc is available, if so, fetch the loadbalancer IP.
    If not, get the Supervisor Cluster Management Network Floating IP (FIP) from the cluster-info configmap. This is to support non-NSX-T development use cases only. If we are unable to find the cluster-info configmap for some reason, we log the error.

Fixes # https://github.com/vmware-tanzu/tanzu-framework/issues/2193

VSphereCPIConfig controller integration tests.

### Release note

```release-note
package-based-lcm: Fix VSphereCPIConfig data values type inconsistency and directly use endpoint IP. 
```

### PR Checklist


vijaykatam:
Change looks okay but needs to be rebased and we need testing section filled out.

HanFa:
added a testcase for fetching API server endpoint from `kube-system/kube-apiserver-lb-svc`

vijaykatam (approved):

PR #2202, opened 2022-04-21
CLI: Add context command

### LINKED ISSUE #2200: Explicit `context` command as to manage contextual information in the Tanzu CLI
A platform operator may need the CLI to target a TMC or a management cluster while an application developer is more likely to target a workload cluster, and an app operator may need to move between a number of workload clusters. Given that Tanzu CLI users may work with more than one end-point they need to switch between them easily. They should understand the context they are currently working against, especially when using plugins that depend on a particular type of end-point.

Currently, the Tanzu CLI supports storing authentication information on a per server basis (management cluster or a global server: TMC). This information is created using the `login` plugin and can be managed using the `config server` command in the core CLI. As we add more server types like a workload cluster, Tanzu Mission Control endpoint, etc. we need a consistent mechanism to manage the configuration for every server or rather for every context.

Also, we need to move away from the notion of `server` to `context`s. To clarify, a server is just a remote endpoint that can be accessed via the Tanzu CLI (or its plugins). A context would include a server endpoint and any related client side configurations like default output format, feature flags enabled, etc. This can also be extended in future to support a local environment without any server if needed.

Hence, we need a `context` command as part of the core CLI like below to create and manage contexts in the CLI.
```
~ tanzu context -h
Configure and manage contexts for the Tanzu CLI

Usage:
  tanzu context [command]

Aliases:
  context, ctx, contexts

Available Commands:
  create      Create a Tanzu CLI context
  delete      Delete a context from the config
  get         Display a context from the config
  list        List contexts
  use         Make a context as the current active context for its type

Flags:
  -h, --help   help for context

Use "tanzu context [command] --help" for more information about a command.
```

1. Currently, we can achieve these operations via `tanzu login` and `tanzu config server` commands. This is not ideal because the operations are scatter across multiple commands. Also, `login` is a plugin whereas `config server` is a sub-command in the core CLI.

2. Using the `KUBECONFIG` env variable for `kubectl` or similar helper commands like `kubectx`. This works for kubernetes clusters, but the Tanzu CLI would need to support other types of servers like TMC, Harbor, etc. in future.

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

giri-varma:

1. Add context types in client config
This is needed to later introduce a context command in the core CLI. Also, add conversion between server and context types for backwards compatibility with older core and older plugins.

2. Add context command
This command is used to create, list, use, and delete Tanzu CLI contexts for various control planes.

3. Hide context command behind a feature flag
The context command is hidden behind the feature flag `features.global.context-target`. This shall be removed once all plugins and successfully tested its working.

Fixes #2200

Added unit tests for the config changes. Tested the `tanzu context` command manually with the following steps:
1. Download an older version of the Tanzu CLI (`v0.21.0`)
2. Login with a management cluster `tanzu login`
3. List clusters `tanzu cluster list` to ensure that the TKG kubeconfig is valid
4. Upgrade to the new CLI with context command `make install-cli` and list the contexts `tanzu context list` to ensure the new CLI translates servers to contexts
5. Get management cluster `tanzu management-cluster get` & list clusters `tanzu cluster list` to ensure that existing plugins are still working
6. Cleanup the management cluster server `tanzu config server delete`
7. Repeat steps 2 & 3 to ensure that login command is still working
8. Repeat step 6
9. Create a context for the management cluster `tanzu context create` to ensure that the new command is working
10. List clusters `tanzu cluster list` to ensure that the new CLI translates contexts to servers

### Release note
```release-note
Currently, the Tanzu CLI supports storing authentication information on a per server basis (management cluster). This information is created using the login plugin and can be managed using the config server command in the core CLI. As we add more server types like a workload cluster, Tanzu Mission Control endpoint, etc. we need a consistent mechanism to manage the configuration for every server or rather for every context.

Also, we need to move away from the notion of server to contexts. To clarify, a server is just a remote endpoint that can be accessed via the Tanzu CLI (or its plugins). A context would include a server endpoint and any related client side configurations like default output format, feature flags enabled, etc. This can also be extended in future to support a local environment without any server if needed.

Hence, added a context command as part of the core CLI like below to create and manage contexts in the CLI.

Note: This is currently behind a feature flag. To enable the flag please run tanzu config set features.global.context-target true

Migration notes:
* CLI users should start using the `context` command instead of the `login` command.
* Plugin developers should use the `config.GetContext` methods in the core CLI library instead of `config.GetServer` when looking for server information.

~ tanzu context -h
Configure and manage contexts for the Tanzu CLI

Usage:
  tanzu context [command]

Aliases:
  context, ctx, contexts

Available Commands:
  create      Create a Tanzu CLI context
  delete      Delete a context from the config
  get         Display a context from the config
  list        List contexts
  use         Make a context as the current active context for its type

Flags:
  -h, --help   help for context

Use "tanzu context [command] --help" for more information about a command.
```

### PR Checklist

### Additional information for the reviewer
* The change is split into three commits to separate the config file changes, the actual command addition and the feature flag. Would prefer to keep this history for future.
* `docs/cli-architecture.md` contains user facing changes and `docs/plugin_implementation_guide.md` contains developer related changes.


vuil:
Thanks for the changes. Even though we are implementing a new command, it does carry with in implications that some existing functionality is being deprecated. 
I have been thinking whether it would make sense to put the context functionality behind a feature flag instead? Would doing so help or complicate the process of plugin authors transitioning their plugins to the context approach?

Also, can you provide additional details to the testing done in the PR description. In particular it would help the reviewers to see that we have demonstrated that the context/login use cases can coexist during this transition period where we are deprecating the latter.

alfredthenarwhal:
> @vuil:  /test upgrade-vc7
Commit: d81a352

Tests failed! Build no: 2157

alfredthenarwhal:
> @vuil:  /test install-aws
Commit: d81a352

Build failed! Build no: 2158

heyjcollins:
This `--help` output copy needs help:
```
use         Make a context as the current active context for its type
```

I propose the following:
```
use         Set the context to be used by default
```

giri-varma:
> Set the context to be used by default

@heyjcollins Thanks for the suggestion! I can change that.

giri-varma:
> Thanks for the changes. Even though we are implementing a new command, it does carry with in implications that some existing functionality is being deprecated. I have been thinking whether it would make sense to put the context functionality behind a feature flag instead? Would doing so help or complicate the process of plugin authors transitioning their plugins to the context approach?
> 
> Also, can you provide additional details to the testing done in the PR description. In particular it would help the reviewers to see that we have demonstrated that the context/login use cases can coexist during this transition period where we are deprecating the latter.

@vuil I intent of deprecating was to ensure that we don't have to support `Server` & `Context` types for a long time for newly introduced endpoints. Also, with them deprecated, not having an alternative would be a bad UX. That's part of the reason why the `context` is not feature flagged.

That said, I can remove the deprecation messages if you think we are not yet ready for the deprecation.

vuil:
Nit: The release notes section in the PR description is actually collated through automation and posted in the notes of the next release which includes the change. Most notes are short. But we have had instances where multiline output has been included, such as in 
https://github.com/vmware-tanzu/tanzu-framework/releases/tag/v0.19.0. 
This is an important change that is definitely worth mentioning. If we retain the text though, please see about wrapping some of the lines to a reasonable width. The alternative would be to just provide a link to another markdown doc in the repo, but in this case it might be useful to be more obvious.

vuil:
cc @rajathagasthya PTAL especially regarding the interaction with the CLI config.

@giri-varma it might be useful to reviewers to mention the intention to include other context types, and the complications https://github.com/vmware-tanzu/tanzu-framework/issues/2652 introduces.

cmd/cli/plugin/login/main.go
```diff
@@ -122,6 +124,7 @@ func login(cmd *cobra.Command, args []string) (err error) {
 		}
 	}
 
+	log.Warning("'tanzu login' is deprecated and shall be removed in a future version. Please use 'tanzu context create' instead.")
```
vuil, line 127: The timing and/or the message content doesn't look quite right when the interactive prompting is involved.

If the user is using interactive mode to add server, it seems weird to show the deprecation notice _after_ all the interaction.
If the user is using interactive mode to select a server, the deprecation message is not quite correct for the situation.
It is tricky to do anything before the prompt since we do not know the user's intention. If we want to, one option is to be general with the message. 
(seeing that you are about to launch into interactive mode...), show:
```
... login deprecated....
  to select endpoint, use ....
  to create server, use ....
```

pkg/v1/config/clientconfig.go
```diff
@@ -416,6 +421,14 @@ func storeConfigToLegacyDir(data []byte) {
 
 // StoreClientConfig stores the config in the local directory.
 func StoreClientConfig(cfg *configv1alpha1.ClientConfig) error {
+	// new plugins would be setting only contexts, so populate servers for backwards compatibility
```
rajathagasthya, line 424: Are there any unit tests that test these compatibility scenarios?

rajathagasthya (commented):
Config code LGTM. I'll let CLI folks give a final approval.

pkg/v1/config/clientconfig.go
rajathagasthya: I think `conversion_test.go` comes pretty close.

pkg/v1/cli/command/core/context.go
```diff
@@ -0,0 +1,643 @@
...
+var createCtxCmd = &cobra.Command{
+	Use:   "create",
+	Short: "Create a Tanzu CLI context",
+	RunE:  createCtx,
+	Example: `
+	# Create a TKG management cluster context using endpoint
+	tanzu context create --endpoint "https://k8s.example.com" --name mgmt-cluster
+
+	# Create a TKG management cluster context using kubeconfig path and context
+	tanzu context create --kubeconfig path/to/kubeconfig --kubecontext kubecontext --name workload-cluster
```
vuil, line 82: nit: the name `workload-cluster` is going to be confusing, maybe just mgmt-cluster?

vuil (commented):
lgtm, thanks for the changes! Just a tiny nit on the usage example.

alfredthenarwhal:
Build failed! Build no: 2273

pkg/v1/cli/command/core/context.go
giri-varma: Fixed

alfredthenarwhal:
Tests failed! Build no: 2279

alfredthenarwhal:
Build failed! Build no: 2294

alfredthenarwhal:
Build failed! Build no: 2314

pkg/v1/config/context.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	return cfg.HasContext(name), nil
+}
+
+// AddContext adds a Context to the config.
+func AddContext(c *configv1alpha1.Context, setCurrent bool) error {
+	cfg, err := GetClientConfig()
```
sathyanarays, line 34: `AcquireTanzuConfigLock()` needs to be added before changing the config. Similarly in case of Remove method as well!

sathyanarays: Also in `SetCurrentContext`

giri-varma: Sure, will add it! Is there a reason why the acquire lock is not part of `StoreClientConfig `? That way not all callers do not have to worry about the lock

giri-varma: @sathyanarays Added the locks!

pkg/v1/config/clientconfig.go
```diff
@@ -103,6 +105,7 @@ const (
 var (
 	DefaultCliFeatureFlags = map[string]bool{
 		FeatureContextAwareCLIForPlugins:                      common.ContextAwareDiscoveryEnabled(),
+		FeatureContextCommand:                                 false,
```
vuil, line 108: nit: 
Not sure how we plan to use the flag, but should we add this feature flag to the config?
The option is definitely there, but so it not advertising it at all.
If the purpose of the flag is to use it for internal/testing, but there is no intention of making it deactivate-able when we turn it on, then there is little reason to introduce it in the config.

vuil (commented):
Thanks @giri-varma for the updates. Changes looks pretty much good to go. I have a couple of questions regarding the deprecation message and inclusion of the cli ff in the config file.

pkg/v1/cli/command/core/config.go
```diff
@@ -37,9 +37,6 @@ func init() {
 	)
 	serversCmd.AddCommand(listServersCmd)
 	addDeleteServersCmd()
-	cli.DeprecateCommandWithAlternative(serversCmd, "v1.0.0", "tanzu context")
```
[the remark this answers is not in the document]
giri-varma: I am not sure if we should deprecate the existing command if the alternative is still behind a feature flag. Adding a deprecation message based on a feature flag does not look like a good UX. The deprecation message can be added when we remove the feature flag.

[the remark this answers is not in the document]
vuil: I view activating the feature flag as effectively introducing new commands while deprecating the use of others. In this situation the deprecation messages seem useful. In addition, a small plus is that it is more unlikely to forgot having them when the ff is removed.

pkg/v1/config/clientconfig.go
vuil: I don't feel strongly one way or the other, especially if we don't plan to retain the use of this flag once the feature is eventually lit up.

pkg/v1/cli/command/core/config.go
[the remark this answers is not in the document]
giri-varma: My concern is that the new commands are not yet available to users. So, seems confusing that they are deprecated. Also, we do not have a decision on when the existing commands will be removed. `v1.0.0` feels misleading.

pkg/v1/config/clientconfig.go
vuil: I am ok with retain the reference in the config on the assumption that do not plan to use it once the functionality is lit up.

pkg/v1/cli/command/core/config.go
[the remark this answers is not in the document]
vuil: What I meant was to show deprecation message only when the feature flag is activated. 

But thinking about it more, we do not need to hold up the PR for this. As long as the deprecation messages are reinstated at the time the context feature is activated for real (also a tad easier to do so then), we should be good.

vuil (approved):
Lgtm, thanks for the patience and the updates.

PR #2247, opened 2022-04-27
Bump cluster-api-azure to v1.2.1

saimanoj01:
Bump cluster-api-azure to v1.2.1

Fixes #

### Release note

```release-note
Bump cluster-api-azure to v1.2.1
```

### PR Checklist


christianang:
I think I would have expected `infrastructure-components.yaml` and `metadata.yaml` to be updated if we bumped this package. We also have additional sync steps because we have both the package and the `pkg/v1/providers` that we would like to keep in sync, which you can find in this [readme](https://github.com/vmware-tanzu/tanzu-framework/tree/main/hack/providers-sync-tools).

I do see you updated the vendir.yaml in the package, which is good. Seems like you need to run `vendir sync` in the package and then run `make all` in the `hack/providers-sync-tools` directory to get the files in sync.

vuil (approved):
lgtm, thanks

Issue #1710, opened 2022-02-17
Embedded certificate bundle should include all Mozilla CA certs, otherwise CLI does not work on Windows


joshrosso:
**Bug description**

_This originated from user feedback we've received in TCE (https://github.com/vmware-tanzu/community-edition/issues/1398)_

Built the `management-cluster` plugin from `0.16.0` and attempted to bootstrap a `managed-cluster`:

![image](https://user-images.githubusercontent.com/6200057/154384486-d59ab4ea-fc2a-4afc-a0e1-baee8e879a6d.png)

You can workaround this using https://github.com/vmware-tanzu/community-edition/issues/1398#issuecomment-902863369. But it should pull without this workaround.

Note that unmanaged cluster pulls a TKR and does **not** have this issue.

![image](https://user-images.githubusercontent.com/6200057/154384611-239ba1ac-6281-4238-96fd-a64f10684522.png)

**Affected product area (please put an X in all that apply)**

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

randomvariable:
Via @stuartpreston :

> Ok, I think I have this figured out and it's pretty bad if you're a Windows user :(
TLDR: We are [embedding in the core CLI](https://github.com/vmware-tanzu/tanzu-framework/blob/main/pkg/v1/cli/clientconfighelpers/windows_trusted_ca.go) the root CA cert for [projects.registry.vmware.com](http://projects.registry.vmware.com/) - however the cert is the Root CA for the internal VIP and not the public access via CDN.

> More: [Tanzu CLI writes the embedded root CA cert file to the home directory on Windows and then adds it to the cert chain](https://github.com/vmware-tanzu/tanzu-framework/blob/fe23cf4d8a1d33251ebf2599b408419122a34808/pkg/v1/cli/clientconfighelpers/helpers.go#L67), which is why no-one using the CLI on Windows other than internal to VMware can download any plugins or do a tanzu init.

> Though the PR will need some consideration because any internal tests may break if they are run on a system that resolves [projects.registry.vmware.com](http://projects.registry.vmware.com/) and [projects-stg.registrty.vmware.com](http://projects-stg.registrty.vmware.com/) to an internal VIP.

randomvariable:
[...]

PR #2283, opened 2022-05-03
Embed PEM of root certificates in Mozilla's root store

saimanoj01:
We are [embedding in the core CLI](https://github.com/vmware-tanzu/tanzu-framework/blob/main/pkg/v1/cli/clientconfighelpers/windows_trusted_ca.go) the root CA cert for [projects.registry.vmware.com](http://projects.registry.vmware.com/) - however the cert is the Root CA for the internal VIP and not the public access via CDN. [Tanzu CLI writes the embedded root CA cert file to the home directory on Windows and then adds it to the cert chain](https://github.com/vmware-tanzu/tanzu-framework/blob/fe23cf4d8a1d33251ebf2599b408419122a34808/pkg/v1/cli/clientconfighelpers/helpers.go#L67), which is why no-one using the CLI on Windows other than internal to VMware can download any plugins or do a tanzu init.

The fix is to include the full mozilla bundle - [PEM of Root Certificates in Mozilla's Root Store with the Websites (TLS/SSL) Trust Bit Enabled](https://ccadb-public.secure.force.com/mozilla/IncludedRootsPEMTxt?TrustBitsInclude=Websites)

Fixes https://github.com/vmware-tanzu/tanzu-framework/issues/1710

1. Made sure that the `tanzu plugin sync` works from a windows machine setup on Azure.
2. Made sure that `tanzu management-cluster create` downloads the bom files necessary to proceed with the cluster creation

### Release note

```release-note

```

### PR Checklist


tenczar (approved):
LGTM. I'm a little annoyed that the Mozilla wiki does not seem to have a hash for downloading this file. Getting a rogue CA slipped into a giant list of trusted root servers seems like something that should be guarded against. In any case the change is straight forward.

pkg/v1/cli/clientconfighelpers/windows_trusted_ca.go
```diff
@@ -10,31 +10,3546 @@ package clientconfighelpers
 //	-showcerts
```
prkalle, line 10: nit: comment to be updated and please mention the url from where we downloaded the mozilla cert bundle and a note to update this bundle periodically would be helpful.

prkalle (approved):
LGTM. Thanks for the changes @saimanoj01 .
I have one minor nit comment.

anujc25 (approved):
LGTM

Issue #2197, opened 2022-04-20
CEIP should be enabled by default except for Tanzu Community Edition


saji-pivotal:
**Bug description**
While creating a management-cluster, the user is opted into CEIP participation even when `ENABLE_CEIP_PARTICIPATION` flag has been explicitly set to false. We also need to ensure that CEIP is set to false by default while creating a TCE cluster, but should be set to true otherwise

**Affected product area (please put an X in all that apply)**

- [ ] APIs
- [ ] Addons
- [X] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**
The user must be opted out of CEIP when ENABLE_CEIP_PARTICIPATION has been set to false

**Steps to reproduce the bug**
* Create a config file (`mc-config.yaml`) for creating the management cluster. 
* In the config file set `ENABLE_CEIP_PARTICIPATION: "false"`
* Run `tanzu management-cluster -f mc-config.yaml`
* The console output shows the following line `CEIP Opt-in status: true`

**Version** (include the SHA if the version is not obvious)

**Environment where the bug was observed (cloud, OS, etc)**
Mac client

**Relevant Debug Output (Logs, manifests, etc)**

PR #2290, opened 2022-05-03
Opted out of CEIP when ENABLE_CEIP_PARTICIPATION is set to false

saji-pivotal:
CEIP should be false for TCE by default, and true for other editions. 
If ENABLE_CEIP_PARTICIPATION is set to false, then CEIP participation must be disabled. This PR is ensure that functionality.

Fixes #https://github.com/vmware-tanzu/tanzu-framework/issues/2197

Manual Testing
Before making the fix
1. Created a management cluster on AWS by setting `ENABLE_CEIP_PARTICIPATION: false`
2. Console output showed CEIP Status: Opt In
3. Noticed that the tkg-telemetry cronjob was running on the cluster
```
$ kubectl get pods -A
NAMESPACE                           NAME                                                                READY   STATUS      RESTARTS   AGE
...
tkg-system-telemetry                tkg-telemetry-27525600-thmkg                                        0/1     Completed   0          16h
tkg-system-telemetry                tkg-telemetry-27525960-wtxfr                                        0/1     Completed   0          10h
tkg-system-telemetry                tkg-telemetry-27526320-rbflp                                        0/1     Completed   0          4h53m
...

$ kubectl get cronjob -A
NAMESPACE              NAME            SCHEDULE      SUSPEND   ACTIVE   LAST SCHEDULE   AGE
tkg-system-telemetry   tkg-telemetry   0 */6 * * *   False     0        4h53m           4d22h
```

After making the fix
1. Created a management cluster on AWS by setting `ENABLE_CEIP_PARTICIPATION: false`
2. Console output showed CEIP Optin: false
```
ENABLE_CEIP_PARTICIPATION=false ./artifacts/darwin/amd64/cli/management-cluster/v0.22.0-dev/tanzu-management-cluster-darwin_amd64 create -f ~/sample-aws-20220502.yaml -v 6
...
CEIP Opt-in status: false
```
3. Verified that the tkg-telemetry cronjob is not running
```
$ kubectl get cronjob -A
No resources found
```
4. Verified the opt-in status using the ceip-plugin
```
$ ./artifacts/darwin/amd64/cli/management-cluster/v0.22.0-dev/tanzu-management-cluster-darwin_amd64 ceip-participation get
  MANAGEMENT-CLUSTER-NAME      CEIP-STATUS
  tkg-mgmt-aws-20220503102629  Opt-out
```

Automated testing:
* Set `ENABLE_CEIP_PARTICIPATION: false` while running the Integration tests that run as part of the Github CI, and added a step to verify that the CEIP Status is Opt-out

## Release note

```release-note
* Fixed bug where ENABLE_CEIP_PARTICIPATION config variable was not being heeded
```

### PR Checklist

### DIFF
--- a/pkg/v1/tkg/tkgctl/init.go
+++ b/pkg/v1/tkg/tkgctl/init.go
@@ -69,7 +69,7 @@ func (t *tkgctl) Init(options InitRegionOptions) error {
 
 	ceipOptIn, err := strconv.ParseBool(options.CeipOptIn)
 	if err != nil {
-		ceipOptIn = false
+		ceipOptIn = true
 	}
 
 	if logPath, err := t.getAuditLogPath(options.ClusterName); err == nil {
@@ -317,6 +317,8 @@ func (t *tkgctl) configureInitManagementClusterOptionsFromConfigFile(iro *InitRe
 		ceipOptIn, err := t.TKGConfigReaderWriter().Get(constants.ConfigVariableEnableCEIPParticipation)
 		if err == nil {
 			iro.CeipOptIn = ceipOptIn
+		} else if iro.Edition == "tce" {
+			iro.CeipOptIn = False
 		} else {
 			iro.CeipOptIn = True
 		}
--- a/pkg/v1/tkg/tkgctl/init.go
+++ b/pkg/v1/tkg/tkgctl/init.go
@@ -48,6 +48,10 @@ type InitRegionOptions struct {
 	GenerateOnly                bool
 }
 
+const (
+	TCEBuildEdition = "tce"
+)
+
 //nolint:gocritic,gocyclo,funlen
 // Init initializes tkg management cluster
 func (t *tkgctl) Init(options InitRegionOptions) error {
@@ -69,7 +73,11 @@ func (t *tkgctl) Init(options InitRegionOptions) error {
 
 	ceipOptIn, err := strconv.ParseBool(options.CeipOptIn)
 	if err != nil {
-		ceipOptIn = true
+		if options.Edition == TCEBuildEdition {
+			ceipOptIn = false
+		} else {
+			ceipOptIn = true
+		}
 	}
 
 	if logPath, err := t.getAuditLogPath(options.ClusterName); err == nil {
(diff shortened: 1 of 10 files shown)


pkg/v1/tkg/tkgctl/utils.go
```diff
@@ -0,0 +1,17 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package tkgctl
+
+import "github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkg/constants"
+
+func setCEIPOptinBasedOnConfigAndBuildEdition(t *tkgctl, edition string) string {
+	ceipOptIn, err := t.TKGConfigReaderWriter().Get(constants.ConfigVariableEnableCEIPParticipation)
```
anujc25, line 9: Do we really need to pass `tkgctl` client as function argument? Couple of options I can suggest are: 
1. It would be better if we can take just the `t.TKGConfigReaderWriter()` as function parameter instead of `tkgctl` client to make function more generic.
2. Or we can make this function part of `tkgctl` client definition so that we do not need to pass `tkgctl` client as function argument and invoke directly with `t,setCEIPOptinBasedOnConfigAndBuildEdition(edition)`

pkg/v1/tkg/tkgctl/init.go
```diff
@@ -69,7 +69,7 @@ func (t *tkgctl) Init(options InitRegionOptions) error {
 
 	ceipOptIn, err := strconv.ParseBool(options.CeipOptIn)
 	if err != nil {
-		ceipOptIn = false
+		ceipOptIn = true
```
anujc25, line 72: What if I am a TCE user and I typed the CEIPOptIn flag value incorrectly. Will it set the value to `true` by default?

saji-pivotal: Good catch. I've added a condition for this

pkg/v1/tkg/tkgctl/utils_test.go
```diff
@@ -0,0 +1,84 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
```
stmcginnis, line 1: Should be 2022 since this is a new file

pkg/v1/tkg/tkgctl/utils.go
```diff
@@ -0,0 +1,17 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
```
stmcginnis, line 1: Should be 2022.

tenczar (approved):
LGTM

pkg/v1/tkg/tkgctl/utils.go
```diff
@@ -0,0 +1,17 @@
...
+
+func (t *tkgctl) setCEIPOptinBasedOnConfigAndBuildEdition(edition string) string {
+	ceipOptIn, err := t.TKGConfigReaderWriter().Get(constants.ConfigVariableEnableCEIPParticipation)
+	if err == nil {
+		return ceipOptIn
+	} else if edition == TCEBuildEdition {
+		return False
+	} else {
+		return True
+	}
```
anujc25, line 16: nit: We can reduce the if... else if ... else and convert it into 4 if statements here as we have `return` in each if statement.

anujc25 (approved):
One minor comment. But looks good otherwise.

pkg/v1/tkg/tkgctl/utils.go
saji-pivotal: Thanks! Updated.

PR #2311, opened 2022-05-04
setting vsphere csi replica count based on controlplane node count

### LINKED ISSUE #2117: VSphereCSIConfigController should set deployment replicas based on control plane node count
**Bug description**
vsphere-csi controller replicas is static which can lead to issues during upgrade of cluster.

Deployment replicas should be set based on control plane count so that scale up/scale down of control plane nodes results in vsphere-csi controller replicas equal to control plane nodes.

Requirements
1. Minimum of one replica
2. Max 3 replicas if control plane node count >=3

- [ ] APIs
- [x] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

peterochodo:
- we should set the vsphere csi replica count based on control plane node count

Fixes #2117 

- env test runs have verified that vsphere-csi count is now based on control-plane count

### Release note

```release-note
package-based-lcm: set vsphere csi replica count based on controlplane node count
```

### PR Checklist

### DIFF
--- a/addons/controllers/csi/vspherecsiconfig_utils.go
+++ b/addons/controllers/csi/vspherecsiconfig_utils.go
@@ -17,6 +17,7 @@ import (
 	capvv1beta1 "sigs.k8s.io/cluster-api-provider-vsphere/apis/v1beta1"
 	capvvmwarev1beta1 "sigs.k8s.io/cluster-api-provider-vsphere/apis/vmware/v1beta1"
 	clusterapiv1beta1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	capictrlpkubeadmv1beta1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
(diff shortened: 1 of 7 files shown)


addons/controllers/csi/vspherecsiconfig_utils.go
```diff
@@ -326,3 +332,49 @@ func (r *VSphereCSIConfigReconciler) overrideMiscValues(dvscsi *DataValuesVSpher
 		dvscsi.InsecureFlag = *config.InsecureFlag
 	}
 }
+
+func (r *VSphereCSIConfigReconciler) constrainNumberOfDeploymentReplicas(ctx context.Context, proposedCount int) int {
```
danniel1205, line 336: Suggestion: wondering if things would be easier to do `proposedCount int32` return `int32` as well? So that caller does not have to do several conversions.

peterochodo: updated

danniel1205 (approved):
lgtm

PR #2313, opened 2022-05-04
addons/pinniped/tanzu-auth-controller-manager: Remove pinniped block if id type none

sabbey37:
We don't need any pinniped-specific data values if `identity_management_type: none`, so let's remove this block if that's the case.  Let's also set the annotation, type, and labels in the `getMutateFn` in `CreateOrPatch`

Fixes #

Updated existing unit tests, added tests, tested on vSphere cluster

### Release note

```release-note

```

### PR Checklist

### DIFF
--- a/addons/pinniped/tanzu-auth-controller-manager/controllers/utils_test.go
+++ b/addons/pinniped/tanzu-auth-controller-manager/controllers/utils_test.go
@@ -93,20 +93,17 @@ func verifySecretFunc(ctx context.Context, cluster *clusterapiv1beta1.Cluster, c
 			identityManagementTypeKey: none,
 			"infrastructure_provider": "vsphere",
 			"tkg_cluster_role":        "workload",
-			"pinniped": map[string]interface{}{
-				supervisorEndpointKey: "",
-				supervisorCABundleKey: "",
-				"concierge": map[string]interface{}{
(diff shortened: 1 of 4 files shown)


addons/pinniped/tanzu-auth-controller-manager/controllers/utils_test.go
```diff
@@ -93,20 +93,17 @@ func verifySecretFunc(ctx context.Context, cluster *clusterapiv1beta1.Cluster, c
 			identityManagementTypeKey: none,
 			"infrastructure_provider": "vsphere",
 			"tkg_cluster_role":        "workload",
-			"pinniped": map[string]interface{}{
```
benjaminapetersen, line 96: Does indeed look removed. 💀

ankeesler (approved):
lgtm (eyeballing it)

addons/pinniped/tanzu-auth-controller-manager/controllers/v1_cascade_controller_test.go
```diff
@@ -178,6 +178,10 @@ var _ = Describe("Controller", func() {
 		})
 
 		When("the secret is deleted", func() {
+			BeforeEach(func() {
+				deleteObject(ctx, secret)
+			})
```
sabbey37, line 183: Wanted to call out that we don't actually need to delete the secret again here since it's deleted in an `aftereach` above, but it seemed good to be extra explicit.

PR #2337, opened 2022-05-09
Expose `AVI_LABELS` in workload cluster config file when performing `tanzu cluster create`

HanFa:

This PR exposes the field `AVI_LABELS` in the workload cluster config file. This allows cluster admin to specify the labels when bootstrapping workload clusters. One actual use case for this flag is the customization for AKO. With this flag provided, the cluster admin doesn't have to label or restart the workload cluster manually after its creation.

Fixes #

Currently testing it with downstream pipelines

### Release note

```release-note
Expose AVI_LABELS for creating workload cluster, for example `"{foo: 'bar'}"`, double quotations are required.
```

### PR Checklist


pkg/v1/providers/ytt/03_customizations/02_avi/add-avi-labels.yaml
```diff
@@ -0,0 +1,20 @@
+#@ load("@ytt:data", "data")
+#@ load("@ytt:json", "json")
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:template", "template")
+
+
+#! Label workload cluster if AVI_LABELS is specified
```
sbhatp, line 7: So we can use AVI_LABELS flag to label cluster during creation? Why we name it with AVI_ prefix?

HanFa: we use the AVI_ prefix to comply with the existing `AVI_LABELS` for mgmt clusters. 

* when `AVI_LABELS` is specified for MC, the install-ako-for-all ADC will have `clusterSelector` set
* when `AVI_LABELS` is specified for WC, the cluster object will be labelled to be selected by the ADC above

pkg/v1/providers/config_default.yaml
```diff
@@ -582,7 +582,7 @@ AVI_MANAGEMENT_CLUSTER_CONTROL_PLANE_VIP_NETWORK_NAME:
 AVI_MANAGEMENT_CLUSTER_CONTROL_PLANE_VIP_NETWORK_CIDR:
 
 AVI_CA_DATA_B64: ""
-AVI_LABELS: ""
```
XudongLiuHarold, line 585: If we change this, will it cause anything changed here:https://github.com/vmware-tanzu/tanzu-framework/blob/main/pkg/v1/providers/ytt/02_addons/avi/ako_operator_secret.yaml#L9?

pkg/v1/providers/ytt/03_customizations/02_avi/add-avi-labels.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+#@ if data.values.PROVIDER_TYPE == "vsphere" and data.values.TKG_CLUSTER_ROLE == "workload" and data.values.AVI_LABELS:
+#@overlay/match missing_ok=True,by=overlay.subset({"kind":"Cluster"})
+---
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: Cluster
+metadata:
+  #@overlay/match-child-defaults missing_ok=True
+  labels:
+    _: #@ template.replace(data.values.AVI_LABELS)
```
XudongLiuHarold, line 17: Just want to confirm that this will append values in `AVI_LABELS` not replace it with `AVI_LABELS`, right?
Cuz we shouldn't replace it with `AVI_LABELS`

HanFa: No it won't replace. instead, it will append the AVI_LABELS to the original cluster label.

XudongLiuHarold (approved):
LGTM

pkg/v1/providers/ytt/03_customizations/02_avi/add-avi-labels.yaml
sbhatp: Got it, thanks

vuil:
one way to ensure the the clustegen diffs produced for this PR shows the effect of avi_add_labels.yaml overlay is to generate some test cases that introduces some label variable values.
take a look at and modify:
cluster_optional.model

vuil:
@HanFa PTAL at the most recent clustergen diffs.
https://storage.googleapis.com/tkg-clustergen/2337/20220524005221/clustergen.diff.txt
It does not look like the generated output is what it is supposed to be.

Ran locally I could see the test cases are generated with the following values for AVI_LABELS:
```
> grep AVI_LABEL tests/clustergen/testdata/*.case
tests/clustergen/testdata/27068254.case:AVI_LABELS: {foo: 'bar'}
tests/clustergen/testdata/27389396.case:AVI_LABELS:
tests/clustergen/testdata/28205612.case:AVI_LABELS: {foo: 'bar'}
tests/clustergen/testdata/33016127.case:AVI_LABELS:
tests/clustergen/testdata/34071049.case:AVI_LABELS:
```
Is that the intention?

HanFa:
@vuil the value of AVI_LABELS is supposed to be something like `{foo: "bar"}` in the cluster config file.

I have verified this manually but not sure how to escape the double quotation in the model file.

pkg/v1/providers/ytt/03_customizations/02_avi/add-avi-labels.yaml
```diff
@@ -0,0 +1,20 @@
+#@ load("@ytt:data", "data")
+#@ load("@ytt:json", "json")
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:template", "template")
+
+
+#! Label workload cluster if AVI_LABELS is specified
+
+#@ if data.values.PROVIDER_TYPE == "vsphere" and data.values.TKG_CLUSTER_ROLE == "workload" and data.values.AVI_LABELS:
```
vuil, line 9: either cluster_option model file's possible values for TKG_CLUSTER_ROLE needs to be updated or the check here should be to allow some test cases to trigger this overlay.
When I tried that, I am noticing that the overlay does not work as expected when the config file has the AVI_LABELS set as this (with no quotes)

```
...
AVI_LABELS: {foo: 'bar'}
```

error seen:
```
- template.replace: expected value to be map, but was string
    in <toplevel>
      03_customizations/02_avi/add-avi-labels.yaml:18 |     _: #@ template.replace(data.values.AVI_LABELS)
```

PTAL to see if the format of the value or overlay needs to be updated.

HanFa: @vuil  I've flipped the logic into `if data.values.TKG_CLUSTER_ROLE != "management"`

Also, for `AVI_LABELS` to work, the value should be surrounded by the double quotation, which the current testcase gen cannot produce. For example,

```
AVI_LABELS: "{foo: 'bar'}"
```

HanFa: Can we merge this one first as it's blocking other tasks? We can have a separate thread on how to fix the testcase gen.

vuil:
> @vuil the value of AVI_LABELS is supposed to be something like `{foo: "bar"}` in the cluster config file.
> 
> I have verified this manually but not sure how to escape the double quotation in the model file. Can I have your help writing this clustergen test case?

If double quoting is required, let's make sure there is clustergen is setup to generate test cases with them.
This requires special handling of double quotes in the model file. You can add the following mods:

```
diff --git a/pkg/v1/providers/tests/clustergen/gencluster_params.py b/pkg/v1/providers/tests/clustergen/gencluster_params.py
index e2577765..2a0f781f 100755
--- a/pkg/v1/providers/tests/clustergen/gencluster_params.py
+++ b/pkg/v1/providers/tests/clustergen/gencluster_params.py
@@ -46,13 +46,14 @@ def write_test_cases(params_file, test_dir):
                         else:
                             cmd_args.append('{} {}'.format(k, v))
                 else:
-                    # hack to workaround the problem with pict where there is no escape char for comma.
+                    # hack to workaround the problem with pict where there is
+                    # no escape char for comma and double quote.
                     # sets "AZURE_CUSTOM_TAGS" to "tagKey1=tagValue1, tagKey2=tagValue2"

                     if k == "AZURE_CUSTOM_TAGS" and v.startswith('tagKey1='):
                         cfg_args.append('{}: {}'.format(k, 'tagKey1=tagValue1, tagKey2=tagValue2'))
                     elif v != "NA":
-                        cfg_args.append('{}: {}'.format(k, v.replace("<comma>", ",")))
+                        cfg_args.append('{}: {}'.format(k, v.replace("<comma>", ",").replace("<qq>", "\"")))


             testid = int(hashlib.sha256(dict_contents.encode('utf-8')).hexdigest(), 16) % 10**8
diff --git a/pkg/v1/providers/tests/clustergen/param_models/cluster_optional.model b/pkg/v1/providers/tests/clustergen/param_models/cluster_optional.model
index a0d34dcd..1d3d768d 100644
--- a/pkg/v1/providers/tests/clustergen/param_models/cluster_optional.model
+++ b/pkg/v1/providers/tests/clustergen/param_models/cluster_optional.model
@@ -118,6 +118,8 @@ AUTOSCALER_SCALE_DOWN_UNNEEDED_TIME: "NA", "50s"
 AUTOSCALER_MAX_NODE_PROVISION_TIME: "NA", "5m"
 AUTOSCALER_MAX_NODES_TOTAL: "NA", "10", "50", "0"

+AVI_LABELS: "NA", "<qq>{foo: 'bar'}<qq>"
+
 # Antrea settings
 ANTREA_NO_SNAT: "false", "true"
 ANTREA_TRAFFIC_ENCAP_MODE: "encap", "noEncap", "hybrid", "networkPolicyOnly"
```

vuil (approved):
lgtm

my only concern is the labels format for the AVI_LABELS value requiring quoting could lead to confusing UX when creation fails due to user forgetting to quote the stringified map

@HanFa please add a release note entry to the PR descriptiont since there is user facing change. This might be one of those case even showing a sample variable value (be sure to include the quotes) in the release notes would be good.

HanFa:
@vuil 
> AVI_LABELS value requiring quoting could lead to confusing UX when creation fails due to user forgetting to quote the stringified map

if double quotations are not provided, the label won't be there and it will be confusing to the user.

Also, I've added the release note. Thanks for the help.

PR #2173, opened 2022-04-18
TKR Source Controller implementation

imikushin:
Fetch TKR BOM, TKR compatibility, TKR package repository images.
Install TKR packages.

### What this PR does / why we need it

The TKR Source controller installs TKRs from TKR packages, and makes sure the TKR Compatible condition reflects TKR objects' compatibility status.

### Which issue(s) this PR fixes

Fixes #1352. 

### Describe testing done for PR

- Unit tests.
- Built the docker image using the added Makefile.
- Run the controller against a TKG management cluster.

### Release note

```release-note
package-based-lcm: TKR Source Controller makes TKRs published as Carvel packages available on the management cluster.
```

### PR Checklist

- [X] Squash the commits into one or a small number of logical commits
      
- [X] Use good commit [messages](https://github.com/vmware-tanzu/tanzu-framework/blob/main/CONTRIBUTING.md)
- [X] Ensure PR contains terms all contributors can understand and links all contributors can access

PR #2343, opened 2022-05-10
TKR Source Controller: Add Cluster API to client scheme

imikushin:

This fixes an issue with initializing scheme for the client. Otherwise the client would not be able to work with the Cluster resource for the management cluster (which it needs to monitor for upgrades).

Follow up to #2173. Fixes an issue with initializing scheme for the client.

### Release note

```release-note
NONE
```

### PR Checklist


prkalle (approved):
LGTM. Thanks!

anujc25 (approved):

PR #2348, opened 2022-05-11
derive all service and pod cidrs from cluster config

adobley:
This fixes an error encountered when trying to delete an IPv6 primary
dualstack cluster.

KinD clusters provisioned when upgrading or deleting dualstack IPv6 primary
clusters expect pod cidrs and service cidrs from both IP families to be
present.

Fixes #

Deployed an IPv6-primary dualstack management cluster and was able to successfully delete it.

### Release note

```release-note
Support deletion of IPv6 primary dualstack management clusters with automatically provisioned cleanup clusters.
```

### PR Checklist


vuil (approved):
lgtm, thanks

PR #2351, opened 2022-05-11
Remove owner reference from PackageInstall on workload cluster

### LINKED ISSUE #2350: OwnerRefs to Cluster on PackageInstall in workload clusters should not exist
**Bug description**
An OwnerRef to Cluster on PackageInstall in workload clusters was added in https://github.com/vmware-tanzu/tanzu-framework/pull/1910. This is not correct because there is no Cluster Kind on a workload cluster. Looks like k8s was not preventing these objects from being created but logging warnings.

The side effect of this is when the Cluster Kind actually exists on a cluster, kubernetes will start garbage collection of these PackageInstall objects because the uid references by the PackageInstall object does not exist.

- [ ] APIs
- [x] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**
Invalid OwnerRefs should not exist

Create bootstrap + management cluster with package-based-lcm

adduarte:
package installs on a workload cluster can not have an owner reference pointing to cluster kind.

- removes owner reference from additional package installs
- adds unit test to verify core and additional package installs do not have owner references 

fixes #2350

### PR Checklist

### Release note
```release-note
package-based-lcm: Remove owner references from PackageInstall on workload clusters. 
```

### DIFF
--- a/addons/controllers/clusterbootstrap_controller_test.go
+++ b/addons/controllers/clusterbootstrap_controller_test.go
@@ -170,26 +170,28 @@ var _ = Describe("ClusterBootstrap Reconciler", func() {
 				Expect(hasPackageInstalls(ctx, k8sClient, cluster, constants.TKGSystemNS,
 					clusterBootstrap.Spec.AdditionalPackages, logr.Logger{})).To(BeTrue())
 
-				By("packageinstalls for additional packages should all have the correct owner reference set")
-				var ownerIsCluster bool
+				By("packageinstalls for core packages should not have owner references")
+				var corePackages []*runtanzuv1alpha3.ClusterBootstrapPackage
+				corePackages = append(corePackages, clusterBootstrap.Spec.CNI, clusterBootstrap.Spec.CPI, clusterBootstrap.Spec.CSI)
 				pkgInstall := &kapppkgiv1alpha1.PackageInstall{}
-				for _, pkg := range clusterBootstrap.Spec.AdditionalPackages {
-					ownerIsCluster = false
+				for _, pkg := range corePackages {
 					pkgInstallName := util.GeneratePackageInstallName(cluster.Name, pkg.RefName)
 					err := k8sClient.Get(ctx, client.ObjectKey{Name: pkgInstallName, Namespace: constants.TKGSystemNS}, pkgInstall)
-					if err == nil {
-						for _, ownerRef := range pkgInstall.OwnerReferences {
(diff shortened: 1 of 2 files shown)


addons/controllers/clusterbootstrap_controller_test.go
```diff
@@ -170,26 +170,6 @@ var _ = Describe("ClusterBootstrap Reconciler", func() {
...
-								ownerIsCluster = true
-								break
-							}
-						}
-						if !ownerIsCluster {
-							break
-						}
-					}
-				}
-				Expect(ownerIsCluster).To(BeTrue())
```
vijaykatam, line 192: Could we add a test to verify that ownerRefs is empty for all packages except kapp?

vijaykatam:
To clarify this change is not specific to just additionalPackages. Everything except kapp.

addons/controllers/clusterbootstrap_controller_test.go
adduarte: done.

addons/controllers/clusterbootstrap_controller_test.go
```diff
@@ -170,26 +170,28 @@ var _ = Describe("ClusterBootstrap Reconciler", func() {
 				Expect(hasPackageInstalls(ctx, k8sClient, cluster, constants.TKGSystemNS,
 					clusterBootstrap.Spec.AdditionalPackages, logr.Logger{})).To(BeTrue())
 
-				By("packageinstalls for additional packages should all have the correct owner reference set")
-				var ownerIsCluster bool
+				By("packageinstalls for core packages should not have owner references")
+				var corePackages []*runtanzuv1alpha3.ClusterBootstrapPackage
+				corePackages = append(corePackages, clusterBootstrap.Spec.CNI, clusterBootstrap.Spec.CPI, clusterBootstrap.Spec.CSI)
```
adduarte, line 175: It would be nice to be able to pull all the PI for a cluster. 
We can do it for additional package, but for core packages the list is hardcoded here: 
https://github.com/vmware-tanzu/tanzu-framework/blob/8e84c7a002814a6711610e11481c653ba48eb770/addons/controllers/clusterbootstrap_controller.go#L241

anujc25 (approved):
Looks good

PR #2378, opened 2022-05-13
Hide TKR_DATA when not running yaml tests

tenczar:
This change adds TKR_DATA to the ccluster.yml when testing in clustergen

This PR updates the clustergen test runner to append fake TKR_DATA to the config file. This data is necessary for the dry run process to generate the expected output classes. We don't want this fake data leak to customer scenarios, so the test runner directly updates the config file with the TKR_DATA

Fixes #

### Release note

```release-note

```

### PR Checklist

### DIFF
--- a/pkg/v1/providers/tests/clustergen/run_tests.sh
+++ b/pkg/v1/providers/tests/clustergen/run_tests.sh
@@ -72,21 +72,60 @@ generate_cluster_configurations() {
     echo "${cmdargs[@]}"
 
     if [ ! -z "$outputdircc" ]; then
-       if [[ $RESULT -eq 0 ]]; then
-         # XXX fixup plan, hard code cluster class
-         cat "$t" | perl -pe 's/--plan (\S+)/--plan $1cc/; s/_PLAN: (\S+)/_PLAN: $1cc/' > /tmp/test_tkg_config_cc
-         echo "CLUSTER_CLASS: tkg-${infra}-default" >> /tmp/test_tkg_config_cc
-         read -r -a cmdargs < <(grep EXE: /tmp/test_tkg_config_cc | cut -d: -f2-)
-         echo $TKG --file /tmp/test_tkg_config_cc --configdir ${TKG_CONFIG_DIR} --log_file /tmp/"$t"_cc.log config cluster "${cmdargs[@]}"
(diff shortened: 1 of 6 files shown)


pkg/v1/providers/tests/clustergen/run_tests.sh
```diff
@@ -77,6 +77,8 @@ generate_cluster_configurations() {
...
          echo "CLUSTER_CLASS: tkg-${infra}-default" >> /tmp/test_tkg_config_cc
          read -r -a cmdargs < <(grep EXE: /tmp/test_tkg_config_cc | cut -d: -f2-)
+         TKR_DATA="  v1.21.2:\n    kubernetesSpec:\n      version: v1.21.2\n      imageRepository: projects-stg.registry.vmware.com\n      etcd:\n        imageTag: v1.0.0-test\n      coredns:\n        imageTag: v1.1.0-test\n      kube-vip:\n        imageTag: v2.0.0-test\n    labels:\n      os-name: ubuntu\n      os-type: linux\n      os-arch: amd64\n  v1.23.5+vmware.1:\n    kubernetesSpec:\n      version: v1.23.5+vmware.1\n      imageRepository: projects-stg.registry.vmware.com\n      etcd:\n        imageTag: v1.0.0-test\n      coredns:\n        imageTag: v1.1.0-test\n      kube-vip:\n        imageTag: v2.0.0-test\n    labels:\n      os-name: ubuntu\n      os-type: linux\n      os-arch: amd64\n    osImageRef:\n      id: test-ami-id\n      region: test-region\n      sku: test-sku\n      publisher: test-publisher\n      offer: test-offer\n      version: test-version\n      thirdPartyImage: test-third-party-image"
```
vuil, line 80: this is pretty hard to read. can we use heredoc instead?
Also, does this need to be done differently for different infras?

tenczar: Added a heredoc. The data can be the same because it is a frankenstein monster of all the data for all the providers. Each provider pulls out the data that they require.

saji-pivotal (approved):

anujc25 (approved):

vuil (commented):
lgtm, thanks.

vuil (approved):

PR #2417, opened 2022-05-19
Enable webhook tls tertificate management

### LINKED ISSUE #1912: Webhook certificate should be mounted as a secret volume
Webhook certificates should also be mounted as a secret so that when rotated the updated tls.crt and tls.key file are remounted. 

Controller-runtime takes care of watching and reloading using https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/certwatcher/certwatcher.go

Used by webhook server https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/server.go#L219

Done criteria
1. Update addons-manager deployment to mount secret volume
2. Verify that an updated certificate will be reloaded by the http server by posting verification details

adduarte:
fixes: 245

PR enables management of webhook tls certificates. 

certificates are watched for expiration and other validation and rotated when applicable. 
a secret is created  which contains the certificates. 

Fixes #1912 

On a management cluster deployed on vsphere platform. 
Compared the contents of webhook-tls secret ca-cert to the contents of the cabundle for the mutating, and validating webhooks.  Verified they were the same.

ssh to supervisor vms which had access to the ip exposed by the tanzu-addons-manager-webhook-service  in the vmware-system-tkg namespace. 
Then used "openssl s_client" command to verify the sertificates being serve by the webhook server. 
Used the ca.pem obtain from webhook-tls secret to validate.  (openssl s_client -CAfile ca-cert.pem)

use openssl s_client --showcerts to verify contents of certificate being provided by the addons-manager.

### Release note

```release-note

```

### PR Checklist

### DIFF
--- a/addons/controllers/certificates_test.go
+++ b/addons/controllers/certificates_test.go
@@ -30,7 +30,8 @@ var _ = Describe("Webhook", func() {
 
 	Context("server's certificate and key", func() {
 		It("should be generated and written to the webhook server CertDir", func() {
-			secret, err := webhooks.NewTLSSecret(ctx, webhookScrtName, webhookServiceName, certPath, keyPath, addonNamespace)
+			print(certPath, keyPath, webhookScrtName, addonNamespace, webhookServiceName, webhookSelectorString)
+			secret, err := webhooks.InstallNewCertificates(ctx, k8sConfig, certPath, keyPath, webhookScrtName, addonNamespace, webhookServiceName, webhookSelectorString)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(secret).NotTo(BeNil())
 			cert, err := cert2.CertsFromFile(certPath)
--- a/addons/main.go
+++ b/addons/main.go
@@ -305,8 +305,20 @@ func enableClusterBootstrapAndConfigControllers(ctx context.Context, mgr ctrl.Ma
 func enableWebhooks(ctx context.Context, mgr ctrl.Manager, flags *addonFlags) {
 	certPath := path.Join(constants.WebhookCertDir, "tls.crt")
 	keyPath := path.Join(constants.WebhookCertDir, "tls.key")
-	if _, err := webhooks.InstallNewCertificates(ctx, mgr.GetConfig(), certPath, keyPath, constants.WebhookScrtName, flags.addonNamespace, constants.WebhookServiceName, constants.AddonWebhookLabelKey+"="+constants.AddonWebhookLabelValue); err != nil {
-		setupLog.Error(err, "unable to install certificates for webhooks")
+	webhookTLS := webhooks.WebhookTLS{
+		Ctx:           ctx,
+		K8sConfig:     mgr.GetConfig(),
+		CertPath:      certPath,
+		KeyPath:       keyPath,
+		Name:          constants.WebhookScrtName,
+		ServiceName:   constants.WebhookServiceName,
+		LabelSelector: constants.AddonWebhookLabelKey + "=" + constants.AddonWebhookLabelValue,
+		Logger:        setupLog,
+		Namespace:     flags.addonNamespace,
+		RotationTime:  constants.WebhookCertLifeTime,
+	}
+	if err := webhookTLS.ManageCertificates(constants.WebhookCertManagementFrequency); err != nil {
+		setupLog.Error(err, "Unable to start webhook tls certificate management")
 		os.Exit(1)
 	}
 	// Set up the webhooks in the manager
--- a/addons/pkg/constants/constants.go
+++ b/addons/pkg/constants/constants.go
@@ -178,6 +178,12 @@ const (
 	// WebhookCertDir is the directory where the certificate and key are stored for webhook server TLS handshake
 	WebhookCertDir = "/tmp/k8s-webhook-server/serving-certs"
 
+	// WebhookCertManagementFrequency is how often the the certificates for webhook server TLS are managed
+	WebhookCertManagementFrequency = time.Second * 60
+
+	// WebhookCertLifeTime is how long the webhook server TLS certificates are good for
+	WebhookCertLifeTime = time.Hour * 24 * 7
+
 	// WebhookServiceName is the name of the k8s service that serves the admission requests
 	WebhookServiceName = "tanzu-addons-manager-webhook-service"
 
--- a/pkg/v1/webhooks/certificates.go
+++ b/pkg/v1/webhooks/certificates.go
@@ -63,52 +63,65 @@ func addCertsToWebhookConfigs(ctx context.Context, k8sclient kubernetes.Interfac
 	return nil
 }
 
-func NewTLSSecret(ctx context.Context, secretName, serviceName, certPath, keyPath, namespace string) (*corev1.Secret, error) {
-	secret, err := resources.MakeSecretInternal(ctx, secretName, namespace, serviceName)
-	if err != nil {
-		return nil, err
-	}
-
+// WriteServerTLSToFileSystem writes servers certificate and key in provided secret to the filesystem paths provided.
+func WriteServerTLSToFileSystem(ctx context.Context, certPath, keyPath string, secret *corev1.Secret) error {
 	if err := certutil.WriteCert(certPath, secret.Data[resources.ServerCert]); err != nil {
-		return secret, err
+		return err
 	}
 	if err := keyutil.WriteKey(keyPath, secret.Data[resources.ServerKey]); err != nil {
-		return secret, err
+		return err
 	}
-
-	return secret, nil
+	return nil
 }
 
+// InstallNewCertificates creates a new set of keys and certificates and saves them to the filesystem paths provided.
+// Adds the CA certificate to webhook configurations matching label selector.
+// Returns a secret containing the server key, sever certificate and CA certificate.
 func InstallNewCertificates(ctx context.Context, k8sConfig *rest.Config, certPath, keyPath, secretName, namespace, serviceName, labelSelector string) (*corev1.Secret, error) {
 	if labelSelector == "" {
 		return nil, fmt.Errorf("label selector not provided for webhook configurations udpate")
 	}
-	secret, err := NewTLSSecret(ctx, secretName, serviceName, certPath, keyPath, namespace)
+	secret, err := resources.MakeSecret(ctx, secretName, namespace, serviceName)
 	if err != nil {
 		return nil, err
 	}
 
-	k8sclient, err := client.New(k8sConfig, client.Options{})
+	err = InstallCertificates(ctx, k8sConfig, secret, certPath, keyPath, labelSelector)
 	if err != nil {
 		return nil, err
 	}
+	return secret, nil
+}
+
+// InstallCertificates saves server certificate and key in provided secret to the filesystem paths provided.
+// Adds the CA certificate to webhook configuration matching label selector.
+func InstallCertificates(ctx context.Context, k8sConfig *rest.Config, secret *corev1.Secret, certPath, keyPath, labelSelector string) error {
+	if err := WriteServerTLSToFileSystem(ctx, certPath, keyPath, secret); err != nil {
+		return err
+	}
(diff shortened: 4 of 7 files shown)


pkg/v1/webhooks/certificates.go
```diff
@@ -151,6 +151,7 @@ func updateOrCreateTLSSecret(ctx context.Context, k8sclient client.Client, tlsSe
 		if err != nil {
 			return err
 		}
+		return nil
```
adduarte, line 154: no need to continue if we got a not found, and successfully created the secret. We can return here.

addons/main.go
```diff
@@ -303,12 +303,29 @@ func enableClusterBootstrapAndConfigControllers(ctx context.Context, mgr ctrl.Ma
 }
 
 func enableWebhooks(ctx context.Context, mgr ctrl.Manager, flags *addonFlags) {
+	managementFrequency := time.Second * 60 // This is how often we manage the webhook tls. This is how often we check certificates and corret if invalid.
```
adduarte, line 306: This could be passed along from one of the addon-manager flags

addons/main.go
```diff
@@ -303,12 +303,29 @@ func enableClusterBootstrapAndConfigControllers(ctx context.Context, mgr ctrl.Ma
...
+		CertPath:      certPath,
+		KeyPath:       keyPath,
+		Name:          constants.WebhookScrtName,
+		ServiceName:   constants.WebhookServiceName,
+		LabelSelector: constants.AddonWebhookLabelKey + "=" + constants.AddonWebhookLabelValue,
+		Logger:        setupLog,
+		Namespace:     flags.addonNamespace,
+		RotationTime:  time.Hour * 24 * 7,
+	}
+	if err := webhookTLS.ManageCertificates(managementFrequency); err != nil {
```
shyaamsn, line 325: What happens if addons-manager restarts ? In the webhook code I don't see the certificate being written to the local filesystem on addons-manager restart.

https://github.com/vmware-tanzu/tanzu-framework/blob/main/pkg/v1/webhooks/webhooks.go#L65

In the case of restart the certificate is valid but its not written to local filesystem. Am i missing something ?

adduarte: on addons-manager start the certificates are installed for the first time in line [308 above ](https://github.com/vmware-tanzu/tanzu-framework/pull/2417/files#diff-b162de15343c757b9843525404e1b18de3623219dee9197c8eca4b6c07fe2de7R308) (call to if _, err := webhooks.InstallNewCertificates..)

everytime the addons-controller binary is restarted new certificates are generated and written to the filesystem. 

@shyaamsn is that what you mean by addons-manager restart?

shyaamsn: Ah ok got it. Is it a good practice to regenerate certificate every time addons-controller restarts ?

addons/main.go
adduarte: decision was made to use a constant defined in addons/pkg/constants

addons/main.go
adduarte: It seems the least path of resistance.  In order to validate existing tls secrets from a previous "existance" of an addons-controller we would have to keep track somehow of the status of the addons-controller.. 
And even if we are able to keep track of state (restart, first start, crash, etc..) 
we have no way of assuring the tls certifictes are valid.  or that all the configured webhooks obtained the correct cabundle. 
for example, if the addons-controller was restarted by a crash right in the middle of updating the webhook configurations, we could end up with somewebhook configurations with a valid ca and some with an invalid. 
that situation would not correct itself until the existing (left over) tls certificates become invalid (time expiration). 

The simplest approach, although a bit brute force, is to assume tls certificates are invalid on addons-controller start, and replace with new ones.

but with that said, I think is worth looking at making the certificates in the webhook-tls secret be the source of true. 

we could validate the content of webhook-tls everytime we start and addons-controller. then write them to file if they are valid.

adduarte: refactored code to better handle restart. The new version, looks at the existing webhook tls secret and validatest the certificates there. If valid it will use them and install it (add to tls.crt, key.crt, and webhook configurations)
If not valid, new certificates will be genreated and installed

pkg/v1/webhooks/certificates.go
```diff
@@ -63,50 +63,64 @@ func addCertsToWebhookConfigs(ctx context.Context, k8sclient kubernetes.Interfac
...
+// func NewTLSSecret(ctx context.Context, secretName, serviceName, certPath, keyPath, namespace string) (*corev1.Secret, error) {
+//	secret, err := resources.MakeSecret(ctx, secretName, namespace, serviceName)
+//	if err != nil {
+//		return nil, err
+//	}
+
+//	return secret, nil
+//}
 
+func WriteServerTLSToFileSystem(ctx context.Context, certPath, keyPath string, secret *corev1.Secret) error {
```
rajathagasthya, line 75: Probably a naive question: why do we need to write the cert to a secret in addition to writing to file system? IIRC, cert-manager creates a secret which you can mount as volume to the controller pod. But we're in full control of cert generation here and we're already writing it to file system (which controller-runtime keeps a watch on).

adduarte: You are correct, we could just keep it in memory and write to the file system, (which btw was our original implementation during our POC) but 
the filesystem is ephemeral, so if the kapp-controller is restarted the certificates would be lost. 
we write the certificates to the secret so we can use the same certificate continously accross restarts. 
and also so it is available from the outside world. 
Theoretically (although not supported) you can upload your own certificates to the secret.

rajathagasthya: Ah, that makes sense. Thanks!

pkg/v1/webhooks/certificates.go
```diff
@@ -63,50 +63,64 @@ func addCertsToWebhookConfigs(ctx context.Context, k8sclient kubernetes.Interfac
...
-		return nil, err
-	}
+// func NewTLSSecret(ctx context.Context, secretName, serviceName, certPath, keyPath, namespace string) (*corev1.Secret, error) {
+//	secret, err := resources.MakeSecret(ctx, secretName, namespace, serviceName)
+//	if err != nil {
+//		return nil, err
+//	}
+
+//	return secret, nil
+//}
```
rajathagasthya, line 74: Could you add godoc comments to all these exported functions please? I'm not sure why the linter is not catching it.

rajathagasthya (commented):
Overall LGTM, just some nits. Will approve once those are addressed.

addons/pkg/constants/constants.go
```diff
@@ -178,6 +178,12 @@ const (
 	// WebhookCertDir is the directory where the certificate and key are stored for webhook server TLS handshake
 	WebhookCertDir = "/tmp/k8s-webhook-server/serving-certs"
 
+	// WebhookCertManagementFrequency is how often the the certificates for webhook server TLS are managed
+	WebhookCertManagmentFrequency = time.Second * 60
```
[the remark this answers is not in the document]
adduarte: done

pkg/v1/webhooks/certificates.go
adduarte: done!

pkg/v1/webhooks/certificates.go
```diff
@@ -63,50 +63,64 @@ func addCertsToWebhookConfigs(ctx context.Context, k8sclient kubernetes.Interfac
 	return nil
 }
 
-func NewTLSSecret(ctx context.Context, secretName, serviceName, certPath, keyPath, namespace string) (*corev1.Secret, error) {
-	secret, err := resources.MakeSecretInternal(ctx, secretName, namespace, serviceName)
-	if err != nil {
-		return nil, err
-	}
+// func NewTLSSecret(ctx context.Context, secretName, serviceName, certPath, keyPath, namespace string) (*corev1.Secret, error) {
```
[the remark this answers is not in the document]
adduarte: done!

pkg/v1/webhooks/certificates.go
```diff
@@ -63,53 +63,70 @@ func addCertsToWebhookConfigs(ctx context.Context, k8sclient kubernetes.Interfac
...
+	// Creates a new set of keys and certificates and saves them to the filesystem paths provided.
+	// Adds the CA certificate to webhook configurations matching label selector.
+	// Returns a secret containing the server key, sever certificate and CA certificate.
+
 	if labelSelector == "" {
 		return nil, fmt.Errorf("label selector not provided for webhook configurations udpate")
 	}
-	secret, err := NewTLSSecret(ctx, secretName, serviceName, certPath, keyPath, namespace)
+	secret, err := resources.MakeSecret(ctx, secretName, namespace, serviceName)
 	if err != nil {
```
shyaamsn, line 87: Why is this lines not covered as part of test ? Seems like this is important functionality that we want to test

pkg/v1/webhooks/certificates.go
```diff
@@ -63,53 +63,70 @@ func addCertsToWebhookConfigs(ctx context.Context, k8sclient kubernetes.Interfac
...
 		return nil, err
 	}
 
-	k8sclient, err := client.New(k8sConfig, client.Options{})
+	err = InstallCertificates(ctx, k8sConfig, secret, certPath, keyPath, labelSelector)
 	if err != nil {
```
shyaamsn, line 92: Why are these lines not covered as part of test ? Seems like this is important functionality that we want to test

addons/controllers/certificates_test.go
```diff
@@ -30,7 +30,8 @@ var _ = Describe("Webhook", func() {
 
 	Context("server's certificate and key", func() {
 		It("should be generated and written to the webhook server CertDir", func() {
-			secret, err := webhooks.NewTLSSecret(ctx, webhookScrtName, webhookServiceName, certPath, keyPath, addonNamespace)
+			print(certPath, keyPath, webhookScrtName, addonNamespace, webhookServiceName, webhookSelectorString)
```
adduarte, line 33: this is needs to be removed

pkg/v1/webhooks/certificates.go
adduarte: The calling function requires a cluster to be running.
The lines are actually covers in testing here: https://github.com/vmware-tanzu/tanzu-framework/pull/2417/files#diff-447441730f38027eb4773a493cd8ed684fa7ef1a7766f8f0f8d7184f41f3be3aR34

pkg/v1/webhooks/certificates.go
adduarte: The calling function requires a cluster to be running.
The lines are actually covers in testing here: https://github.com/vmware-tanzu/tanzu-framework/pull/2417/files#diff-447441730f38027eb4773a493cd8ed684fa7ef1a7766f8f0f8d7184f41f3be3aR34

pkg/v1/webhooks/certificates.go
```diff
@@ -63,53 +63,70 @@ func addCertsToWebhookConfigs(ctx context.Context, k8sclient kubernetes.Interfac
 	return nil
 }
 
-func NewTLSSecret(ctx context.Context, secretName, serviceName, certPath, keyPath, namespace string) (*corev1.Secret, error) {
-	secret, err := resources.MakeSecretInternal(ctx, secretName, namespace, serviceName)
-	if err != nil {
-		return nil, err
-	}
+func WriteServerTLSToFileSystem(ctx context.Context, certPath, keyPath string, secret *corev1.Secret) error {
+	// Writes servers certificate and key in provided secret to the filesystem paths provided
```
rajathagasthya, line 67: Godoc needs to be before the function signature to be rendered properly in https://pkg.go.dev/github.com/vmware-tanzu/tanzu-framework. Should be of the form:

```go
// FunctionFoo does something.
func FunctionFoo() {}
```

Example: https://github.com/vmware-tanzu/tanzu-framework/blob/db9d5fb7dfbbbca8667bcc9ab0d16b0874d7057f/pkg/v1/sdk/features/featuregate/featuregate.go#L96-L97

adduarte: done

rajathagasthya (approved):

PR #2423, opened 2022-05-19
Update Azure TKR image data processing

tenczar:
Update templates that set Azure image data from TKR_DATA.

Azure image data can come either in the form of a marketplace image or a sharedGallery image. This change updates the Azure clusterclass to support either of these image formats. Also updates the clustergen test to provide specific TKR_DATA fakes for each provider. Additionally it switches fake TKR_DATA for Azure between marketplace and sharedGallery depending no the provided config file.

Fixes #

Checked clustergen diffs to ensure that the marketplace and sharedGallery images are being generated correctly. Also made sure that the vsphere and aws versions haven't changed because of the change to the fake TKR_DATA.

### Release note

```release-note

```

### PR Checklist

### DIFF
--- a/pkg/v1/providers/tests/clustergen/run_tests.sh
+++ b/pkg/v1/providers/tests/clustergen/run_tests.sh
@@ -77,22 +77,10 @@ generate_cluster_configurations() {
         cat "$t" | perl -pe 's/--plan (\S+)/--plan $1cc/; s/_PLAN: (\S+)/_PLAN: $1cc/' > /tmp/test_tkg_config_cc
         echo "CLUSTER_CLASS: tkg-${infra}-default" >> /tmp/test_tkg_config_cc
         read -r -a cmdargs < <(grep EXE: /tmp/test_tkg_config_cc | cut -d: -f2-)
-        cat <<- EOF >> /tmp/test_tkg_config_cc
+
+        if [[ "${infra}" == "aws" ]]; then
+          cat <<- EOF >> /tmp/test_tkg_config_cc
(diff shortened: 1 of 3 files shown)


pkg/v1/providers/tests/clustergen/run_tests.sh
```diff
@@ -115,6 +115,10 @@ TKR_DATA: |-
       offer: test-offer
       version: test-version
       thirdPartyImage: test-third-party-image
+      gallery: test-gallery
```
vuil, line 118: Since we are essentially hacking in locally a TKR_DATA that really would be been server-injected, I would suggest setting it up in a way that is representative of what we can expect in reality. Since we don't particularly care which test case is fed which kind of azure image metadata, something simple to enhance this hack with is to use SharedGallery info for half the testcases's TKR_DATA and use Marketplace info the rest. wdyt?

vuil:
Thanks for the adding the TKR_DATA conditionally.
@tenczar 
PTAL at  https://storage.googleapis.com/tkg-clustergen/2423/20220524000720/clustergen_cc.diff.txt
I am seeing a couple of test cases (06364262.case, 27925943.case, could be more) that now fails generation after the change

pkg/v1/providers/tests/clustergen/run_tests.sh
tenczar: Updated to break out cases for each provider and special case the sharedGallery image.

vuil (approved):
lgtm, thanks for the updates!

PR #2425, opened 2022-05-20
Prevent finalizer from being added to management cluster

### LINKED ISSUE #2426: Management clusters should not have addons finalizers added to them

adduarte:
Management clusters should not have addons finalizers adde to them. 
If the finalizers is present, it will cause problems for the cluster lifecycle. 
For example, when kind clean up clusters is used to delete the management cluster, the clean up cluster does not 
have kapp controller running (or addons-manager) so there is nothing to remove the finalizer from it. 

This pr prevents finalizers from being added to clusters which are labeled as management clusters.

Fixes #2426

make test. 

### Release note

```release-note
addons finalizers will not be added to clusters labeled as management clusters
```

### PR Checklist

### DIFF
--- a/addons/controllers/clusterbootstrap_controller.go
+++ b/addons/controllers/clusterbootstrap_controller.go
@@ -279,10 +279,13 @@ func (r *ClusterBootstrapReconciler) createOrPatchResourcesForAdditionalPackages
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


danniel1205 (approved):
LGTM. Thanks for the PR.

maralavi (approved):

PR #2494, opened 2022-05-26
Create aggregate cluster role for CAPV

### LINKED ISSUE #2367: CPI and CSI config controller should add RBAC rules required for CAPV Provider Service Account to use
**Bug description**
CAPV creates a Role (with the permissions listed above), a ServiceAccount and binds the two using a RoleBinding to a particular namespace. CAPV controller manager ClustreRole needs to have the permissions specified in the ProviderServiceAccount.spec.Rules before it is able to create that Role object. Hence on startup, (ideally) the tanzu addons manager should create a ClusterRole with the label [capv.infrastucture.cluster.x-k8s.io/aggregate-to-manager](http://capv.infrastucture.cluster.x-k8s.io/aggregate-to-manager): "true" and the necessary permissions.
That should enable the reconciliation of the ProviderSvcAccount successfully.

- [ ] APIs
- [x] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**

vijaykatam:
Create aggregated cluster role that will be inherited by CAPV. CAPV needs to hold these rules before it can grant it to serviceAccount for CPI and CSI.

Create aggregated cluster role that will be inherited by CAPV. 

Fixes #2367 

Env tests.

### Release note

```release-note
package-based-lcm: Create aggregate cluster role for CAPV
```

### PR Checklist

### Additional information
Relates to https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1529

### DIFF
--- a/addons/controllers/csi/vspherecsiconfig_controller.go
+++ b/addons/controllers/csi/vspherecsiconfig_controller.go
@@ -10,8 +10,9 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
-	yaml "gopkg.in/yaml.v3"
+	"gopkg.in/yaml.v3"
 	v1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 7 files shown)


addons/controllers/csi/vspherecsiconfig_controller.go
```diff
@@ -39,6 +40,39 @@ type VSphereCSIConfigReconciler struct {
 	Scheme *runtime.Scheme
 }
 
+var providerServiceAccountRBACRules = []rbacv1.PolicyRule{
```
danniel1205, line 43: nit(non-blocking): Was thinking what if we put this block inside of `vspherecpiconfig_utils.go` ? But defining it here is totally fine.

addons/controllers/csi/vspherecsiconfig_utils.go
```diff
@@ -172,39 +171,8 @@ func (r *VSphereCSIConfigReconciler) mapCSIConfigToProviderServiceAccount(vspher
...
-				},
-				{
-					APIGroups: []string{""},
-					Resources: []string{"events"},
-					Verbs:     []string{"list"},
-				},
-			},
+			Ref:              &v1.ObjectReference{Name: vsphereCluster.Name, Namespace: vsphereCluster.Namespace},
+			Rules:            providerServiceAccountRBACRules,
 			TargetNamespace:  "vmware-system-csi",
```
danniel1205, line 176: question(non-blocking): This is orthogonal, but I guess using a constant variable is better than hardcoding. We could follow up on a separate transaction.

danniel1205 (approved):

addons/controllers/csi/vspherecsiconfig_controller.go
vijaykatam: I added it here because originally I wanted to create the clusterole as part of controller init but ran into issues determining if TKGS or not so the create is in the reconcile loop now.

addons/controllers/csi/vspherecsiconfig_utils.go
vijaykatam: yup the csi controller needs some refactoring

PR #2524, opened 2022-05-31
TKR Source Controller: Install TKR package contents directly

imikushin:
#### Install TKR package contents directly
TKR package contents get automatically installed without the use of Carvel PackageInstall resource (which doesn't support shared resources across packages). This is necessary because a TKR package may contain resources that are reused by other TKR packages, e.g. different TKRs shipping the same Kubernetes version may share some OSImages or addon packages (addon packages may even be reused across K8s versions).

#### Inject Registry as an interface into components that need it
This improves testability and modularity of affected components, and reuse of Registry as a component. Now, both Fetcher and TKR Package Reconciler re-use the same Registry instance provided externally.

#### Only fetch compatible TKR BOMs and TKR packages
We're also now using compatibility information to download only compatible TKRs into the management cluster.

#### PatchSet improvements
We're making PatchSet more generic and efficient.

Before this change, PatchSet would not work with ConfigMaps. We're also removing a dependency on util/patch from Cluster API.

PatchSet is also now safe for concurrent use.

N/A

- Added/updated unit tests.
- Manual testing with an AWS management cluster.

### Release note

```release-note
Fixed TKR Source Controller (pkg/v2/tkr/controller/tkr-source): it now installs 
TKR package contents.

Changed TKR Source Controller (pkg/v2/tkr/controller/tkr-source): only compatible TKR BOMs and TKR packages are being fetched now.

Improved util/patchset.PatchSet: it is now safe for concurrent use and more efficient.
```

### PR Checklist


pkg/v2/tkr/controller/tkr-source/fetcher/fetcher.go
```diff
@@ -305,24 +305,38 @@ func (f *Fetcher) fetchTKRPackages(ctx context.Context) error {
...
+	for tag := range imageTagsToPull {
 		if err := f.createTKRPackages(ctx, tag); err != nil {
 			errs = append(errs, errors.Wrapf(err, "failed to create TKR Package for image %s", fmt.Sprintf("%s:%s", f.Config.TKRRepoImagePath, tag)))
 		}
 	}
-	if len(errs) != 0 {
-		return errs
-	}
 
 	f.Log.Info("Done fetching TKR Packages", "image", f.Config.TKRRepoImagePath)
```
prkalle, line 328: nit: This log in case of errors would be misleading?

prkalle (approved):
LGTM. Thanks @imikushin

PR #2584, opened 2022-06-08
Do not update certificates on filesystem if content is identical.

### LINKED ISSUE #2570: Webhook management causes cuntroller-runtime to reload certs too often
controller-runtime reloads TLS certificates everytime the webhook 
certificates are managed. Currently that is every 1 minute, and this is causing connection drops
and possibly interrupting cluster deployment. 

This is because when installing new certificates, even if the file contents
are identical, it will retrigger a a reload of TLS certificates
by the controller-runtime. This causes the controller-runtime
to reload TLS certificates everytime the webhook tls are managed.

adduarte:
When installing new certificates, even if the file contents
are identical, it will retrigger a a reload of TLS certificates
by the controller-runtime. This causes the controller-runtime
to reload TLS certificates everytime the webhook tls are managed.

This pr fixes that by only writing the certificates to the filesystem
if the contents of the file (certificates) have actually changed.

This pr also increases the management frequency from 1 minute to
90 minutes

Fixes #2570

local unit test 
deployed aws cluster and pointed addons-manager to it. then manually removed tls.crt and tls.key files and observed the files succesfully be re-written, repeated same test but instead of removing file, changed contents of file and observed files being re-written. 
Also observed that if files where unchanged, they were NOT rewritten at management time. 
 

### Release note

```release-note
package-based-lcm: Fix constant reloading of certificates of controller-runtime by only changing certs on filesystem if contents have changed.
```

### PR Checklist


vijaykatam (commented):
Minor comments. Could you update issue with more details? It is completely empty.

pkg/v1/webhooks/certificates.go
```diff
@@ -74,6 +84,39 @@ func WriteServerTLSToFileSystem(ctx context.Context, certPath, keyPath string, s
...
+	_, err := os.Stat(certPath)
+	if err != nil && !os.IsNotExist(err) {
+		return false, err
+	}
+	if os.IsNotExist(err) {
+		return false, nil
+	}
+
+	_, err = os.Stat(keyPath)
+	if os.IsNotExist(err) {
```
vijaykatam, line 97: Same comment applies here.

vijaykatam (approved):
LGTM. Rebase for CI fix.

PR #2598, opened 2022-06-09
clusterboostrap controller should ignore clusters without a CBT

### LINKED ISSUE #2546: UTKG Addons manager needs to only process TKRs which do NOT have "run.tanzu.vmware.com/legacy-tkr" label
**Bug description**
Currently, based on https://github.com/vmware-tanzu/tanzu-framework/issues/1674, utkg addons controller only processes clusters with label run.tanzu.vmware.com/tkr. However, additionally we need to also check that the related TKR does *not* have label "run.tanzu.vmware.com/legacy-tkr"

- [ ] APIs
- [X] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**

adduarte:
Some type of clusters, like Legacy tkr clusters, do not have
CB template object associated with them. These clusters
should not be processed by the clusterboostrap controller

Fixes #2546 

unit test executed locally

### Release note

```release-note
cluters without an associated clusterboostraptemplate are not reconciled by the 
clusterboostrap_controller
```

### PR Checklist

### DIFF
--- a/addons/controllers/clusterbootstrap_controller.go
+++ b/addons/controllers/clusterbootstrap_controller.go
@@ -364,10 +364,13 @@ func (r *ClusterBootstrapReconciler) createOrPatchClusterBootstrapFromTemplate(c
 	clusterBootstrapTemplate := &runtanzuv1alpha3.ClusterBootstrapTemplate{}
 	key := client.ObjectKey{Namespace: r.Config.SystemNamespace, Name: tkrName}
 	if err := r.Client.Get(r.context, key, clusterBootstrapTemplate); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("no ClusterBootstrapTemplate associated with cluster")
+			return nil, nil
+		}
 		log.Error(err, "unable to fetch ClusterBootstrapTemplate", "objectkey", key)
 		return nil, err
 	}
-
 	clusterBootstrap := &runtanzuv1alpha3.ClusterBootstrap{}
 	err := r.Client.Get(r.context, client.ObjectKeyFromObject(cluster), clusterBootstrap)
 	if err != nil && !apierrors.IsNotFound(err) {
--- a/addons/controllers/clusterbootstrap_controller_test.go
+++ b/addons/controllers/clusterbootstrap_controller_test.go
@@ -922,6 +922,22 @@ var _ = Describe("ClusterBootstrap Reconciler", func() {
 		})
 	})
 
+	When("Legacy cluster is created", func() {
+		BeforeEach(func() {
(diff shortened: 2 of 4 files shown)


addons/controllers/clusterbootstrap_controller.go
```diff
@@ -364,10 +364,13 @@ func (r *ClusterBootstrapReconciler) createOrPatchClusterBootstrapFromTemplate(c
 	clusterBootstrapTemplate := &runtanzuv1alpha3.ClusterBootstrapTemplate{}
 	key := client.ObjectKey{Namespace: r.Config.SystemNamespace, Name: tkrName}
 	if err := r.Client.Get(r.context, key, clusterBootstrapTemplate); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("no ClusterBootstrapTemplate associated with cluster")
```
shyaamsn, line 368: **suggestion**: Can we log cluster name and namespace.

adduarte: I believe they are logged on already, but, I will check.

shyaamsn: Ah ok. Good then

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -364,10 +364,13 @@ func (r *ClusterBootstrapReconciler) createOrPatchClusterBootstrapFromTemplate(c
...
+			return nil, nil
```
shyaamsn, line 369: Could we add a testcase for this please. Thanks

addons/controllers/clusterbootstrap_controller.go
adduarte: Yes, the cluster name and namespace are added to the logs here: https://github.com/vmware-tanzu/tanzu-framework/blob/50c237ec19e948f4999679ec44d5eb6b8ca407dd/addons/controllers/clusterbootstrap_controller.go#L141

vijaykatam: I think this is an error. If there is a classy cluster it must have a CBT unless we want to filter out TKGS legacy-tkr which we should do explicitly instead of ignoring all errors.

adduarte:
proposing different solution here: https://github.com/vmware-tanzu/tanzu-framework/pull/2624

adduarte:
Preferable solution on https://github.com/vmware-tanzu/tanzu-framework/pull/2624

PR #2610, opened 2022-06-10
WIP: DNC : Test cgen build

vuil:
test cgen build


alfredthenarwhal:
> @vuil:  /test install-vc7
Commit: c1ab013

Tests passed! Build no: 2151

PR #2640, opened 2022-06-13
Ensure a non-empty array is present for preKubeAdmCommands

saji-pivotal:
Ensure a non-empty array is present for preKubeadmCommands in the KubeAdmControlPlaneTemplate to facilitate patching

Fixes #

### Release note

```release-note

```

### PR Checklist

### DIFF
--- a/packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml
+++ b/packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml
@@ -1027,7 +1027,9 @@ spec:
     spec:
       kubeadmConfigSpec:
         files: []
-        preKubeadmCommands: []
+        preKubeadmCommands:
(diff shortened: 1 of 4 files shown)


packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml
```diff
@@ -1027,7 +1027,9 @@ spec:
     spec:
       kubeadmConfigSpec:
         files: []
-        preKubeadmCommands: []
+        preKubeadmCommands:
+        # Ensure a non-empty preKubeadmCommands array is present to facilitate patching
```
anujc25, line 1031: In all other places we are using `#!` to avoid ytt validation errors. We should use the same here as well.
```suggestion
        #! Ensure a non-empty preKubeadmCommands array is present to facilitate patching
```

anujc25 (approved):
Looks good

PR #2667, opened 2022-06-15
UI/CLI: Set AviControlPlaneEndpointProvider false by default

miclettej:
Resolves issue when configuring vsphere management cluster via UI, using default
Kube VIP would result in AviControlPlaneEndpointProvider never getting set to a
boolean value. This caused AVI_CONTROL_PLANE_HA_PROVIDER value to be written as a
blank string in the config yaml.

The fix is to add omitempty as a catch-all, but also to intialize AviControlPlaneEndpointProvider
to a boolean false value by default.

Fixes #

confirmed config yaml AVI_CONTROL_PLANE_HA_PROVIDER value is written as false by default

### Release note

```release-note
none
```

### PR Checklist


vuil (approved):
lgtm, thanks!

PR #2653, opened 2022-06-14
Remove use of deprecated ioutil module

stmcginnis:
### What this PR does / why we need it

The ioutil module was deprecated in go 1.16. There have been previous
PRs to clean up its usage to use the recommended os and io replacements,
but new instances have since snuck in.

To try to prevent this from happening repeatedly, this also adds a
simple check to the go-lint target to check for the presence of any
ioutil imports in any of the go source code.

### Describe testing done for PR

Ran `make go-lint` before updating the source files to make sure instances of `ioutils` were properly detected. Then updates the source and ran again to make sure the job passed successfully.

Ran `make build-cli` to make sure everything still compiled correctly.

PR #2669, opened 2022-06-15
Fix temp dir creation in client_test.go

stmcginnis:

Changes were recently made to remove the use of the deprecated ioutil
module. In `pkg/v1/tkg/tkgctl/client_test.go` there was a mistake made
switching from creating a temporary directory to creating a temporary
file. This corrects that error to use the proper replacement for the
original ioutil call.

Fixes #2653 

Running `make test` locally.


vuil (approved):
lgtm, thanks!

PR #2672, opened 2022-06-15
Remove omitEmpty from boolean values to ensure all the bool values get displayed regardless of true or false

### LINKED ISSUE #2663: Antrea secret shows only data values which are true, false data values are not displayed
**Bug description**
Antrea secret shows only data values which are true, false data values are not displayed
```
➜  tanzu-framework git:(package-based-lcm) ✗ kubectl get secret mc-jun10-cluster-antrea-data-values -n tkg-system -o jsonpath='{.data.values\.yaml}' | base64 -d
infraProvider: aws
antrea:
    config:
        serviceCIDR: 100.64.0.0/13
        trafficEncapMode: encap
        tlsCipherSuites: TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_256_GCM_SHA384
        featureGates:
            AntreaProxy: true
            EndpointSlice: true
            AntreaPolicy: true
            AntreaTraceflow: true
```            

- [x] APIs
- [x] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**
Behaviour shud be to display both true and false data values

shivaani0505:
Remove omitEmpty from boolean values to ensure all the bool values get displayed regardless of true or false

Fixes #2663 

### Release note

```release-note
package-based-lcm: fix issue for Marshal function from yaml module which treats false booleans as if they were nil values, and thus omits them from the resulted json for data values secret
```

### PR Checklist

### DIFF
--- a/addons/controllers/antrea/antreaconfig_util.go
+++ b/addons/controllers/antrea/antreaconfig_util.go
@@ -52,25 +52,25 @@ type antreaConfigDataValue struct {
 	ServiceCIDR             string              `yaml:"serviceCIDR,omitempty"`
 	ServiceCIDRv6           string              `yaml:"serviceCIDRv6,omitempty"`
 	TrafficEncapMode        string              `yaml:"trafficEncapMode,omitempty"`
-	NoSNAT                  bool                `yaml:"noSNAT,omitempty"`
-	DisableUDPTunnelOffload bool                `yaml:"disableUdpTunnelOffload,omitempty"`
+	NoSNAT                  bool                `yaml:"noSNAT"`
+	DisableUDPTunnelOffload bool                `yaml:"disableUdpTunnelOffload"`
 	DefaultMTU              string              `yaml:"defaultMTU,omitempty"`
 	TLSCipherSuites         string              `yaml:"tlsCipherSuites,omitempty"`
 	FeatureGates            antreaFeatureGates  `yaml:"featureGates,omitempty"`
 }
 
 type antreaFeatureGates struct {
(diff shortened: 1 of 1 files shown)


vijaykatam:
Add release note and reasoning on why we are removing `omitEmpty` would be good since convention is `omitEmpty`. We should also followup with a CRD sync in the package.

shivaani0505:
> Add release note and reasoning on why we are removing `omitEmpty` would be good since convention is `omitEmpty`. We should also followup with a CRD sync in the package.

@vijaykatam , added Release notes.
Reasoning: fix issue for Marshal function from yaml module which treats false booleans as if they were nil values, and thus omits them from the resulted json for data values secret

Do you also want me to update types file. If yes, then I will regenerate the CRDs

knabben:
would be nice to have some unit tests for these marshalling functions

vijaykatam (commented):
Change looks good. As discussed earlier today, check if we need to be consistent for  serializations in calico and other addons. If not, please fix them here or log issue(s).

vijaykatam (approved):

PR #2673, opened 2022-06-15
Remove redundant vet and fmt calls

stmcginnis:

This removes the `vet` target as that (and more) is covered by our
golangci-lint configuration. Running vet would add a large delay, so
this speeds things up for any targets that would call that.

A few dependencies on the `fmt` target were removed where `generate` was
also listed. `generate` calls `fmt` internally at the end of its
execution, so this was just causing `fmt` to be run twice in a row.

Also cleaned up an extra golangci-lint call in the `go-lint` target.
This was in-flight at the same time that the target was updated to
automatically detect all modules, so when it merged it created a second
run for the same module.

Ran `make test ENVS=darwin-amd64` locally and verified there were no errors. Further validation will be done as part of the GitHub Actions that will run when this PR is submitted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,6 @@ TOOLS_BIN_DIR := $(TOOLS_DIR)/bin
 BIN_DIR := bin
 ADDONS_DIR := addons
 YTT_TESTS_DIR := pkg/v1/providers/tests
-PACKAGE_TOOLING_DIR := hack/packages/package-tools
 PACKAGES_SCRIPTS_DIR := $(abspath hack/packages/scripts)
 UI_DIR := pkg/v1/tkg/web
 GO_MODULES=$(shell find . -path "*/go.mod" | grep -v "^./pinniped" | xargs -I _ dirname _)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -503,12 +501,6 @@ test: generate fmt vet manifests build-cli-mocks ## Run tests
 test-cli: build-cli-mocks ## Run tests
 	$(GO) test ./pkg/v1/cli/... ./pkg/v1/auth/... ./pkg/v1/builder/... ./pkg/v1/config/... ./pkg/v1/encoding/... ./pkg/v1/grpc/...
 
-fmt: tools ## Run goimports
-	$(GOIMPORTS) -w -local github.com/vmware-tanzu ./
```
rajathagasthya, line 507: We should keep this target as it's actually sorting (writing) imports and enforcing a `-local` rule. The linter only checks if imports are sorted. This is handy when linter complains and you can just run `make fmt` ([example](https://github.com/vmware-tanzu/tanzu-framework/pull/2530#issuecomment-1146858106)).

stmcginnis: Done!

rajathagasthya (approved):
Thanks!

PR #2693, opened 2022-06-17
Add tests for MHC on classy clusters

tenczar:
This PR adds MachineHealthCheck tests to the CC test suite. The tests exercise the new behavior of the machinehealthcheck API when using classy clusters. In particular, create, update, and delete are not supported for classy clusters and those those operations now return errors. Machine health checks must be modified at the clusterclass level for classy clusters. This change also updates the tanzu cli to properly retrieve machine health checks for nodes in either legacy clusters or classy clusters.

Fixes #

E2E tests retrieve machine health checks and then verifies errors on set/delete operations. The test also causes a node failure state and waits for the cluster remediation to finish.

### Release note

```release-note
Modifying machinehealthchecks on classy clusters is not supported. The API will return an error when performing these operations on classy clusters.
```

### PR Checklist

### DIFF
--- a/pkg/v1/tkg/test/tkgctl/shared/mhc_cc.go
+++ b/pkg/v1/tkg/test/tkgctl/shared/mhc_cc.go
@@ -1,0 +1,184 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// nolint:typecheck,goconst,gocritic,stylecheck,nolintlint
+package shared
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+	"os"
+	"path/filepath"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkg/test/framework"
+	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkg/tkgctl"
+)
+
+func E2EMhcCCSpec(context context.Context, inputGetter func() E2ECommonSpecInput) { //nolint:funlen
+	var (
+		input         E2ECommonSpecInput
+		tkgCtlClient  tkgctl.TKGClient
+		tkgCtlClient2 tkgctl.TKGClient
+		logsDir       string
+		clusterName   string
+		namespace     string
+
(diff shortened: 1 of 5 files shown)


pkg/v1/tkg/test/tkgctl/shared/mhc_cc.go
```diff
@@ -0,0 +1,182 @@
...
+		if tkgCtlClient2 == nil {
+			_, _ = GinkgoWriter.Write([]byte("tkgCtlClient is nil"))
+		}
+		err = tkgCtlClient2.DeleteMachineHealthCheck(tkgctl.DeleteMachineHealthCheckOptions{
+			ClusterName:            clusterName,
+			MachinehealthCheckName: mhc.Name,
+			Namespace:              namespace,
+			SkipPrompt:             true,
+		})
+		Expect(err).To(HaveOccurred())
```
anujc25, line 140: We should also add a check to make sure we are getting the error message here by comparing the output with the error message that we are expecting.
```
Expect(err.Error()).To(ContainSubstring("deleting machine health checks on clusterclass based clusters is not supported"))
```

pkg/v1/tkg/test/tkgctl/shared/mhc_cc.go
```diff
@@ -0,0 +1,182 @@
...
+
+		// Set MHC and verify if it is set
+		By(fmt.Sprintf("Updating MHC for cluster %q", clusterName))
+		err = tkgCtlClient2.SetMachineHealthCheck(tkgctl.SetMachineHealthCheckOptions{
+			ClusterName:            clusterName,
+			Namespace:              namespace,
+			MachineHealthCheckName: mhc.Name,
+			UnhealthyConditions:    fmt.Sprintf("%s:%s:%s", string(corev1.NodeReady), string(corev1.ConditionFalse), "5m"),
+		})
+		Expect(err).To(HaveOccurred())
```
anujc25, line 150: Same as above
```
Expect(err.Error()).To(ContainSubstring("setting machine health checks on clusterclass based clusters is not supported"))
```

anujc25 (approved):
Looks good.

Couple of minor comments to also test the error message as part of e2e test to make sure we are seeing the error that we are expecting.

pkg/v1/tkg/test/tkgctl/shared/mhc_cc.go
tenczar: done

PR #2725, opened 2022-06-23
Skip reconcile delete for clusters already deleted.

### LINKED ISSUE #2756: cluster delete request stays in queue indefinitely if cluster has already been deleted.
**Bug description**
The clusterboostrap controller prepares  workload clusters for deletion by removing some of the packageInstalls and removing finalizers from the resources related to the cluster.  

In order to do so it must obtain a k8s client to execute commands against the cluster. 

If the cluster has already been deleted, when the controller revisits the delete request it fails to obtain the k8s client, errors out, and put the delete request back in the queue.  This results on the cluster delete request never being  fully processed and finalizers are not removed from the resources associated with the cluster. 

The delete requests stays in a loop in the queue. 

- [ ] APIs
- [x ] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**

adduarte:
If a cluster is already delete by the time the
controller is ready to reconcile delete, the controller
should skip reconciliation to prevent loop lock.

The k8s remote client should be created only once it is
absolutely necessary.

PackageInstalls with DeleteFailed status are treated as deleted
    
Added log info lines to make debugging easier

Fixes #2756

Deployed cluster on vsphere platform. Deployed workload cluster using kubectl and yaml file. 
Deleted workload cluster to verify it is correctly deleted and 
Used a packageInstall which would go into DeleteFailed state when deleted to verify controller assumes it is deleted and continues with cluster deletion. 

### Release note

``` 
package-based-lcm: Clusterboostrap controller removes cluster delete request from queue if cluster no longer exists.  

It will also treat packageInstalls in DeleteFailed state as deleted and continue with cluster deletion as if the packageInstall has been deleted.

```

### PR Checklist

### DIFF
--- a/addons/controllers/clusterbootstrap_controller.go
+++ b/addons/controllers/clusterbootstrap_controller.go
@@ -1262,6 +1262,7 @@ func (r *ClusterBootstrapReconciler) reconcileClusterProxyAndNetworkSettings(clu
 }
 
 func (r *ClusterBootstrapReconciler) makeClusterIsReadyForDeletion(cluster *clusterapiv1beta1.Cluster, remoteClient client.Client, log logr.Logger) (bool, error) {
+	log.Info("preparing cluster for deletion")
 	clusterBootstrap := &runtanzuv1alpha3.ClusterBootstrap{}
 	err := r.Client.Get(r.context, client.ObjectKeyFromObject(cluster), clusterBootstrap)
 	if err != nil && !apierrors.IsNotFound(err) {
@@ -1273,6 +1274,7 @@ func (r *ClusterBootstrapReconciler) makeClusterIsReadyForDeletion(cluster *clus
 	}
 
 	if hasPackageInstalls(r.context, remoteClient, cluster, r.Config.SystemNamespace, clusterBootstrap.Spec.AdditionalPackages, log) {
+		log.Info("cluster has additional packageInstalls that need to be deleted")
 		err = r.removeAdditionalPackageInstalls(remoteClient, cluster, clusterBootstrap, log)
 		if err != nil {
 			return false, err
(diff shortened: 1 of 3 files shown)


addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1314,15 +1314,25 @@ func (r *ClusterBootstrapReconciler) reconcileDelete(cluster *clusterapiv1beta1.
...
-	remoteClient, err := util.GetClusterClient(r.context, r.Client, r.Scheme, clusterapiutil.ObjectKey(cluster))
+	existingCluster := &clusterapiv1beta1.Cluster{}
+	err = r.Client.Get(r.context, client.ObjectKeyFromObject(cluster), existingCluster)
+	if apierrors.IsNotFound(err) {
+		log.Info("cluster not found. Skipping reconciling")
+		return ctrl.Result{}, nil
+	}
 	if err != nil {
-		return ctrl.Result{RequeueAfter: constants.RequeueAfterDuration}, fmt.Errorf("failed to get remote cluster client: %w", err)
+		log.Error(err, "failed to lookup cluster")
```
adduarte, line 1324: better wording?

shyaamsn:
@adduarte Is this ready for review ? Could you please create an issue and add it as Fixes # here as well as add release notes please. Thanks

adduarte:
This is ready to merge (with any review comments addressed and build test completed). 
already tested locally on a vsphere platform

adduarte:
latest build failures were due to using wrong logger in the unit tests.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1320,25 +1322,39 @@ func (r *ClusterBootstrapReconciler) reconcileDelete(cluster *clusterapiv1beta1.
 	okToRemoveFinalizers := false
 	var err error
 
-	remoteClient, err := util.GetClusterClient(r.context, r.Client, r.Scheme, clusterapiutil.ObjectKey(cluster))
+	log.Info("reconciling cluster delete")
+	existingCluster := &clusterapiv1beta1.Cluster{}
+	err = r.Client.Get(r.context, client.ObjectKeyFromObject(cluster), existingCluster)
+	if apierrors.IsNotFound(err) {
+		log.Info("cluster not found. Skipping reconciling")
```
adduarte, line 1329: nit ->"Skipping reconciliation"

shyaamsn (approved):
LGTM. Please merge after @vijaykatam takes a look.

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1359,30 +1375,37 @@ func hasPackageInstalls(ctx context.Context, remoteClient client.Client,
 
 	for _, pkg := range packages {
 		pkgInstallName := util.GeneratePackageInstallName(cluster.Name, pkg.RefName)
-		if packageInstallExists(ctx, pkgInstallName, namespace, remoteClient, log) {
+		if packageInstallExistsAndCanBeDeleted(ctx, pkgInstallName, namespace, remoteClient, log) {
+			log.Info("found " + pkgInstallName + " packageInstall on cluster")
```
maralavi, line 1379: Why not using Sprintf instead?

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -1359,30 +1375,37 @@ func hasPackageInstalls(ctx context.Context, remoteClient client.Client,
...
+func packageInstallExistsAndCanBeDeleted(ctx context.Context, pkgInstallName, pkgInstallNamespace string,
 	remoteClient client.Client, log logr.Logger) bool {
 
 	pkgInstall := &kapppkgiv1alpha1.PackageInstall{}
 	if err := remoteClient.Get(ctx, client.ObjectKey{Name: pkgInstallName, Namespace: pkgInstallNamespace}, pkgInstall); err != nil {
 		if apierrors.IsNotFound(err) {
 			return false
 		}
-		log.Error(err, "could not verify if package install is present")
+		log.Error(err, "could not verify status of "+pkgInstallNamespace+"/"+pkgInstallName)
```
maralavi, line 1394: Could use Sprintf instead?

vijaykatam (approved):

Issue #2769, opened 2022-06-27
Flaky tests in vspherecpiconfig_controller_test.go


vijaykatam:
**Bug description**

There are a few tests in vspherecpiconfig_controller_test that are flaky and get resolved by re running CI. 

```
Summarizing 4 Failures:
[Fail] VSphereCPIConfig Reconciler [JustBeforeEach] reconcile VSphereCPIConfig manifests in paravirtual mode, with floating IP Should reconcile VSphereCPIConfig and create data values secret for VSphereCPIConfig on management cluster 
/home/runner/work/tanzu-framework/tanzu-framework/addons/controllers/vspherecpiconfig_controller_test.go:104
[Fail] VSphereCPIConfig Reconciler [JustBeforeEach] reconcile VSphereCPIConfig manifests in paravirtual mode, with floating IP Should reconcile ProviderServiceAccount 
/home/runner/work/tanzu-framework/tanzu-framework/addons/controllers/vspherecpiconfig_controller_test.go:104
[Fail] VSphereCPIConfig Reconciler [JustBeforeEach] reconcile VSphereCPIConfig manifests in paravirtual mode, with floating IP Should reconcile aggregated cluster role 
/home/runner/work/tanzu-framework/tanzu-framework/addons/controllers/vspherecpiconfig_controller_test.go:104
[Fail] VSphereCPIConfig Reconciler [JustBeforeEach] reconcile VSphereCPIConfig manifests in paravirtual mode, with floating IP when Headless service and endpoints already exists Should use headless service's endpoints 
/home/runner/work/tanzu-framework/tanzu-framework/addons/controllers/vspherecpiconfig_controller_test.go:104

```
**Affected product area (please put an X in all that apply)**

- [ ] APIs
- [x] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [x] Test and Release
- [ ] User Experience

**Expected behavior**

Tests should not be flaky

PR #2771, opened 2022-06-27
Fix flaky CPIConfig controller test.

HanFa:

Fix flaky CPIConfig controller test caused by the cluster object still being deleted, when the next test case tries to recreate it.

Fixes # https://github.com/vmware-tanzu/tanzu-framework/issues/2769

### Release note

```release-note
Fix flaky CPIConfig controller test.
```

### PR Checklist

### DIFF
--- a/addons/controllers/testdata/test-vsphere-cpi-paravirtual-clusters-same-name.yaml
+++ b/addons/controllers/testdata/test-vsphere-cpi-paravirtual-clusters-same-name.yaml
@@ -1,7 +1,7 @@
 apiVersion: cluster.x-k8s.io/v1beta1
 kind: Cluster
 metadata:
-  name: test-cluster-cpi-paravirtual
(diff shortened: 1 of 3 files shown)


HL-EverGreen:
Thanks for the quick fix! Do you have idea about why the resources created in test `VSphereCPIConfig Reconciler multi clusters` can't be deleted within the timeout? 

I tried to add `AfterEach` block in this test to delete created resources, but it just timed out. Is there anything need to be reconfigured in your test?

HanFa:
Thanks @HL-EverGreen I don't know the exact reason for the deletion timeout. My guess is it is related to some finalizer not getting cleaned up properly, even with the `DeleteResources` in the `afterEach` function. @vijaykatam are you aware of any possibility?

HanFa:
solved the issue by using different names of clusters for each test case.

vijaykatam (approved):

PR #2882, opened 2022-07-08
Use fslock to acquire lock when reading/writing the tanzu config file for update

### LINKED ISSUE #2357: Implement file locking during CLI config writes
**Bug description**

Currently, there is no file locking mechanism when writing CLI config `~/.config/tanzu/config.yaml`. This means multiple processes/plugins interacting with the CLI can issue commands that result in concurrent writes. This can get the config file into a bad state.

We've seen this issue a few times in internal tests where the config ended up looking like this (an invalid YAML):

```yaml
clientOptions:
  cli:
    bomRepo: projects.registry.vmware.com/tkg
    discoverySources:
    - oci:
        image: projects.registry.vmware.com/tkg/packages/standalone/standalone-plugins:v0.21.0-dev-112-ga404323d_vmware.1
        name: default
    edition: tkg
  features:
    cluster:
      custom-nameservers: "false"
    global:
      package-based-lcm-beta: "false"
    management-cluster:
      custom-nameservers: "false"
      export-from-confirm: "true"
      import: "false"
      network-separation-beta: "false"
kind: ClientConfig
metadata:
  creationTimestamp: null
lse"
      network-separation-beta: "false"
current: tkg-az-multi-mgmt-azure
kind: ClientConfig
metadata:
  creationTimestamp: null
servers:
- managementClusterOpts:
    context: tkg-az-multi-mgmt-azure-admin@tkg-az-multi-mgmt-azure
    path: /home/kubo/.kube-tkg/config
  name: tkg-az-multi-mgmt-azure
  type: managementcluster
```

To avoid this, we can do two things:

1. It appears that we are [_updating_ the config during a get operation](https://github.com/vmware-tanzu/tanzu-framework/blob/dcd5c0c2fb8b87ab9f899f01d391d0a2ff08431c/pkg/v1/config/clientconfig.go#L311-L319). A config get ideally shouldn't do a write, so we need find a different place to do those updates and let multiple processes read concurrently (with something like an RWLock). 
2. Implement file locking [during store operation](https://github.com/vmware-tanzu/tanzu-framework/blob/dcd5c0c2fb8b87ab9f899f01d391d0a2ff08431c/pkg/v1/config/clientconfig.go#L406), so concurrent writes are safe.

There are probably some libraries in Go that lets us do file locking. There was also a [proposal](https://go.googlesource.com/proposal/+/master/design/33974-add-public-lockedfile-pkg.md) to make it part of the standard library, but it appears to be incomplete.

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**

Config writes should be concurrency-safe.

anujc25:
- Use fslock to acquire lock when reading the tanzu config file for update
- This PR implements wrapper functions `AcquireTanzuConfigLock` and `ReleaseTanzuConfigLock` using fslock.

Fixes #2357

- Added unit tests to read and updated the configuration file in parallel.
- Verified that the unit test fails without this locking mechanism and it is successful with this change.

### Release note

```release-note
Use fslock to acquire lock when reading/writing the tanzu config file for update
```

### PR Checklist

### DIFF
--- a/pkg/v1/cli/command/core/config.go
+++ b/pkg/v1/cli/command/core/config.go
@@ -83,6 +83,11 @@ var setConfigCmd = &cobra.Command{
 		if len(args) > 2 {
 			return errors.Errorf("only path and value are allowed")
 		}
+
+		// Acquire tanzu config lock
+		config.AcquireTanzuConfigLock()
+		defer config.ReleaseTanzuConfigLock()
+
 		cfg, err := config.GetClientConfig()
 		if err != nil {
 			return err
@@ -201,6 +206,10 @@ var initConfigCmd = &cobra.Command{
 	Use:   "init",
 	Short: "Initialize config with defaults",
 	RunE: func(cmd *cobra.Command, args []string) error {
+		// Acquire tanzu config lock
+		config.AcquireTanzuConfigLock()
+		defer config.ReleaseTanzuConfigLock()
+
 		cfg, err := config.GetClientConfig()
 		if err != nil {
 			return err
@@ -325,6 +334,11 @@ var unsetConfigCmd = &cobra.Command{
 		if len(args) > 1 {
 			return errors.Errorf("only path is allowed")
 		}
+
+		// Acquire tanzu config lock
+		config.AcquireTanzuConfigLock()
+		defer config.ReleaseTanzuConfigLock()
+
 		cfg, err := config.GetClientConfig()
 		if err != nil {
 			return err
--- a/pkg/v1/cli/pluginmanager/test/config.yaml
+++ b/pkg/v1/cli/pluginmanager/test/config.yaml
@@ -3,6 +3,9 @@ clientOptions:
   cli:
     discoverySources:
     - local:
+        name: default-local
+        path: default
+    - local:
         name: fake
         path: standalone
     useContextAwareDiscovery: true
--- a/pkg/v1/config/clientconfig.go
+++ b/pkg/v1/config/clientconfig.go
@@ -22,6 +22,27 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
+func init() {
+	// Acquire tanzu config lock
+	AcquireTanzuConfigLock()
+	defer ReleaseTanzuConfigLock()
+
+	c, err := GetClientConfig()
+	if err != nil {
+		log.Warningf("unable to get client config: %v", err)
+	}
+
+	addedDefaultDiscovery := populateDefaultStandaloneDiscovery(c)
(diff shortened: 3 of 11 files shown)


pkg/v1/config/clientconfig.go
```diff
@@ -22,6 +22,27 @@ import (
...
+
+	addedDefaultDiscovery := populateDefaultStandaloneDiscovery(c)
+	addedFeatureFlags := addDefaultFeatureFlagsIfMissing(c, DefaultCliFeatureFlags)
+	addedEdition := addDefaultEditionIfMissing(c)
+	addedBomRepo := addBomRepoIfMissing(c)
+	addedCompatabilityFile := addCompatibilityFileIfMissing(c)
+
+	if addedFeatureFlags || addedDefaultDiscovery || addedEdition || addedCompatabilityFile || addedBomRepo {
+		_ = StoreClientConfig(c)
+	}
```
anujc25, line 43: Instead of doing this update on each `GetClientConfig` call, do it as part of `init` call. The `GetClientConfig` should only be responsible for getting the configuration and not updating it.

pkg/v1/cli/pluginmanager/test/config.yaml
```diff
@@ -2,6 +2,9 @@ apiVersion: config.tanzu.vmware.com/v1alpha1
 clientOptions:
   cli:
     discoverySources:
+    - local:
+        name: default-local
+        path: default
```
anujc25, line 7: Because now we are not adding default discovery source add part of `GetClientConfig` adding it in the config file so that existing unit tests works as expected

tenczar (approved):
LGTM

vuil (approved):
lgtm, thanks for the updates

pkg/v1/cli/command/core/config.go
```diff
@@ -83,6 +83,11 @@ var setConfigCmd = &cobra.Command{
 		if len(args) > 2 {
 			return errors.Errorf("only path and value are allowed")
 		}
+
+		// Acquire tanzu config lock
+		config.AcquireTanzuConfigLock()
+		defer config.ReleaseTanzuConfigLock()
```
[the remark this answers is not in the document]
anujc25: Unfortunately, I don't think it is possible without refactoring how we are calling this functions because these external functions use `config.GetClientConfig` to get the config, update the config as part of the function and use `config.StoreClientConfig` to update the config. We would like to have the RW lock for this entire read/write operation.

I am not sure we can reliably add locking mechanism as part of `config.GetClientConfig` and `config.StoreClientConfig` individually.

PR #2883, opened 2022-07-08
Concurrency guard addons/pkg/util.gvrHelper

imikushin:

gvrHelper is used as a singleton in a concurrent manner. Additionally, it runs a go-routine that flushes its cache periodically.

Go maps are unsafe for concurrent use: uncontrolled concurrent writes can potentially crash the program. This change guards against these concurrency issues.

N/A

Add a unit test case for concurrent use. Run unit tests with `-race` flag. Without the fix, `WARNING: DATA RACE` is printed out.

### Release note

```release-note
NONE
```

### PR Checklist

### DIFF
--- a/addons/pkg/util/gvr_helper.go
+++ b/addons/pkg/util/gvr_helper.go
@@ -3,6 +3,7 @@ package util
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


vijaykatam (approved):
Thanks for the change.

PR #2986, opened 2022-07-20
Remove feature-gate validation for TKGS cluster and make it based on vSphere version

### LINKED ISSUE #2985: Devops users doesn't have ability to list featuregates and hence `tanzu cluster list/create` fails
**Bug description**
In some of the environments, (specifically TKGS clusters) devops users doesn't have ability to list featuregates and hence `tanzu cluster list/create` fails.

Make necessary changes in the `tanzu cluster create` and `tanzu cluster list` calls to not rely on listing of featuregate.

Possible solution:
- As both ClassyCluster feature-gate will always be enabled for vSphere8 rely on vSphere version to determine ClusterCluster support.

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**
- `tanzu cluster create` and `tanzu cluster list` should work for devops users

anujc25:

In some of the environments, (specifically TKGS clusters) devops users doesn't have ability to list featuregates and hence tanzu cluster list/create fails. This PR solves this problem in the following way.

- As both ClassyCluster feature-gate will always be enabled for vSphere8 rely on vSphere version to determine ClusterCluster support.

Signed-off-by: Anuj Chaudhari <anujc@vmware.com>

Fixes #2985

- Manually tested the `tanzu cluster list` command with the devops user and was able to list the clusters successfully

### Release note

```release-note
Remove feature-gate validation for TKGS cluster and make it based on vSphere version
```

### DIFF
--- a/pkg/v1/tkg/clusterclient/clusterclient.go
+++ b/pkg/v1/tkg/clusterclient/clusterclient.go
@@ -63,6 +63,7 @@ import (
 	kappipkg "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/packaging/v1alpha1"
 
 	cliv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/cli/v1alpha1"
+	configv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/config/v1alpha1"
 	runv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/run/v1alpha1"
 	runv1alpha3 "github.com/vmware-tanzu/tanzu-framework/apis/run/v1alpha3"
 	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/buildinfo"
@@ -330,6 +331,8 @@ type Client interface {
 	GetClusterStatusInfo(clusterName, namespace string, workloadClusterClient Client) ClusterStatusInfo
 	// GetCLIPluginImageRepositoryOverride returns map of image repository override
 	GetCLIPluginImageRepositoryOverride() (map[string]string, error)
+	// VerifyExistenceOfCRD returns true if CRD exists else return false
+	VerifyExistenceOfCRD(resourceName, resourceGroup string) (bool, error)
 }
 
 // PollOptions is options for polling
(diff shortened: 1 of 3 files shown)


pkg/v1/tkg/clusterclient/clusterclient.go
```diff
@@ -1575,6 +1579,26 @@ func (c *client) GetPacificTKCAPIVersion() (string, error) {
...
+// For vSphere 7 and older version of TKGS cluster this API will not exists but it will be available in
+// vSphere 8 based TKGS cluster
+func (c *client) IsVSphere8BasedTKGSCluster() (bool, error) {
+	// Since we're looking up API types via discovery, we don't need the dynamic client.
+	clusterQueryClient, err := capdiscovery.NewClusterQueryClient(c.dynamicClient, c.discoveryClient)
+	if err != nil {
+		return false, err
+	}
+
+	var queryObject = capdiscovery.Group("featuregates", configv1alpha1.GroupVersion.Group).WithResource("featuregates")
```
rajathagasthya, line 1593: suggestion: I would be specific about versions using `WithVersions("v1alpha1")` for some future proofing in case FeatureGates has an API version bump.

anujc25: Do we really need to care considering we just want to check if the CRD is present or not? 
The current function name seems generic, so adding `WithVersions("v1alpha1")` will require change in the CLI whenever we bump the API version.

And if the purpose for using this function is to just check presence of CRD, does it make sense to add version?

rajathagasthya: I think you're right and it's not really necessary since vSphere 7 won't ever have these CRDs.

chandrareddyp (commented):
As discussed, better update featuregate_helper.FeatureActivatedInNamespace() method to identify the tkc-api and clusterclass feature's are enabled or not by calling tkgctl. IsVSphere8BasedTKGSCluster() and don't change any of the code and test cases. thanks.

pkg/v1/tkg/clusterclient/clusterclient.go
```diff
@@ -1575,6 +1579,23 @@ func (c *client) GetPacificTKCAPIVersion() (string, error) {
 	return strings.TrimSpace(match[1]), nil
 }
 
+// VerifyFeaturegateCRD returns true if CRD exists else return false
+func (c *client) VerifyFeaturegateCRD() (bool, error) {
```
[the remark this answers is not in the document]
anujc25: Updated

chandrareddyp:
Changes looks good.

vuil (approved):
lgtm, thanks

PR #2987, opened 2022-07-20
validate workload cluster existence

### LINKED ISSUE #2988: CLI does not validate workload cluster existence for tkgs use case
**Bug description**
CLI command `tanzu cluster create` to create tkgs workload cluster, does not validate whether the workload cluster already exists or not, and if exists its failing at tkgs supervisor with a misleading error message 

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**
CLI should validate the given input cluster name already exists or not, if exists throw error and fail-fast.
- Create a tkgs workload cluster
- Create a tkgs workload cluster with same name as in previous step

chandrareddyp:
For CLI command `tanzu cluster create`, validates whether the given input workload cluster name already exists or not, if exists then throws error

Fixes #2988 

- Unit Test cases
- Tested to create tkgs workload cluster which already exists
- Tested to create tkgs workload cluster which is already not exists

### Release note

```release-note
Fixes to validate workload cluster existence and throws an error if exists
```

### DIFF
--- a/pkg/v1/tkg/tkgctl/get_cluster.go
+++ b/pkg/v1/tkg/tkgctl/get_cluster.go
@@ -57,3 +57,19 @@ func (t *tkgctl) GetClusters(options ListTKGClustersOptions) ([]client.ClusterIn
 
 	return clusters, nil
 }
+
+func (t *tkgctl) IsClusterExists(clustername, namespace string) (bool, error) {
+	clusters, err := t.GetClusters(ListTKGClustersOptions{
+		Namespace: namespace,
+		IncludeMC: true,
+	})
+	if err != nil {
+		return false, err
+	}
+	for _, cluster := range clusters {
+		if cluster.Name == clustername && cluster.Namespace == namespace {
+			return true, nil
+		}
+	}
+	return false, nil
+}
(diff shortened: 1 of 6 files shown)


pkg/v1/tkg/tkgctl/get_cluster.go
```diff
@@ -57,3 +57,19 @@ func (t *tkgctl) GetClusters(options ListTKGClustersOptions) ([]client.ClusterIn
...
+func (t *tkgctl) IsClusterExists(clustername, namespace string) (bool, error) {
+	clusters, err := t.GetClusters(ListTKGClustersOptions{
+		Namespace: namespace,
+		IncludeMC: true,
+	})
+	if err != nil {
+		return false, err
+	}
+	for _, cluster := range clusters {
+		if cluster.Name == clustername && cluster.Namespace == namespace {
```
anujc25, line 70: Just noticed that we have updated condition to also check the `namespace` as well. Previously we were just checking the cluster name irrespective of the namespace.

anujc25 (approved):
Looks good.

Please Add the `Release Notes` as part of PR description.

pkg/v1/tkg/tkgctl/get_cluster.go
chandrareddyp: okay, thanks.

PR #2995, opened 2022-07-21
enable rolling upgrade for csi daemonset

jvrahav:
Adds overlays and schema changes to set UpgradeStrategy for csi daemonset to RollingUpgrade. This is required as we upgrade tkc from pre-tkr/tkgs-tkr to unified tkr, the csi-node pods are stuck in ImagePullBackoff as they are referring to images on the older version of csi.
This change should help recycle the old node and replace the new one

Fixes #

### Release note

```release-note

```


DimpleRajaVamsi (approved):
LGTM

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -40,6 +40,7 @@ import (
 	kapppkgiv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/packaging/v1alpha1"
 	kapppkgv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apiserver/apis/datapackaging/v1alpha1"
 	versions "github.com/vmware-tanzu/carvel-vendir/pkg/vendir/versions/v1alpha1"
+
```
adduarte, line 43: gratuitous line

vijaykatam:
Post discussion with a few folks, we decided that this change is not needed and the upgrade issue for TKGS is being handled differently.

PR #2996, opened 2022-07-21
Exposes addons-manager cluster-delete-timeout as template parameter

### LINKED ISSUE #2991: Make cluster-delete-timeout param in addons controller available as a package value
TKGS doesn’t want to wait for deletion of additional packages for cluster delete. We do have a param that controls the timeout cluster-delete-timeout that is available but not exposed all the way up to the package. Provide a package value that can be propagated to the controller.

N/A

- ( ) APIs
- (x) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Additional context**

adduarte:
cluster-delete-timeout controls how log the addons-manager
waits before timing out a cluster deletion. Once the timeout
is reached, the cluster is deleted unconditionally.

The variable is exposed as tanzuAddonsManager.deployment.clusterDeleteTimeout

Fixes #2991

Deployed tkg cluster on vsphere.
Created addons-manager package and pushed to OCI registry. 
Edited the Package:  kubectl edit package -n vmware-system-pkgs   addons-manager.tanzu.vmware.com.0.25.0-dev-11-g87115377+vmware.1 

Verified new deployment is  created with new --cluster-delete-timeout=10m

Verify tanzu-addons-controller-manager-#### starts and logs. 

### Release note

```release-note
addons-manager "--cluster-delete-timeout" is exposed to templates as tanzuAddonsManager.deployment.clusterDeleteTimeout
```

### DIFF
--- a/packages/addons-manager/bundle/config/upstream/addons-manager.yaml
+++ b/packages/addons-manager/bundle/config/upstream/addons-manager.yaml
@@ -347,6 +347,7 @@ spec:
         - #@ "--metrics-bind-addr={}".format(data.values.tanzuAddonsManager.deployment.metricsBindAddress)
         - #@ "--leader-elect={}".format(data.values.tanzuAddonsManager.deployment.leaderElection)
         - #@ "--health-addr=:{}".format(data.values.tanzuAddonsManager.deployment.healthzPort)
+        - #@ "--cluster-delete-timeout={}".format(data.values.tanzuAddonsManager.deployment.clusterDeleteTimeout)
         - #@ "--webhook-server-port={}".format(data.values.tanzuAddonsManager.deployment.webhookServerPort)
         - #@ "--addon-namespace={}".format(data.values.tanzuAddonsManager.namespace)
         #@ if/end data.values.tanzuAddonsManager.featureGates.clusterBootstrapController:
--- a/packages/addons-manager/bundle/config/values.yaml
+++ b/packages/addons-manager/bundle/config/values.yaml
@@ -14,6 +14,7 @@ tanzuAddonsManager:
     tolerations: []
     webhookServerPort: 9865
     healthzPort: 18316
+    clusterDeleteTimeout: "10m"
     metricsBindAddress: "localhost:18317"
   featureGates:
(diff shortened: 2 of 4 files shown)


packages/addons-manager/bundle/config/values.yaml
```diff
@@ -14,6 +14,7 @@ tanzuAddonsManager:
     tolerations: []
     webhookServerPort: 9865
     healthzPort: 18316
+    clusterDeleteTimeout: "5m"
```
shyaamsn, line 17: The default in the code is 10m. Lets stick to that

packages/addons-manager/bundle/config/upstream/addons-manager.yaml
```diff
@@ -347,6 +347,7 @@ spec:
         - #@ "--metrics-bind-addr={}".format(data.values.tanzuAddonsManager.deployment.metricsBindAddress)
         - #@ "--leader-elect={}".format(data.values.tanzuAddonsManager.deployment.leaderElection)
         - #@ "--health-addr=:{}".format(data.values.tanzuAddonsManager.deployment.healthzPort)
+        - #@ "--cluster-delete-timeout=:{}".format(data.values.tanzuAddonsManager.deployment.clusterDeleteTimeout)
```
shyaamsn, line 350: I dont think there should be a colon (:) here. Thats only for health-addr since its refering to a addr:port

packages/addons-manager/bundle/config/values.yaml
adduarte: done

packages/addons-manager/bundle/config/upstream/addons-manager.yaml
adduarte: done

shyaamsn:
@adduarte Could you please add a release note. Thanks

vijaykatam (approved):
LGTM. Re run build to fix flake.

PR #2999, opened 2022-07-21
Update Scale API to support ClusterClass clusters

tenczar:
Adds a new DoScaleCCCluster function to handle scaling the control plane
and worker nodes of a clusterclass based cluster. Operates like the
legacy API when passing in a node pool name. Scales the mds in a round
robin fashion when no node pool name is given.

Fixes #

### Release note

```release-note
Updates scale API to support clusterclass based clusters.
```


prkalle (approved):
LGTM. Thanks!

anujc25 (approved):

Issue #3000, opened 2022-07-21
Spaces or newline in TKG_NO_PROXY causing export command to behave wrongly


lubronzhan:
**Bug description**

**Affected product area (please put an X in all that apply)**

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**
NO_PROXY configration should be correct even there are spaces inside TKG_NO_PROXY
**Steps to reproduce the bug**
input `TKG_NO_PROXY` like
```
export TKG_NO_PROXY='192.168.111.0/24,192.168.116.0/24,192.168.117.0/24,192.168.118.0/24,192.168.119.0/24,192.168.120.0/24,
172.18.0.0/16,fc00:f853:ccd:e793::/64,100.64.0.0/13,100.96.0.0/11, tanzu.lab,.svc,
172.19., tanzu-vc01.tanzu.lab, 172.19.80.6,127.0.0.1, alb-proxy.tanzu.lab,
.svc.cluster.local,.svc.cluster.local,localhost, 172.19.10.10, 172.19.0.0/16,tkg-kind-cb8j4276hkosmfto284g-control-plane,100.64.0.0/13,100.96.0.0/11,tkg-kind-cb8j4276hkosmfto284g-control-plane,.svc,.svc.cluster,.svc.cluster.local'
```
Create cluster
Then the NO_PROXY in kube-apiserver spec will lose some of the information
```
env:
name: NO_PROXY
value: .svc.cluster.local,tkg-kind-cb8j4276hkosmfto284g-control-plane,.svc,192.168.111.0/24,192.168.118.0/24,
name: HTTPS_PROXY
value: http://192.168.116.1:3128/
name: HTTP_PROXY
value: http://192.168.116.1:3128/
image: projects.registry.vmware.com/tkg/kube-apiserver:v1.22.9_vmware.1
```
**Version** (include the SHA if the version is not obvious)
Any tkg version
**Environment where the bug was observed (cloud, OS, etc)**

**Relevant Debug Output (Logs, manifests, etc)**

PR #3001, opened 2022-07-21
Trim spaces in no_proxy, also quote the assign value of export command

lubronzhan:
Spaces or newline in TKG_NO_PROXY causing export command to behave wrongly

Fixes #https://github.com/vmware-tanzu/tanzu-framework/issues/3000

### Release note

```release-note
Fix the issue that if there are spaces and newline in TKG_NO_PROXY will cause cluster creation to fail
```

### DIFF
--- a/pkg/v1/providers/tests/clustergen/param_models/cluster_optional.model
+++ b/pkg/v1/providers/tests/clustergen/param_models/cluster_optional.model
@@ -93,7 +93,7 @@ VSPHERE_CLONE_MODE: "fullClone","linkedClone"
 # http proxy (too many "NA" leads to overwhelmingly negative cases)
 TKG_HTTP_PROXY: "http://10.0.200.100", "http://[fc00:f853:ccd:e793::1]:3128"
 TKG_HTTPS_PROXY: "NA", "http://10.0.200.100", "http://[fc00:f853:ccd:e793::1]:3128"
-TKG_NO_PROXY: "NA", "10.0.200.1/24", "10.184.90.80"
+TKG_NO_PROXY: "NA", "10.0.200.1/24", "10.184.90.80<comma>        10.0.0.1/24"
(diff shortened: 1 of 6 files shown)


vuil (approved):
lgtm, thanks

vuil:
Could we tweak cluster_optional.model to include sample data with some spacing? That will produce clustergen diffs that can highlight the effect of the change.

Also, we probably need similar treatment in the yttcc side of the overlays.

vuil (commented):
Some additional thoughts/suggestions above

pkg/v1/providers/tests/clustergen/param_models/cluster_optional.model
```diff
@@ -93,7 +93,9 @@ VSPHERE_CLONE_MODE: "fullClone","linkedClone"
 # http proxy (too many "NA" leads to overwhelmingly negative cases)
 TKG_HTTP_PROXY: "http://10.0.200.100", "http://[fc00:f853:ccd:e793::1]:3128"
 TKG_HTTPS_PROXY: "NA", "http://10.0.200.100", "http://[fc00:f853:ccd:e793::1]:3128"
-TKG_NO_PROXY: "NA", "10.0.200.1/24", "10.184.90.80"
+TKG_NO_PROXY: '"NA", "10.0.200.1/24", "10.184.90.80"
```
vuil, line 96: this doesn't look right, possible values should belong on one line. Comma is a special character that needs to be escaped, so if the third case is to introduce a string with space, it should be:
```
TKG_NO_PROXY: '"NA", "10.0.200.1/24", "10.184.90.80<comma> 10.0.0.1/24"
```

lubronzhan: I want to add a new line. Sorry didnt know the syntax

randomvariable:
Please rebase this one.

PR #3002, opened 2022-07-21
Fix: object-Propagation controller failing to copy resources

imikushin:

Without this change, the Object-Propagation Controller fails to copy
ClusterClass resources with errors like this in the log:
`spec.workers.machineDeployments[0].template.bootstrap.ref.namespace:
Invalid value: "tkg-system": template must be in the same namespace as
the ClusterClass (default)`.

N/A

Added a unit test to reproduce the failure. Fixed the code to make tests pass. 

### Release note

```release-note
NONE
```


anujc25 (approved):
Looks good.

PR #3006, opened 2022-07-22
OPC should remove ownerReferences from propagated objects

imikushin:

Without this change, the OPC (Object-Propagation Controller) copies ClusterClass-related resources to other namespaces, but because they have an ownerReference with UID from the original ClusterClass (in the source namespace), Kubernetes immediately deletes them (as the original ClusterClass is not present in the target namespace, only a propagated copy of it).

N/A

Added unit tests covering the fix.

### Release note

```release-note
NONE
```


anujc25 (approved):
Looks good

PR #3007, opened 2022-07-22
Propagate only clusterclass related resources to other namespaces based on label

### LINKED ISSUE #3005: Object propagation controller propagates Cluster specific `AWSMachineTemplate` and `KubeadmConfigTemplate` resources as well
**Bug description**

Object propagation controller is responsible for coping all clusterclass related resources to all the namespaces and we are providing the configuration as part of object propagation deployment.

However, controller also propagates Cluster specific `AWSMachineTemplate` and `KubeadmConfigTemplate` resources as well to all the namespaces.

We should only propagate ClusterClass owned resources to all namespaces and not Cluster owned resources.

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**
Cluster specific resources should not be copied to all namespaces

anujc25:
- Propagate only clusterclass related resources to other namespaces which doesn't have `topology.cluster.x-k8s.io/owned`
- `MachineTemplate` and `KubeadmConfigTemplate` resources that gets created by the topology controller for the clusters will include `topology.cluster.x-k8s.io/owned` label. So, we need to skip resources with this labels from coping to other namespaces.
- Also skip propagating ClusterClass resources to Cluster-API provider namespaces based on`cluster.x-k8s.io/provider` label

Fixes #3005

- Created AWS based management-cluster
- Verified that resources created for management-cluster created in `tkg-system` doesn't get copied over to all other namespace.

### Release note

```release-note
Fixed object-propagation-controller copying cluster owned `MachineTemplate` and `KubeadmConfigTemplate` resources to other namespaces
```

### DIFF
--- a/packages/tkg-clusterclass/bundle/config/object-propagation/deployment.yaml
+++ b/packages/tkg-clusterclass/bundle/config/object-propagation/deployment.yaml
@@ -10,7 +10,7 @@
     namespace: #@ data.values.namespaceForPackageInstallation
     labelSelector: ''
   target:
-    namespaceLabelSelector: ''
+    namespaceLabelSelector: '!cluster.x-k8s.io/provider'
     detectAndReplaceSourceNSRef: true
 - source:
     apiVersion: controlplane.cluster.x-k8s.io/v1beta1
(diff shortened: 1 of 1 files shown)


imikushin (approved):
Looks great!

PR #3009, opened 2022-07-22
Run vendir sync for packages concurrently

### LINKED ISSUE #3010: Run `vendir-sync` concurrently for all packages

`vendir sync` is run serially for each package in `packages/` currently when run using `make package-vendir-sync`, which is slow. Run it concurrently for better performance.

Keep it as is and experience slower `vendor sync`s.

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- (X) Developer Experience

**Additional context**

rajathagasthya:

Use `errgroup` for running `vendir sync` concurrently. Also uses
`vendir`'s `--chdir` instead of `os.Chdir` to remove race.

Fixes #3010 

```
$ make package-vendir-sync
...
cd hack/packages/package-tools && go run main.go vendir sync --repository=
Syncing package "addons-manager"
Syncing package "capabilities"
Syncing package "cliplugins"
Syncing package "core-management-plugins"
Syncing package "featuregates"
Syncing package "framework"
Syncing package "standalone-plugins"
Syncing package "tanzu-auth"
Syncing package "tkg"
Syncing package "tkg-autoscaler"
Syncing package "tkg-clusterclass"
Syncing package "tkg-clusterclass-aws"
Syncing package "tkg-clusterclass-azure"
Syncing package "tkg-clusterclass-docker"
Syncing package "tkg-clusterclass-vsphere"
No vendir.yml found in package "tkg-storageclass". Skipping vendir sync...
Syncing package "tkr-aws-machine-webhook"
Syncing package "tkr-azure-machine-webhook"
Syncing package "tkr-docker-machine-webhook"
Syncing package "tkr-service"
Syncing package "tkr-source-controller"
Syncing package "tkr-vsphere-machine-webhook"
```

### Release note

```release-note
NONE
```

### DIFF
--- a/hack/packages/package-tools/cmd/vendir-sync.go
+++ b/hack/packages/package-tools/cmd/vendir-sync.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 
 	"github.com/spf13/cobra"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/vmware-tanzu/tanzu-framework/hack/packages/package-tools/constants"
 	"github.com/vmware-tanzu/tanzu-framework/hack/packages/package-tools/utils"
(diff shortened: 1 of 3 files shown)


hack/packages/package-tools/cmd/vendir-sync.go
```diff
@@ -43,40 +44,50 @@ func runPackageVendirSync(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	var g errgroup.Group
```
codegold79, line 47: I know we Go programmers try to keep variables as short as possible, but the caveat is that it needs to be clear what it means. I think `g` is used far enough from its declaration, it should be named something like `errGrp` to be clear.

codegold79, line 47: I know we Go programmers try to keep variables as short as possible, but the caveat is that it needs to be clear what it means. I think g is used far enough from its declaration, it should be named something like errGrp to be clear.

codegold79 (changes requested):
Changes make sense. Only had a minor suggestion.

hack/packages/package-tools/cmd/vendir-sync.go
rajathagasthya: I can also just move this declaration to L55. Would that be clearer or would you still like it to be renamed? `var errGrp errgroup.Group` is too much stutter.

codegold79: L55 would be better. But what do you think about naming it `errs`?
Would `var errs errgroup.Group` be more confusing or more clear?

codegold79: After looking at docs, I think `g` is ok. Maybe it's common enough that it will not be confusing.
Screen Shot 2022-07-22 at 9 07 11 AM

codegold79 (approved):

hack/packages/package-tools/cmd/vendir-sync.go
rajathagasthya: `errs` indicate a slice of errors, so it would be confusing. Typically, people use `g` or `wg` for error/waitgroups like this.

codegold79 (approved):

vijaykatam (approved):

PR #3021, opened 2022-07-26
Modify logic to determine stretch supervisor

jvrahav:
Modify logic to determine AZ based on the following 1. if there is more than one AZ assume stretched supervisor 2. if there is only one AZ or none assume non-stretched supervisor This logic is arrived at based on the fact that regardless of stretch supervisor or single supervisor, in 8.0+ we always create a zone.

Fixes #

### Release note

```release-note

```

### DIFF
--- a/addons/controllers/csi/vspherecsiconfig_utils.go
+++ b/addons/controllers/csi/vspherecsiconfig_utils.go
@@ -380,11 +380,7 @@ func (r *VSphereCSIConfigReconciler) isStretchedSupervisorCluster(ctx context.Co
 		return false, err
 	}
 
-	if len(azList.Items) == 0 {
-		return false, nil
-	}
-
-	if len(azList.Items) == 1 {
+	if len(azList.Items) <= 1 {
 		// by default every non-stretched cluster contains a single zone,
 		// if there is only one AZ we assume it is non-stretched cluster
 		return false, nil
--- a/addons/controllers/csi/vspherecsiconfig_utils.go
+++ b/addons/controllers/csi/vspherecsiconfig_utils.go
@@ -19,11 +19,12 @@ import (
 	capictrlpkubeadmv1beta1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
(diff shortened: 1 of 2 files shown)


addons/controllers/csi/vspherecsiconfig_utils.go
```diff
@@ -385,11 +385,9 @@ func (r *VSphereCSIConfigReconciler) isStretchedSupervisorCluster(ctx context.Co
...
 
 	if len(azList.Items) == 1 {
-		// by default every cluster contains legacy availability zone,
-		// if only the legacy AZ is present, return false
-		if azList.Items[0].Name == LegacyZoneName {
-			return false, nil
-		}
+		// by default every non-stretched cluster contains a single zone,
+		// if there is only one AZ we assume it is non-stretched cluster
+		return false, nil
```
danniel1205, line 390: nit: Would it be a good idea to combine L387 with L383 by using just one IF statement ? That helps prevent `cyclomatic complexity` from exceeding the limit (I know it is not at the moment, but recommend we do our best to simplify the function).

jvrahav: makes sense. Handled it ...

alfredthenarwhal:
> @jvrahav: /test

Tests can be triggered only by the trusted reviewers (users mentioned in CODEOWNERS file).

addons/controllers/vspherecsiconfig_controller_test.go
```diff
@@ -274,26 +274,34 @@ var _ = Describe("VSphereCSIConfig Reconciler", func() {
...
 			// zone and set zone to true
-			availabilityZone = &topologyv1alpha1.AvailabilityZone{
+			availabilityZone1 = &topologyv1alpha1.AvailabilityZone{
 				ObjectMeta: metav1.ObjectMeta{
 					Name: "test-az",
 				},
 			}
 
-			Expect(k8sClient.Create(ctx, availabilityZone)).Should(Succeed())
+			availabilityZone1 = &topologyv1alpha1.AvailabilityZone{
```
maralavi, line 289: did you mean ```availabilityZone2``` here instead?

jvrahav: i just fixed this in the latest push. can you please check

maralavi (approved):

PR #3026, opened 2022-07-27
Fix the issue of tkg-clusterclass packageinstall reconcile failed after 15 minutes of management-cluster creation

### LINKED ISSUE #3027: `tkg-clusterclass` package shows reconciled failed after 15 minutes of management-cluster creation
**Bug description**
When creating management-cluster using `package-based-lcm-beta` feature flag enabled. We are seeing the management-cluster gets created successfully.

Even the management packages are reconciled successfully at the time of management cluster creation. However, after few minutes (10-15 minutes) `tkg-clusterclass`, `tkg-clusterclass-aws` and `tkg-pkg` shows reconciled failed with error message:

Resource 'awsclustertemplate/aws-cluster-template (infrastructure.cluster.x-k8s.io/v1beta1) namespace: tkg-system' is already associated with a different label 'kapp.k14s.io/app=1655969821199577275'
 
This is because, after the clusterctl move clusterclass resources gets over written. Because of that we are seeing this error in the PackageInstall reconciliation.

This can be fixed by removing the [kapp.k14s.io/app=1655969821199577275](http://kapp.k14s.io/app=1655969821199577275) label from the clusterclass and other infratemplate resources.

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**
Should reconcile all packages successfully

anujc25:
Remove kapp labels from clusterclass resources after move during management-cluster creation

- Solves the issue of clusterclass packageinstall reconciliation failures after 15 minutes of move because of the invalid kapp-controller labels that got updated during move.
- This change try to solve this PR by removing kapp-controller labels from clusterclass resources so that clusterclass package can override the resources and manages the resources through packageinstalls

Signed-off-by: Anuj Chaudhari <anujc@vmware.com>

Fixes #3027

- Added unit tests for the core function of updating labels
- Create AWS management-cluster using `package-based-lcm-beta` feature-flag enabled
- Once the management-cluster creation is successful waited for 30 minutes and verified the packageinstall statuses.
- All management-packages were in `Reconcile Succeeded` status and clusterclass resources had correct labels.

### Release note

```release-note
package-based-lcm: Fix the issue of tkg-clusterclass packageinstall reconcile failed after 15 minutes of management-cluster creation
```

### DIFF
--- a/pkg/v1/tkg/client/init.go
+++ b/pkg/v1/tkg/client/init.go
@@ -14,8 +14,16 @@ import (
 	"github.com/go-openapi/swag"
 	"github.com/juju/fslock"
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	capav1beta1 "sigs.k8s.io/cluster-api-provider-aws/api/v1beta1"
+	capzv1beta1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
+	capvv1beta1 "sigs.k8s.io/cluster-api-provider-vsphere/apis/v1beta1"
+	capi "sigs.k8s.io/cluster-api/api/v1beta1"
+	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
 	clusterctlv1 "sigs.k8s.io/cluster-api/cmd/clusterctl/api/v1alpha3"
 	clusterctl "sigs.k8s.io/cluster-api/cmd/clusterctl/client"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	crtclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/config"
@@ -415,6 +423,14 @@ func (c *TkgClient) PatchClusterInitOperations(regionalClusterClient clusterclie
 		return errors.Wrap(err, "unable to patch optional metadata under labels")
 	}
 
+	if config.IsFeatureActivated(config.FeatureFlagPackageBasedLCM) {
+		// Patch and remove kapp-controller labels from clusterclass resources
+		err = c.removeKappControllerLabelsFromClusterClassResources(regionalClusterClient)
+		if err != nil {
+			return errors.Wrap(err, "unable to remove kapp-controller labels from the clusterclass resources")
+		}
+	}
+
 	return err
 }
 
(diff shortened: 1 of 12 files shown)


pkg/v1/tkg/client/init.go
```diff
@@ -768,3 +784,29 @@ func (c *TkgClient) safelyAddFeatureFlag(featureFlags map[string]string, feature
 	}
 	return featureFlags
 }
+
+// removeKappControllerLabelsFromClusterClassResources removes kapp-controller labels from clusterclass resources
+func (c *TkgClient) removeKappControllerLabelsFromClusterClassResources(regionalClusterClient clusterclient.Client) error {
+	errList := []error{}
+	labelsToBeDeleted := []string{"kapp.k14s.io/app", "kapp.k14s.io/association"}
```
saji-pivotal, line 791: nit-ish: There are a lot of harcoded strings here (up to line 801). Can we either make these string constants or use existing constants?

saji-pivotal (changes requested):
Looks good overall. I do have somewhat of a nit though

pkg/v1/tkg/client/init.go
```diff
@@ -149,8 +157,8 @@
~		log.Infof("Using custom image repository: %s", customImageRepo)
~	}
~
~	c.ensureClusterTopologyConfiguration()
~
~	providerName, _, err := ParseProviderName(options.InfrastructureProvider)
~	if err != nil {
~		return errors.Wrap(err, "unable to parse provider name")
```
prkalle: should we update the error message?

pkg/v1/tkg/client/init.go
```diff
@@ -415,6 +423,14 @@ func (c *TkgClient) PatchClusterInitOperations(regionalClusterClient clusterclie
 		return errors.Wrap(err, "unable to patch optional metadata under labels")
 	}
 
+	if config.IsFeatureActivated(config.FeatureFlagPackageBasedLCM) {
+		// Patch and remove kapp-controller labels from clusterclass resources
+		err = c.removeKappControllerLabelsFromClusterClassResources(regionalClusterClient)
+		if err != nil {
+			return errors.Wrap(err, "unable to patch optional metadata under labels")
```
prkalle, line 430: should we update the error message?

prkalle (commented):
Overall the changes look good. I have one minor nit comment.

tenczar (approved):
I agree with Sudarshan about the magic strings. Otherwise looks good to me

pkg/v1/tkg/client/init.go
anujc25: Updated the PR to use constants. Also updated few more places to use consistent constants across the board.

saji-pivotal (approved):

miclettej (approved):

PR #3031, opened 2022-07-28
Add calico e2e test to aws clusterclass suite

### LINKED ISSUE #2460: TC11: AWS Cluster - Ability to create CC guest cluster on AWS with calico CNI and verify it is successful

codegold79:
Signed-off-by: F. Gold <fgold@vmware.com>, Leo Li  <lhaoran@vmware.com>

- Add Calico CNI E2E tests to AWS_CC suite

Fixes #2460

- E2E tests passed locally

### Release note

```release-note
Add Calico CNI E2E tests to AWS_CC suite
```

### DIFF
--- a/pkg/v1/tkg/test/tkgctl/shared/e2e_common.go
+++ b/pkg/v1/tkg/test/tkgctl/shared/e2e_common.go
@@ -220,7 +220,7 @@ func E2ECommonSpec(ctx context.Context, inputGetter func() E2ECommonSpecInput) {
 				By(fmt.Sprintf("Verify workload cluster %q resources have been deleted", clusterName))
 				Eventually(func() bool {
 					return clusterResourcesDeleted(ctx, mngClient, clusterResources)
-				}, waitTimeout, pollingInterval).Should(BeTrue())
+				}, resourceDeletionWaitTimeout, pollingInterval).Should(BeTrue())
 			}
 		}
 
--- a/pkg/v1/tkg/test/tkgctl/shared/package.go
+++ b/pkg/v1/tkg/test/tkgctl/shared/package.go
@@ -54,10 +54,11 @@ import (
 )
 
 const (
-	getResourceTimeout  = time.Minute * 1
-	waitForReadyTimeout = time.Minute * 10
-	waitTimeout         = time.Minute * 10
-	pollingInterval     = time.Second * 30
+	getResourceTimeout          = time.Minute * 1
+	waitForReadyTimeout         = time.Minute * 10
+	waitTimeout                 = time.Minute * 10
+	resourceDeletionWaitTimeout = time.Minute * 20
+	pollingInterval             = time.Second * 30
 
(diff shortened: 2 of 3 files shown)


pkg/v1/tkg/test/tkgctl/shared/package.go
```diff
@@ -267,7 +267,7 @@ func verifyClusterBootstrap(ctx context.Context, c client.Client, clusterBootstr
 		expectedClusterBootstrap.Spec.CNI.ValuesFrom.ProviderRef = &corev1.TypedLocalObjectReference{
 			APIGroup: &apiGroup,
 			Kind:     "CalicoConfig",
-			Name:     GeneratePackageSecretName(clusterBootstrap.Name, "calico"),
```
maralavi, line 270: for clusters made with custom CB, the name of the config object is the same as the CB. But for those CB created from CBT, the above name is used. So both these names can be valid for different scenarios. 
Now why this change is needed?

pkg/v1/tkg/test/tkgctl/shared/e2e_common.go
```diff
@@ -216,7 +216,7 @@ func E2ECommonSpec(ctx context.Context, inputGetter func() E2ECommonSpecInput) {
 				By(fmt.Sprintf("Verify workload cluster %q resources have been deleted", clusterName))
 				Eventually(func() bool {
 					return clusterResourcesDeleted(ctx, mngClient, clusterResources)
-				}, waitTimeout, pollingInterval).Should(BeTrue())
+				}, 2*waitTimeout, pollingInterval).Should(BeTrue())
```
maralavi, line 219: Why increasing the timeout, and what guarantees this is enough? Why current timeout used to work before?

pkg/v1/tkg/test/tkgctl/shared/package.go
HL-EverGreen: Thanks for the review! As Shyaam mentioned in the channel, this change is in the path of Calico CNI. And since Calico is the case of custom CB, I think it is safe to change here. The CB created from CBT won't go in this path as the CNI won't be Calico.

pkg/v1/tkg/test/tkgctl/shared/e2e_common.go
HL-EverGreen: Hi @maralavi, the old timeout didn't work for Calico tests before. The test was flaky due to timeout in resource deletion. After increasing the timeout I didn't see that issue for several runs. So I think it helps mitigate the flaky scenario.

pkg/v1/tkg/test/tkgctl/shared/package.go
maralavi: Ok, so just noticed ```strings.Contains(clusterBootstrap.Spec.CNI.RefName, "calico")``` check, and calico always use custom CB. We should be good then. Thanks for the clarification.

maralavi (approved):

PR #3033, opened 2022-07-29
Validate cluster kubernetes version upgrade check as initial step

### LINKED ISSUE #3034: cluster TKR label should not be updated if user tries to downgrade the cluster
**Bug description**
If user tries to downgrade a cluster created with latest TKR version to older TKR version using TKG library, CLI is returning error after comparing the k8s version, however the cluster TKR label is updated with the older TKR version causing the issue since Add-ons depends on the cluster TKR label to update the Add-ons and add-ons reconcilation is failing.

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**
We should check k8s version downgrade before updating the cluster TKR label and reject the downgrade operation.

prkalle:
Signed-off-by: Prem Kumar Kalle <pkalle@vmware.com>

This PR adds cluster kubernetes verison upgrade check as an Initial step for upgrade. This would address the issue where the cluster TKR label is updated to old TKR version though the cluster kubernetes version downgrade is blocked later in the upgrade process causing Add-ons reconcilation issue because Add-ons manager relies on cluster TKR label for Add-ons reconcilation. This issue would only be seen if user uses tkgctl library.

Fixes #3034 

Unit tested the changes

### Release note

```release-note
Validate cluster kubernetes version upgrade check as initial step
```

### DIFF
--- a/pkg/v1/tkg/client/upgrade_cluster.go
+++ b/pkg/v1/tkg/client/upgrade_cluster.go
@@ -179,7 +179,13 @@ func (c *TkgClient) UpgradeCluster(options *UpgradeClusterOptions) error {
 }
 
 func (c *TkgClient) DoLegacyClusterUpgrade(regionalClusterClient, currentClusterClient clusterclient.Client, options *UpgradeClusterOptions) error {
-	err := c.addKubernetesReleaseLabel(regionalClusterClient, options)
+	log.Info("Verifying kubernetes version...")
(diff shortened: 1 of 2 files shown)


anujc25 (approved):
Looks good. Thanks for the fix.

anujc25:
Please also file Github issue describing the bug.

PR #3053, opened 2022-08-02
Display correct error message when passing Cluster based resource without ClusterClass as part of cluster create

### LINKED ISSUE #3054: Add validation and throw proper error when Cluster resource without ClusterClass is passed with cluster create command
**Bug description**
Currently, when user passes Cluster resource as part of `tanzu cluster create --file <input.yaml>` and the cluster resource doesn't include ClusterClass as part of `.spec.topology.cluster` a wrong error message is thrown which can be misleading to the user.

As part of the error message, throw meaningful error message as this usecase is not yet supported.

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**
Should throw meaningful error message as this usecase is not yet supported.

anujc25:
- Display correct error message when passing Cluster based resource without ClusterClass as part of cluster create

Currently, when user passes `Cluster` resource as part of `tanzu cluster create --file <input.yaml>` and the cluster resource doesn't include ClusterClass as part of `.spec.topology.cluster` thrown proper error message.

Fixes #3054

- Added unit tests
- Manually verified by passing `--dru-run` output to `tanzu cluster create --file <manifest.yaml>` and verified the error message

### Release note

```release-note
Display correct error message when passing Cluster resource without ClusterClass as part of cluster create
```

### DIFF
--- a/pkg/v1/tkg/constants/clusterclass_mapping.go
+++ b/pkg/v1/tkg/constants/clusterclass_mapping.go
@@ -279,7 +279,8 @@ const (
 
 	SPEC = "spec"
 
-	TopologyClassIncorrectValueErrMsg = "input cluster class file, attribute spec.topology.class has no value or incorrect value or not following correct naming convension"
+	TopologyClassIncorrectValueErrMsg                = "input cluster class file, attribute spec.topology.class has no value or incorrect value or not following correct naming convention"
+	ClusterResourceWithoutTopologyNotSupportedErrMsg = "input file contains Cluster resource which doesn't have ClusterClass specified. Passing Cluster resource without ClusterClass specification is not supported"
 )
 
 // InfrastructureSpecificVariableMappingMap has, infra name to variable mapping map, which makes easy to get infra specific mapping map
--- a/pkg/v1/tkg/tkgctl/helpers_test.go
+++ b/pkg/v1/tkg/tkgctl/helpers_test.go
@@ -11,6 +11,7 @@ import (
 	. "github.com/onsi/gomega"
 
 	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkg/constants"
+	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkg/utils"
 )
 
 var _ = Describe("Cluster Class - IP Family Validation related test cases: ", func() {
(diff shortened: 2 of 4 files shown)


miclettej (approved):

pkg/v1/tkg/constants/clusterclass_mapping.go
```diff
@@ -279,7 +279,8 @@ const (
 
 	SPEC = "spec"
 
-	TopologyClassIncorrectValueErrMsg = "input cluster class file, attribute spec.topology.class has no value or incorrect value or not following correct naming convension"
+	TopologyClassIncorrectValueErrMsg                = "input cluster class file, attribute spec.topology.class has no value or incorrect value or not following correct naming convension"
```
miclettej, line 282: misspelling 'convension' should be 'convention'

anujc25: Thanks for catching this. fixed.

tenczar (approved):
LGTM

prkalle (approved):
LGTM

pkg/v1/tkg/constants/clusterclass_mapping.go
```diff
@@ -279,7 +279,8 @@ const (
 
 	SPEC = "spec"
 
-	TopologyClassIncorrectValueErrMsg = "input cluster class file, attribute spec.topology.class has no value or incorrect value or not following correct naming convension"
+	TopologyClassIncorrectValueErrMsg                = "input cluster class file, attribute spec.topology.class has no value or incorrect value or not following correct naming convention"
+	ClusterResourceWithoutTopologyNotSupportedErrMsg = "Input file contains Cluster resource which doesn't have ClusterClass specified. Passing Cluster resource without ClusterClass specification is not supported"
```
marckhouzam, line 283: Nit: the first error message starts with a lower-case (`input`) while the new one start with uppercase (`Input`).

I don't yet know how to trigger the error, so I don't have an opinion on which format is best but I was thinking we should be consistent.

pkg/v1/tkg/tkgctl/helpers_test.go
```diff
@@ -211,3 +212,66 @@ var _ = Describe("Cluster Class - IP Family Validation related test cases: ", fu
...
+spec:
+  clusterNetwork:
+    pods:
+      cidrBlocks: []
+`
+			})
+			It("should return false with error", func() {
+				Expect(isClusterClassBased).To(BeFalse())
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring(""))
```
marckhouzam, line 264: Should the new error be specified in `ContainSubstring()` here?

pkg/v1/tkg/tkgctl/helpers_test.go
```diff
@@ -211,3 +212,66 @@ var _ = Describe("Cluster Class - IP Family Validation related test cases: ", fu
...
+			})
+		})
+		When("File doesn't contain cluster resource", func() {
+			BeforeEach(func() {
+				configFileContent = ``
+			})
+			It("should return false with error", func() {
```
marckhouzam, line 271: Should the description of the test say "without error"?  That is what the check below seems to expect.

pkg/v1/tkg/tkgctl/helpers_test.go
chandrareddyp: +1

chandrareddyp:
LGTM

pkg/v1/tkg/tkgctl/helpers_test.go
anujc25: Thanks for catching this. Update this to use `ClusterResourceWithoutTopologyNotSupportedErrMsg`

marckhouzam:
LGTM

PR #3067, opened 2022-08-03
Fixes PCI validation to use decimal format

srm09:
As, CAPV expects the PCI inputs of deviceID:vendorID in a Decimal format, this patch fixes the validation to compare against the Decimal values of the Tesla T4 Hex values. 
It also sets the deviceID vendorID values in the map as an integer instead of a string.

Fixes #

- Created a cluster with GPU enabled worker nodes via PCI passthrough in a vSphere testbed.

### Release note
```release-note
NONE
```


jayunit100:
approveeeddd

jayunit100 (approved):

prkalle (approved):
LGTM. Thanks!

pkg/v1/providers/infrastructure-vsphere/v1.3.1/ytt/overlay.yaml
```diff
@@ -69,6 +38,41 @@
...
 #@ end
 
+
+#@ valid_pci_devices = [[0x10DE,0x1EB8]]
+
+#@ def valid_pci_devices_list():
+#@       return ",".join(map(lambda x:  "0x%X:0x%X" % (x[0], x[1]), valid_pci_devices))
+#@ end
+
+#@ def get_pci_devices(pci_devices_string, pci_ignore_pci_device_validation):
```
sathyanarays, line 48: NIT: The variable name could be either pci_ignore_device_validation or ignore_pci_device_validation.

pkg/v1/providers/infrastructure-vsphere/v1.3.1/ytt/overlay.yaml
```diff
@@ -69,6 +38,41 @@
...
+#@         if len(val) <= 0:
+#@             assert.fail(help_error)
+#@         end
+#@         kv = val.split(":")
+#@         if len(kv) != 2:
+#@             assert.fail(help_error)
+#@         end
+#@         id_pair = dict()
+#@         vendor_id = int(kv[0], 16)
+#@         device_id = int(kv[1], 16)
```
sathyanarays, line 61: Should we return specific error message if the deviceId or vendorId provided is not a valid hexadecimal?

pkg/v1/providers/infrastructure-vsphere/v1.3.1/ytt/overlay.yaml
```diff
@@ -69,6 +38,41 @@
...
+#@         if not pci_ignore_pci_device_validation:
+#@             for x in valid_pci_devices:
+#@                 if vendor_id != x[0] or device_id != x[1]:
+#@                     assert.fail("Device 0x%X:0x%X is not supported. Supported devices are: %s. Set VSPHERE_IGNORE_PCI_DEVICES_ALLOW_LIST to ignore this check." % (vendor_id, device_id, valid_pci_devices_list()))
+#@                 end
+#@             end
+#@         end
```
sathyanarays, line 68: Shouldn't the logic be like this?

```
#@        if not pci_ignore_pci_device_validation:
#@.            matchesAny = false
#@             for x in valid_pci_devices:
#@                 if vendor_id == x[0] or device_id == x[1]:
#@                     matchesAny = true
#@                 end
#@             end
#@             if not matchesAny:
#@                 assert.fail("Device 0x%X:0x%X is not supported. Supported devices are: %s. Set VSPHERE_IGNORE_PCI_DEVICES_ALLOW_LIST to ignore this check." % (vendor_id, device_id, valid_pci_devices_list()))
#@             end
#@         end
```

pkg/v1/providers/infrastructure-vsphere/v1.3.1/ytt/overlay.yaml
randomvariable: There's no try-catch support for this in Starlark unfortunately.

pkg/v1/providers/infrastructure-vsphere/v1.3.1/ytt/overlay.yaml
randomvariable: Apart from `if vendor_id == x[0] or device_id == x[1]:`
needing to be `if vendor_id == x[0] and device_id == x[1]:`, and for which i've now added a test case just now, yup. Changed.

pkg/v1/providers/infrastructure-vsphere/v1.3.1/ytt/overlay.yaml
randomvariable: Done

pkg/v1/providers/infrastructure-vsphere/v1.3.1/ytt/overlay.yaml
randomvariable: Added a comment that this needs to be covered in documentation.

sathyanarays (approved):

yastij (approved):

PR #3073, opened 2022-08-03
Automate create/delete CC Guest cluster on supervisor using Custom ClusterBootstrap and verify if all addons are successfully deployed

### LINKED ISSUE #2451: TC03: Able to create CC Guest cluster on supervisor using Custom ClusterBootstrap and verify if all addons are successfully deployed
Custom Clusterbootstrap deployment (Guest Cluster)

shivaani0505:

- Modified the helper function library (package.go) to work for TKGS and TKGm. 
- Created test case for custom bootstrap. Used the modified helper functions to validate custom bootstrap and packages
- Added test case for custom Antrea package 

Fixes #2451 

1. Verified by creating a classy cluster
2. Verified Antrea data values secret and output as follows:
3. Verified the packageInstalls for addons.
```
infraProvider: vsphere
antrea:
    config:
        serviceCIDR: 198.51.100.0/12
        trafficEncapMode: encap
        noSNAT: false
        disableUdpTunnelOffload: false
        tlsCipherSuites: TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_256_GCM_SHA384
        featureGates:
            AntreaProxy: true
            EndpointSlice: true
            AntreaPolicy: true
            FlowExporter: false
            Egress: false
            NodePortLocal: false
            AntreaTraceflow: false
            NetworkPolicyStats: false
            AntreaIPAM: false
            ServiceExternalIP: false
            Multicast: false
```

### Release note

```release-note
Automate create/delete CC Guest cluster on supervisor using Custom ClusterBootstrap and verify if all addons are successfully deployed
```

### DIFF
--- a/pkg/v1/tkg/test/data/testdata/customClusterBootstrapAntrea.yaml
+++ b/pkg/v1/tkg/test/data/testdata/customClusterBootstrapAntrea.yaml
@@ -1,0 +1,78 @@
+apiVersion: cni.tanzu.vmware.com/v1alpha1
+kind: AntreaConfig
+metadata:
+  name: cc-cb-antrea-package
+  namespace: customcb-ns
+spec:
+  antrea:
+    config:
+      disableUdpTunnelOffload: false
+      featureGates:
+        AntreaPolicy: true
+        AntreaProxy: true
+        AntreaTraceflow: false
+        Egress: true
+        EndpointSlice: true
+        FlowExporter: false
+        NodePortLocal: true
+      noSNAT: false
+      trafficEncapMode: encap
+---
+apiVersion: run.tanzu.vmware.com/v1alpha3
+kind: ClusterBootstrap
+metadata:
+  annotations:
+    tkg.tanzu.vmware.com/add-missing-fields-from-tkr: v1.23.5---vmware.1-tkg.1-zshippable
+  name: cc-cb
+  namespace: customcb-ns
+spec:
+  additionalPackages:
+    - refName: metrics-server*
+    - refName: secretgen-controller*
+    - refName: pinniped*
+  cni:
+    refName: antrea*
+    valuesFrom:
+      providerRef:
+        apiGroup: cni.tanzu.vmware.com
+        kind: AntreaConfig
+        name: cc-cb-antrea-package
+  kapp:
+    refName: kapp-controller*
+---
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: Cluster
+metadata:
+  name: cc-cb
+  namespace: customcb-ns
+spec:
+  clusterNetwork:
+    pods:
+      cidrBlocks:
+        - 192.0.2.0/16
+    serviceDomain: cluster.local
+    services:
+      cidrBlocks:
+        - 198.51.100.0/12
+  topology:
+    class: tanzukubernetescluster
+    controlPlane:
+      metadata: {}
+      replicas: 1
+    variables:
+      - name: storageClasses
+        value:
+          - wcpglobalstorageprofile
+      - name: ntp
+        value: time1.vmware.com
+      - name: vmClass
+        value: best-effort-small
+      - name: storageClass
+        value: wcpglobalstorageprofile
+    version: v1.23.5+vmware.1-tkg.1-zshippable
+    workers:
+      machineDeployments:
+        - class: node-pool
+          metadata: {}
+          name: np-2
+          replicas: 1
--- a/pkg/v1/tkg/test/data/testdata/customClusterBootstrapAntrea.yaml
+++ b/pkg/v1/tkg/test/data/testdata/customClusterBootstrapAntrea.yaml
@@ -1,7 +1,7 @@
 apiVersion: cni.tanzu.vmware.com/v1alpha1
 kind: AntreaConfig
 metadata:
-  name: cc-cb-antrea-package
+  name: cc-cb
   namespace: customcb-ns
 spec:
   antrea:
@@ -36,7 +36,7 @@ spec:
       providerRef:
         apiGroup: cni.tanzu.vmware.com
         kind: AntreaConfig
-        name: cc-cb-antrea-package
+        name: cc-cb
   kapp:
     refName: kapp-controller*
 ---
--- a/pkg/v1/tkg/test/tkgctl/shared/package.go
+++ b/pkg/v1/tkg/test/tkgctl/shared/package.go
@@ -7,6 +7,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"reflect"
+	"strings"
+	"time"
 
 	"golang.org/x/text/cases"
 	"golang.org/x/text/language"
@@ -20,10 +23,6 @@ import (
 
 	kapppkgv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apiserver/apis/datapackaging/v1alpha1"
 	runtanzuv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/run/v1alpha1"
-
-	"reflect"
-	"strings"
-	"time"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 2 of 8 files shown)


pkg/v1/tkg/test/tkgctl/shared/package.go
```diff
@@ -158,27 +162,33 @@ func getPackagesFromCB(ctx context.Context, clusterBootstrap *runtanzuv1alpha3.C
...
 
 	var clusterBootstrap *runtanzuv1alpha3.ClusterBootstrap
+
+	systemNamespace := constants.TkgNamespace
+	if infrastructureName == "tkgs" {
+		systemNamespace = constants.TKGSClusterClassNamespace
+	}
+
 	if isManagementCluster {
 		clusterBootstrap = getClusterBootstrap(ctx, mccl, mcClusterNamespace, mcClusterName)
```
maralavi, line 185: As there is no CB for supervisor cluster, here in line 184 you need to also do this check as well, so that fetching CB for the supervisor cluster is not attempted.
```
if isManagementCluster && infrastructureName != "tkgs" {..}
```

pkg/v1/tkg/test/tkgctl/shared/package.go
```diff
@@ -158,27 +162,33 @@ func getPackagesFromCB(ctx context.Context, clusterBootstrap *runtanzuv1alpha3.C
...
 	} else {
 		clusterBootstrap = getClusterBootstrap(ctx, mccl, wcClusterNamespace, wcClusterName)
 	}
 
 	By(fmt.Sprintf("Verify clusterbootstrap matches clusterbootstraptemplate"))
-	verifyClusterBootstrap(ctx, mccl, clusterBootstrap, clusterBootstrap.Status.ResolvedTKR)
+	verifyClusterBootstrap(ctx, mccl, clusterBootstrap, clusterBootstrap.Status.ResolvedTKR, systemNamespace)
 
 	packages, err := getPackagesFromCB(ctx, clusterBootstrap, mccl, wccl, mcClusterName, mcClusterNamespace, wcClusterName, wcClusterNamespace, infrastructureName, isManagementCluster)
 	Expect(err).NotTo(HaveOccurred())
```
maralavi, line 194:
```suggestion
if clusterBootstrap == nil {
     return
}

By(fmt.Sprintf("Verify clusterbootstrap matches clusterbootstraptemplate"))
verifyClusterBootstrap(ctx, mccl, clusterBootstrap, clusterBootstrap.Status.ResolvedTKR, systemNamespace)
packages, err := getPackagesFromCB(ctx, clusterBootstrap, mccl, wccl, mcClusterName, mcClusterNamespace, wcClusterName, wcClusterNamespace, infrastructureName, isManagementCluster)
Expect(err).NotTo(HaveOccurred())

Comment (maralavi) [reply pkg/v1/tkg/test/tkgctl/shared/package.go:194]: The reason for the above suggestion is that for supervisor cluster specifically, there is no CB, and the following checks would panic for that case

Comment (maralavi) [reply pkg/v1/tkg/test/tkgctl/shared/package.go:194]: Instead of checking for the nilness of ```clusterBootstrap```, alternatively you can do the following check. You decide which one you prefer:
```
  if isManagementCluster && infrastructureName == "tkgs" {
        return nil 
 }
```

File: pkg/v1/tkg/test/data/testdata/customClusterBootstrapAntrea.yaml [hunk 3]
```diff
@@ -0,0 +1,78 @@
+apiVersion: cni.tanzu.vmware.com/v1alpha1
+kind: AntreaConfig
+metadata:
+  name: cc20-antrea-package
```
maralavi, line 4: nit: why the name has 20 suffix?:)

pkg/v1/tkg/test/tkgctl/shared/tkr_cc_misc.go
```diff
@@ -146,7 +147,7 @@ func ValidateTKRsRelatedObjectsAvailability(ctx context.Context, mcProxy *framew
 		tkr := &runv1.TanzuKubernetesRelease{}
 		err := crClient.Get(ctx, client.ObjectKey{Name: expectedTKRs[i]}, tkr)
 		Expect(err).To(BeNil(), fmt.Sprintf("failed to get TKR :%s", expectedTKRs[i]))
-		cbt := getClusterBootstrapTemplate(ctx, crClient, tkr.Name)
+		cbt := getClusterBootstrapTemplate(ctx, crClient, tkr.Name, constants.TKGNamespace)
```
maralavi, line 150: Not related to your change, but wondering why this check is in ```tkr_cc_misc.go``` file?!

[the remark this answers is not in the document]
shivaani0505: I am preserving the behavior that was before. Previously, the call to getClusterBootstrapTemplate() would result in using constants.TKGNamespace here:
https://github.com/vmware-tanzu/tanzu-framework/blob/main/pkg/v1/tkg/test/tkgctl/shared/package.go#L388

Now, we are explicitly passing the same constants.TKGNamespace from the caller because we have modified the shared function to handle tkgs as well. So, I want to keep the behavior the same as before for ValidateTKRsRelatedObjectsAvailability()

shivaani0505: refer to first comment

pkg/v1/tkg/test/data/testdata/customClusterBootstrapAntrea.yaml
shivaani0505: I can use another name => cc-cb. Let me know if thats fine or Please suggest a name you like.

pkg/v1/tkg/test/tkgctl/shared/package.go
```diff
@@ -202,12 +212,15 @@ func checkClusterCB(ctx context.Context, mccl, wccl client.Client, mcClusterName
 		if strings.Contains(pkg.Name, "kapp-controller") {
 			verifyPackageInstall(ctx, mccl, pkg.Namespace, GeneratePackageInstallName(clusterName, pkg.Name), pkg.Spec.RefName, pkg.Spec.Version)
 		} else {
-			verifyPackageInstall(ctx, client, pkg.Namespace, GeneratePackageInstallName(clusterName, pkg.Name), pkg.Spec.RefName, pkg.Spec.Version)
+			// TODO: Verify the capabilities package install when it is fixed to reconcile
+			if infrastructureName != "tkgs" || pkg.Name != "capabilities" {
+				verifyPackageInstall(ctx, client, pkg.Namespace, GeneratePackageInstallName(clusterName, pkg.Name), pkg.Spec.RefName, pkg.Spec.Version)
```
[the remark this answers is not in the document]
shivaani0505: It does check for tkgs as well. It only skips the capabilities package for tkgs.

pkg/v1/tkg/test/tkgctl/shared/package.go
shivaani0505: Given CB only exists on supervisor cluster in workload cluster namespace for tkgs, from the test file itself we are not calling this for management cluster. Only the else is exercised.

pkg/v1/tkg/test/tkgctl/shared/package.go
shivaani0505: As mentioned before, for tkgs tests we are not calling this for supervisor cluster.

pkg/v1/tkg/test/data/testdata/customClusterBootstrapAntrea.yaml
maralavi: that sounds good!

pkg/v1/tkg/test/tkgctl/shared/package.go
maralavi: Oh ok, we are good then

pkg/v1/tkg/test/tkgctl/shared/package.go
maralavi: I see. If we are sure that to-be-added BYOT tests also don't ever run these on supervisor cluster, then we are good.

pkg/v1/tkg/test/tkgctl/shared/package.go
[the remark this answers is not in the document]
maralavi: Oh yes, you are right

maralavi:
@shivaani0505 please remember to add release note as well

pkg/v1/tkg/test/data/testdata/customClusterBootstrapAntrea.yaml
shivaani0505: Changed it to cc-cb

maralavi (approved):

PR #3071, opened 2022-08-03
Update makefile to download arm64 version of vale

seemiller:
### What this PR does / why we need it

The command to download the `vale` application that is used in the `lint` make task needs to support `arm64`.

### Which issue(s) this PR fixes

Fixes #3070 

### Describe testing done for PR

[vuil] on amd and apple silicon macs:
```
git clean -dfx .
make tools
```

### Release note

```release-note

```

PR #3088, opened 2022-08-05
Remove installation of kubeval

### LINKED ISSUE #3075: kubeval linter not used
**Bug description**

When running `make lint`, the tool `kubeval` is being installed but it does not seem to ever be used.
It is a bit inefficient but mostly, it prevents `make lint` from running on `darwin-arm64` because `kubeval` is not available for that architecture; therefore, `make lint` fails when trying to install `kubeval` even though the tool is not needed.

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- (x) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

Ultimately, the goal is to have `make lint` pass on `darwin-arm64`,

On a Mac M1, run `make lint` and you should see:
```
make -C /Users/kmarc/git/tanzu-framework/hack/tools kubeval
mkdir -p bin
curl -sL https://github.com/instrumenta/kubeval/releases/download/0.15.0/kubeval-darwin-arm64.tar.gz | tar -xz -C /tmp/
tar: Error opening archive: Unrecognized archive format
make[1]: *** [bin/kubeval] Error 1
make: *** [/Users/kmarc/git/tanzu-framework/hack/tools/bin/kubeval] Error 2
```

`main` branch: sha 46b4f80a

It seems that the use of `kukbeval` was removed in #204, but the installation of it was not removed.

Locally on Mac M1

marckhouzam:

This PR removes the installation of `kubeval` which is not used anymore.
Trying to install it breaks the `make lint` target on `arm64` since there is no `kubeval` release for that architecture.

Note that the use of `kubeval` was removed in #204.

Fixes #3075 

On a Mac ARM machine, ran `make tools` to see that the installation of `kubeval` failed before the PR and passes after applying the PR.

Also ran `make lint` to see that it no longer failed when trying to intall `kubeval`.  Note that this requires some tweaking for `vale` and `misspell` which also fail on ARM (there is a PR for vale already #3071).

/cc @vuil

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -82,12 +82,11 @@ KBLD               := $(TOOLS_BIN_DIR)/kbld
 VENDIR             := $(TOOLS_BIN_DIR)/vendir
 IMGPKG             := $(TOOLS_BIN_DIR)/imgpkg
 KAPP               := $(TOOLS_BIN_DIR)/kapp
-KUBEVAL            := $(TOOLS_BIN_DIR)/kubeval
 GINKGO             := $(TOOLS_BIN_DIR)/ginkgo
 VALE               := $(TOOLS_BIN_DIR)/vale
 YQ                 := $(TOOLS_BIN_DIR)/yq
 CONVERSION_GEN     := $(TOOLS_BIN_DIR)/conversion-gen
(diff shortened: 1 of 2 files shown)


vuil (approved):
Thanks for the changes! lgtm.

PR #3094, opened 2022-08-05
update error message while checking management cluster infra provider details

### LINKED ISSUE #3095: misleading error message while creating non-tkgs workload cluster
**Bug description**
While creating the non-tkgs cluster, there is an error message `unable to determine if management cluster is on vSphere with Tanzu` which is misleading end user. It says checking `the cluster is vSphere with Tanzu` even though the user is not on vSphere with Tanzu.

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**
The error message should be more generic
Create a management cluster (non-tkgs)
Update the context of the mc in ~/.config/tanzu/config.yaml with some incorrect value
Now create workload cluster, see the error message.

chandrareddyp:
Updates the error message as a generic error message while checking the infrastructure provider

Current error message specific to tkgs (vSphere with Tanzu) even though the user might not be on tkgs cluster which misleads the end user.

Fixes #3095 
Unite test cases are updated
Tested for TKGS cluster creation

### Release note

```release-note

```

### DIFF
--- a/pkg/v1/tkg/client/cluster.go
+++ b/pkg/v1/tkg/client/cluster.go
@@ -248,7 +248,7 @@ func (c *TkgClient) waitForClusterCreation(regionalClusterClient clusterclient.C
 	}
 	isTKGSCluster, err := regionalClusterClient.IsPacificRegionalCluster()
 	if err != nil {
-		return errors.Wrap(err, constants.ErrorMsgIsTKGSCluster)
+		return err
 	}
 	if isClusterClassBased {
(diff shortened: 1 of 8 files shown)


anujc25 (approved):
LGTM

PR #3103, opened 2022-08-09
Handle plugin errors properly

### LINKED ISSUE #44: unnecessary msg "Error: exit status 1" when a Tanzu CLI command failed
**Bug description**

The following tanzu command prints `Error: accepts 1 arg(s), received 0` and the help content, but also `Error: exit status 1` at the end. Seems there is no need to print `Error: exit status 1`.  I also noticed when `tanzu management-cluster create ...` prints the help context when its execution failed.

```
$ tanzu cluster kubeconfig get
Error: accepts 1 arg(s), received 0
Get kubeconfig of a cluster and merge the context into the default kubeconfig file

Usage:
  tanzu cluster kubeconfig get CLUSTER_NAME [flags]

Examples:
  
    # Get workload cluster kubeconfig
    tanzu cluster kubeconfig get CLUSTER_NAME    # Get workload cluster admin kubeconfig
    tanzu cluster kubeconfig get CLUSTER_NAME --adminFlags:
      --admin                Get admin kubeconfig of the workload cluster
      --export-file string   File path to export a standalone kubeconfig for workload cluster
  -h, --help                 help for get
  -n, --namespace string     The namespace where the workload cluster was created. Assumes 'default' if not specified.Global Flags:
        --log-file string   Log file path
  -v, --verbose int32     Number for the log level verbosity(0-9)

Error: exit status 1
✖  exit status 1
```

For example, 'kubectl get' doesn't print the exit status.
```
$ kubectl get 
You must specify the type of resource to get. Use "kubectl api-resources" for a complete list of supported resources.

error: Required resource not specified.
Use "kubectl explain <resource>" for a detailed description of that resource (e.g. kubectl explain pods).
See 'kubectl get -h' for help and examples
```

[ ] APIs
[ ] Addons
[X] CLI
[ ] Docs
[ ] Installation
[ ] Plugin
[ ] Security
[ ] Test and Release
[ ] User Experience

**Expected behavior**

$ tanzu version
version: v1.3.0-rc.2
buildDate: 2021-02-26
sha: bc3e781-dirty

### Collated Context

Context from 2021-03-10 07:23:20
User: jessehu
Here is the log when `tanzu management-cluster create ...` failed and it prints the usage content after the detail error msg. It's not expected to print the usage content in this command failure case.

```
[2021-03-09T16:53:50.149Z] Success waiting on all providers.
[2021-03-09T16:53:50.149Z] Start creating management cluster...
[2021-03-09T16:53:50.149Z] 
[2021-03-09T16:53:50.149Z] 
[2021-03-09T16:53:50.149Z] Failure while deploying management cluster, Here are some steps to investigate the cause:
[2021-03-09T16:53:50.149Z] 
[2021-03-09T16:53:50.149Z] Debug:
[2021-03-09T16:53:50.149Z]     kubectl get po,deploy,cluster,kubeadmcontrolplane,machine,machinedeployment -A --kubeconfig /home/capv/.kube-tkg/tmp/config_L5etLzc8
[2021-03-09T16:53:50.149Z]     kubectl logs deployment.apps/<deployment-name> -n <deployment-namespace> manager --kubeconfig /home/capv/.kube-tkg/tmp/config_L5etLzc8
[2021-03-09T16:53:50.149Z] 
[2021-03-09T16:53:50.149Z] To clean up the resources created by the management cluster:
[2021-03-09T16:53:50.149Z] 	  tanzu management-cluster delete
[2021-03-09T16:53:50.149Z] Error: unable to set up management cluster: unable to wait for cluster and get the cluster kubeconfig: error waiting for cluster to be provisioned (this may take a few minutes): cluster creation failed, reason:'VMProvisionFailed @ Machine/tkgextensions-bee09a7c0b33-control-plane-xmh5k', message:'1 of 2 completed'

[2021-03-09T16:53:50.149Z] Usage:
[2021-03-09T16:53:50.149Z]   tanzu management-cluster create [flags]
[2021-03-09T16:53:50.149Z] 
[2021-03-09T16:53:50.149Z] Examples:
[2021-03-09T16:53:50.149Z]   
[2021-03-09T16:53:50.149Z]     # Create a management cluster on AWS infrastructure, initializing it with
[2021-03-09T16:53:50.149Z]     # components required to create workload clusters through it on the same infrastructure
[2021-03-09T16:53:50.149Z]     # by bootstrapping through a self-provisioned bootstrap cluster.
[2021-03-09T16:53:50.149Z]     tanzu management-cluster create --file ~/clusterconfigs/aws-mc-1.yaml
[2021-03-09T16:53:50.149Z]     # Launch an interactive UI to configure the settings necessary to create a
[2021-03-09T16:53:50.149Z]     # management cluster
[2021-03-09T16:53:50.149Z]     tanzu management-cluster create --ui
[2021-03-09T16:53:50.149Z]     # Create a management cluster on vSphere infrastructure by using an existing
[2021-03-09T16:53:50.149Z]     # bootstrapper cluster. The current kube context should point to that
[2021-03-09T16:53:50.149Z]     # of the existing bootstrap cluster.
[2021-03-09T16:53:50.149Z]     tanzu management-cluster create --use-existing-bootstrap-cluster --file vsphere-mc-1.yaml
[2021-03-09T16:53:50.149Z] 
[2021-03-09T16:53:50.149Z] Flags:
[2021-03-09T16:53:50.149Z]   -b, --bind string                      Specify the IP and port to bind the Kickstart UI against (e.g. 127.0.0.1:8080). (default "127.0.0.1:8080")
[2021-03-09T16:53:50.149Z]       --browser string                   Specify the browser to open the Kickstart UI on. Use 'none' for no browser. Defaults to OS default browser. Supported: ['chrome', 'firefox', 'safari', 'ie', 'edge', 'none']
[2021-03-09T16:53:50.149Z]   -f, --file string                      Configuration file from which to create a management cluster
[2021-03-09T16:53:50.149Z]   -h, --help                             help for create
[2021-03-09T16:53:50.149Z]   -t, --timeout duration                 Time duration to wait for an operation before timeout. Timeout duration in hours(h)/minutes(m)/seconds(s) units or as some combination of them (e.g. 2h, 30m, 2h30m10s) (default 30m0s)
[2021-03-09T16:53:50.149Z]   -u, --ui                               Launch interactive management cluster provisioning UI
[2021-03-09T16:53:50.149Z]   -e, --use-existing-bootstrap-cluster   Use an existing bootstrap cluster to deploy the management cluster
[2021-03-09T16:53:50.149Z]   -y, --yes                              Create management cluster without asking for confirmation
[2021-03-09T16:53:50.149Z] 
[2021-03-09T16:53:50.149Z] Global Flags:
[2021-03-09T16:53:50.149Z]         --log-file string   Log file path
[2021-03-09T16:53:50.149Z]   -v, --verbose int32     Number for the log level verbosity(0-9)
[2021-03-09T16:53:50.150Z] 
[2021-03-09T16:53:50.150Z] Error: exit status 1
[2021-03-09T16:53:50.150Z] ✖  exit status 1 
```

marckhouzam:

Currently, when a plugin exists with an error the CLI will print `✖  exit status 1`.
However, in such a case, we shouldn't print the plugin exit code as a string but instead should use it as our own exit code.

This commit catches the fact that the error is due to a plugin exiting with error and behaves accordingly.

Furthermore, to avoid duplicating error message, this commit tells Cobra not to print final error messages since the CLI prints them itself.

Fixes #44 

I first recompile the `cluster` plugin to return an exit code of 2 just for testing.
Then I ran the following before applying the PR.  The first command uses the `cluster` plugin while the second command uses a native CLI command, so we can compare.
We can see:
1. for a plugin the `exit status 2` string is printed (twice)
2. the exit code of a plugin is not used as the CLI exit code (see the `===== errcode 1` which should be a value of 2)
3. for a native command, the error is printed twice

```
$ tz cluster kubeconfig get; echo "===== errcode $?"
Error: accepts 1 arg(s), received 0
Error: exit status 2

✖  exit status 2
===== errcode 1
$ tz plugin delete
Error: must provide plugin name as positional argument

✖  must provide plugin name as positional argument
```

Then I ran the same tests with the PR applied.  We can see that:
1. the CLI no longer prints the `exit status 1` string
2. upon error, the exit code of the plugin is also the exit code of the CLI
3. error strings are only printed once 
```
$ tz cluster kubeconfig get; echo "===== errcode $?"
Error: accepts 1 arg(s), received 0
===== errcode 2
$ tz plugin delete
✖  must provide plugin name as positional argument
```

### Release note

```release-note
A plugin error code is propagated to the core CLI.
When a command fails the error messages are no longer duplicated.
```

When dealing with native commands, it is the CLI that prints the error using "github.com/aunum/log".  The format is nicer as it uses colors and icons.  However it prints the error to `stdout` and there does not seem to be a way to print it to `stderr`.  I believe such errors should go to stderr.  I will open a separate issue to track this.

### DIFF
--- a/cmd/cli/tanzu/main.go
+++ b/cmd/cli/tanzu/main.go
@@ -4,7 +4,8 @@
 package main
 
 import (
-	"fmt"
+	"os"
+	"os/exec"
 
 	"github.com/aunum/log"
 
@@ -13,7 +14,14 @@ import (
 
 func main() {
 	if err := core.Execute(); err != nil {
-		fmt.Println("")
-		log.Fatal(err)
+		if errStr, ok := err.(*exec.ExitError); ok {
+			// If a plugin exited with an error, we don't want to print its
+			// exit status as a string, but want to use it as our own exit code.
+			os.Exit(errStr.ExitCode())
+		} else {
+			// We got an error other than a plugin exiting with an error, let's
+			// print the error message.
+			log.Fatal(err)
+		}
 	}
 }
--- a/pkg/v1/cli/command/core/root.go
+++ b/pkg/v1/cli/command/core/root.go
@@ -23,6 +23,9 @@ import (
 // RootCmd is the core root Tanzu command
 var RootCmd = &cobra.Command{
 	Use: "tanzu",
+	// Don't have Cobra print the error message, the CLI will
+	// print it itself in a nicer format.
+	SilenceErrors: true,
 }
 
 var (


cmd/cli/tanzu/main.go
```diff
@@ -13,7 +14,14 @@ import (
 
 func main() {
 	if err := core.Execute(); err != nil {
-		fmt.Println("")
-		log.Fatal(err)
+		if errStr, ok := err.(*exec.ExitError); ok {
+			// If a plugin exited with an error, we don't want to print its
```
anujc25, line 18: It's not just the plugin but applies to all commands. Plugin is also another cobra command which we have some mechanism written to invoke as a separate binary.
So, Basically Tanzu CLI is a combination of its own commands like `tanzu plugin` and some plugins like `tanzu cluster`. We might want to use command as a generic word to include all core commands as well as plugin invocation. This also applies to other comments as well as PR title and description.

```suggestion
			// If a command exited with an error, we don't want to print its
```

anujc25 (commented):
Changes looks great. Thanks for fixing this.

Just had a nit comment.

cmd/cli/tanzu/main.go
marckhouzam: Thanks for the clarification @anujc25!

In this particular case, IIUC, it is just plugins that will exit with an error of type `exec.ExitError` and provide an exit code because the plugin command calls the plugin as an external process.
For native cobra commands of the tanzu cli (like `tanzu plugin delete`) an error will be returned but of a different type and fall into the `else` case.

I tried to test both code paths by using `tz cluster kubeconfig get` to generate an error from a plugin and `tz plugin delete` to generate an error from a command.

Does that make sense, or did I mis-understand?

anujc25: Ahh. This makes sense. Thanks for the explanation.

anujc25 (approved):

marckhouzam:
The last forced-push was simply to re-trigger the code-coverage test which had failed for some reason...  It passed now.

PR #3106, opened 2022-08-09
add test for aws cc with custom clusterbootstrap

### LINKED ISSUE #2458: TC09: AWS Cluster - Able to create CC Guest cluster on AWS using Custom ClusterBootstrap and verify if all addons are successfully deployed

ggpaue:
add test for aws cc with custom clusterbootstrap

Fixes #2458 

### Release note

```release-note
Add cc cluster E2E tests with custom clusterbootstrap on AWS
```


pkg/v1/tkg/test/config/aws.yaml
```diff
@@ -19,3 +19,5 @@ tkg_config_variables:
   AWS_PRIVATE_SUBNET_ID_2: subnet-016eb6a45203c010e
 management_cluster_options:
   plan: dev
+workload_cluster_options:
```
shyaamsn, line 22: You can add a testcase in https://github.com/vmware-tanzu/tanzu-framework/tree/main/pkg/v1/tkg/test/tkgctl/aws_cc for customclusterbootstrap so that the test is run as part of the pipeline.

i.e. add 
```
var _ = Describe("Functional tests for aws (clusterclass) - Antrea", func() {
	E2ECommonSpec(context.TODO(), func() E2ECommonSpecInput {
		return E2ECommonSpecInput{
			E2EConfig:       e2eConfig,
			ArtifactsFolder: artifactsFolder,
			Cni:             "antrea",
			Plan:            "dev",
			Namespace:       "tkg-system",
			CustomClusterBootstrapFile: <filepath>,
			OtherConfigs:    map[string]string{"clusterclass": "true"},
		}
	})
})
```

maralavi: @ggpaue Please make  sure to update the description above, as ```aws_cc_antrea_test.go``` already uses this description

pkg/v1/tkg/test/tkgctl/aws_cc/aws_cc_ccb_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+var _ = Describe("Functional tests to create AWS cluster with custom ClusterBootstrap", func() {
+	E2ECommonSpec(context.TODO(), func() E2ECommonSpecInput {
+		return E2ECommonSpecInput{
+			E2EConfig:       e2eConfig,
+			ArtifactsFolder: artifactsFolder,
+			Cni:             "antrea",
+			Plan:            "dev",
+			Namespace:       "tkg-system",
+			OtherConfigs:    map[string]string{"clusterclass": "true"},
```
shyaamsn, line 23: You need to pass in something to say to execute custom clusterbootstrap. May be a boolean flag or pass in the customboostrap cc file

pkg/v1/tkg/test/config/aws.yaml
ggpaue: added test case for custom clusterbootstrap

pkg/v1/tkg/test/tkgctl/aws_cc/aws_cc_ccb_test.go
ggpaue: added a boolean flag to control

pkg/v1/tkg/test/config/aws.yaml
```diff
@@ -19,3 +19,5 @@ tkg_config_variables:
...
+  CLUSTER_CLASS_CB_FILE_PATH: ../../../test/data/testdata/customClusterBootstrapAntrea-aws.yaml
```
maralavi, line 23: Shouldn't this be ```../data/testdata/customClusterBootstrapAntrea-aws.yaml```, as these two files are both under ```test``` already and its only their subdirectories ```config``` vs ```data``` is different

maralavi:
@ggpaue you need to add release note and make sure all e2e tests pass (exception is azure).

pkg/v1/tkg/test/tkgctl/aws_cc/aws_cc_ccb_test.go
```diff
@@ -0,0 +1,27 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
```
maralavi, line 1: I suggest renaming this file to ```aws_cc_custom_cb_test.go``` cause ```ccb``` doesn't help those unfamiliar with the code, what the the test file is about.

pkg/v1/tkg/test/tkgctl/shared/package.go
```diff
@@ -693,3 +693,14 @@ func CheckTKGSAddons(ctx context.Context, tkgctlClient tkgctl.TKGClient, svClust
 
 	return nil
 }
+
+func ValidateClusterClassConfigFile(clusterclassConfigFilePath string) (string, string) {
```
maralavi, line 697: Question: We have same file name under ```pkg/v1/tkg/test/tkgctl/tkgs```. Which of these two is now duplicate and redundant and can be removed?

maralavi: potentially we could remove the other two instances and make sure they all use this one in a shared directory

ggpaue: @maralavi yes I added here because don't want to call functions from other tests, let me clean up the redundancies in this PR and only defined in shared/package.go

pkg/v1/tkg/test/config/aws.yaml
ggpaue: updated this file path

pkg/v1/tkg/test/tkgctl/aws_cc/aws_cc_ccb_test.go
ggpaue: changed file name to aws_cc_custom_cb_test.go

pkg/v1/tkg/test/tkgctl/shared/package.go
ggpaue: removed duplication function and let other test suite use package.go only

maralavi:
@ggpaue I guess your test for Antrea package would fails. Shouldn't ```CheckClusterCB``` be called with its ```isCustomCB``` parameter set to true? Toy can simply pass your ```IsCCB``` instead false when calling ```CheckClusterCB```?

ggpaue:
> @ggpaue I guess your test for Antrea package would fails. Shouldn't `CheckClusterCB` be called with its `isCustomCB` parameter set to true? Toy can simply pass your `IsCCB` instead false when calling `CheckClusterCB`?

thanks for the advice, I updated my code to pass ccb boolean when calling  CheckClusterCB

vijaykatam:
Change in https://github.com/vmware-tanzu/tanzu-framework/pull/3106/commits/c41676117538349476c02edb4a5e0ccea0f914a9 looks good to me

maralavi (approved):

PR #3120, opened 2022-08-11
explictly set the criSocket field in AWS and Azure to avoid race conditions

yastij:
Signed-off-by: Yassine TIJANI <ytijani@vmware.com>

This PR fixes an issue where we weren't setting the `criSocket` explicitly for Azure and AWS. What happens:

- containerd startup is slow, the containerd sock is not yet there
- kubeadm init sees that in our config in Azure and AWS we’re not specifying the CRISocketPath and tries to default that dynamically in-memory https://github.com/kubernetes/kubernetes/blob/b1aa1bd3088fad184cbb4fe36bd156dde7605ee4/cmd/kubeadm/app/util/config/initconfiguration.go#L111
- no socket is in there because containerd slowness, kubeadm defaults to the docker socket for backward compatibility
- kubeadm assumes docker for preflight checks, but fails since the binary isn’t installed

- tested that specifying the socket fixes this
- generated the ytt output for the management cluster, it shows that `kubeadmControlPlane` and `KubeadmConfigTemplate` had the field in the right place

### Release note

```release-note
explicitly set criSocket for KubeadmControlPlane and KubeadmConfigTemplate in AWS and Azure to avoid race conditions at the node level
```

cc @randomvariable


randomvariable:
run `packages/tkg-clusterclass/hack/sync-cc.sh` and we should be good.

randomvariable (approved):

PR #3127, opened 2022-08-12
UpdatesAvailable not showing all available TKRs

imikushin:
Make sure we include TKR and OSImage objects that are newer than the cluster into its UpdatesAvailable status condition calculation.

Without this fix, when new TKRs (and their OSImages) are added to the system, they are not going to be considered for UpdatesAvailable calculation for already existing clusters, until the clusters are modified (e.g. updating an annotation) or the TKR Status Controller is
restarted.

N/A

Manually on a vSphere Supervisor Cluster.

### Release note

```release-note
Fixed an issue with Cluster UpdatesAvailable condition calculation: 
when new TKRs (and their OSImages) are added to the system, they are not being considered 
for UpdatesAvailable calculation for already existing clusters, until the clusters are 
modified (e.g. updating an annotation) or the TKR Status Controller is restarted.
```


prkalle (changes requested):

prkalle (approved):
LGTM. Thanks @imikushin !

anujc25 (approved):

PR #3132, opened 2022-08-15
Dev experience fixes for package-tools

randomvariable:

- package-tools: Allow local registry to be overridden
- package-tools: Make commands verbose

I'm using the default install of docker-distribution from the Fedora repos locally which sets up a systemd unit for the registry to run on localhost:5000, not 5001.

Local development

### Release note

```release-note
NONE
```


vuil (approved):
lgtm, thanks!

yharish991:
@randomvariable these changes are causing the downstream package bundle building to fail.

PR #3134, opened 2022-08-15
Clarify error when discovery source is invalid

### LINKED ISSUE #3078: Misleading error message for unknown discovery source type
**Bug description**

When creating a discovery source with an invalid type (like a type named "random" or "mytype"), the error message says `Error: discovery source type 'random' is not yet supported`.

This is misleading as this discovery source type will never be supported 😄 

- (x) CLI

**Expected behavior**

Differentiate between discovery source types that are not supported and invalid ones by having a more appropriate error message for the case of an invalid discovery source type.  Something like:
```
$ tz plugin source add -n test -u test -t gcp
Error: discovery source type 'gcp' is not yet supported

$ tz plugin source add -n test -u test -t random
Error: unknown discovery source type 'random'
```

```
$ tanzu plugin source add -n test -u test -t random
Error: discovery source type 'random' is not yet supported
```

`main` branch, latest sha: c67deb70

Local

marckhouzam:

Clarify error when a plugin discovery source is invalid versus when it is not yet supported.

Fixes #3078

Before the PR we can see that an invalid discovery source of type `fish` gives the same error as an valid but unsupported discovery source type of `GCP`:
```
$ tanzu plugin source add -n test -u example.com/test --type fish
✖  discovery source type 'fish' is not yet supported
$ tanzu plugin source add -n test -u example.com/test --type GCP
✖  discovery source type 'GCP' is not yet supported
```
With the PR applied, only valid discovery source types give the `unsupported` error, while others give an `invalid` error:
```
$ tz plugin source add -n test -u example.com/test --type fish
✖  unknown discovery source type 'fish'
$ tz plugin source add -n test -u example.com/test --type GCP
✖  discovery source type 'GCP' is not yet supported
$ tz plugin source add -n test -u example.com/test --type kubernetes
✖  discovery source type 'kubernetes' is not yet supported
$ tz plugin source add -n test -u example.com/test --type rest
✖  discovery source type 'rest' is not yet supported
```

### Release note

```release-note
Improved error messages when the user attempts to add a discovery source of an invalid type.
```

### DIFF
--- a/pkg/v1/cli/command/core/discovery_source.go
+++ b/pkg/v1/cli/command/core/discovery_source.go
@@ -229,8 +229,10 @@ func createDiscoverySource(dsType, dsName, uri string) (configv1alpha1.PluginDis
 		pluginDiscoverySource.Local = createLocalDiscoverySource(dsName, uri)
 	case common.DiscoveryTypeOCI:
 		pluginDiscoverySource.OCI = createOCIDiscoverySource(dsName, uri)
-	default:
+	case common.DiscoveryTypeGCP, common.DiscoveryTypeKubernetes, common.DiscoveryTypeREST:
 		return pluginDiscoverySource, errors.Errorf("discovery source type '%s' is not yet supported", dsType)
+	default:
+		return pluginDiscoverySource, errors.Errorf("unknown discovery source type '%s'", dsType)
 	}
 	return pluginDiscoverySource, nil
 }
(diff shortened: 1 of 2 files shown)


anujc25 (approved):
LGTM. Thanks for the fix.

marckhouzam:
Thanks for the quick review @anujc25 !

Issue #2398, opened 2022-05-17
Go modules for apis/cni and apis/cpi breaks the controller-gen logic


12345lcr:
**Bug description**
https://github.com/vmware-tanzu/tanzu-framework/commit/8d668eb8b0963e54e841a59e5d04b7b2c38fbeb8
After the go modules are injected in apis, the `make manifests` can no longer generate the CRD.

**Affected product area (please put an X in all that apply)**

- [x] APIs
- [x] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [x] Test and Release
- [ ] User Experience

**Expected behavior**
The `make manifests` can generate CRDs.

**Steps to reproduce the bug**
1. Delete all in `config/crd/bases`
2. Run `make manifests`
3. All CRDs are added back except antrea, calico and cpi.
4. Remove the go.mod and go.sum in `apis/cni` and `apis/cpi`, run `make manifests` again, the missing CRDs are added back.

**Version** (include the SHA if the version is not obvious)

**Environment where the bug was observed (cloud, OS, etc)**
MacOS

**Relevant Debug Output (Logs, manifests, etc)**

stmcginnis:
May need to do something like this: https://github.com/kubernetes-sigs/cluster-api/pull/4442

Issue #3140, opened 2022-08-16
Create separate go modules for apis/config


anujc25:
**Describe the feature request**
Create separate go module for apis/config

- An api group should be not be split across go modules
- A go module can have multiple api groups

**Describe alternative(s) you've considered**
None

**Affected product area (please put an X in all that apply)**

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Additional context**

PR #3141, opened 2022-08-16
Create separate go modules for `apis/cli` and `apis/config`

### LINKED ISSUE #3139: Create separate go modules for apis/cli
Create separate go module for apis/cli

- An api group should be not be split across go modules
- A go module can have multiple api groups

The group `apis/cli` will have resources `catalogs` and `cliplugins`

None

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Additional context**

anujc25:
- Create separate go modules for `apis/cli`
- Create separate go modules for `apis/config`

Fixes #3139 and #3140

### Release note

```release-note
Create separate go modules for `apis/cli` and `apis/config`
```


marckhouzam:
This looks great @anujc25.

Comparing the `apis/cli` and `apis/config` with the existing `apis/cni` and `apis/cpi`, I notice that the latter both have a `Makefile` and are part of the main `Makefile` `manifests` target. These `Makefile`s were added to fix #2398.  I don't understand the `controller-gen` concept at the moment, but I thought I'd mention it, in case we need the same thing for the new go modules.

anujc25:
> Comparing the `apis/cli` and `apis/config` with the existing `apis/cni` and `apis/cpi`, I notice that the latter both have a `Makefile` and are part of the main `Makefile` `manifests` target. These `Makefile`s were added to fix #2398. I don't understand the `controller-gen` concept at the moment, but I thought I'd mention it, in case we need the same thing for the new go modules.

Thanks. Great catch. Updated the PR to update Makefiles.

pkg/v1/sdk/capabilities/Dockerfile
```diff
@@ -13,6 +13,7 @@ WORKDIR /workspace
 # Copy the Go Modules manifests
 COPY go.mod go.mod
 COPY go.sum go.sum
+COPY apis/ apis/
```
marckhouzam, line 16: Weird.  Are you seeing a failure when doing `make docker-build` because the new `go.mod` files are missing?  I do see such an error.  But why wasn't it a problem for the `apis/cni` and `apis/cpi` directories, which also have `go.mod` files?  I am confused.

apis/config/v1alpha1/clientconfig.go
```diff
@@ -155,55 +143,3 @@ func (c *ClientConfig) SetEditionSelector(edition EditionSelector) {
 	}
 	c.ClientOptions.CLI.UnstableVersionSelector = EditionStandard
 }
-
-// SetCompatibilityFile changes the compatibility file for the edition.
-func (c *ClientConfig) SetCompatibilityFile(edition EditionSelector) error {
```
anujc25, line 160: This function should not be part of the api definition and hence it was removed from here and moved to `pkg/v1/tkg/tkgctl/compatibility.go` as `SetCompatibilityFileBasedOnEdition`.

pkg/v1/cli/command/core/config.go
```diff
@@ -190,12 +190,6 @@ func setEdition(cfg *configv1alpha1.ClientConfig, edition string) error {
 	switch editionOption {
 	case configv1alpha1.EditionCommunity, configv1alpha1.EditionStandard:
 		cfg.SetEditionSelector(editionOption)
-		// when community edition is set, configure the compatibility file to use
-		// community edition's.
-		err := cfg.SetCompatibilityFile(editionOption)
```
anujc25, line 195: This was removed from core-cli code and moved to individual plugin command invocations.

marckhouzam: Ok, now I'm completely out of my element, but please be patient with me.

I was curious about the change in logic about `SetCompatibilityFile`.  Looking at the change, I'm guessing that this compatibility file is only used when creating a cluster?  That is why you only need to trigger it for the `managed-cluster` and `cluster` plugins?

But somehow (probably me doing something wrong), the main configuration file does not get updated properly anymore.  This is a simplification of what I did:
```
# Check that the main config file is set to tkg (BOM and edition)
$ egrep 'bom|edition' ~/.config/tanzu/config.yaml
    bomRepo: projects-stg.registry.vmware.com/tkg
    edition: tkg

# Set the edition to TCE
$ tanzu config set cli.edition tce
# =====> Notice the BOM is no longer set properly
$ egrep 'bom|edition' ~/.config/tanzu/config.yaml
    bomRepo: projects-stg.registry.vmware.com/tkg
    edition: tce

# Remove the BOM line completely
$ grep -v bom ~/.config/tanzu/config.yaml > tt && \mv tt ~/.config/tanzu/config.yaml && grep bom ~/.config/tanzu/config.yaml

# Regenerate the config file by running any command
$ tanzu
Tanzu CLI
Usage:
  tanzu [command]
[...]
# =====> Notice the BOM is added but does not match the edition
$ egrep 'bom|edition' ~/.config/tanzu/config.yaml
    bomRepo: projects-stg.registry.vmware.com/tkg
    edition: tce
```
I don't know if this is actually a problem but it does not happen on the main branch.

marckhouzam (commented):
Sorry if I'm all over the place, this is still pretty new to me...

prkalle (commented):
The changes looks good

pkg/v1/cli/command/core/config.go
anujc25: Thanks for catching that. I have fixed the problem and added unit test around `SetCompatibilityFileBasedOnEdition` function as well. 

Please note that `tanzu config set cli.edition tce` command will now just update the `edition` field in config and `BomRepo` and `Compatibility` information is updated when user runs `cluster` or `management-cluster` plugin command. Ideally, `BomRepo` and `Compatibility` information should not be part of `ClientConfig` at the first place and I have filled #3149 for that.

apis/config/v1alpha1/featuregate_webhook_test.go
```diff
@@ -331,3 +329,17 @@ func TestNamespaceConflicts(t *testing.T) {
...
+//
+// This function is intended to be used in tests for comparing expected and actual values, and printing the diff for
+// users to debug:
+//
+//   if diff := SliceDiffIgnoreOrder(got, want); diff != "" {
+//       t.Errorf("got: %v, want: %v, diff: %s", got, want, diff)
+//   }
+func SliceDiffIgnoreOrder(a, b []string) string {
+	return cmp.Diff(a, b, cmpopts.EquateEmpty(), cmpopts.SortSlices(func(x, y string) bool { return x < y }))
+}
```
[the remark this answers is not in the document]
anujc25: currently I don't think we have any package like that. `apis` are kind of core and we would like to minimize the dependency `apis` package has on other packages. But I agree with the observation that if there are more functions like this that can be used by other packages we should think about creating a top level library for that to avoid code duplication.

pkg/v1/sdk/capabilities/Dockerfile
anujc25: That's a really good question. I am not super sure on this but my guess is, none of the internal go modules is using `apis/cni` and `apis/cpi` as a direct dependency where as `apis/config` or `apis/cli` might be getting used as direct dependency.

marckhouzam: Ah yes, that makes sense.  I can see the `api/cli` and `api/config` in the `require` section of the main `go.mod` file but not the other two.  Good deduction 👍

cmd/cli/plugin/managementcluster/delete.go
```diff
@@ -45,6 +46,11 @@ func init() {
 }
 
 func runDeleteRegion(server *v1alpha1.Server) error {
+	err := tkgctl.SetCompatibilityFileBasedOnEdition()
+	if err != nil {
+		log.V(3).Infof("%v", err.Error())
+	}
+
```
marckhouzam, line 53: I don't know when this compatibility file is needed, so I can't review if all cases are covered.
However I notice that it is being set when deleting a management cluster but not a normal cluster.  Is that ok?

config/crd/bases/cni.tanzu.vmware.com_antreaconfigs.yaml
```diff
@@ -153,6 +153,18 @@ spec:
...
+                        properties:
+                          activeFlowTimeout:
+                            type: string
+                          collectorAddress:
+                            type: string
+                          idleFlowTimeout:
+                            type: string
+                          pollInterval:
+                            type: string
+                        type: object
```
marckhouzam, line 167: These seem to be valid but unrelated changes.  Are you purposefully including them or did they slip in silently in a `git add .` command?

(I personally am of the school of thought that different changes should be part of different PRs/commits, but that is just my opinion)

marckhouzam (approved):
Thanks for your patience and efforts @anujc25.  I think this will pave the way to easier future refactorings for everyone else.

The new tests have to be fixed (they don't compile), but the rest looks great 👍

cmd/cli/plugin/managementcluster/delete.go
anujc25: Yeah. Good observation. We need BoM file when deleting a management cluster because we need to spin-up cleanup kind cluster and deploy providers and etc to the cleanup kind cluster. But to delete the workload cluster we do not need to use BoM file and hence it is not been added for workload cluster deletion.

config/crd/bases/cni.tanzu.vmware.com_antreaconfigs.yaml
anujc25: Thanks. I will add it as part of separate commit. I think it was generated when I was testing `make generate` and `make manifest`. And I agree this should be added as part of separate commit. I will update the PR.

vuil (approved):
lgtm, thanks!

PR #3159, opened 2022-08-18
Makefile: Fix `make generate-go` to compile counterfeiter if missing

randomvariable:

`make generate-go` didn't actually work if counterfeiter wasn't available. This ensures it's a `make` target and gets auto installed prior to running `go generate ./...`

Fixes #

```
rm hack/tools/bin/*
make generate-go
```

### Release note

```release-note
NONE
```

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -66,9 +66,10 @@ endif
 
 
 # Directories
-TOOLS_DIR := $(abspath $(ROOT_DIR)/hack/tools)
+TOOLS_DIR_RELATIVE := hack/tools
+TOOLS_DIR := $(abspath $(ROOT_DIR)/$(TOOLS_DIR_RELATIVE))
 TOOLS_BIN_DIR := $(TOOLS_DIR)/bin
-
+TOOLS_DIR_DEPS := $(TOOLS_DIR)/go.sum $(TOOLS_DIR)/go.mod $(TOOLS_DIR)/Makefile
 
 # Add tooling binaries here and in hack/tools/Makefile
 CONTROLLER_GEN     := $(TOOLS_BIN_DIR)/controller-gen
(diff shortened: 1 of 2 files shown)


common.mk
```diff
@@ -114,3 +116,6 @@ generate-manifests:
 
 fmt: $(GOIMPORTS) ## Run goimports
 	$(GOIMPORTS) -w -local github.com/vmware-tanzu ./
+
+$(TOOLS_DIR_RELATIVE)/bin/%: $(TOOLS_DIR_DEPS)
```
marckhouzam, line 120: Could you confirm why you are adding these targets?
Is it to allow to build individual tools more easily? For example, instead of having to do `make tools` we can now just build `counterfeiter` by doing `make hack/tools/bin/counterfeiter`?

That is nice.

hack/tools/Makefile
```diff
@@ -74,7 +75,7 @@ $(GINKGO): go.mod
 	mkdir -p $(BIN_DIR)
 	go build -tags=tools -o $(BIN_DIR)/ginkgo github.com/onsi/ginkgo/ginkgo
 
-COUNTERFEITER := $(BIN_DIR)/counterfeiter
+counterfeiter: $(COUNTERFEITER)
 $(COUNTERFEITER): $(BIN_DIR) go.mod go.sum
```
marckhouzam, line 79: I realize you didn't change this but I wonder why there is a dependency on `$(BIN_DIR)`?  It may look reasonable in case the whole directory is removed, but in practice it actually breaks things in that case because there is no target for `$(BIN_DIR)`.

```
$ cd hack/tools;make clean;cd -
rm -rf bin
~git/tanzu-framework
$ make hack/tools/bin/counterfeiter
make -C /Users/kmarc/git/tanzu-framework/hack/tools bin/counterfeiter
make[1]: *** No rule to make target `bin', needed by `bin/counterfeiter'.  Stop.
make: *** [hack/tools/bin/counterfeiter] Error 2
```
but if removing that dependency, it will pass because the `go build` will create the directory.

Just a thought 😄

marckhouzam (approved):
Nice!  Just a couple of questions if you permit me.

LGTM

common.mk
randomvariable: Yes, it's based on the same stuff I did here https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/Makefile#L407-L409

randomvariable:
Lack of activity.

PR #3157, opened 2022-08-18
Update to Cluster API v1.2.3 in go dependencies

randomvariable:
### What this PR does / why we need it
- Update Cluster API and related dependencies to v1.2.3

The providers itself will be updated in a separate PR (#3513). The API changes are compatible.

Fixes # N/A

### Release note

```release-note
Update Cluster API go dependencies to v1.2.3, controller-runtime to 0.12 and Kubernetes to 1.24
For AWS, the following actions have been added to the policy for Cluster API AWS controllers:
* ec2:ReplaceRoute
* elasticloadbalancing:DeleteTargetGroup
```

PR #3161, opened 2022-08-18
Github actions: Replace old go get with go mod

randomvariable:

Updating dependencies in #3157 seems to fail on `go get -v -t -d ./...` in an infinite loop or errors that don't make sense. Given that we should be using `go mod download` instead, this updates old legacy behaviour to that.

N/A - This only happens in a Github action

### Release note

```release-note
NONE
```


randomvariable:
Closing for lack of interest

PR #3170, opened 2022-08-22
Fix Question mark prompt color

### LINKED ISSUE #292: Question mark prompt color
**Bug description**

The [styleguide](https://github.com/vmware-tanzu/tanzu-framework/blob/main/docs/cli/style_guide.md#color) calls for question prompts in cyan (informational), currently they are showing up green (success)

example:
screenshot of cli output showing a cyan question mark

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [x] User Experience

**Expected behavior**

mpanchajanya:

Fix the Question mark prompt color as per style guide and refactor the code to use survey pkg directly.

Fixes #292 

Screen Shot 2022-08-23 at 22 22 02

### Release note

```release-note
Any component using prompt.go will have the question mark color changed from green to cyan.

```


marckhouzam:
Hi @mpanchajanya.  Could you explain why there is a refactoring of using the survey package?  I haven't looked at the code, but it would help me to understand the reason behind the refactoring.  Thanks 👍

anujc25:
@mpanchajanya We should continue to use component library for this functionality and fix things in the `component` package itself rather than updating it in individual plugins. Another reason to continue use `component` package is that, not all plugins are developed within this repository but plugin developers uses this `component` package as part of plugin development.

anujc25:
@mpanchajanya can you add screenshots or sample output of how prompt looks like with the recent change as part of PR description under `Describe testing done for PR` section.

marckhouzam (approved):
LGTM

Nice work @mpanchajanya

anujc25 (approved):
Changes looks good.

@mpanchajanya can you please fill release note section as well in the PR description?

pkg/v1/cli/component/prompt_test.go
```diff
@@ -66,3 +66,17 @@ func Test_translatePromptConfig_Input(t *testing.T) {
...
+	assert := assert.New(t)
+
+	options := defaultPromptOptions()
+	opts := translatePromptOpts(options)
+
+	assert.NotNil(options)
+	assert.NotNil(opts)
+	assert.Equal("?", options.Icons.Question.Text)
+	assert.Equal("cyan+b", options.Icons.Question.Format)
+	assert.Equal(2, len(opts))
```
marckhouzam, line 80: I think it would be a good test to make sure that that `translatePromptOpts()` does not miss any fields.
I think you could create an empty `AskOptions` variable and apply both `opts` to it and then make sure the contents match what is expected.

marckhouzam (commented):
Great that you added a test!

PR #3171, opened 2022-08-22
Fix incorrect log message when creating management-cluster

anujc25:
- Fix incorrect log message when creating management-cluster
- Instead of using `ClusterClass based management cluster config file has been generated` should use `Management cluster config file has been generated`

Fixes #

### Release note

```release-note
Fix incorrect log message when creating management-cluster
```


vuil (approved):
lgtm, thanks.

PR #3173, opened 2022-08-22
Fix template issues in vsphere clusterclass implementation

### LINKED ISSUE #3439: ClusterClass based capv Clusters don't provision successfully.
**Bug description**

Running `tanzu management-cluster create` with a vsphere cluster config file and the `package-based-lcm-beta` flag set to true fails to provision a classy capv cluster.

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- (x) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

Expect to successfully deploy a classy capv cluster.

tenczar:
This PR updates ytt and go text templates for vsphere clusterclass that prevented successful deployment of clusterclass based capv clusters. Changes include:
- Fix indentation issue in VsphereMachineTemplate when updating network settings
- Fix the kube-vip image and tag declaration to properly access resources from TKR_DATA
- Adds a regex matcher for applying node labels. This prevents values from the TKR_DATA being used as node-labels if those values are not valid node-label values
- Fix machineHealthCheck definition to only add this block if a cluster definition specifies mhcs.
- Fixes credential values
- Fixes an issue with incorrectly typed config values being passed to the clusterbootstrap
- 

Fixes #3439 

Manually deployed a clusterclass based management cluster using Antrea and validated that the cluster is created successfully.

### Release note

```release-note
Fixes vsphere clusterclass definition to successfully deploy clusterclass capv clusters.
```

### DIFF
--- a/packages/tkg-clusterclass-vsphere/bundle/config/upstream/base.yaml
+++ b/packages/tkg-clusterclass-vsphere/bundle/config/upstream/base.yaml
@@ -75,11 +75,6 @@ metadata:
 spec:
   template:
     spec:
-      machineTemplate:
-        infrastructureRef:
-          apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
-          kind: VSphereMachineTemplate
-          name: 'tkg-vsphere-control-plane'
       kubeadmConfigSpec:
         useExperimentalRetryJoin: true
         clusterConfiguration:
@@ -161,7 +156,8 @@ spec:
       - echo "127.0.0.1   localhost" >>/etc/hosts
       - echo "127.0.0.1   {{ ds.meta_data.hostname }}" >>/etc/hosts
       - echo "{{ ds.meta_data.hostname }}" >/etc/hostname
-      postKubeadmCommands: []
+      postKubeadmCommands:
+      - echo "running postKubeadmCommands…"
       files:
       - path: /tmp/empty.txt
         owner: "root:root"
@@ -202,7 +198,6 @@ spec:
         kind: VSphereMachineTemplate
         apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
         name: 'tkg-vsphere-control-plane'
-    machineHealthCheck:
   infrastructure:
     ref:
       apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
@@ -212,7 +207,6 @@ spec:
   workers:
     machineDeployments:
     - class: tkg-worker
-      machineHealthCheck:
       template:
         bootstrap:
           ref:
@@ -227,7 +221,6 @@ spec:
             name: 'tkg-vsphere-worker'
             namespace: #@ data.values.NAMESPACE
     - class: tkg-worker-windows
-      machineHealthCheck:
       template:
         bootstrap:
           ref:
@@ -712,7 +705,7 @@ spec:
                     value: "20"
                   - name: vip_retryperiod
                     value: "4"
-                  image: {{(index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec.imageRepository}}/kube-vip:{{(index (index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec "kube-vip").imageTag}}
+                  image: {{(index (index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec "kube-vip").imageRepository}}/kube-vip:{{(index (index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec "kube-vip").imageTag}}
                   imagePullPolicy: IfNotPresent
                   name: kube-vip
                   resources: {}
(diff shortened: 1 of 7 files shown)


pkg/v1/providers/config_default.yaml
```diff
@@ -174,9 +174,9 @@ NSXT_PASSWORD: ""
 #! NSX-T host
 NSXT_MANAGER_HOST: ""
 #! set this to true if NSX-T uses self-signed cert
-NSXT_ALLOW_UNVERIFIED_SSL: "false"
+NSXT_ALLOW_UNVERIFIED_SSL: false
```
vuil, line 177: @tenczar : this change enables legacy config file not specifying this value to work correctly with the vsphere CC, but  this changes the behavior when we need to use the equivalent legacy config file in fallback (pure client side ytt) mode. 

Either existing logic that assumed string instead of bool will have to be updated (see ytt/02_addons/cpi/cpi_addon_data.lib.yaml and associated vendir-ed code)

or we should consider retaining the definition/semantics of the variables in config_default.yaml

(same for NSXT_REMOTE_AUTH)

lubronzhan: >Either existing logic that assumed string instead of bool will have to be updated (see ytt/02_addons/cpi/cpi_addon_data.lib.yaml and associated vendir-ed code)
If we change the existing logic, will the upgrade break, since customer's previous value is still string

DanielXiao: Thanks for reminding me. We will look for convertion for CC to avoid breaking existing logic.

DanielXiao:
@tenczar PR to fix CPI/CSI issue is https://github.com/vmware-tanzu/tanzu-framework/pull/3320

pkg/v1/providers/yttcb/clusterbootstrap.yaml
```diff
@@ -181,6 +181,7 @@ metadata:
   annotations:
     tkg.tanzu.vmware.com/add-missing-fields-from-tkr: #@ data.values.KUBERNETES_RELEASE
 spec:
+  #@ if data.values.CNI != "antrea" or antrea_configs_exist():
```
wjun, line 184: What is this condition for?

tenczar: removed.

packages/tkg-clusterclass-vsphere/bundle/config/upstream/base.yaml
```diff
@@ -749,14 +742,14 @@ spec:
           template: |
             devices:
             - networkName: {{ .vcenter.network }}
-            {{ if .controlPlane.network.nameservers -}}
-              nameservers:
-              {{- range .controlPlane.network.nameservers }}
-            - {{ . }}
+              {{ if .controlPlane.network.nameservers -}}
+                nameservers:
```
ydp, line 746: Hi Nick, are you intensionally adding 2 spaces before `nameservers` or it's a typo?

packages/tkg-clusterclass-vsphere/bundle/config/upstream/base.yaml
```diff
@@ -892,13 +885,15 @@ spec:
           template: |
             {{ $first := true }}
             {{- range $key, $val := (index .TKR_DATA .builtin.controlPlane.version).labels }}
+            {{- if regexMatch "^(?:[a-zA-z])(?:[-\\w\\.]*[a-zA-z])$" $val }}
```
ydp, line 888: Hi Nick, I'm trying to understand the code, this part of jsonPath has nothing to do with the custom Node Labels feature https://docs.vmware.com/en/VMware-Tanzu-Kubernetes-Grid/1.5/vmware-tanzu-kubernetes-grid-15/GUID-tanzu-k8s-clusters-config-plans.html#custom-node-labels-8 , right? It's specifically for Node Pool.

tenczar: It is both. This allows customers to define custom node labels and they can do it per node pool.

packages/tkg-clusterclass-vsphere/bundle/config/upstream/base.yaml
tenczar: This is intentional. This whole block was not correctly intended previously.

imikushin (approved):
Looks good. Thanks @tenczar! (for the change and for your patience)

packages/tkg-clusterclass-vsphere/bundle/config/upstream/base.yaml
ydp: Thank you @tenczar , though it looks strange that line 746 is not align with 744 and 748, suppose it should also work.
```
 744           - networkName: {{ .vcenter.network }}
 745             {{ if .controlPlane.network.nameservers -}}
 746               nameservers:
 747               {{- range .controlPlane.network.nameservers }}
 748             - {{ . }}
 749               {{- end }}
 750             {{- end }}
```

PR #3178, opened 2022-08-23
Add automation tests for upgrade aws cluster with custom clusterbootstrap

### LINKED ISSUE #2459: TC10: AWS Cluster - Ability to upgrade CC workload cluster on AWS using Custom ClusterBootstrap and verify if all addons are upgraded successfully

maralavi:

Add automation tests for upgrade aws workload cluster with custom clusterbootstrap

Fixes #2459 

Added automation tests as part of the PR

### Release note

```release-note
Add automation tests for upgrade aws workload cluster with custom clusterbootstrap
```


pkg/v1/tkg/test/tkgctl/shared/e2e_common.go
```diff
@@ -13,17 +13,15 @@ import (
 	"strings"
 	"time"
 
-	"sigs.k8s.io/cluster-api/util"
```
maralavi, line 16: Note: as ```e2e_common.go``` is used by all pre-utkg tests as well, its better not to add utkg specific tests here, as it cause clashes when certain functionality get called from there, for example when resources functionalities interacting with specific version of YanzuKubernetesRelease be called, etc.

Because of that reverting this file to almost what it was prior to #3106 and define a new file for utkg related tests

adduarte:
@maralavi did you run the test locally? if so can you add details to message. Thanks.

adduarte:
LGTM, lets wait for test to finish and look good.

pkg/v1/tkg/test/tkgctl/shared/e2e_tkrresolver.go
```diff
@@ -236,10 +227,6 @@ func E2ETKRResolverValidationForClusterCRUDSpec(context context.Context, inputGe
 
 		// verify addons are deployed successfully in clusterclass mode
 		if isClusterClass, ok := input.OtherConfigs["clusterclass"]; ok && isClusterClass == "true" {
-			By(fmt.Sprintf("Verify addon packages on management cluster %q matches clusterBootstrap info on management cluster %q", input.E2EConfig.ManagementClusterName, input.E2EConfig.ManagementClusterName))
```
HL-EverGreen, line 239: question: why can we skip checking addons on the management cluster?

pkg/v1/tkg/test/tkgctl/shared/e2e_common_cc.go
```diff
@@ -0,0 +1,280 @@
...
+		By(fmt.Sprintf("Deleting workload cluster %q", clusterName))
+		err = tkgCtlClient.DeleteCluster(tkgctl.DeleteClustersOptions{
+			ClusterName: clusterName,
+			Namespace:   namespace,
+			SkipPrompt:  true,
+		})
+		Expect(err).To(BeNil())
+
+		// verify addon resources are deleted successfully in clusterclass mode
+		if isClusterClass, ok := input.OtherConfigs["clusterclass"]; ok && isClusterClass == "true" {
```
HL-EverGreen, line 151: nit: shall we get rid of `input.OtherConfigs["clusterclass"]` here since we have individual test suite for CC? Also might be able to remove it from the test entries in `aws_cc` folder.

haoranleo (approved):
LGTM. Thanks for separating the cluster based tests!

pkg/v1/tkg/test/tkgctl/shared/e2e_tkrresolver.go
maralavi: Cause the individual tests are only deploying a workload cluster and no need to check addons deployments on the otherwise same management cluster. This check doesn't cause any problems in most test cases, but in the ```aws_cc_custom_cb_test.go```  testcase, which uses custom CB, the following check only holds for workload cluster and would fail for an existing managment cluster: https://github.com/vmware-tanzu/tanzu-framework/blob/main/pkg/v1/tkg/test/tkgctl/shared/package.go#L283

I can add a flag check for ```IsCustomCB``` before doing that test, but will make the code a bit dirty

pkg/v1/tkg/test/tkgctl/shared/e2e_common_cc.go
maralavi: Fair enough, I'll see what I can drop here

maralavi:
> @maralavi did you run the test locally? if so can you add details to message. Thanks.

Yes @adduarte  I modified only AWS related files and ran the test locally as well.

pkg/v1/tkg/test/tkgctl/shared/e2e_tkrresolver.go
HL-EverGreen: That makes sense, thanks for the explaination! I'm ok with removing the check on management cluster in this case

maralavi: I kept the check as much as possible; Instead, I made changes in [this line](https://github.com/vmware-tanzu/tanzu-framework/blob/main/pkg/v1/tkg/test/tkgctl/shared/package.go#L283) of code which had not considered management cluster before (as it was originally written with only tkgs in mind maybe)

ggpaue (approved):
thank you for creating the separate e2e common for cc tests, LGTM

haoranleo (approved):
Latest commit looks good. Thanks for incorporating the change!

vijaykatam (approved):

PR #3185, opened 2022-08-24
Turns on metadata versioning for addons-manager package

### LINKED ISSUE #3186: addons-manager package in management repo needs metadata version annotation
**Bug description**
Since the addons-manager package will exists in two repositories (core and management), the metadata file for the package bundle needs to be versioned in order for repository reconciliation to be succeful, because if the metadata files are not identical (which they are not)  reconciliation will fail. 

- ( ) APIs
- ( x) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

To reproduce, deploy a cluster and enable managment repostiory which has addons-manager package. Reconciliation of the management repository will fail with warning about metadata files not being identical. 

adduarte:
Metadata versioning allows a package to move from one repo to another
without having to have idential metadata

From carvel tools documentation at:
https://carvel.dev/kapp-controller/docs/develop/packaging/#combinations-and-special-cases

If a Package[Metadata] is provided by multiple PKGRs and despite having the same name/refname/version it is known that one of the resources is an update or later revision, the annotation key “packaging.carvel.dev/revision” can be used to provide an ordering.
- Resources without the annotation are considered to have version -1
- Resources with the revision annotation should have a value of the form “int” or “int.int” or “int.int.int.int.int….”
- Revision 0 will be considered < revision 0.0 (longer revision is “greater than” shorter revision of the same numbers)
- Revision 1.1.0 will be considered < revision 1.2.0 (semver sensibilities, but we don’t parse ‘+’ or other extensions)

Fixes #3186 

Deployed 1.6 testbed and uploaded new management repository with the changes. 
The repository was reconciled correctly. 

### Release note

```release-note
Adds metadata versioning to the addons-manager package bundle
```

### DIFF
--- a/packages/addons-manager/metadata.yaml
+++ b/packages/addons-manager/metadata.yaml
@@ -3,6 +3,26 @@ kind: PackageMetadata
 metadata:
   name: addons-manager.tanzu.vmware.com
   namespace: addons-manager
+
+  # The packaging.carvel.dev/revision annotation is added to help Kapp-controller
+  # decide which version of the file to select as source of truth in the case that
+  # a package is defined in multiple repositories.
+  # See here for more details: https://carvel.dev/kapp-controller/docs/develop/packaging/#combinations-and-special-cases
+  #
+  # The  string "2.0" was chosen because it was during the creation of tkg 2.0 that we introduced the annotation.
+  # The string in this file should never require changing. It is not tied to a release or required to be tied to a
+  # release or any other "version" number. See cited documentation for string format requirements.
+  #
+  # The string "2.0" was used in this file because it was during tkg 2.0 that the annotation was first introduced. Because
+  # it was the first time it was used in this package, the string could have been chosen to be anything.
+  # If at any other time, a new package-bundle with the same name, addons-manager.tanzu.vmware.com, is defined in a DIFFERENT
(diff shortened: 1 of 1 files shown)


ggpaue (approved):
LGTM

packages/addons-manager/metadata.yaml
```diff
@@ -3,6 +3,8 @@ kind: PackageMetadata
 metadata:
   name: addons-manager.tanzu.vmware.com
   namespace: addons-manager
+  annotations:
+    packaging.carvel.dev/revision: "1.0"
```
maralavi, line 7: Why do we start with revision 1.0 to begin with? And does this versioning applies to all packages in the repository?

ggpaue: I think this is the first annotation we add for revision, and other pacakges without this annotation will be considered to have version -1 to avoid conflict

adduarte: correct. Since this is the first time we introduce this versionning, we use 1,.0 because it will allow any future changes to go higher or lower. We can't predict why the change would happen but we can allow for all options.

PR #3201, opened 2022-08-26
Create separate directory for cli integration library and make it a go module

### LINKED ISSUE #3213: Create a separate go module for tanzu cli plugin integration library
At present, the tanzu framework is a combination of many different components and it shares a common root level go.mod file for all the components. To develop a simple plugin user needs to import some common structures and interfaces as well as some generic functions for accessing the Tanzu configuration file (~/.config/tanzu/config.yaml). This requires plugin developers to import the entire tanzu-framework go module which brings lots of other dependencies that plugin developers do not need. Sometimes, dependencies present in tanzu-framework conflict with the go dependencies that the plugin wants to use and it becomes an issue for plugin developers.

To avoid all this type of issues, the proposal is to have a core cli integration library as an independent go module that plugin authors can import. This independent go module for the integration library should contain very specific functionalities that almost all plugins will need and should contain a minimal set of dependencies within it.

None

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Additional context**

anujc25:
This PR moves some code base that every plugin requires from core-cli `pkg/v1` to `cli/runtime` module.

- Create separate directory `cli/runtime` for cli integration library and make it a go module.
    - Implements `plugin` development related helpers
    - Implements `component` library related helpers
        - Additional function `AskForConfirmation` is also migrated to this package from `pkg/v1/cli → cli/runtime/component`
    - Implements `client config` related helpers
    - Implement `test` package that includes test plugin related functions
        - `pkg/v1/test/cli/framework.go → cli/runtime/test`
        - This package also includes `NewTestFor` function migrated from `pkg/v1/cli → cli/runtime/test`
    - Implements `command` package that includes command specific generic helper function
        - Currently this package includes Deprecation related helper functions migrated from `pkg/v1/cli → cli/runtime/command`

- Update references in the repository to use `cli/runtime` module for all existing dependencies

Fixes #3213

- All existing build and test pipelines should be successful.

### Release note

```release-note
Create separate directory `cli/runtime` for cli integration library and make it a go module.
```

### Additional information

**Impact on existing refactoring PRs that might need rebase:**
1. PR authors might need to rebase their PRs considering `import` path references changes.
2. As you can see in [this commit](https://github.com/vmware-tanzu/tanzu-framework/pull/3201/commits/0f76bfaee1d3f593b35170795f638ecf0bdbe5c3) there are some import path changes as below that might create conflicts and we should try to use new paths when resolving the conflicts:
    - `github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli/component` -> `github.com/vmware-tanzu/tanzu-framework/cli/runtime/component`
    - `github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli/command/plugin` -> `github.com/vmware-tanzu/tanzu-framework/cli/runtime/plugin`
    - `github.com/vmware-tanzu/tanzu-framework/pkg/v1/config` -> `github.com/vmware-tanzu/tanzu-framework/cli/runtime/config`
3. If you are creating a new go module after refactoring, you will likely need to update `go.mod` of your module to include and point `cli/runtime` module to local directory path. `github.com/vmware-tanzu/tanzu-framework/cli/runtime => ../cli/runtime` Sample:https://github.com/vmware-tanzu/tanzu-framework/pull/3201/commits/0f76bfaee1d3f593b35170795f638ecf0bdbe5c3#diff-0e79a95786ea0956ef098d5e33af2ff66a8b2e277e8afa22512a3948b8a18e41R12


vuil:
This has the potential to affect a good number of pending PRs. @anujc25 Can you point out  more explicitly (in the PR description perhaps) the kind of changes one may expect to have to make to their PRs to account for this change?

prkalle (commented):
The changes LGTM. 
I would wait for others to review as well. Thanks!

cli/runtime/plugin/plugin.go
```diff
@@ -0,0 +1,113 @@
...
+	return
+}
+
+// ApplyDefaultConfig applies default configurations to plugin descriptor.
+func ApplyDefaultConfig(p *cliv1alpha1.PluginDescriptor) {
+	// TODO(anuj): Removed next lines as buildsha and version is not relevant to library and needs to be configured per plugin
+	// p.BuildSHA = buildinfo.SHA
+	// if p.Version == "" {
+	// 	p.Version = BuildVersion
+	// }
```
marckhouzam, line 83: Since you updated the plugin descriptors for every plugin to set the version and sha, do you want to remove the above commented code?

cli/runtime/plugin/plugin.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return desc, errors.Wrap(err, "could not unmarshal plugin descriptor")
+	}
+	return
+}
+
+// ApplyDefaultConfig applies default configurations to plugin descriptor.
+func ApplyDefaultConfig(p *cliv1alpha1.PluginDescriptor) {
+	// TODO(anuj): Removed next lines as buildsha and version is not relevant to library and needs to be configured per plugin
+	// p.BuildSHA = buildinfo.SHA
```
marckhouzam, line 80: Since we won't be setting the sha automatically, what do you think about adding it in the template used to generate plugin code?
https://github.com/vmware-tanzu/tanzu-framework/blob/0605dfc98500015c474ca3709e78d8cb434ac607/pkg/v1/builder/template/plugintemplates/main.go.tmpl#L12-L17

cli/runtime/plugin/plugin.go
```diff
@@ -0,0 +1,113 @@
...
+	// skip builder plugin for bootstrapping
+	if p.Name == "builder" {
+		return nil
+	}
+	if p.Name == "" {
+		err = multierr.Append(err, fmt.Errorf("plugin %q name cannot be empty", p.Name))
+	}
+	if p.Version == "" {
+		err = multierr.Append(err, fmt.Errorf("plugin %q version cannot be empty", p.Name))
+	}
```
marckhouzam, line 102: Are there plugins outside of the Framework repositories?  I'm asking because if they don't set the version, they will stop working now that we don't set it.  But if all plugins we care about are in Framework, I see you fixed them all 👍

marckhouzam (commented):
This is a big change so I'm not finished and I'm hoping to continue tonight or tomorrow, but here are a few comments from what I have looked at.

Overall, this is a nice and thorough change 👍

cli/runtime/plugin/plugin.go
anujc25: Yes. Makes sense

cli/runtime/plugin/plugin.go
anujc25: There are plugins outside of the tanzu-framework repository. I don't think any of those plugins are not configuring `Version` as part of plugin descriptor. If they are not then those might be temp or some development plugins that users are using for POC and testing. So, I don't think that would be the concern.

cli/runtime/plugin/doc.go
```diff
@@ -35,8 +36,9 @@ var genDocsCmd = &cobra.Command{
 		}
 		emptyStr := func(s string) string { return "" }
 
+		// TODO(anuj): This dependency was removed from core. Check what can be done to fix this
 		// Necessary to generate correct output
-		core.RootCmd.AddCommand(cmd.Parent())
+		// core.RootCmd.AddCommand(cmd.Parent())
```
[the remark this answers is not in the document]
anujc25: I think this is a great suggestion. I missed TODO here and hence missed fixing this. But I like you suggestion and I think it totally makes sense to implement this as you suggested.

pkg/v1/config/doc.go
```diff
@@ -1,5 +0,0 @@
-// Copyright 2021 VMware, Inc. All Rights Reserved.
-// SPDX-License-Identifier: Apache-2.0
-
-// Package config provides functions for reading and writing client configuration.
-package config
```
marckhouzam, line 5: What are these empty `doc.go` files for?

pkg/v1/sdk/features/Dockerfile
```diff
@@ -14,6 +14,7 @@ WORKDIR /workspace
 COPY go.mod go.mod
 COPY go.sum go.sum
 COPY apis/ apis/
+COPY cli/runtime cli/runtime
 RUN  go mod download
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
```
marckhouzam, line 20: Do we need to copy `cli/runtime` in the list of `COPY`s below?

marckhouzam (approved):
Wow!  Good job @anujc25!  This was massive.

I couple minor comments in-line but after than LGTM.

pkg/v1/config/doc.go
anujc25: I think based on my understanding doc.go file within a package is used to write go documentation for specific to the package.

pkg/v1/sdk/features/Dockerfile
anujc25: I saw failures in some package builds where this `runtime` library was getting used and hence added this line here.

cli/runtime/config/lock.go
```diff
@@ -5,13 +5,19 @@ package config
...
+	"time"
 
 	"github.com/juju/fslock"
+	"github.com/pkg/errors"
+)
 
-	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkg/constants"
-	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkg/utils"
+const (
+	LocalTanzuFileLock = ".tanzu.lock"
```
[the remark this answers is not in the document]
anujc25: Ack. We need to centralize config file update and locking mechanism and more discussion is needed but it might be out of scope of this PR.

cli/runtime/test/framework.go
```diff
@@ -20,9 +20,25 @@ import (
...
-	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli"
+	cliv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/cli/v1alpha1"
 )
 
+const (
+	CLIName = "tanzu"
+)
+
+// NewTestFor creates a plugin descriptor for a test plugin.
+func NewTestFor(pluginName string) *cliv1alpha1.PluginDescriptor {
```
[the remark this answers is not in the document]
anujc25: Updated PR with builder plugin template changes.

marckhouzam (approved):

cli/runtime/Makefile
```diff
@@ -0,0 +1,55 @@
...
+.PHONY: test
+test: fmt vet ## Run Tests
+	${GO} test ./... -timeout 60m -race -coverprofile coverage.txt -v
+
+.PHONY: vet
+vet: ## Vet codebase
+	${GO} vet ./...
+
+fmt: $(GOIMPORTS) ## Run goimports
+	$(GOIMPORTS) -w -local github.com/vmware-tanzu ./
```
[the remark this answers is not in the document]
anujc25: I have added `go fmt ./...` to this target as well

cli/runtime/Makefile
```diff
@@ -0,0 +1,55 @@
...
+	${GO} fmt ./...
```
vuil, line 54: doesn't goimport do what gofmt does already and more?

anujc25: I am not sure. I will take a look.

pkg/v1/builder/template/plugintemplates/Makefile.tmpl
```diff
@@ -11,10 +11,6 @@ GOARCH ?= $(shell go env GOARCH)
 GOHOSTOS ?= $(shell go env GOHOSTOS)
 GOHOSTARCH ?= $(shell go env GOHOSTARCH)
 
-LD_FLAGS = -X 'github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli.BuildDate=$(BUILD_DATE)'
-LD_FLAGS += -X 'github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli.BuildSHA=$(BUILD_SHA)'
-LD_FLAGS += -X 'github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli.BuildVersion=$(BUILD_VERSION)'
```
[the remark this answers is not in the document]
anujc25: Yeah. That would super useful if you can propose better option.

cli/runtime/Makefile
```diff
@@ -0,0 +1,55 @@
...
+
+.PHONY: test
+test: fmt vet ## Run Tests
+	${GO} test ./... -timeout 60m -race -coverprofile coverage.txt -v
+
+.PHONY: vet
+vet: ## Vet codebase
+	${GO} vet ./...
+
+fmt: $(GOIMPORTS) ## Run goimports
```
vuil, line 52: minor: (maybe something to address in a followup):
there is nothing to produce/obtain goimports if it it is not already available

vuil (approved):
Left a few minor comments, but otherwise lgtm.

cli/runtime/Makefile
anujc25: I will address this as part of followup PR.

cli/runtime/Makefile
anujc25: Sure. I will address this as part of followup PR.

PR #3205, opened 2022-08-26
Kickstart UI: Fixes issue with blank AVI/Cluster labels in payload

miclettej:
Fixes issue where Cluster and AVI key/value pair labels are including a pair of blank strings. This resulted in a failure when setting up AVI LB.

The fix now sets the value of labels to an empty object, ommitting any blank string key/value pairs.

Issue was raised in Slack

Fixes #

Have confirmed that UI payload during cluster creation includes an empty object for Cluster and Avi labels when user has not specify labels. Have also verified that key/value pair of empty strings is no longer included when user does provide labels.

### Release note

```release-note
none
```

### DIFF
--- a/pkg/v1/tkg/web/src/app/views/landing/wizard/shared/wizard-base/wizard-base.ts
+++ b/pkg/v1/tkg/web/src/app/views/landing/wizard/shared/wizard-base/wizard-base.ts
@@ -349,7 +349,7 @@ export abstract class WizardBaseDirective extends BasicSubscriber implements Wiz
      * @param arrObj of type [ {key: 'a', value: '1}, {key : 'b, value:'2}]
      * @param params.key<Function> To Identify the ObjectKey
      * @param params.value<Function> To Identify the ObjectValue
-     * @return Object {'a': 1 , 'b': '1'}
+     * @return Object {'a': 1 , 'b': '1'} and drops values with empty string on key
      */
     arrayOfObjectsToObject<T>(arrObj: T[], params: Params<T>): KeyValueObject {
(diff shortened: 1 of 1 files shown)


pkg/v1/tkg/web/src/app/views/landing/wizard/shared/wizard-base/wizard-base.ts
```diff
@@ -343,14 +347,23 @@ export abstract class WizardBaseDirective extends BasicSubscriber implements Wiz
...
-    arrayOfObjectsToObject<T>(arrObj: T[], params: Params<T>): { [key: string]: string; } {
-        return arrObj && arrObj instanceof Array
-            ? arrObj.reduce((obj, item) => ((obj[params.key(item)] = params.value(item)), obj), {})
-            : {};
+    arrayOfObjectsToObject<T>(arrObj: T[], params: Params<T>): KeyValueObject {
+        if (!arrObj || !(arrObj instanceof Array)) {
+            return {};
+        }
+
+        return arrObj.reduce<KeyValueObject>(
```
swalner-vmware, line 359: nit: Either here and/or in the comment block above, might be good to add a comment that we DROP any object with a blank key

swalner-vmware (approved):

pkg/v1/tkg/web/src/app/views/landing/wizard/shared/wizard-base/wizard-base.ts
miclettej: Good idea, will do. Thank you!

vuil (approved):

PR #3217, opened 2022-08-30
Refactor Pinniped component from addons directory to be under project root

danniel1205:

This PR copies Pinniped code from addons directory to be under project root. This is part of the Tanzu Framework code cleanup and refactoring. 

Naming the root level Pinniped directory to be `pinniped-components` is because there is a make target (`ensure-pinniped-repo` uses `pinniped` as the name of directory when cloning upstream pinniped). 

Fixes #NA

- GH action covers most of the local tests needed. And this PR does not remove the original pinniped code from `addons` directory, so it won't break any build/release flow.

### Release note

```release-note

```


CODEOWNERS
```diff
@@ -1,6 +1,7 @@
 * @vmware-tanzu/tanzu-framework-reviewers
 /addons @vmware-tanzu/tanzu-framework-reviewers @vmware-tanzu/tkg-addons-owners
 /addons/pinniped @vmware-tanzu/tanzu-framework-reviewers @vmware-tanzu/tkg-iam-owners
+/internal-pinniped @vmware-tanzu/tanzu-framework-reviewers @vmware-tanzu/tkg-iam-owners
```
benjaminapetersen, line 4: How would you feel about either of these as alternatives to `internal-pinniped`:

```bash
/pinniped-internal
/pinniped-components  <-- I lean slightly towards this
```
If this is short-term and will eventually be just "pinniped" then it doesn't matter, but that would mean we would plan to change the `ensure-pinniped-repo` make target.

If this is longer-term, I think the directory being some form of `pinniped-*` will make it easier to find. 
Additionally, after `ensure-pinniped-repo` runs, the directories would be siblings:

```
/pinniped                          # from the ensure-pinniped-repo make target
/pinniped-components   # the rest of our pinniped bits
```
There is some value in this kind of organization, unless `internal-*` is the pattern for other addons being evicted.

danniel1205: `internal-pinniped` is just some random temporary name I picked to distinguish from the folder name `ensure-pinniped-repo` uses. `pinniped-components` sounds good to me, will do the change. Thanks

benjaminapetersen (approved):
Nice. Onward!

vijaykatam (approved):

PR #3218, opened 2022-08-30
WIP: add pkgi delete function in management components

ggpaue:
add pkgi delete function in utils

Fixes #

### Release note

```release-note

```

### DIFF
--- a/addons/pkg/util/addon_util.go
+++ b/addons/pkg/util/addon_util.go
@@ -14,11 +14,8 @@ import (
 	"gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clusterapiv1beta1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	kappctrl "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
 	pkgiv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/packaging/v1alpha1"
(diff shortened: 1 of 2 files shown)


addons/pkg/util/addon_util.go
```diff
@@ -217,6 +218,23 @@ func GetPackageInstallFromAddonSecret(ctx context.Context,
...
+// DeletePackageInstall deletes the PackageInstall CR
+func DeletePackageInstall(ctx context.Context, c client.Client, pkgiName, pkgiNamespace string) error {
+	pkgi := &pkgiv1alpha1.PackageInstall{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      pkgiName,
+			Namespace: pkgiNamespace,
+		},
+	}
+	pkgi.Spec.NoopDelete = true
+
```
adduarte, line 230: Should we update the package install before deleting it? 
i.e. here the nooopdelete is set in the pkgi in memory. 
Then you call delete imeditaley. Not sure if the noopdelete ever makes it to the actual database. It could be that the Delete just looks at the name and namespace. 
I think we should patch the object with noopdelete first, then delete it.

shyaamsn:
@adduarte @ggpaue The code to upgrade addons-manager should be in Tanzu CLI i.e. in tanzu-framework/pkg/v1/ . Not sure why we are adding utility in addons folder. 

Also, there is tkgpackageclient available to do many of the package install CRUD operations. We should try to see if we can leverage those

addons/pkg/util/addon_util.go
adduarte: need to do a 
key = metava1.objectkey(name: pkginma, namespace: pkginamespace)
error := c.get(ctx, key, pkgi)
pkgi.spec.noopdelete
then patch
then call tkgclient.packageuninstall.

ggpaue: moved to management components under pkg/v1/tkg

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -196,3 +198,30 @@ func WaitForManagementPackages(clusterClient clusterclient.Client, packageInstal
 	}
 	return nil
 }
+
+// DeletePackageInstall deletes the PackageInstall CR
+func noopDeletePackageInstall(ctx context.Context, clusterClient clusterclient.Client, pkgClient tkgpackageclient.TKGPackageClient, pkgiName, pkgiNamespace string) error {
+	pkgi := &kappipkg.PackageInstall{}
+	err := clusterClient.GetResource(pkgi, pkgiName, pkgiNamespace, nil, nil)
```
adduarte, line 205: Need to handle "not found" which is not an error. if not found nothing to do so, return nil

PR #3227, opened 2022-08-31
Fix make build-cli-with-oci-discovery

### LINKED ISSUE #1791: Makefile missing build-cli-oci-* targets
**Bug description**

PR https://github.com/vmware-tanzu/tanzu-framework/pull/1144 added pattern creation for these targets and set them as dependencies for other targets. But no targets were added to match the generated patterns, so attempting to call one of the targets that sets these as a dependency like `make build-cli-with-oci-discovery` fails with an error:

```
make: *** No rule to make target `build-plugin-oci-linux-amd64', needed by `build-cli-with-oci-discovery'.  Stop
```

- [ ] APIs
- [ ] Addons
- [ ] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [ ] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Expected behavior**

mpanchajanya:

Fixes #1791 

Screen Shot 2022-08-31 at 12 29 37
Screen Shot 2022-08-31 at 12 29 30

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -206,7 +206,7 @@ endif
 
 # Dynamically generate the OS-ARCH targets to allow for parallel execution
 CLI_JOBS_OCI_DISCOVERY := $(addprefix build-cli-oci-,${ENVS})
-CLI_ADMIN_JOBS_OCI_DISCOVERY := $(addprefix build-plugin-oci-,${ENVS})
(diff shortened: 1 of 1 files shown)


marckhouzam (approved):
LGTM

anujc25 (approved):

PR #3229, opened 2022-08-31
Graceful upgrade of addons-manager

### LINKED ISSUE #3265: Logic and code are needed to upgrade addons-manager from core to manage repository

- ( ) APIs
- (x ) Addons
- (x ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Additional context**

adduarte:
Graceful upgrade of addons-manager
    
    - Pauses lifecycle management of addons-manager package
    - NoopDelete addons-manager packageinstall
    - Deletes adoons-manager addon secret

Adds necessary functions and logic to gracefully upgrade addons-manager package from the package provided by core repository to the addons-manager provided by the management repository.

logic flow: 
pause lifecycle management for addons-manager
  pause reconciliation for existing pkgi  
  pause reconciliation for existing addonsSecret
noop delete of existing addons-manger pkgi
wait for succefull deployment of management packages (inluding new addons-manager)
remove old addonsSecret for old addons-manager

Fixes #3265

- Manual testing of PauseAddonLifecycleManagement, NoopDeletePackageInstall, and DeleteAddonSecret
- Unit testing of PauseAddonLifecycleManagement, NoopDeletePackageInstall, and DeleteAddonSecret

### Release note

```release-note
Upgrade to using addons-manager package from management repository
During upgrade previous addons-managerr lifecycle is paused, and if upgrade is succesful  previous addons-manager addon secret is remove to prevent it being reinstalled by reconciliation logic. 
```

#### Testing done
Deployed 1.6 GA on to vpshere

set the following env variables 
```
export _MANAGEMENT_PACKAGE_REPO_IMAGE=gcr.io/eminent-nation-87317/tanzu_framework/github-actions/main/management:v0.21.0
export _MANAGEMENT_PACKAGE_VERSION=0.21.0
```
ran "tanzu mc upgrade -y"
results: 

Results: 

```
aduarte@aduarte-a01 poc1 % kubectl get pkgi -n tkg-system tanzu-addons-manager
NAME                   PACKAGE NAME                      PACKAGE VERSION   DESCRIPTION           AGE
tanzu-addons-manager   addons-manager.tanzu.vmware.com   0.21.0            Reconcile succeeded   36m

and pods where generated with matching image

### DIFF
--- a/tkg/managementcomponents/management_component_install.go
+++ b/tkg/managementcomponents/management_component_install.go
@@ -5,14 +5,21 @@ package managementcomponents
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 	"os"
 	"time"
 
 	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 	crtclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	kappipkg "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/packaging/v1alpha1"
 	"github.com/vmware-tanzu/tanzu-framework/packageclients/pkg/packageclient"
@@ -20,6 +27,11 @@ import (
 	"github.com/vmware-tanzu/tanzu-framework/tkg/clusterclient"
 	"github.com/vmware-tanzu/tanzu-framework/tkg/constants"
 	"github.com/vmware-tanzu/tanzu-framework/tkg/log"
+)
+
+const (
+	addonsManagerName = "addons-manager"
+	addonFinalizer    = "tkg.tanzu.vmware.com/addon"
 )
 
 // ClusterOptions specifies cluster configuration
@@ -48,11 +60,194 @@ type ManagementComponentsInstallOptions struct {
 	ManagementPackageRepositoryOptions ManagementPackageRepositoryOptions
 }
 
+func generateAddonSecretName(clusterName, addonName string) string {
+	return fmt.Sprintf("%s-tanzu-%s-addon", clusterName, addonName)
+}
+
+func pauseAddonSecretReconciliation(clusterClient clusterclient.Client, addonSecreteName, namespace string) error {
+	log.Infof("Pausing reconciliation for %s/%s secret", namespace, addonSecreteName)
+	secret := &corev1.Secret{}
+	jsonPatch := []map[string]interface{}{
+		{
+			"op":    "add",
+			"path":  fmt.Sprintf("/metadata/annotations/tkg.tanzu.vmware.com~1addon-paused"),
+			"value": "",
+		},
+	}
+	payloadBytes, err := json.Marshal(jsonPatch)
+	if err != nil {
+		return errors.Wrap(err, "unable to generate json patch")
+	}
+
+	err = clusterClient.PatchResource(secret, addonSecreteName, namespace, string(payloadBytes), types.JSONPatchType, nil)
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return errors.Wrapf(err, "failed to pause %s secret reconciliation", addonSecreteName)
+	}
+	return nil
+}
+
+func pausePackageInstallReconciliation(clusterClient clusterclient.Client, pkgiName, namespace string) error {
+	log.Infof("Pausing reconciliation for %s/%s packageinstall", namespace, pkgiName)
+	pkgi := &kappipkg.PackageInstall{}
+	jsonPatch := []map[string]interface{}{
+		{
+			"op":    "add",
+			"path":  "/spec/paused",
+			"value": true,
+		},
+	}
+	payloadBytes, err := json.Marshal(jsonPatch)
+	if err != nil {
+		return errors.Wrap(err, "unable to generate json patch")
+	}
+	err = clusterClient.PatchResource(pkgi, pkgiName, namespace, string(payloadBytes), types.JSONPatchType, nil)
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return errors.Wrapf(err, "failed to pause %s packageinstall reconciliation", pkgiName)
+	}
+	return nil
+}
+
+// PausedAddonLifecycleManagement pauses/unpauses the lifecycle management of addon package with given name and namespace
+func PauseAddonLifecycleManagement(clusterClient clusterclient.Client, clusterName, addonName, namespace string) error {
+	log.Infof("Pausing lifecycle management for %s", addonName)
+	pkgiName := fmt.Sprintf("tanzu-%s", addonsManagerName)
+	addonSecretName := generateAddonSecretName(clusterName, addonsManagerName)
+
+	err := pauseAddonSecretReconciliation(clusterClient, addonSecretName, namespace)
+	if err != nil {
+		return err
+	}
+
+	err = pausePackageInstallReconciliation(clusterClient, pkgiName, namespace)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// NoopDeletePackageInstall sets spec.noopdelete = true before deleting the package install
+func NoopDeletePackageInstall(clusterClient clusterclient.Client, addonName, namespace string) error {
+	log.Infof("Deleting %s/%s packageinstall with noopdelete", namespace, addonName)
+	pkgiName := fmt.Sprintf("tanzu-%s", addonName)
+
+	jsonPatch := []map[string]interface{}{
+		{
+			"op":    "add",
+			"path":  "/spec/noopDelete",
+			"value": true,
+		},
+	}
+	payloadBytes, err := json.Marshal(jsonPatch)
+	if err != nil {
+		return errors.Wrap(err, "unable to generate json patch")
+	}
+	pkgi := &kappipkg.PackageInstall{}
+	err = clusterClient.PatchResource(pkgi, pkgiName, namespace, string(payloadBytes), types.JSONPatchType, nil)
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return errors.Wrapf(err, "failed to patch %s packageinstall", pkgiName)
+	}
+
+	pkgi.Name = pkgiName
+	pkgi.Namespace = namespace
+	err = clusterClient.DeleteResource(pkgi)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrapf(err, "failed to delete PackageInstall resource %s", pkgiName)
+	}
+	return nil
+}
+
+// DeleteAddonSecret deletes the secrete associated with the addon if present. Return no error if secret not found.
+func DeleteAddonSecret(clusterClient clusterclient.Client, clusterName, addonName, namespace string) error {
+	addonSecret := &corev1.Secret{}
+	addonSecret.Name = generateAddonSecretName(clusterName, addonName)
+	addonSecret.Namespace = constants.TkgNamespace
+	log.Infof("Deleting %s/%s secret", addonSecret.Namespace, addonSecret.Name)
+	err := clusterClient.GetResource(addonSecret, addonSecret.Name, addonSecret.Namespace, nil, nil)
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	if controllerutil.ContainsFinalizer(addonSecret, addonFinalizer) {
+		controllerutil.RemoveFinalizer(addonSecret, addonFinalizer)
+	}
+
+	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		return clusterClient.UpdateResource(addonSecret, addonSecret.Name, addonSecret.Namespace)
+	})
+
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	err = clusterClient.DeleteResource(addonSecret)
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return errors.Wrapf(err, "failed to delete addon addonSecret %s", addonSecret)
+	}
+	return nil
+}
+
+// AddonSecretExists returns true if given addon is present and was installed from core repository.
+func AddonSecretExists(clusterClient clusterclient.Client, clusterName, addonName, namespace string) (bool, error) {
+	addonSecret := &corev1.Secret{}
+	addonSecret.Name = generateAddonSecretName(clusterName, addonName)
+	addonSecret.Namespace = constants.TkgNamespace
+
+	err := clusterClient.GetResource(addonSecret, addonSecret.Name, addonSecret.Namespace, nil, nil)
+	if apierrors.IsNotFound(err) {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
 // InstallManagementComponents installs the management component to cluster
 func InstallManagementComponents(mcip *ManagementComponentsInstallOptions) error {
 	clusterClient, err := clusterclient.NewClient(mcip.ClusterOptions.Kubeconfig, mcip.ClusterOptions.Kubecontext, clusterclient.Options{})
 	if err != nil {
 		return errors.Wrap(err, "unable to get cluster client")
+	}
+	clusterName, err := clusterClient.GetCurrentClusterName(mcip.ClusterOptions.Kubecontext)
+	if err != nil {
+		return errors.Wrap(err, "unable to get cluster name")
+	}
+
+	// If the addons-manager is moving from core repository to management repository, its lifecycle management
+	// needs to be paused.
+	previousAddonsManagerIsFromCoreRepo, err := AddonSecretExists(clusterClient, clusterName, addonsManagerName, constants.TkgNamespace)
+	if err != nil {
+		return err
+	}
+	if previousAddonsManagerIsFromCoreRepo {
+		err = PauseAddonLifecycleManagement(clusterClient, clusterName, addonsManagerName, constants.TkgNamespace)
+		if err != nil {
+			return err
+		}
+
+		err = NoopDeletePackageInstall(clusterClient, addonsManagerName, constants.TkgNamespace)
+		if err != nil {
+			return err
+		}
 	}
 
 	// create package client
@@ -78,6 +273,13 @@ func InstallManagementComponents(mcip *ManagementComponentsInstallOptions) error
 	if resouceFile != "" {
 		log.Infof("Appling additional management component configuration from %q", resouceFile)
 		err := clusterClient.ApplyFile(resouceFile)
+		if err != nil {
+			return err
+		}
+	}
+
+	if previousAddonsManagerIsFromCoreRepo {
+		err = DeleteAddonSecret(clusterClient, clusterName, addonsManagerName, constants.TkgNamespace)
 		if err != nil {
 			return err
 		}
--- a/tkg/managementcomponents/management_component_install_test.go
+++ b/tkg/managementcomponents/management_component_install_test.go
@@ -11,7 +11,11 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
 	packageclientfakes "github.com/vmware-tanzu/tanzu-framework/packageclients/pkg/fakes"
+	"github.com/vmware-tanzu/tanzu-framework/tkg/constants"
 	"github.com/vmware-tanzu/tanzu-framework/tkg/fakes"
 	. "github.com/vmware-tanzu/tanzu-framework/tkg/managementcomponents"
 )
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: pkg/v1/tkg/managementcomponents/management_component_install.go [hunk 1]
```diff
@@ -49,18 +60,183 @@ type ManagementComponentsInstallOptions struct {
...
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	if controllerutil.ContainsFinalizer(addonSecret, addonFinalizer) {
+		controllerutil.RemoveFinalizer(addonSecret, addonFinalizer)
+	}
+	err = clusterClient.UpdateResource(addonSecret, addonSecret.Name, addonSecret.Namespace)
```
maralavi, line 190: Is this update to add finalizer changes? Wondering if it shouldn't be patch instead of update?

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -49,18 +60,183 @@ type ManagementComponentsInstallOptions struct {
...
 func InstallManagementComponents(mcip *ManagementComponentsInstallOptions) error {
 	clusterClient, err := clusterclient.NewClient(mcip.ClusterOptions.Kubeconfig, mcip.ClusterOptions.Kubecontext, clusterclient.Options{})
 	if err != nil {
 		return errors.Wrap(err, "unable to get cluster client")
 	}
+	clusterName, err := clusterClient.GetCurrentClusterName(mcip.ClusterOptions.Kubecontext)
+	if err != nil {
+		return errors.Wrap(err, "unable to get cluster name")
+	}
+	if err != nil {
```
maralavi, line 218: This is a redundant check

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -49,18 +60,183 @@ type ManagementComponentsInstallOptions struct {
...
+		return err
+	}
 
-	// create package client
-	pkgClient, err := tkgpackageclient.NewTKGPackageClientForContext(mcip.ClusterOptions.Kubeconfig, mcip.ClusterOptions.Kubecontext)
+	err = PauseAddonLifecycleManagement(clusterName, addonsManagerName, constants.TkgNamespace, clusterClient)
+	if err != nil {
+		panic(err)
```
maralavi, line 224: Why should this one panic instead of returning an error?

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -49,18 +60,183 @@ type ManagementComponentsInstallOptions struct {
...
+	}
+
+	err = NoopDeletePackageInstall(addonsManagerName, constants.TkgNamespace, clusterClient)
+	if err != nil {
+		return nil
```
maralavi, line 229: Should return ```err``` not ```nil```

pkg/v1/tkg/managementcomponents/management_component_install_test.go
```diff
@@ -169,3 +172,208 @@ var _ = Describe("Test WaitForManagementPackages", func() {
...
+		It("should return error", func() {
+			Expect(err).To(HaveOccurred())
+		})
+	})
+	Context("Patching resources returns not found", func() {
+		BeforeEach(func() {
+			clusterClient.PatchResourceReturns(notFoundError)
+		})
+		It("should return no error", func() {
+			Expect(err).NotTo(HaveOccurred())
```
maralavi, line 222: But according your code, you had returned ```nil``` on ```NotFound``` return err value for ```PatchResource```. Wondering why error is expected here?

pkg/v1/tkg/managementcomponents/management_component_install_test.go
```diff
@@ -169,3 +172,208 @@ var _ = Describe("Test WaitForManagementPackages", func() {
...
+		It("should return error", func() {
+			Expect(err).To(HaveOccurred())
+		})
+	})
+	Context("Deleting resources returns not found", func() {
+		BeforeEach(func() {
+			clusterClient.DeleteResourceReturns(notFoundError)
+		})
+		It("should return no error", func() {
+			Expect(err).NotTo(HaveOccurred())
```
maralavi, line 289: Same as above comment:
According your code, you had returned nil on NotFound return err value for PatchResource. Wondering why error is expected here?

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -49,18 +60,183 @@ type ManagementComponentsInstallOptions struct {
...
+func generateAddonSecretName(clusterName, addonName string) string {
+	return fmt.Sprintf("%s-%s-addon", clusterName, generateAddonsPkgiName(addonName))
+}
+
+func pauseAddonSecretReconciliation(ctx context.Context, clusterClient clusterclient.Client, clusterName, addonSecreteName, namespace string) error {
+	secret := &corev1.Secret{}
+	jsonPatch := []map[string]interface{}{
+		{
+			"op":    "add",
+			"path":  fmt.Sprintf("/metadata/annotations/tkg.tanzu.vmware.com~1addon-paused"),
```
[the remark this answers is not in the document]
adduarte: The "~1" is  as special character used to "scape" the "/"  int he real path. 
See https://www.rfc-editor.org/rfc/rfc6901#section-3

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -49,18 +60,183 @@ type ManagementComponentsInstallOptions struct {
...
 
+func generateAddonsPkgiName(addonName string) string {
+	return fmt.Sprintf("tanzu-%s", addonName)
+}
+
+func generateAddonSecretName(clusterName, addonName string) string {
+	return fmt.Sprintf("%s-%s-addon", clusterName, generateAddonsPkgiName(addonName))
+}
+
+func pauseAddonSecretReconciliation(ctx context.Context, clusterClient clusterclient.Client, clusterName, addonSecreteName, namespace string) error {
```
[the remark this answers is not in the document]
adduarte: we don't need clusterName. Left over from previous version of code. wil remove.

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -49,18 +60,183 @@ type ManagementComponentsInstallOptions struct {
...
+	secret := &corev1.Secret{}
+	jsonPatch := []map[string]interface{}{
+		{
+			"op":    "add",
```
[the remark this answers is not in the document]
adduarte: no.  Patch is the operations. "add" here means the patch will add the line. 
a patch operation requires an operation ("op"), the operation can be "remove","add", "replace" see https://jamesdefabia.github.io/docs/user-guide/kubectl/kubectl_patch/

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -49,18 +60,183 @@ type ManagementComponentsInstallOptions struct {
...
+		return nil
+	}
+	if err != nil {
+		return errors.Wrapf(err, "failed to pause %s packageinstall reconciliation", pkgiName)
+	}
+	return nil
+}
+
+// PausedAdoonLifecycleManagement pauses/unpauses the lifecycle management of addon package with given name and namespace
+func PauseAddonLifecycleManagement(clusterName, addonName, namespace string, clusterClient clusterclient.Client) error {
```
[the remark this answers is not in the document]
adduarte: will do

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -49,18 +60,183 @@ type ManagementComponentsInstallOptions struct {
...
+	pkgiName := generateAddonsPkgiName(addonsManagerName)
```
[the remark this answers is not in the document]
adduarte: Cannot use constant. The name includes the clustername, so it changes for each cluster.

pkg/v1/tkg/managementcomponents/management_component_install.go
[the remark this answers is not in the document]
adduarte: fixed

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: removed

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: fixed

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: fixed

pkg/v1/tkg/managementcomponents/management_component_install_test.go
adduarte: ?? - Expect(err).NotTo(HaveOccurred()) . means error is not expected

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -49,18 +60,183 @@ type ManagementComponentsInstallOptions struct {
...
+		return errors.Wrapf(err, "failed to patch %s packageinstall", pkgiName)
+	}
+
+	pkgi.Name = pkgiName
+	pkgi.Namespace = namespace
+	err = clusterClient.DeleteResource(pkgi)
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
```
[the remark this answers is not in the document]
adduarte: fixed.

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -49,18 +60,183 @@ type ManagementComponentsInstallOptions struct {
...
+		return errors.Wrapf(err, "failed to delete PackageInstall resource %s", pkgiName)
+	}
+	if apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
adduarte: fixed

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -49,18 +60,183 @@ type ManagementComponentsInstallOptions struct {
...
+		return nil
+	}
+	return nil
+}
+
+func DeleteAddonSecret(clusterName, addonName, namespace string, clusterClient clusterclient.Client) error {
```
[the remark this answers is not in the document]
adduarte: done

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: yes. Patch seems a better choice

pkg/v1/tkg/managementcomponents/management_component_install.go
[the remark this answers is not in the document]
adduarte: fixed

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: Unfortunately jsonpatch does not support removing an element from an array list in the yaml: 
https://github.com/json-patch/json-patch2/issues/18
finalizers is an array and there is no way to predict which index the addons finalizer will be in.  
The only option is to remove all finalizers, and that seems like a bad option.

adduarte: instead added retryonconflict

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -48,18 +60,182 @@ type ManagementComponentsInstallOptions struct {
 	ManagementPackageRepositoryOptions ManagementPackageRepositoryOptions
 }
 
+func generateAddonSecretName(clusterName, addonName string) string {
+	return fmt.Sprintf("%s-%s-addon", clusterName, fmt.Sprintf("tanzu-%s", addonName))
```
maralavi, line 64: Question: Are you sure addon secret name used to have ```tanzu``` in it? What is an example addone secret name?

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -48,18 +60,182 @@ type ManagementComponentsInstallOptions struct {
...
+
+	if controllerutil.ContainsFinalizer(addonSecret, addonFinalizer) {
+		controllerutil.RemoveFinalizer(addonSecret, addonFinalizer)
+	}
+
+	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		return clusterClient.UpdateResource(addonSecret, addonSecret.Name, addonSecret.Namespace)
+	})
+
+	if apierrors.IsNotFound(err) {
```
maralavi, line 190: Isn't this one supposed to be returned error for ```UpdateResourec``` inside the block and not ```RetryOnConflict``` function?

maralavi:
I just left two questions as comments. The PR looks good otherwise to me. It would be good to get an approval from @vijaykatam as well.

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: yes: 
https://github.com/vmware-tanzu/tanzu-framework/blob/458ccb00877360826fbef34c429ca08259793195/pkg/v1/providers/ytt/02_addons/tanzu-addons-manager/add_tanzu-addons-manager.yaml#L49

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: My understanding of https://pkg.go.dev/k8s.io/client-go/util/retry@v0.23.5#RetryOnConflict is that if the error is a non-conflict, it will just return the error. so i a not found error, err should have "notfound"

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -48,18 +61,207 @@ type ManagementComponentsInstallOptions struct {
...
 		return errors.Wrap(err, "unable to get cluster client")
 	}
+	clusterName, err := clusterClient.GetCurrentClusterName(mcip.ClusterOptions.Kubecontext)
+	if err != nil {
+		return errors.Wrap(err, "unable to get cluster name")
+	}
+
+	// If the addons-manager is moving from core repository to management repository, its lifecycle management
+	// needs to be paused.
+	pauseAddonsManagerLifecycle, err := addonsPackageIsMovingRepositories(clusterClient, clusterName, addonsManagerName, constants.TkgNamespace)
```
adduarte, line 238: We add this check here because we only want to pause the lifecycle management of the addons-manager if necessary.  Currenly it is necessary when the pacakge is moving from core repository to management repository: from 1.6...x to -> utkg.

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -48,18 +61,207 @@ type ManagementComponentsInstallOptions struct {
...
+		if err != nil {
+			return err
+		}
+
+		err = NoopDeletePackageInstall(clusterClient, addonsManagerName, constants.TkgNamespace)
+		if err != nil {
+			return err
+		}
+
+		err = DeleteAddonSecret(clusterClient, clusterName, addonsManagerName, constants.TkgNamespace)
```
adduarte, line 253: we should delete the addons secret because if we do not the "legacy" controller will reinstall the old package.

maralavi (approved):

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -48,18 +61,207 @@ type ManagementComponentsInstallOptions struct {
...
+		return nil
+	}
+	if err != nil {
+		return errors.Wrapf(err, "failed to delete addon addonSecret %s", addonSecret)
+	}
+	return nil
+}
+
+func addonsPackageIsMovingRepositories(clusterClient clusterclient.Client, clusterName, addonName, namespace string) (bool, error) {
+	// We can check if the addons package has been installed from the core repository (https://gitlab.eng.vmware.com/core-build/tkg-packages/-/tree/main/core/addons-manager)
```
vijaykatam, line 210: remove references to internal repositories

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -48,18 +61,207 @@ type ManagementComponentsInstallOptions struct {
...
+	return fmt.Sprintf("%s-%s-addon", clusterName, fmt.Sprintf("tanzu-%s", addonName))
+}
+
+func pauseAddonSecretReconciliation(ctx context.Context, clusterClient clusterclient.Client, addonSecreteName, namespace string) error {
+	log.Infof("Pausing reconciliation for %s/%s secret", namespace, addonSecreteName)
+	secret := &corev1.Secret{}
+	jsonPatch := []map[string]interface{}{
+		{
+			"op":    "add",
+			"path":  fmt.Sprintf("/metadata/annotations/tkg.tanzu.vmware.com~1addon-paused"),
```
vijaykatam, line 73: What does `~1` mean here?

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -48,18 +61,207 @@ type ManagementComponentsInstallOptions struct {
...
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return errors.Wrapf(err, "failed to pause %s packageinstall reconciliation", pkgiName)
+	}
+	return nil
+}
+
+// PausedAdoonLifecycleManagement pauses/unpauses the lifecycle management of addon package with given name and namespace
```
vijaykatam, line 116: typo in doc

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -48,18 +61,207 @@ type ManagementComponentsInstallOptions struct {
 	ManagementPackageRepositoryOptions ManagementPackageRepositoryOptions
 }
 
+func generateAddonSecretName(clusterName, addonName string) string {
+	return fmt.Sprintf("%s-%s-addon", clusterName, fmt.Sprintf("tanzu-%s", addonName))
+}
+
+func pauseAddonSecretReconciliation(ctx context.Context, clusterClient clusterclient.Client, addonSecreteName, namespace string) error {
+	log.Infof("Pausing reconciliation for %s/%s secret", namespace, addonSecreteName)
```
vijaykatam, line 68: unused context

pkg/v1/tkg/managementcomponents/management_component_install.go
vijaykatam: why the double Sprintf? Could we do `%s-tanzu-addons-manager-addon`

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -48,18 +61,207 @@ type ManagementComponentsInstallOptions struct {
...
+		return nil
+	}
+	if err != nil {
+		return errors.Wrapf(err, "failed to pause %s secret reconciliation", addonSecreteName)
+	}
+	return nil
+}
+
+func pausePackageInstallReconciliation(ctx context.Context, clusterClient clusterclient.Client, pkgiName, namespace string) error {
+	log.Infof("Pausing reconciliation for %s/%s packageinstall", namespace, pkgiName)
```
vijaykatam, line 93: unused context var

pkg/v1/tkg/managementcomponents/management_component_install.go
vijaykatam: Even after pausing?

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -48,18 +60,209 @@ type ManagementComponentsInstallOptions struct {
...
+			"value": "",
+		},
+	}
+	payloadBytes, err := json.Marshal(jsonPatch)
+	if err != nil {
+		return errors.Wrap(err, "unable to generate json patch")
+	}
+
+	err = clusterClient.PatchResource(secret, addonSecreteName, namespace, string(payloadBytes), types.JSONPatchType, nil)
+	if apierrors.IsNotFound(err) {
```
vijaykatam, line 83: Question: Are we ignoring not found errors everywhere because we want the cluster upgrade to be repeatable in case of errors?

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -48,18 +60,209 @@ type ManagementComponentsInstallOptions struct {
...
+		return nil
+	}
+	if err != nil {
+		return errors.Wrapf(err, "failed to delete addon addonSecret %s", addonSecret)
+	}
+	return nil
+}
+
+// IsFromCoreRepository returns true if given addon is present and was installed from core repository.
+func IsFromCoreRepository(clusterClient clusterclient.Client, clusterName, addonName, namespace string) (bool, error) {
```
vijaykatam, line 209: I think this name is misleading, we are only checking if a secret exists. Perhaps a better name to reflect what it is actually doing?

vijaykatam (commented):
Minor comments. I think cluster upgrade should be repeatable in case of errors that happen midway(network timeouts, what not) so lets ensure this code is idempotent.

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: ack

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: The "~1" is as special character used to "scape" the "/" int he real path.
See https://www.rfc-editor.org/rfc/rfc6901#section-3

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: ack

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: ack

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: The name is on purpose. What we want to really know is if the package is from the core repository. 
It just happens that right now the only way we know how is to check for the existance of the secret. 
There might be a better way, or we might need to change it.

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: yes. Idempotency. If we hit a "not found" it means the resource does not exists so we have to do nothing. 

The last thing we do is delete the secret which will make a new execution skip all of this.

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: fixed

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: fixed

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: fixed

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: fixed

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: fixed

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: ... if we don't delete then we have to somehow make sure it never gets "unpaused"..

adduarte: Perhaps moving to InstallManagementPakcages will provide better impdepotency.

adduarte: Moved deleteaddonsecret until end of "installManagementPackage" 
manual testing showed the code is still idempotent.

adduarte:
Succesfully tested for upgrade repeatable in case of failure.

vijaykatam (approved):

vijaykatam (commented):
Can you check out why capd is failing?

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -48,18 +60,202 @@ type ManagementComponentsInstallOptions struct {
...
+	if err != nil {
+		return err
+	}
+	if previousAddonsManagerIsFromCoreRepo {
+		err = PauseAddonLifecycleManagement(clusterClient, clusterName, addonsManagerName, constants.TkgNamespace)
+		if err != nil {
+			return err
+		}
+
+		err = NoopDeletePackageInstall(clusterClient, addonsManagerName, constants.TkgNamespace)
```
vuil, line 247: question: how severe is failure in any of the 3 steps (PauseAddon.. NoopDelete.., DeleteAddon...)? I wonder if there is any that is harmless enough to just log and allow (likely the MC upgrade operation) to succeed?

vuil:
posted a question.
cc @anujc25 you may want to comment on the changes to the InstallManagementComponents flow

pkg/v1/tkg/managementcomponents/management_component_install.go
```diff
@@ -48,18 +60,202 @@ type ManagementComponentsInstallOptions struct {
...
+	addonSecretName := generateAddonSecretName(clusterName, addonsManagerName)
+
+	err := pauseAddonSecretReconciliation(clusterClient, addonSecretName, namespace)
+	if err != nil {
+		return err
+	}
+
+	err = pausePackageInstallReconciliation(clusterClient, pkgiName, namespace)
+	if err != nil {
+		return err
```
tenczar, line 129: If we get here and this error occurs, that means the secret reconciliation has been paused, but presumably on the package install? Are there remediation steps? what happens if someone tries the operation again? Will this fail at the pauseAddonSecret step?

tenczar (approved):
One minor comment from me.

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: If we have a failure here but we continue with upgrade, it will leave us with a system in an unknown state.  
as best we can predict we will would end up with the wrong version of addons-manager (at best) at worst we end up with two... Its really dificult to predict thus why we though it best to stop. 
If a failure occurs and we have to stop upgrade, the upgrade can be restarted. The code is meant  to be idempotent, and local testing on a vsphere cluster, has so far shown that it is...i.e. I can run the upgrade over and over in case of a failure, and the code results in the correct deployment of pkgis. 

I think is riskier to continue if we fail because the user would have a lot of cleanning up to do if we don't stop the upgrade.

pkg/v1/tkg/managementcomponents/management_component_install.go
adduarte: If we fail at this point, then the remediation step would be to retry the upgrade. 
If total failure happens and user changes their mind about upgrade, they will have to unpause the addons secret reconciliation by hand.  We could try to "roll back" the addons secret pause, but it is difficult to predict the state of the system with enough accuracy to ensure a succesfull "roll back" . Perhaps is best to let the user explore why the failure and unpause  by hand. We do put out a log that we are pausing the addons secret.  Perhaps we should put a log here saying that user needs to check system and unpause secret. 

What will happen then is that the addonsecret is paused so there is no reconcilation of old addons-manager. 
This would not block any new attempts to upgrade. We tested locally that scenario and the upgrade is "succesful" if you restart it after a failure at this point. 

The only other scenario is if the user decides to not do the upgrade after a failure here, which means they would have to manually unpause the lifecycle.

tenczar: Ok, then it probably makes sense to add that log message when this second pause fails, and then filing a docs bug to make sure that the steps for recovery are documented when a user doesn't just retry the upgrade.

Issue #3247, opened 2022-09-01
"tanzu cluster describe" returns a default value, and doesnt process input.


jayunit100:
**Bug description**

`tanzu cluster describe` is a valid command but it doesn't actually do anything :).

```
kubo@ubuntu-2004:~$ tanzu cluster describe
Kubernetes cluster operations
```
This seems to be , basically a no-op command.  I think thats "ok", but its a little confusing when it takes input:

```
kubo@ubuntu-2004:~$ tanzu cluster describe tkg-vc-antreas
Kubernetes cluster operations
```

And still, is a no-op....

PR #3248, opened 2022-09-01
Add some clarity to the tanzu cluster describe return value

jayunit100:
Quick fix to make `tanzu cluster describe` output more self describing so it doesnt get confused w/ `kubectl describe ...` #3247


randomvariable:
At least as far as the current cobra implementation goes, i think it's clear enough?

PR #3251, opened 2022-09-01
Create new top level directory for package  related clients

### LINKED ISSUE #3287: Extract package clients from TKG to top level packageclients directory as a separate go module
Restructure the clients needed by both package/secret plugins.
Create a top level directory with its own go module to facilitate subsequent move of it to its own repository.

There are 3 go packages ```tkgpackageclient```, ```kappclient```, ```tkgpackagedatamodel```, that need to be separated out into its own top level directory with a dedicated go module.
secret plugin, package plugin and management cluster plugin will import this as a dependency.
move integration tests to the cmd layer and update github workflows and test-infra code.

- ( ) APIs
- (X) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- (X) Plugin
- ( ) Security
- (X) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Additional context**

maralavi:
Create new top level directory for package related clients

Fixes #3287 

### Release note

```release-note
None
```

### Additional information

Moved the following directories from pkg/v1/tkg to packageclients/ module:

- tkgpackageclient
- tkgpackagedatamodel
- kappclient


anujc25 (approved):
Thanks for the changes. This looks good to me.

cc @vuil @tenczar @vijaykatam

cmd/cli/plugin/package/test/package_plugin_integration_test.go
```diff
@@ -1,7 +1,7 @@
 // Copyright 2021 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-package tkgpackageclient_test
+package main
```
vijaykatam, line 4: Question: Why is the test package renamed to main?

packageclients/pkg/fakes/crtclient.go
```diff
@@ -0,0 +1,776 @@
+// Code generated by counterfeiter. DO NOT EDIT.
```
vijaykatam, line 1: Question: Where are code gen markers for these fakes located? We will also need a make target within packageclients `Makefile` that generates them.

pkg/v1/tkr/Dockerfile
```diff
@@ -15,6 +15,7 @@ COPY go.mod go.mod
 COPY go.sum go.sum
 COPY apis/ apis/
 COPY cli/runtime cli/runtime
+COPY packageclients/ packageclients/
```
vijaykatam, line 18: We discussed offline that is an interim step required for tkr components to be able to build. Should go away when tkr components are refactored to their own go mod.

vijaykatam (commented):
Change looks good for the most part. We are missing Makefile for packageclients go module that should at a minimum do vet, fmt, test and code gen for fakes.

cmd/cli/plugin/package/test/package_plugin_integration_test.go
maralavi: Cause I moved it to the ```main``` package of ```cmd/cli/plugin/package/test``` from ```pkg/v1/tkg/test/packageclients``` and the package name needed to match  ```cmd/cli/plugin/package/test``` and wouldn't accept multiple package names.

packageclients/pkg/fakes/crtclient.go
maralavi: > Question: Where are code gen markers for these fakes located?

It is in the interface definitions for each client

maralavi: > We will also need a make target within packageclients Makefile that generates them.
done.

packageclients/pkg/packageclient/image_utils.go
```diff
@@ -1,7 +1,7 @@
 // Copyright 2021 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-package tkgpackageclient
+package packageclient
```
codegold79, line 4: Usually, package names are not in camel case. Also, to be consistent with others, shouldn't this be `packageclient`?

packageclients/pkg/packageclient/interface.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
```
codegold79, line 1: Should this be 2022?

codegold79 (changes requested):
Great work @maralavi. Only a couple very minor comments.

packageclients/pkg/packageclient/interface.go
maralavi: I had originally written this file in 2021, I just moved it to a new directory. So I think date remains as it was.

packageclients/pkg/packageclient/image_utils.go
maralavi: This is not camelcase. And I've already renamed it to ```packageclient```, so it is correct.

codegold79 (approved):

packageclients/pkg/packageclient/image_utils.go
codegold79: I might have been seeing the old version. When I left the review comment, I saw the package name as `packageClient` (camel case). But I see now it is how I had suggested, `packageclient`.

vijaykatam (approved):

PR #3254, opened 2022-09-01
Adding CAPD CC docker tests to CI workflow

saji-pivotal:
Adds clusterclass based CAPD tests to the Github CI

Fixes #

- Added pipeline for CAPD CC tests and verification is done through pipeline run

### Release note

```release-note
None
```

### DIFF
--- a/.github/workflows/tkg_integration_tests_clusterclass.yaml
+++ b/.github/workflows/tkg_integration_tests_clusterclass.yaml
@@ -187,6 +187,69 @@ jobs:
           sudo rm -rf /home/tanzu/.local/share/tanzu-cli/*
           sudo sh -c 'find /tmp -type d -name "go-build*" -mtime +3 | xargs rm -rf'
 
+      - name: show environment
+        uses: hmarr/debug-action@v2
+
+      - name: Extract PR Number
+        uses: Dovyski/payload-info-action@master
+        id: get_pr_num
+        with:
+          # when event is a pull request, obtaining the PR number is obvious
+          filter_pull_request: '.number'
+          # when event is a push (merge of PR), since we require linear history,
+          # we are not even generating a merge commit that can help identify the
+          # PR number, so don't even try. Instead we just hard-code to a dummy
+          # value.
+          filter_push: 1009
+
+      - name: Determine registry directory
+        uses: Dovyski/payload-info-action@master
+        id: get_registry_dir
+        with:
+          # when event is a pull request, use the PR number as the registry dir
+          filter_pull_request: '.number'
+          # when event is a push, use branch name instead
+          filter_push: '.ref | sub("^refs/heads/"; "")'
+
+      - name: Extract registry directory
+        shell: bash
+        run: |
+          export REGDIR=$(echo ${{ steps.get_registry_dir.outputs.value }} | tr -d '"')
+          echo "##[set-output name=regdir;]$(echo $REGDIR)"
+        id: extract_reg_dir
+
+      - name: Set registry
+        id: set_registry
+        shell: bash
+        run: |
+          if [ -z "${gcp_bucket_sa}" ]; then
+            echo "##[set-output name=registry;]localhost:5001"
+          else
+            echo "##[set-output name=registry;]gcr.io/eminent-nation-87317"
+          fi
+
+      - name: Find Comment
+        uses: peter-evans/find-comment@v1
+        id: findcomment
+        with:
+          issue-number: ${{ steps.get_pr_num.outputs.value }}
+          body-includes: //usebom
+          direction: last
+
+      - name: Extract Alternate BOM
+        shell: bash
+        run: |
+          export BOMCOMMENT=$(echo -e "${{ steps.findcomment.outputs.comment-body }}" | tr -d "[:space:]")
+          echo "##[set-output name=bompath;]$(echo "$BOMCOMMENT" | awk -F : '{print $2}')"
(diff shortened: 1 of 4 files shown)


.github/workflows/tkg_integration_tests_clusterclass.yaml
```diff
@@ -40,6 +40,87 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  integ-tests-docker-cc:
+    name: TKG Integration Tests CAPD CC
```
anujc25, line 44: We should use consistent naming here similar to what we are using for CAPA?


```suggestion
    name: TKG Integration Tests CAPD (ClusterClass)
```

anujc25:
I see we are not building packages here and not testing the changes through package installation. I think we need to do similar to what we are doing for CAPA CC test. https://github.com/vmware-tanzu/tanzu-framework/blob/main/.github/workflows/tkg_integration_tests_clusterclass.yaml#L144-L175

.github/workflows/tkg_integration_tests_clusterclass.yaml
```diff
@@ -40,6 +40,120 @@ concurrency:
...
+        run: rm -rf ~/.tanzu ~/.tkg ~/.config; docker kill $(docker ps -q) | true; docker system prune -a --volumes -f
+
+      - name: Run TKG integration tests
+        run: |
+          if [[ ! -z "${{ steps.extract_bom.outputs.bompath }}" ]]; then
+            export TKG_DEFAULT_COMPATIBILITY_IMAGE_PATH=${{ steps.extract_bom.outputs.bompath }}
+          fi
+          make configure-bom
+          make generate-embedproviders
+          E2E_CONFIG=$PWD/pkg/v1/tkg/test/config/cc/docker/docker.yaml hack/tools/bin/ginkgo -nodes=3 -v -trace -tags embedproviders pkg/v1/tkg/test/tkgctl/docker_cc
```
anujc25, line 155: We need to set this env that we are setting in AWS CC tests to use management package repository that we are building.
```
          _MANAGEMENT_PACKAGE_REPO_IMAGE: ${{ steps.set_registry.outputs.registry }}/tanzu_framework/github-actions/${{ steps.extract_reg_dir.outputs.regdir }}/management:v0.21.0
          _MANAGEMENT_PACKAGE_VERSION: "0.21.0"
          TKG_CUSTOM_IMAGE_REPOSITORY: "projects-stg.registry.vmware.com/tkg"
```

.github/workflows/tkg_integration_tests_clusterclass.yaml
```diff
@@ -40,6 +40,120 @@ concurrency:
...
+        uses: docker/login-action@v2
+        with:
+          registry: gcr.io
+          username: _json_key
+          password: ${{ secrets.GCP_BUCKET_SA }}
+
+      - name: Build management packages
+        env:
+          NUM: ${{ github.run_id }}
+          OCI_REGISTRY: ${{ steps.set_registry.outputs.registry }}/tanzu_framework/github-actions/${{ steps.extract_reg_dir.outputs.regdir }}
```
anujc25, line 123: We should use different path for CAPD and CAPA package building. May be something like appending `capd` for capd tests?

```suggestion
          OCI_REGISTRY: ${{ steps.set_registry.outputs.registry }}/tanzu_framework/github-actions/${{ steps.extract_reg_dir.outputs.regdir }}/capd
```

And for existing CAPA tests, we can append `capa` to the registry path so that both pipeline doesn't use same path to publish packages? cc @vuil

vuil: thinking out loud here:
we are talking about the same workflow file here right? 
why would we need to build and publish twice?

saji-pivotal: This kinda makes sense to me. Will wait for @vuil 's input on adding this

vuil: My input is not to build publish twice so we don't even have this problem to begin with.
Factor out the part where we build/publish the management repo, then have other jobs depend on it.
Check out https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds

.github/workflows/tkg_integration_tests_clusterclass.yaml
```diff
@@ -149,30 +280,123 @@ jobs:
...
+        with:
+          # when event is a pull request, obtaining the PR number is obvious
+          filter_pull_request: '.number'
+          # when event is a push (merge of PR), since we require linear history,
+          # we are not even generating a merge commit that can help identify the
+          # PR number, so don't even try. Instead we just hard-code to a dummy
+          # value.
+          filter_push: 1009
+
+      - name: Determine registry directory
```
vuil, line 331: I think all the steps in the integ-tests-xxx that involves figuring out the registry related variables can also be removed. Instead we should be able to reference the variables computed in the build-and-publish-packages job instead. See if something like jobs._jobid_.steps.get_registry_dir.outputs.value works. (You will want to set a friendly id: value for each of the jobs, but especially the first one)

saji-pivotal: This was a great suggestion. It needed something like `needs.<jobID>.outputs.<outputID>`

anujc25 (approved):

Issue #3149, opened 2022-08-17
Plugin specific configuration like `BOMRepo` and `CompatibilityFilePath` should not be part of `ClientConfig` API


anujc25:
**Bug description**

Currently, ClientConfig API is used to store core tanzu configuration at `~/.config/tanzu/config.yaml`  However, there are few configurations like `bomRepo` and `compatibilityFilePath` was also added to this API as part of #1605 which is more for specific plugins and has nothing to do with the core tanzu CLI.

```
apiVersion: config.tanzu.vmware.com/v1alpha1
clientOptions:
  cli:
    bomRepo: projects.registry.vmware.com/tce
    compatibilityFilePath: tkg-compatibility
```

We should remove this configuration from the `ClientConfig` API and put it to plugin specific configuration file.

**Affected product area (please put an X in all that apply)**

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- (x) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

Remove plugin specific configuration like `BOMRepo` and `CompatibilityFilePath` from `ClientConfig` API

**Steps to reproduce the bug**
N/A

**Version** (include the SHA if the version is not obvious)

**Environment where the bug was observed (cloud, OS, etc)**

**Relevant Debug Output (Logs, manifests, etc)**

PR #3264, opened 2022-09-02
Remove `tkgconfigpaths.DefaultTKGBOMRepo` and `tkgconfigpaths.DefaultTKGCompatibilityPath`  dependency from `pkg/v1/config`

anujc25:
- This PR updates the the `pkg/v1/config` module to not rely on `tkgconfigpaths` package.
- Currently, `pkg/v1/config` was using to fetch `DefaultTKGBOMRepo` and `DefaultTKGCompatibilityPath` information from `tkgconfigpaths`. Considering we are do not want to store plugins specific configuration into the Tanzu Config file we can remove this dependency and we do not need to dynamically set this information in the tkg client config.
- As part of the PR, I am removing this dependency and hard-coding values that Core CLI will be configuring for `clientOptions.cli. bomRepo` and `clientOptions.cli. compatibilityFilePath ` to default TKG production registry to make sure existing TKG plugins continue to work as it is.

Related to #3149

### Release note

```release-note
Remove configuring `clientOptions.cli.bomRepo` and `clientOptions.cli.compatibilityFilePath` from Tanzu-Config and use hard-coded values to support legacy plugins. This configuration is deprecated and will be removed from future version.
```


vuil (approved):
lgtm.
We should add to the release notes that these settings will be removed.

Also, we should find a place to explain (I think a .md in-tree that walks through how BOMs are processed should be the one to cover this) that if the default compat/bom locations are not what is required for the CLI/plugins, that env vars should continue to be used to override them.

anujc25:
> We should add to the release notes that these settings will be removed.
Updated the Release-Notes.


>Also, we should find a place to explain (I think a .md in-tree that walks through how BOMs are processed should be the one to cover this) that if the default compat/bom locations are not what is required for the CLI/plugins, that env vars should continue to be used to override them.

I will file separate issue to track the documentation on how BoMs are processed and the changes regarding default compat/bom locations are not what is required for CLI/plugins.

PR #3276, opened 2022-09-07
Support test plugins with `context-aware-cli-for-plugins` feature-flag enabled

### LINKED ISSUE #1164: Support test plugins with context-aware cli discovery
With the implementation of #1144 as part of #74, Tanzu CLI will switch to using API-driven and context-aware plugin discovery. However, the current approach of Plugin Discovery and Distribution does not take `test` plugins into consideration.

And the `Plugin Tests` pipeline is still using a non-context-aware way (` ENABLE_CONTEXT_AWARE_PLUGIN_DISCOVERY=false`) of installing test plugins for the tests.

We need to support test plugins as part of context-aware discovery work and completely switch to using this new of plugin discovery and installation.

- Continue using non-context-aware discovery for `Plugin Tests`

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [x] Plugin
- [ ] Security
- [x] Test and Release
- [ ] User Experience

**Additional context**

anujc25:

This PR adds support for installing test plugin along with plugin installation in local mode with `tanzu test fetch --local` command.

- The change also implements `FetchTest` interface for `local` artifacts and throws error when using the same with `oci` and `http` artifacts.

- This change also updates `Plugin Test` pipeline to not rely on old implementation and uses new `context-aware-cli-for-plugins: true` to deploy and run test plugins.

Fixes #1164

- Build the cli and plugins locally. Generates artifacts under `artifacts/darwin/amd64/cli/` directory
- Use `tanzu plugin install all --local artifacts/darwin/amd64/cli/` to install plugins along with test plugins
- Run `tanzu test plugin --help` to verify all plugins are available under the command to run tests
- Run `tanzu test plugin <plugin-name>` to run the plugin tests

### Release note

```release-note
Support test plugins with `context-aware-cli-for-plugins` feature-flag enabled
```

### DIFF
--- a/cli/core/pkg/artifact/http.go
+++ b/cli/core/pkg/artifact/http.go
@@ -9,6 +9,8 @@ import (
 	"io"
 	"net/http"
 	"time"
+
+	"github.com/pkg/errors"
 )
 
 const (
@@ -71,4 +73,9 @@ func (g *HTTPArtifact) Fetch() ([]byte, error) {
 	}
 
 	return out, nil
+}
+
+// FetchTest returns test artifact
+func (g *HTTPArtifact) FetchTest() ([]byte, error) {
+	return nil, errors.New("fetching test plugin from HTTP source is not yet supported")
 }
--- a/cli/core/pkg/artifact/local.go
+++ b/cli/core/pkg/artifact/local.go
@@ -58,13 +58,13 @@ func (l *LocalArtifact) FetchTest() ([]byte, error) {
 		return nil, errors.Wrap(err, "unable to read test plugin directory")
 	}
 	if len(dirEntries) != 1 {
-		return nil, errors.Errorf("Expected only 1 file under the '%v' directory, but found %v files", testPluginDir, len(dirEntries))
+		return nil, errors.Errorf("expected only 1 file under the '%v' directory, but found %v files", testPluginDir, len(dirEntries))
 	}
 
 	// Assuming the file under the test directory is the test plugin binary
 	testPluginFile := dirEntries[0]
 	if testPluginFile.IsDir() {
-		return nil, errors.Errorf("Expected to find test plugin binary but found directory '%v'", testPluginFile)
+		return nil, errors.Errorf("expected to find test plugin binary but found directory %q", testPluginFile.Name())
 	}
 
 	testPluginPath := filepath.Join(testPluginDir, testPluginFile.Name())
--- a/cli/core/pkg/artifact/local.go
+++ b/cli/core/pkg/artifact/local.go
@@ -45,3 +45,32 @@ func (l *LocalArtifact) Fetch() ([]byte, error) {
 	}
 	return b, nil
 }
+
+// FetchTest reads test plugin artifact based on the local plugin artifact path
+// To fetch the test binary from the plugin we are using plugin binary path and creating test plugin path from it
+// If the plugin binary path is `artifacts/darwin/amd64/cli/cluster/v0.27.0-dev/tanzu-cluster-darwin_amd64`
+// then test plugin binary will be under `artifacts/darwin/amd64/cli/cluster/v0.27.0-dev/test/` directory
+func (l *LocalArtifact) FetchTest() ([]byte, error) {
+	// test plugin directory based on the plugin binary location
+	testPluginDir := filepath.Join(filepath.Dir(l.Path), "test")
+	dirEntries, err := os.ReadDir(testPluginDir)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to read test plugin directory")
+	}
+	if len(dirEntries) != 1 {
+		return nil, errors.Errorf("Expected only 1 file under the '%v' directory, but found %v files", testPluginDir, len(dirEntries))
+	}
+
+	// Assuming the file under the test directory is the test plugin binary
+	testPluginFile := dirEntries[0]
+	if testPluginFile.IsDir() {
+		return nil, errors.Errorf("Expected to find test plugin binary but found directory '%v'", testPluginFile)
+	}
+
+	testPluginPath := filepath.Join(testPluginDir, testPluginFile.Name())
+	b, err := os.ReadFile(testPluginPath)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error while reading test artifact")
+	}
+	return b, nil
+}
--- a/cli/core/pkg/artifact/local_test.go
+++ b/cli/core/pkg/artifact/local_test.go
@@ -1,0 +1,48 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package artifact
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// When local artifact directory exists and test directory exists
+func TestLocalArtifactWhenArtifactAndTestDirExists(t *testing.T) {
+	assert := assert.New(t)
+
+	testPluginPath, err := filepath.Abs(filepath.Join("test", "local", "v0.0.1", "tanzu-plugin-darwin_amd64"))
+	assert.NoError(err)
+	artifact := NewLocalArtifact(testPluginPath)
+
+	b, err := artifact.Fetch()
+	assert.NoError(err)
+	assert.Contains(string(b), "plugin binary")
+
+	b, err = artifact.FetchTest()
+	assert.NoError(err)
+	assert.Contains(string(b), "test plugin binary")
+}
+
+// When local artifact doesn't exists and multiple files exists within test directory
+func TestLocalArtifactWhenArtifactDoesntExistsAndMultipleFilesUnderTest(t *testing.T) {
+	assert := assert.New(t)
+
+	testPluginPath, err := filepath.Abs(filepath.Join("test", "local", "v0.0.2", "plugin_binary_does_not_exists"))
+	assert.NoError(err)
+
+	artifact := NewLocalArtifact(testPluginPath)
+
+	// When local artifact doesn't exists
+	_, err = artifact.Fetch()
+	assert.Error(err)
+	assert.ErrorContains(err, "error while reading artifact")
+
+	// When multiple files exists under test directory
+	_, err = artifact.FetchTest()
+	assert.Error(err)
+	assert.ErrorContains(err, "Expected only 1 file under the")
+}
--- a/cli/core/pkg/artifact/local_test.go
+++ b/cli/core/pkg/artifact/local_test.go
@@ -44,5 +44,5 @@ func TestLocalArtifactWhenArtifactDoesntExistsAndMultipleFilesUnderTest(t *testi
 	// When multiple files exists under test directory
 	_, err = artifact.FetchTest()
 	assert.Error(err)
-	assert.ErrorContains(err, "Expected only 1 file under the")
+	assert.ErrorContains(err, "expected only 1 file under the")
 }
--- a/cli/core/pkg/distribution/artifact.go
+++ b/cli/core/pkg/distribution/artifact.go
@@ -80,6 +80,27 @@ func (aMap Artifacts) Fetch(version, os, arch string) ([]byte, error) {
 		"arch:%s", version, os, arch)
 }
 
+// FetchTest the test binary for a plugin version.
+func (aMap Artifacts) FetchTest(version, os, arch string) ([]byte, error) {
+	a, err := aMap.GetArtifact(version, os, arch)
+	if err != nil {
+		return nil, err
+	}
+
+	if a.Image != "" {
+		return artifact.NewOCIArtifact(a.Image).FetchTest()
+	}
+	if a.URI != "" {
+		u, err := artifact.NewURIArtifact(a.URI)
+		if err != nil {
+			return nil, err
+		}
+		return u.FetchTest()
+	}
+
+	return nil, errors.Errorf("invalid artifact for version:%s, os:%s, arch:%s", version, os, arch)
+}
+
 // GetDigest returns the SHA256 hash of the binary for a plugin version.
 func (aMap Artifacts) GetDigest(version, os, arch string) (string, error) {
 	a, err := aMap.GetArtifact(version, os, arch)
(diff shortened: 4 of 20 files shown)


cli/core/pkg/artifact/local.go
```diff
@@ -45,3 +45,27 @@ func (l *LocalArtifact) Fetch() ([]byte, error) {
 	}
 	return b, nil
 }
+
+// FetchTest reads test plugin artifact based on the local plugin artifact path
+// To fetch the test binary from the plugin we are using plugin binary path and creating test plugin path from it
+// If the plugin binary path is `artifacts/darwin/amd64/cli/cluster/v0.27.0-dev/tanzu-cluster-darwin_amd64`
+// then test plugin binary will be under `artifacts/darwin/amd64/cli/cluster/v0.27.0-dev/test/` directory
```
saji-pivotal, line 52: q: Will the binary will be under `artifaces/darwin/amd64/cli/cluster/vX.Y.Z-dev/test` directory for any future versions of framework?

cli/core/pkg/artifact/local.go
```diff
@@ -45,3 +45,27 @@ func (l *LocalArtifact) Fetch() ([]byte, error) {
...
+func (l *LocalArtifact) FetchTest() ([]byte, error) {
+	// test plugin directory based on the plugin binary location
+	testPluginDir := filepath.Join(filepath.Dir(l.Path), "test")
+	dirEntries, err := os.ReadDir(testPluginDir)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to read test plugin directory")
+	}
+	if len(dirEntries) != 1 {
+		return nil, errors.Errorf("unable to find test plugin under test directory(%v). Found %v files under that directory", testPluginDir, len(dirEntries))
```
saji-pivotal, line 61: nit: The error might be a little misleading, we could probably just say "Expected only 1 file under the directory, but found ..." . I'm saying this to cover the case where the test plugin is one of many files in the directory.

cli/core/pkg/artifact/local.go
```diff
@@ -45,3 +45,27 @@ func (l *LocalArtifact) Fetch() ([]byte, error) {
...
+	}
+
+	// Assuming the file under the test directory is the test plugin binary
```
saji-pivotal, line 64: Q: Is there any way of verifying that the file under the test directory is the test plugin binary?

cli/core/pkg/artifact/local.go
```diff
@@ -45,3 +45,27 @@ func (l *LocalArtifact) Fetch() ([]byte, error) {
...
+	testPluginPath := filepath.Join(testPluginDir, dirEntries[0].Name())
```
saji-pivotal, line 65: Can we extract dirEntries[0] into another variable? That way we can potentially check if it is `nil` or run any other checks on it we can think of

cli/core/pkg/artifact/local_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLocalArtifact(t *testing.T) {
+	assert := assert.New(t)
+
+	// When local artifact directory exists and test directory exists
```
saji-pivotal, line 16: nit-ish: IMO we should make these helper functions instead of commenting out what we expect from the subsequent lines

cli/core/pkg/distribution/artifact.go
```diff
@@ -80,6 +80,28 @@ func (aMap Artifacts) Fetch(version, os, arch string) ([]byte, error) {
...
+	}
+	if a.URI != "" {
+		u, err := artifact.NewURIArtifact(a.URI)
+		if err != nil {
+			return nil, err
+		}
+		return u.FetchTest()
+	}
+
+	return nil, errors.Errorf("invalid artifact for version:%s, os:%s, "+
```
saji-pivotal, line 101: I'm just curious: Are you splitting the content of Errorf into two lines just for readability?

cli/core/pkg/pluginmanager/manager.go
```diff
@@ -488,6 +488,16 @@ func installOrUpgradePlugin(serverName string, p *plugin.Discovered, version str
 	descriptor.Discovery = p.Source
 	descriptor.DiscoveredRecommendedVersion = p.RecommendedVersion
 
+	b, err = p.Distribution.FetchTest(version, runtime.GOOS, runtime.GOARCH)
+	if err == nil {
+		testpluginPath := filepath.Join(common.DefaultPluginRoot, pluginName, fmt.Sprintf("test-%s", version))
+		err = os.WriteFile(testpluginPath, b, 0755)
+		if err != nil {
+			return errors.Wrap(err, "could not write file")
```
saji-pivotal, line 496: nit: We should make the error message a bit more descriptive IMO

saji-pivotal (changes requested):
Great work! I have a few questions though

cli/core/pkg/artifact/local.go
anujc25: `artifaces/darwin/amd64/cli` is just the directory that we pass with `--artifacts` flag with `tanzu builder cli compile` command. If we need to change the directory we pass with `--artifacts` flag this might change the root level directory but assumption here is on how builder plugin places plugin and test plugin binaries. Assuming we are passing ${ARTIFACTS_DIR} as part of `--artifacts` flag with compile command...
```
tanzu builder cli compile --path ./cmd/cli/plugin --artifacts ${ARTIFACTS_DIR}
```
the plugin binary will be under `${ARTIFACTS_DIR}/<plugin-name>/tanzu-<plugin-name>-darwin_amd64` and test plugin binary will be under `${ARTIFACTS_DIR}/<plugin-name>/test/` directory.

cli/core/pkg/distribution/artifact.go
anujc25: Nope. Might be an oversight. I will fix this as part of addressing comments.

cli/core/pkg/artifact/local.go
anujc25: I was thinking about this and one thing that came into my mind was to run `<plugin-binary> info` command and check the `group` information from the output. But I wasn't sure about the how important is this validation here considering if it is incorrect we will see the error with `tanzu test plugin <plugin-name>` command invocation. It would be helpful if you can provide your thoughts on this.

```
{"name":"cluster-test","description":"test for cluster","version":"v0.0.1","buildSHA":"","digest":"","group":"Test","docURL":"","completionType":0,"aliases":["cluster-alias"],"installationPath":"","testPluginInstallationPath":"","discovery":"","scope":"","status":"","discoveredRecommendedVersion":""}
```

chandrareddyp:
I ran below commands, and got output as `Plugin tests`
```
❯ tanzu test plugin --help
Plugin tests
❯ tanzu test plugin cluster
Plugin tests

```
IMO, as an end-user, I would expect more details!
 this is first time I ran `tanzu test`, I don't know if its a relevant question!

chandrareddyp:
I ran the below commands to test plugins, but I don't see any output; it might be successful, but I think it would be better if we print output.
```
❯ tanzu test plugin feature
❯ tanzu test plugin secret
❯ tanzu test plugin package
```

cli/core/pkg/artifact/local.go
```diff
@@ -45,3 +45,27 @@ func (l *LocalArtifact) Fetch() ([]byte, error) {
...
+// FetchTest reads test plugin artifact based on the local plugin artifact path
+// To fetch the test binary from the plugin we are using plugin binary path and creating test plugin path from it
+// If the plugin binary path is `artifacts/darwin/amd64/cli/cluster/v0.27.0-dev/tanzu-cluster-darwin_amd64`
+// then test plugin binary will be under `artifacts/darwin/amd64/cli/cluster/v0.27.0-dev/test/` directory
+func (l *LocalArtifact) FetchTest() ([]byte, error) {
+	// test plugin directory based on the plugin binary location
+	testPluginDir := filepath.Join(filepath.Dir(l.Path), "test")
+	dirEntries, err := os.ReadDir(testPluginDir)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to read test plugin directory")
```
chandrareddyp, line 58: I think it would be better to print `testPluginDir` in the error message.

anujc25: I think `err` will already list the plugin directory name in this case considering we are wrapping the error.

anujc25:
> I ran the below commands to test plugins, but I don't see any output; it might be successful, but I think it would be better if we print output.

I was looking into this and found out that as part of invocation, Tanzu CLI just invokes `test` plugin binary for a specified plugin when we do `tanzu test plugin <plugin-name>` and the test plugin is responsible for printing the status on the stdout. 

As part of the plugin initialization that `builder` plugin does, we bootstrap `test/main.go` file for the plugin so that we can show some output of the test run. It turns out that plugins like `package`, `secret`, `feature` has removed that piece of the code from its test plugin codebase and hence we are not seeing anything in the output.

I feel that is a plugin specific code and Tanzu CLI might not be able to show any output in that case if plugin author decide not to print anything on the terminal because the test plugin is not getting used.

anujc25:
> I ran below commands, and got output as `Plugin tests`
> 
> ```
> ❯ tanzu test plugin --help
> Plugin tests
> ❯ tanzu test plugin cluster
> Plugin tests
> ```
> 
> IMO, as an end-user, I would expect more details as output! this is first time I ran `tanzu test`, I don't know if its a relevant question!

Regarding `tanzu test plugin cluster` when not installed showing helptext. I have filled this bug (https://github.com/vmware-tanzu/tanzu-framework/issues/3473) that we should address separately.

cli/core/pkg/artifact/local.go
saji-pivotal: Thanks for the explanation! I'd be ok with`tanzu test plugin <plugin-name>` throwing a meaningful error in order to verify the binary.

saji-pivotal (approved):

cli/runtime/apis/cli/v1alpha1/catalog_types.go
```diff
@@ -141,6 +141,10 @@ type PluginDescriptor struct {
 	// E.g., cluster/v0.3.2@sha256:...
 	InstallationPath string `json:"installationPath"`
 
+	// TestPluginInstallationPath is a installation path for a test plugin binary.
+	// E.g., cluster/test-v0.3.2
+	TestPluginInstallationPath string `json:"testPluginInstallationPath"`
```
marckhouzam, line 146: Why do we need this to be in the `PluginDescriptor`?  The `TestPluginInstallationPath` is built using a fixed pattern so do we really need to store it like this instead of building it when we need it?  Could we instead adjust `runner.RunTest()` and keep using it like before?

Unless you want to allow a plugin to override this value?

marckhouzam (commented):
Thanks @anujc25 for the nice improvement.
It is hard for me to understand the subtleties since I don't know what was the old behaviour, so my questions may not make much sense, but I'll try anyway.

The (very small) README of the `test` plugin mentions that to install all the tests plugins, we should use `tanzu test fetch`.  In your PR description you used `tanzu plugin install all` instead.  What do you prefer users to do between the two approaches?  And if we use `tanzu test fetch`, would it allow us to avoid installing the tests plugins when doing a `tanzu plugin install`?

marckhouzam:
Sorry, by previous comment should have used `tanzu test fetch` (and not `tanzu plugin fetch`).  I have corrected it.

cli/runtime/apis/cli/v1alpha1/catalog_types.go
anujc25: Initially my thought was to make it flexible on where we store the `test` plugin binary. You brought a really good point here, considering we are following a very generic approach on where and how we are storing test plugin binary, I think we do not need to add `TestPluginInstallationPath`  in the `PluginDescriptor`. Also we can use `RunTest` function to determine the location. I will update the code based on this.

anujc25:
> The (very small) README of the test plugin mentions that to install all the tests plugins, we should use tanzu test fetch. In your PR description you used tanzu plugin install all instead. What do you prefer users to do between the two approaches? And if we use tanzu test fetch, would it allow us to avoid installing the tests plugins when doing a tanzu plugin install?

Good point. I think we should not install test plugin as part of plugin installation but just install the test plugin when we do `tanzu test fetch --local`. I am not sure about the previous behavior either but this makes more sense. I have updated the implementation accordingly. cc @marckhouzam

vuil (approved):
lgtm, thanks!

marckhouzam (approved):
LGTM

cli/core/pkg/pluginmanager/manager.go
```diff
@@ -488,6 +488,20 @@ func installOrUpgradePlugin(serverName string, p *plugin.Discovered, version str
 	descriptor.Discovery = p.Source
 	descriptor.DiscoveredRecommendedVersion = p.RecommendedVersion
 
+	if installTestPlugin {
+		log.Infof("Installing test plugin for '%v:%v'", p.Name, version)
+		b, err = p.Distribution.FetchTest(version, runtime.GOOS, runtime.GOARCH)
+		if err != nil {
+			log.Infof("Unable to install test plugin for '%v:%v'. Error: %v", p.Name, version, err)
```
rajathagasthya, line 495: We just want to log if test plugin install errors out?

rajathagasthya (commented):
Just minor comments. No blockers.

cli/core/pkg/artifact/http.go
```diff
@@ -72,3 +74,8 @@ func (g *HTTPArtifact) Fetch() ([]byte, error) {
 
 	return out, nil
 }
+
+// FetchTest returns test artifact
+func (g *HTTPArtifact) FetchTest() ([]byte, error) {
+	return nil, errors.New("fetching test plugin from HTTP source is not yet supported")
```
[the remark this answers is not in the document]
anujc25: I feel it would be better to use `errors` package consistently across all the files. Most of the other places we are using `errors.New`, `errors.Wrap`, `errors.Wrapf` to handle the error.

cli/core/pkg/pluginmanager/manager.go
anujc25: Thanks for catching this. Updated.

cli/core/pkg/artifact/http.go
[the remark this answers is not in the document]
rajathagasthya: Now that `fmt.Errorf` supports wrapping with `%w`, I don't feel `pkg/errors` is necessary anymore.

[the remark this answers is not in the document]
anujc25: Thanks for the feedback. This is good to know. 

But if you don't mind, I would prefer to do that change as a separate PR and not hold this PR for that.

PR #3280, opened 2022-09-07
Decouple addons module from tanzu-framework

### LINKED ISSUE #3290: Decouple addons go module from tanzu-framework
As part of tanzu-framework refactoring we are breaking up repo into multiple go modules. Addons go module should be decoupled from tanzu-framework go module

N/A

- ( ) APIs
- (x) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Additional context**

vijaykatam:
* Move bom types to apis/run
* Move webhook package to addons
* Update constants
* Copy buildinfo type to addons
* Add fakegen for cluster client

Fixes #3290

Existing unit tests and integration tests

### Release note

```release-note
None
```


addons/hack/fakegen.go
```diff
@@ -0,0 +1,12 @@
+package hack
+
+import (
+	crtclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//go:generate counterfeiter -o ../pkg/fakeclusterclient/crtclusterclient.go --fake-name CRTClusterClient . CrtClient
```
maralavi, line 7: If ```counterfeiter``` is going to be used here, I think Makefile needs to have a way to download it if not existing in order for ```generate-fakes``` to work. For example, for ```packageclients```, I used the root ```hack/tools/bin``` for now (otherwise, a ```hack/tools/bin``` can be introduced under addons):

https://github.com/vmware-tanzu/tanzu-framework/blob/main/packageclients/Makefile#L17
https://github.com/vmware-tanzu/tanzu-framework/blob/main/packageclients/Makefile#L12

maralavi, line 7: Also wouldn't it be better for a ```fakes``` directory be created in order to group of all future addons fake files (instead of current ```fakeclusterclient```) and better code readability:  e.g.```../pkg/fakes/crtclusterclient.go```?

addons/test/testutil/test_helpers.go
```diff
@@ -40,7 +40,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
-	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/webhooks"
+	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/webhooks"
```
maralavi, line 43: Considering that with this PR, now have both ```addons/webhooks``` and ```addons/pkg/webhooks```, would it make sense to merge them & use a same directory for both?

addons/hack/fakegen.go
vijaykatam: I added hack/tools for addons

vijaykatam: I think we can restructure if we have more fakes in the future

addons/test/testutil/test_helpers.go
vijaykatam: pkg/webhooks is external API for webhook utilities where addons/webhooks are the webhooks for various APIs

maralavi (approved):

PR #3282, opened 2022-09-08
Azure cc test

wjun:
Make azure cc dev plan creation work

Fixes #

### Release note

```release-note

```


xiaoping8385 (approved):

packages/tkg-clusterclass-azure/bundle/config/upstream/base.yaml
```diff
@@ -1255,22 +1267,19 @@ metadata:
...
+      location: #@ data.values.AZURE_LOCATION
       networkSpec:
         vnet:
-          name: "${AZURE_VNET_NAME:=${CLUSTER_NAME}-vnet}"
-          resourceGroup: ${AZURE_VNET_RESOURCE_GROUP:=${CLUSTER_NAME}}
-          cidrBlock: ${AZURE_VNET_CIDR}
+          cidrBlock: #@ data.values.AZURE_VNET_CIDR
         subnets:
-        - name: ${AZURE_CONTROL_PLANE_SUBNET_NAME}
+        - name: #@ data.values.AZURE_CONTROL_PLANE_SUBNET_NAME
```
ydp, line 1276: Hi Jun, just want to learn the code, what's the difference between `${var}` and `#@ data.value.var` ?

ydp (approved):

DanielXiao (approved):

packages/tkg-clusterclass-azure/bundle/config/upstream/base.yaml
DanielXiao: ${var} is deprecated and should not be used in ClusterClass because ytt cli can not interpolate it.

wjun: The ${var} will not be replaced in the package until cc topology controller processes it, while data.value.var will be replaced with value when the package is installed. ${var} is blocked by azureclustertemplate admission controller which has some subnet name regex check not allowing $ and {}, For this case, we must use data.values.var format. Ideally, we should remove ${var} from all cc related yamls.

PR #3285, opened 2022-09-08
Correct go module name in cli/runtime/hack/tools

randomvariable:

Although the module name is valid, there is already another module with that name, so this change is important.

Fixes #

### Release note

```release-note
NONE
```


marckhouzam (approved):
LGTM

Although the module name is valid, there is already another module with that name, so this change is important.

anujc25 (approved):
Thanks for the fix.

Issue #899, opened 2021-10-15
How to build a single built-in plugin for local iteration?


mamachanko:
Hi, I am trying to get myself set up for contributing. However, I struggle with building the one plugin that I am making changes to, installing it and testing my changes.

I understand that there're a couple of make targets that should help with this and there's `tanzu builder`, but I fail to set up a workflow.

Let's say I want to work on the `package` plugin:
1. As documented, on a clean checkout I do `make build-install-cli-local` which builds and installs all plugins for my local architecture.
1. Then I am making changes to e.g. `cmd/cli/plugin/package/package_install.go`
1. Then I am trying to rebuild the `package` plugin again without having to wait for all of `make build-install-cli-local`, but that fails:

```shell
$ tanzu builder cli compile --match package --version $(make version)
ℹ  building local repository at ./artifacts
ℹ  🐷 - building plugin at path "cmd/cli/plugin/package"
✖  🐷 - error: exit status 1
✖  🐷 - output: ✖  plugin "package" version cannot be empty; version "package" "" is not a valid semantic version
exit status 1

========
✖  🐷 - building plugin "cmd/cli/plugin/package" failed - exit status 1
Error: exit status 1

✖  exit status 1
```

`$(make version)` was my best guess for the `--version` parameter because it asks for the version of the root.

I am aware that tests and documentation are part of the process, but I would like to discover and get a feel for all the machinery before.

saji-pivotal:
@mamachanko For now, can you try running `make build-cli-local` and see if the plugin you want is installed under `./artifacts/darwin/amd64/cli` ? You can then use `./artifacts/darwin/amd64/cli/<plugin_name>/<version>/tanzu-<plugin>-<arch>` to use the plugin. Please let me know if you test out the workaround I suggested.

mamachanko:
> ./artifacts/darwin/amd64/cli/<plugin_name>/<version>/tanzu-<plugin>-<arch>

Thanks @saji-pivotal for the workaround. It does work!

However, `make build-cli-local` takes ~3mins on my machine. It builds all the plugins. Is that the recommended way of iterating on a built-in plugin? Is there a way to only built, say the `package` plugin without hand-crafting a `go build ...` command?

chandrareddyp:
- This is not a bug, we need to pass the Version info as input value to `--ldflags` flag, here is eg:
[...]

PR #3289, opened 2022-09-08
Update the builder plugin to make version flag optional and add new 'build-plugin-local' make target 

chandrareddyp:

1. Updates the command `tanzu builder cli compile --match <package_name> --version` logic to make the flag "--version" as optional and the help text of the flag.
2. New make targets `build-plugin-local` and `install-plugin-local` has added to give flexibility to users to build and install the given plugin locally.

Fixes #https://github.com/vmware-tanzu/tanzu-framework/issues/899

1. The command ` tanzu builder cli compile --match package --version $(make version) --ldflags "-X 'github.com/vmware-tanzu/tanzu-framework/pkg/v1/buildinfo.Version=v0.27.0-dev'"` tested with --ldflags having Version info, works fine.
2. The command ` tanzu builder cli compile --match package ---ldflags "-X 'github.com/vmware-tanzu/tanzu-framework/pkg/v1/buildinfo.Version=v0.27.0-dev'"` tested without --version flag, works fine.
3. The new make target `build-plugin-local` tested: With valid plugin name `make build-plugin-local PLUGIN_NAME=package` works fine, builds the given plugin. Also tested with empty and in-valid PLUGIN_NAME, it does throw an error saying: `The PLUGIN_NAME: '' is not valid or not exists or empty, please provide valid PLUGIN_NAME.`
4. The new make target `install-plugin-local` tested: With valid plugin name `make install-plugin-local PLUGIN_NAME=login` works fine, build and install the given plugin. Also tested with empty and in-valid PLUGIN_NAME, it does throw an error saying: `The PLUGIN_NAME: '' is not valid or not exists or empty, please provide valid PLUGIN_NAME.`

### Release note

```release-note
The command `tanzu builder cli compile` has been updated to make the flag `--version` as optional
New make target  `build-plugin-local` has been added to build the given plugin locally
New make target `install-plugin-local` has been added to build and install the given plugin locally
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,7 @@ endif
 # Package tooling related variables
 PACKAGE_VERSION ?= ${BUILD_VERSION}
 REPO_BUNDLE_VERSION ?= ${BUILD_VERSION}
+PLUGIN_PATH ?= ./cmd/cli/plugin
 
 DOCKER_DIR := /app
 SWAGGER=docker run --rm -v ${PWD}:${DOCKER_DIR}:$(DOCKER_VOL_OPTS) quay.io/goswagger/swagger:v0.21.0
@@ -239,6 +240,20 @@ build-plugin-admin-with-oci-discovery: ${CLI_ADMIN_JOBS_OCI_DISCOVERY} publish-a
 
 .PHONY: build-plugin-admin-with-local-discovery
 build-plugin-admin-with-local-discovery: ${CLI_ADMIN_JOBS_LOCAL_DISCOVERY} publish-admin-plugins-all-local ## Build Tanzu CLI admin plugins with Local standalone discovery
+
+.PHONY: build-plugin-local
+build-plugin-local: prepare-builder ## Build given CLI Plugin locally, needs PLUGIN_NAME
+	@if [ "${PLUGIN_NAME}" = "" ] || [ ! -d ${PLUGIN_PATH}/${PLUGIN_NAME} ]; then \
+		echo "The PLUGIN_NAME: '${PLUGIN_NAME}' is not valid or not exists or empty, please provide valid PLUGIN_NAME." ; \
+	else \
+		$(BUILDER) cli compile --match "$(PLUGIN_NAME)" --version $(BUILD_VERSION) --ldflags "$(LD_FLAGS)" --path $(PLUGIN_PATH) --target local --artifacts artifacts/${GOHOSTOS}/${GOHOSTARCH}/cli ; \
+	fi
+
+.PHONY: install-plugin-local
+install-plugin-local: build-plugin-local ## Build and Install given CLI Plugin locally, needs PLUGIN_NAME
+	@if [ "${PLUGIN_NAME}" != "" ] && [ -d ${PLUGIN_PATH}/${PLUGIN_NAME} ]; then \
+		tanzu plugin install ${PLUGIN_NAME} --local $(ARTIFACTS_DIR)/$(GOHOSTOS)/$(GOHOSTARCH)/cli ; \
+	fi
 
 .PHONY: build-plugin-admin-%
 build-plugin-admin-%: prepare-builder
--- a/cmd/cli/plugin-admin/builder/cli.go
+++ b/cmd/cli/plugin-admin/builder/cli.go
@@ -45,7 +45,7 @@ func newCompileCmd() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVar(&compileArgs.Version, "version", "", "Version of the root cli (required)")
+	cmd.Flags().StringVar(&compileArgs.Version, "version", "", "Version of the core tanzu cli")
 	cmd.Flags().StringVar(&compileArgs.LDFlags, "ldflags", "", "ldflags to set on build")
 	cmd.Flags().StringVar(&compileArgs.Tags, "tags", "", "Tags to set on build")
 	cmd.Flags().StringVar(&compileArgs.Match, "match", compileArgs.Match, "Match a plugin name to build, supports globbing")
@@ -54,8 +54,6 @@ func newCompileCmd() *cobra.Command {
 	cmd.Flags().StringVar(&compileArgs.ArtifactsDir, "artifacts", compileArgs.ArtifactsDir, "Path to output artifacts")
 	cmd.Flags().StringVar(&compileArgs.CorePath, "corepath", "", "Path for core binary")
 	cmd.Flags().StringVar(&compileArgs.GoPrivate, "goprivate", "", "Comma-separated list of glob patterns of module path prefixes to set as GOPRIVATE on build")
-
-	_ = cmd.MarkFlagRequired("version")
 
 	return cmd
 }
--- a/cmd/cli/plugin-admin/builder/command/cli_compile.go
+++ b/cmd/cli/plugin-admin/builder/command/cli_compile.go
@@ -131,10 +131,6 @@ type errInfo struct {
 // setGlobals initializes a set of global variables used throughout the compile
 // process, based on the arguments passed in.
 func setGlobals(compileArgs *PluginCompileArgs) {
-	if compileArgs.Version == "" {
-		log.Fatal("version value must be set")
-	}
-
 	version = compileArgs.Version
 	artifactsDir = compileArgs.ArtifactsDir
 	ldflags = compileArgs.LDFlags


marckhouzam (commented):
This works for me.  One comment in-line.

pkg/v1/builder/command/cli_compile.go
```diff
@@ -239,6 +239,7 @@ func buildPlugin(path string, arch cli.Arch, id string) (plugin, error) {
 
 	var modPath string
 
+	ldflags += " -X 'github.com/vmware-tanzu/tanzu-framework/pkg/v1/buildinfo.Version=" + version + "' "
```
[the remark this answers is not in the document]
chandrareddyp: That's a good catch; we will definitely use https://github.com/vmware-tanzu/tanzu-framework/blob/main/cli/runtime/buildinfo/buildinfo.go after the refactoring, for now, I will add a comment and also add the Version from  https://github.com/vmware-tanzu/tanzu-framework/blob/main/cli/runtime/buildinfo/buildinfo.go, so that we can avoid bug if in case refactoring ignores adding Version from cli/runtime/buildinfo/buildinfo.go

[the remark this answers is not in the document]
anujc25: I am not sure we should do this hardcoded string here because it requires change in builder plugin everytime this `buildinfo` gets moved to different location.

Also, this is just general recommendation that plugins use to define their versions but plugin developer can hardcode the plugin version or rely on some other means to define plugin version in the PluginDescriptor. So, assuming this might not be good idea. 

We need to think about passing this type of version is even possible to build a specific plugin or not.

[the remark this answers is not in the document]
saji-pivotal: @chandrareddyp I don't think we should be hardcoding an ldflag in the code. IMO we should be able to set `buildinfo.Version` by doing something like what's being done in the Makefile - https://github.com/vmware-tanzu/tanzu-framework/blob/main/Makefile#L87-L115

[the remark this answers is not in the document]
marckhouzam: > @chandrareddyp I don't think we should be hardcoding an ldflag in the code. IMO we should be able to set `buildinfo.Version` by doing something like what's being done in the Makefile 

The problem with that is that if you just call `tanzu builder cli compile --match package --version v0.1.2` the `--version` flag will be ignored, which seems wrong from a user's point-of-view.  This is related to why @chandrareddyp was asking what this `--version` flag is really meant to do.

> I am not sure we should do this hardcoded string here because it requires change in builder plugin everytime this `buildinfo` gets moved to different location.

True.

> Also, this is just general recommendation that plugins use to define their versions but plugin developer can hardcode the plugin version or rely on some other means to define plugin version in the PluginDescriptor. So, assuming this might not be good idea.
> We need to think about passing this type of version is even possible to build a specific plugin or not.

Yeah, that is tricky.  It comes back to "what is the `--version` flag supposed to do?"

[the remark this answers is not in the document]
anujc25: > The problem with that is that if you just call tanzu builder cli compile --match package --version v0.1.2 the --version flag will be ignored, which seems wrong from a user's point-of-view.

I was not involved in the development of this plugin/command, however, Based on the help-text for version, it seems it is meant only for core-cli version and not for plugin version. ` --version string       Version of the root cli (required)`

I think, the main issue here is that `--version` flag should not be a required flag as we expect user to just build the plugin without core-cli.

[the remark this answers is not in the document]
chandrareddyp: The  `--version` flag is not required to compile the plugin, so we have made changes to make `--version` flag as optional.

anujc25 (commented):
`build-plugin-local` will compile the plugin and put it in the `artifacts` directory within the repository.

We should add `install-plugin-local` that installs the plugin as well. Because as a end user I would like to do some change in the plugin, then I want to build as well as install the plugin to test my change.

marckhouzam (commented):
Nit: 
* git recommends that the commit title (the first line) be 50 characters or less.  This is useful because that line is used in many different git outputs (see the first paragraph from https://git-scm.com/docs/git-commit#_discussion)
* as for the text of the commit, it is recommended that it wraps at 72 characters.  If not, it makes `git log` hard to read, among other git output.

Makefile
```diff
@@ -229,6 +230,10 @@ build-plugin-admin-with-oci-discovery: ${CLI_ADMIN_JOBS_OCI_DISCOVERY} publish-a
 .PHONY: build-plugin-admin-with-local-discovery
 build-plugin-admin-with-local-discovery: ${CLI_ADMIN_JOBS_LOCAL_DISCOVERY} publish-admin-plugins-all-local ## Build Tanzu CLI admin plugins with Local standalone discovery
 
+.PHONY: build-plugin-local
+build-plugin-local: ## Build plugin's locally for given plugin name (match-expression) and path
+	tanzu builder cli compile --match "$(PLUGIN_NAME)" --version $(BUILD_VERSION) --ldflags "$(LD_FLAGS)" --path $(PLUGIN_PATH) --target local --artifacts artifacts/${GOHOSTOS}/${GOHOSTARCH}/cli
```
marckhouzam, line 235: 1. I don't think we should use `tanzu builder` because the `builder` plugin may not be built, or up to date.
Elsewhere we use `$(GO) run ./cmd/cli/plugin-admin/builder` which will compile and run the `builder` plugin on the spot. Note that the refactoring of the `builder` changes that a little but I can adapt my PR on top of yours: #3349
 
1. I'm a little worried that it is not clear that the user needs to set PLUGIN_NAME.  Without it nothing gets built but it give the impression that it may have built something.  Maybe we need an explicit logic check to handle this case?  I'm not sure.

pkg/v1/builder/template/plugintemplates/Makefile.tmpl
```diff
@@ -11,6 +11,8 @@ GOARCH ?= $(shell go env GOARCH)
 GOHOSTOS ?= $(shell go env GOHOSTOS)
 GOHOSTARCH ?= $(shell go env GOHOSTARCH)
 
+PLUGIN_PATH ?= ./cmd/cli/plugin
```
marckhouzam, line 14: I think this should be `./cmd/plugin` for the generated plugins

pkg/v1/builder/template/plugintemplates/Makefile.tmpl
```diff
@@ -39,7 +41,11 @@ go.mod go.sum: $(GO_SRCS)
 
 .PHONY: build-local
 build-local: ## Build the plugin
-	tanzu builder cli compile --version $(BUILD_VERSION) --ldflags "$(LD_FLAGS)" --path ./cmd/plugin --target local --artifacts artifacts/${GOHOSTOS}/${GOHOSTARCH}/cli
+	tanzu builder cli compile --version $(BUILD_VERSION) --ldflags "$(LD_FLAGS)" --path $(PLUGIN_PATH) --target local --artifacts artifacts/${GOHOSTOS}/${GOHOSTARCH}/cli
+
+.PHONY: build-plugin-local
+build-plugin-local: ## Build plugin's locally for given plugin name (match-expression) and path
+	tanzu builder cli compile --match "$(PLUGIN_NAME)" --version $(BUILD_VERSION) --ldflags "$(LD_FLAGS)" --path $(PLUGIN_PATH) --target local --artifacts artifacts/${GOHOSTOS}/${GOHOSTARCH}/cli
```
marckhouzam, line 48: Why do we want this target for the generated plugins?  It looks the same as `build-local` except for the `--match` flag.  However, I think there is usually only a single plugin in these generated repos, so using `--match` may not be useful.  But I'm just learning about this stuff so I can very well be wrong.

marckhouzam (changes requested):
A couple of things to clarify

pkg/v1/builder/template/plugintemplates/Makefile.tmpl
chandrareddyp: PLUGIN_PATH represents where it needs to search for plugin files.

Makefile
chandrareddyp: > I'm a little worried that it is not clear that the user needs to set PLUGIN_NAME. Without it nothing gets built but it give the impression that it may have built something. Maybe we need an explicit logic check to handle this case? I'm not sure.
I don't think there is a way to add help text for the make target, as it's being used by developers they must be aware of usage.

chandrareddyp: > I don't think we should use tanzu builder because the builder plugin may not be built, or up to date.
> Elsewhere we use $(GO) run ./cmd/cli/plugin-admin/builder which will compile and run the builder plugin on the spot. Note that the refactoring of the builder changes that a little but I can adapt my PR on top of yours: https://github.com/vmware-tanzu/tanzu-framework/pull/3349

@anujc25 any comments on this?

pkg/v1/builder/template/plugintemplates/Makefile.tmpl
chandrareddyp: After our discuss, i agree that it should be ./cmd/plugin

chandrareddyp: changes done.

pkg/v1/builder/template/plugintemplates/Makefile.tmpl
chandrareddyp: After discussing with Anuj, we decided not to add new make targets in Makefile.tmpl

Makefile
chandrareddyp: We have added the validation to validate the PLUGIN_NAME, it builds and install only given plugin, if plugin name is empty then it throws an error.

chandrareddyp:
> `build-plugin-local` will compile the plugin and put it in the `artifacts` directory within the repository.
> 
> We should add `install-plugin-local` that installs the plugin as well. Because as a end user I would like to do some change in the plugin, then I want to build as well as install the plugin to test my change.

Implementation has updated to add new make target `install-plugin-local` to build and update given plugin locally.

Makefile
```diff
@@ -240,6 +241,20 @@ build-plugin-admin-with-oci-discovery: ${CLI_ADMIN_JOBS_OCI_DISCOVERY} publish-a
 .PHONY: build-plugin-admin-with-local-discovery
 build-plugin-admin-with-local-discovery: ${CLI_ADMIN_JOBS_LOCAL_DISCOVERY} publish-admin-plugins-all-local ## Build Tanzu CLI admin plugins with Local standalone discovery
 
+.PHONY: build-plugin-local
+build-plugin-local: ## Build given CLI Plugin locally, needs PLUGIN_NAME
```
anujc25, line 245: This target needs to be linked with `prepare-builder`. I am seeing below error otherwise

```
$ make build-plugin-local PLUGIN_NAME=login
bash: /Users/anujc/Documents/code/tanzu-framework/bin/builder: No such file or directory
make: *** [build-plugin-local] Error 127
```

anujc25 (commented):
One comment. But otherwise looks good.

Makefile
chandrareddyp: for some reason, i don't see the error even in the new terminal, anyhow I have updated the target with prepare-builder

marckhouzam (approved):
LGTM

Thanks!

Makefile
marckhouzam: The builder used by the Makefile is in `./bin/builder`.  If you have built it anytime by another make target, it will stay there.  This is probably why Chandra didn't see the error.

anujc25 (approved):
Thanks for the change. LGTM

PR #3300, opened 2022-09-09
VSphereCPIConfig derives the insecure flags from the TLS thumbprint in cluster variable.

HanFa:

This PR changes  the data values reconcile logics in the VSphereCPIConfig controller. 

First, in the reconcile loop, the `tlsthumbprint` will be resolved in the following priority

* `spec.vsphereCPI.thumbprint` in the VSphereCPIConfig CR that user specified
* `thumbprint` derived from the cluster variables

Second, the controller then determine the value of `insecureFlag`, and make it possible to be overriden if user specifies in `spec.vsphereCPI.insecure`.

https://github.com/vmware-tanzu/tanzu-framework/issues/3381

This PR aims to fix cpi package reconcilation error `vsphereCPI tlsThumbprint should be provided when insecureFlag is False`

### Release note

```release-note
package-based-lcm: VSphereCPIConfig derives the insecure flags from the TLS thumbprint in cluster variable.
```

### DIFF
--- a/addons/controllers/cpi/vspherecpiconfig_utils.go
+++ b/addons/controllers/cpi/vspherecpiconfig_utils.go
@@ -120,8 +120,16 @@ func (r *VSphereCPIConfigReconciler) mapCPIConfigToDataValuesNonParavirtual( // 
 
 	d.Region = tryParseString(d.Region, c.Region)
 	d.Zone = tryParseString(d.Zone, c.Zone)
+
+	d.InsecureFlag = d.TLSThumbprint == ""
 	if c.Insecure != nil {
 		d.InsecureFlag = *c.Insecure
+	}
+	if !d.InsecureFlag && d.TLSThumbprint == "" {
+		return nil, errors.New("empty thumbprint is provided while TLS peer verification is enabled")
+	}
+	if d.InsecureFlag && d.TLSThumbprint != "" {
+		r.Log.Info("ignore provided thumbprint because TLS peer verification is disabled", "thumbprint", d.TLSThumbprint)
 	}
 
 	if c.VMNetwork != nil {
--- a/apis/addonconfigs/config/crd/bases/cpi.tanzu.vmware.com_vspherecpiconfigs.yaml
+++ b/apis/addonconfigs/config/crd/bases/cpi.tanzu.vmware.com_vspherecpiconfigs.yaml
@@ -57,7 +57,6 @@ spec:
                     description: The datacenter in which VMs are created/located
                     type: string
                   insecure:
-                    default: false
                     description: The flag that disables TLS peer verification
                     type: boolean
                   ipFamily:
--- a/apis/addonconfigs/cpi/v1alpha1/vspherecpiconfig_types.go
+++ b/apis/addonconfigs/cpi/v1alpha1/vspherecpiconfig_types.go
@@ -128,7 +128,6 @@ type NonParavirtualConfig struct {
 
 	// The flag that disables TLS peer verification
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default:=false
 	Insecure *bool `json:"insecure,omitempty"`
 
 	// The IP family configuration
--- a/packages/addons-manager/bundle/config/upstream/addonconfigscrds/cpi.tanzu.vmware.com_vspherecpiconfigs.yaml
+++ b/packages/addons-manager/bundle/config/upstream/addonconfigscrds/cpi.tanzu.vmware.com_vspherecpiconfigs.yaml
@@ -57,7 +57,6 @@ spec:
                     description: The datacenter in which VMs are created/located
                     type: string
                   insecure:
-                    default: false
                     description: The flag that disables TLS peer verification
                     type: boolean
                   ipFamily:


lubronzhan (approved):

apis/addonconfigs/config/crd/bases/cpi.tanzu.vmware.com_vspherecpiconfigs.yaml
```diff
@@ -57,7 +57,6 @@ spec:
                     description: The datacenter in which VMs are created/located
                     type: string
                   insecure:
-                    default: false
```
adduarte, line 60: Why remove the default? 
won't the added code set this to true if tlsthumbprint is provided.

apis/addonconfigs/cpi/v1alpha1/vspherecpiconfig_types.go
```diff
@@ -128,7 +128,6 @@ type NonParavirtualConfig struct {
 
 	// The flag that disables TLS peer verification
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default:=false
```
adduarte, line 131: same here

packages/addons-manager/bundle/config/upstream/addonconfigscrds/cpi.tanzu.vmware.com_vspherecpiconfigs.yaml
```diff
@@ -57,7 +57,6 @@ spec:
                     description: The datacenter in which VMs are created/located
                     type: string
                   insecure:
-                    default: false
```
adduarte, line 60: and here

adduarte (commented):
Please add link to isue number this fixes with description of original problem. 
Also not sure what will happen if 

spec.vsphereCPI.thumbprint  in the VSphereCPIConfig CR that user specified ="" (not provided)
&
thumbprint derived from the cluster variables = "" (not provided)
&
and  user did not specify  spec.vsphereCPI.insecure.

HanFa:
> Also not sure what will happen if

In that case, the final values of thumbprint is resolved to be empty. Hence, the insecure flag will be set to true with `d.InsecureFlag = d.TLSThumbprint == ""`

addons/controllers/cpi/vspherecpiconfig_utils.go
```diff
@@ -120,9 +120,17 @@ func (r *VSphereCPIConfigReconciler) mapCPIConfigToDataValuesNonParavirtual( //
...
 	d.Zone = tryParseString(d.Zone, c.Zone)
+
+	d.InsecureFlag = d.TLSThumbprint == ""
 	if c.Insecure != nil {
 		d.InsecureFlag = *c.Insecure
 	}
+	if !d.InsecureFlag && d.TLSThumbprint == "" {
+		return nil, errors.New("empty thumbprint is provided while TLS peer verification is enabled")
+	}
+	if d.InsecureFlag && d.TLSThumbprint != "" {
```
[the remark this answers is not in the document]
HanFa: we do need to at least log it as warning. 

the value of `d.InsecureFlag` can be optionally specified and override by the Config user in `d.InsecureFlag = *c.Insecure`.

so it's possible that insecure flag is on while having an non-empty thumbprint.

apis/addonconfigs/cpi/v1alpha1/vspherecpiconfig_types.go
HanFa: those two other changes are auto-generated from here

addons/controllers/cpi/vspherecpiconfig_utils.go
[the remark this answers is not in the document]
adduarte: That's what I thought you might mean. 
but if you do this: 
```
f d.InsecureFlag {
         	r.Log.Info("ignore thumbprint while TLS peer verification is disabled")
	}
```
The use _will_ get the error if they specify tlsthumbprint and d.insecureFlag=True. 

i.e. if the user sets tlsthumbprint="somevalue" but d.insecureFlag=True, they will see the error above. 

But if you really want to specifically say that the user has provided a tlsthumbprint but it will be ignore because inseureFlag = true then change the wording of the error, because right now it does not read that way. 

small changes but perhaps make it: 
" ignoring provided tlsthumbprint because TLS peer verification is disabled"

apis/addonconfigs/config/crd/bases/cpi.tanzu.vmware.com_vspherecpiconfigs.yaml
adduarte: I see, so now thumbprint must be provided. as longs thumbprintTLS = "" is the default then it should be fine.

adduarte (changes requested):
Please make the error message a bit clearer, right now they don't quite match the logic. thanks.

adduarte:
Please provide testing details to verify this patches fixes problem.

adduarte:
Currently this patch will turn insecureFlag=True if thumbprintTLS is not available. 
This is probably not a behavior we want.  I could result in the connection being insecure without user knowing.. 
i.e. the absense of thumbprintTLS could be an oversight or bug.

addons/controllers/cpi/vspherecpiconfig_utils.go
```diff
@@ -120,9 +120,17 @@ func (r *VSphereCPIConfigReconciler) mapCPIConfigToDataValuesNonParavirtual( //
 
 	d.Region = tryParseString(d.Region, c.Region)
 	d.Zone = tryParseString(d.Zone, c.Zone)
+
```
lubronzhan, line 123: Hi fang, could you remove this file's change? Use this PR to only remove the default value. But we should still assume the default value of insecureFlag should be explicitly passed from Cluster object

addons/controllers/cpi/vspherecpiconfig_utils.go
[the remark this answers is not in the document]
HanFa: sure, changed the error message to 

```
		r.Log.Info("ignore provided thumbprint because TLS peer verification is disabled", "thumbprint", d.TLSThumbprint)

```

addons/controllers/cpi/vspherecpiconfig_utils.go
lubronzhan: It's expected that insecureFlag is deprected and will treat empty tlsThumbprint as insecure

adduarte (approved):

HanFa:
per @srm09, the insecure flag from CAPV will be deprecated. An empty string in thumbprint should imply to use insecure connection.

vuil (approved):
lgtm

PR #3319, opened 2022-09-13
Add legacy 2 cc storageclass config

m1zzx2:

Fixes #

### Release note

```release-note

```

### DIFF
--- a/packages/tkg-storageclass/bundle/config/upstream/base.yaml
+++ b/packages/tkg-storageclass/bundle/config/upstream/base.yaml
@@ -1,7 +1,20 @@
 #@ load("@ytt:overlay", "overlay")
 #@ load("@ytt:yaml", "yaml")
 #@ load("@ytt:data", "data")
+#@ load("@ytt:regexp", "regexp")
 
+#@ def compare_semver_versions(a, b):
+#@   a_array = regexp.replace("v?(\d+\.\d+\.\d+).*", a, "$1").split(".")
+#@   b_array = regexp.replace("v?(\d+\.\d+\.\d+).*", b, "$1").split(".")
+#@   for i in range(len(a_array)):
+#@     if int(a_array[i]) > int(b_array[i]):
+#@       return 1
+#@     elif int(a_array[i]) < int(b_array[i]):
+#@       return -1
+#@     end
+#@   end
+#@   return 0
+#@ end
 
 #@ if data.values.metadata.infraProvider == "aws":
 ---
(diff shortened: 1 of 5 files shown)


pkg/v1/providers/yttcb/clusterbootstrap.yaml
```diff
@@ -221,5 +274,8 @@ spec:
     - refName: metrics-server*
     - refName: secretgen-controller*
     - refName: pinniped*
+    - refName: tkg-storageclass*
+      valuesFrom:
```
wjun, line 278: @vijaykatam This additional package is only defined in a clusterbootstrap, but not clusterbotstraptemplate, as it will generate secret name and content according to a few data values. Is this a best practice?

pkg/v1/providers/yttcb/clusterbootstrap.yaml
```diff
@@ -171,6 +177,52 @@ spec:
...
+
+
+#@ def azure_storage_class():
+---
+metadata:
+  infraProvider: azure
+  enableCsiDriver: #@ enable_csi_driver()
+#@ end
+
+#@ if data.values.PROVIDER_TYPE != "tkg-service-vsphere" and data.values.PROVIDER_TYPE != "docker":
```
wjun, line 206: reuse storage_class_configs_exist()?

pkg/v1/providers/yttcc/lib/helpers.star
```diff
@@ -333,3 +333,24 @@ def get_labels_map_from_string(labelString):
...
+    elif int(a_array[i]) < int(b_array[i]):
+      return -1
+    end
+  end
+  return 0
+end
+
+def enable_csi_driver():
+  tkrVersion = get_tkr_version_from_tkr_name(data.values.KUBERNETES_RELEASE)
+  if compare_semver_versions(tkrVersion, "v1.23.0") >= 0:
```
wjun, line 352: Just wonder why we select v1.23.0 as the comparison to decide if enable or not enable csi driver?

jeffwubj: azuredisk & awsebs should be enabled >= 1.23; azurefile should be enabled >=1.24

pkg/v1/providers/yttcb/clusterbootstrap.yaml
vijaykatam: This is fine, but when users set static secrets we expect them to manage their lifecycle via explict ownerRef's because addons-manager does not know if that secret is in use elsewhere. 

Also consider using `inline` option if there is no sensitive data in the secret as this removes the need to set ownerRef's

m1zzx2:
//usebom:sandbox/v1.7.0-zshippable/427806996654235526/tkg-compatibility:v11

m1zzx2:
//usebom:sandbox/v1.7.0-zshippable/429535539864018376/tkg-compatibility:v12

m1zzx2:
//usebom:sandbox/v1.7.0-zshippable/427806996654235526/tkg-compatibility:v12

m1zzx2:
//usebom:sandbox/v1.7.0-zshippable/429535539864018376/tkg-compatibility:v12

jeffwubj (approved):

wjun (approved):
LGTM.

PR #3330, opened 2022-09-13
Implement NO_COLOR environment variable to enable/disable Aurora package Color property

### LINKED ISSUE #43: Disable color and symbols when TTY not present
Add a TTY checker for the CLI to disable the use of color, symbols, animation, and emojis when TTY not present or when env var TERM=dumb is set.

Add a checker for the CLI to disable the use of color when the NO_COLOR env var is set to true.

Making machine-reading easier by simplifying the outputs when we can detect non-TTY action or when explicitly requested by the user. 

[ ] APIs
[ ] Addons
[x] CLI
[ ] Docs
[ ] Installation
[ ] Plugin
[ ] Security
[ ] Test and Release
[ ] User Experience

**Additional context**

mpanchajanya:

Fixes #43 

### Release note

```release-note

```

### DIFF
--- a/pkg/v1/cli/command/core/root.go
+++ b/pkg/v1/cli/command/core/root.go
@@ -47,7 +47,7 @@ func NewRootCmd() (*cobra.Command, error) {
 	if ni != "" || strings.EqualFold(forceNoInit, "true") {
 		noInit = true
 	}
-	if os.Getenv("TANZU_CLI_NO_COLOR") != "" {
+	if os.Getenv("TANZU_CLI_NO_COLOR") != "" || os.Getenv("NO_COLOR") != "" {
 		color = false
 	}
 


marckhouzam:
First, let me ask if this PR is supposed to completely address #43, or is only one step towards that goal?  See below why I ask.

I don't know how to test this on a Mac so this is hard to figure out.
Looking at #352 we can notice that when running `tanzu --help`, the bold text does cause garbage chars to show up.
So we can start by looking at bold text to evaluate this PR.

What I noticed is that when we compare the output of `tanzu --help` and `TANZU_CLI_NO_COLOR=1 tanzu --help` only the first line goes from bold to not bold.  Of course the same happens with the new handling of `NO_COLOR`.

I think if we want to fix #43 completely, we need to find a way to disable all special characters all the time.

mpanchajanya:
> First, let me ask if this PR is supposed to completely address #43, or is only one step towards that goal? See below why I ask.
> 
> I don't know how to test this on a Mac so this is hard to figure out. Looking at #352 we can notice that when running `tanzu --help`, the bold text does cause garbage chars to show up. So we can start by looking at bold text to evaluate this PR.
> 
> What I noticed is that when we compare the output of `tanzu --help` and `TANZU_CLI_NO_COLOR=1 tanzu --help` only the first line goes from bold to not bold. Of course the same happens with the new handling of `NO_COLOR`.
> 
> I think if we want to fix #43 completely, we need to find a way to disable all special characters all the time.

As per discussions with the team. we will handle the color in this PR and We will be creating another issue for rest of the TTY related issues.

cli/core/pkg/command/root.go
```diff
@@ -47,7 +48,7 @@ func NewRootCmd() (*cobra.Command, error) {
 	if ni != "" || strings.EqualFold(forceNoInit, "true") {
 		noInit = true
 	}
-	if os.Getenv("TANZU_CLI_NO_COLOR") != "" {
+	if component.IsTTYEnabled() {
 		color = false
 	}
```
marckhouzam, line 53: How about we remove this code and adjust the below to use the new cli.Bold(..) you just adapted?
We'd have to make those functions public though.

cli/runtime/plugin/usage.go
```diff
@@ -23,6 +24,12 @@ var UsageFunc = func(c *cobra.Command) error {
 	return t.Execute(os.Stdout, c)
 }
 
+var au aurora.Aurora
+
+func init() {
+	au = aurora.NewAurora(!component.IsTTYEnabled())
+}
```
marckhouzam, line 31: To avoid having this code duplicated, I wonder if we could move it to `component` along with the 5 functions at the bottom, make the functions public and use them in `cli/core/pkg/cli/usage.go`?

cli/runtime/component/tty.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"github.com/mattn/go-isatty"
+	"os"
+	"strings"
+)
+
+func IsTTYEnabled() bool {
+	ttyEnabled := false
+	if os.Getenv("TANZU_CLI_NO_COLOR") != "" || os.Getenv("NO_COLOR") != "" || strings.EqualFold(os.Getenv("TERM"), "DUMB") || !isatty.IsTerminal(os.Stdout.Fd()) {
+		ttyEnabled = true
```
marckhouzam, line 12: I'm not well versed in the concept of a `TTY`, but I think the enable/disable is reversed in this function.
To print colours I believe you need a tty.  So when we don't print colours, we probably want to return `false`.

Maybe we can clarify things by changing the name of the function and file?  I don't have a good suggestion though.  Maybe another reviewer can chime in?

marckhouzam (commented):
Thanks for the wider change, it works nicely.
I just have some implementation suggestions.

cli/core/pkg/cli/usage.go
```diff
@@ -5,15 +5,12 @@ package cli
 
 import (
 	"fmt"
+	"github.com/spf13/cobra"
```
marckhouzam, line 8: So, for imports, go packages need to be grouped together alphabetically at the top of the import list, and github packages should follow under them.

You can run `make fmt` and it will detect and fix these automatically.

marckhouzam (commented):
This looks great.  Thanks @mpanchajanya !
Have you been able to test it on Windows successfully?

mpanchajanya:
> This looks great. Thanks @mpanchajanya ! Have you been able to test it on Windows successfully?

Yes, I have attached the screenshot in the description

PR #3328, opened 2022-09-13
Refactor test plugin

marckhouzam:
### What this PR does / why we need it

Refactor the `test` admin plugin into its own go module.

Fixes #3346.

PR #3284, opened 2022-09-08
Refactor codegen plugin

yharish991:
### What this PR does / why we need it
This PR refactors the codegen plugin and makes it a go module

### Which issue(s) this PR fixes

Fixes https://github.com/vmware-tanzu/tanzu-framework/issues/3298

### Describe testing done for PR

Ran the tests and codegen plugin separately to verify the functionality

### Release note

```release-note
Refactored codegen plugin and made it a go module
```

PR #3261, opened 2022-09-02
Provide automatic buildinfo for plugins

marckhouzam:
### What this PR does / why we need it

This is kind of a follow-up of the refactoring of #3201.
The PR simply copies `pkg/v1/buildinfo/buildvar.go` to a new `cli/runtime/buildinfo` package.
This will allow plugins to get access to `buildinfo.Version`, `buildinfo.SHA` and `buildinfo.Date` without having to import Framework.

The PR also updates the `Makefile.tmpl` template to inject those values and the `main.go.tmpl` to use this mechanism for new plugins.

### Which issue(s) this PR fixes

Fixes #3230

### Describe testing done for PR

Created a new plugin and built it and verified that Version and SHA were injected automatically:
```
tanzu builder init myplugin --repo-type github
cd myplugin
tanzu builder cli add-plugin myplugin --description myplugin
git add .
git commit -m start
make init || true

# point the go.mod to my local Framework repo
echo "replace github.com/vmware-tanzu/tanzu-framework/cli/runtime => /Users/$USER/git/tanzu-framework/cli/runtime" >> go.mod
go mod tidy -compat=1.17
make build-local

# Call the built binary which depends on your machine arch and os
artifacts/darwin/arm64/cli/myplugin/v0.0.1/tanzu-myplugin-darwin_arm64 info

# Check that the Version and SHA are set properly
```
### Additional information

We would need some doc to teach this to plugin developers.  I can add it once #3259 is merged.

PR #3345, opened 2022-09-14
Fill plugin buildinfo variables

### LINKED ISSUE #3344: Makefile should set the new plugin buildinfo variables
**Bug description**

#3261 added `buildinfo` variable to the new `cli/runtime` library that is meant to be used by plugins.
Since Framework's plugins are all built through the `Makefile`, those new `buildinfo` variables should be set by the `Makefile` so plugins can start using them transparently.

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- (x) Plugin
- ( ) Security
- (x) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

When plugins start using `github.com/vmware-tanzu/tanzu-framework/cli/runtime/buildinfo` they should have proper values set when being built though the `Makefile`.

Start using `github.com/vmware-tanzu/tanzu-framework/cli/runtime/buildinfo` in any plugin and then build the plugin using the `Makefile`.  The build will fail because the version of the plugin will not be set.

Today's `main`: 645af22f

Locally

marckhouzam:

#3261 added `buildinfo` variable to the new `cli/runtime` library that is meant to be used by plugins.
Since Framework's plugins are all built through the `Makefile`, those new `buildinfo` variables should be set by the `Makefile` so plugins can start using them transparently.

This PR sets those variables.

Fixes #3344

Before the PR:
1. Update the `test` plugin to use the new `github.com/vmware-tanzu/tanzu-framework/cli/runtime/buildinfo` (see #3328)
1. Build using `make build-plugin-admin-with-local-discovery` => this will fail with `output: ✖  plugin "test" version cannot be empty; version "test" "" is not a valid semantic version`
1. Apply this PR
2. `make build-plugin-admin-with-local-discovery` => this will now pass

### Additional information

As discussed directly with @yharish991, this change is pulled out of #3284 as it is needed by all plugins.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -50,6 +50,11 @@ LD_FLAGS += -X 'github.com/vmware-tanzu/tanzu-framework/pkg/v1/buildinfo.Date=$(
 LD_FLAGS += -X 'github.com/vmware-tanzu/tanzu-framework/pkg/v1/buildinfo.SHA=$(BUILD_SHA)'
 LD_FLAGS += -X 'github.com/vmware-tanzu/tanzu-framework/pkg/v1/buildinfo.Version=$(BUILD_VERSION)'
 
+# Set buildinfo for plugins
+LD_FLAGS += -X 'github.com/vmware-tanzu/tanzu-framework/cli/runtime/buildinfo.Date=$(BUILD_DATE)'
+LD_FLAGS += -X 'github.com/vmware-tanzu/tanzu-framework/cli/runtime/buildinfo.SHA=$(BUILD_SHA)'
+LD_FLAGS += -X 'github.com/vmware-tanzu/tanzu-framework/cli/runtime/buildinfo.Version=$(BUILD_VERSION)'
+
 # Add supported OS-ARCHITECTURE combinations here
 ENVS ?= linux-amd64 windows-amd64 darwin-amd64
 STANDALONE_PLUGINS ?= login management-cluster package pinniped-auth secret telemetry


yharish991 (approved):

anujc25 (approved):

common.mk
```diff
@@ -50,6 +50,11 @@ LD_FLAGS += -X 'github.com/vmware-tanzu/tanzu-framework/pkg/v1/buildinfo.Date=$(
 LD_FLAGS += -X 'github.com/vmware-tanzu/tanzu-framework/pkg/v1/buildinfo.SHA=$(BUILD_SHA)'
 LD_FLAGS += -X 'github.com/vmware-tanzu/tanzu-framework/pkg/v1/buildinfo.Version=$(BUILD_VERSION)'
```
maralavi, line 51: Do we need a TODO comment to remove these once got obsolete?

mpanchajanya (approved):

common.mk
marckhouzam: Good question.  I'm not sure if these will be removed or changed.  The new `buildinfo` are just for plugins, so I expect the existing `buildinfo` will be moved to some other location.

I'm expecting/hoping that the refactoring of the `pkg/v1/buildinfo` will know to correct the lines above as they should still be needing them.

Does that makes sense, or did I misunderstand the next steps?

maralavi: Yes, sounds good:)

maralavi (approved):

PR #3349, opened 2022-09-14
Refactor builder plugin

### LINKED ISSUE #3347: Create "builder" plugin as a separate module

Currently the "builder" plugin is part of the main tanzu-framework module. We would like to decouple the two and create a new module for the plugin while keeping the plugin code in the same location.

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- (x) Plugin
- ( ) Security
- (x) Test and Release
- ( ) User Experience
- ( ) Developer Experience

marckhouzam:

This PR refactors the `builder` plugin into its own go module.

With this change, it becomes impossible to use `go run` to use the `builder` plugin to build other parts of Framework.  This is because the `go` command cannot access two different modules at the same time (go workspaces could fix that but they require go 1.18, which we do not want to force).

Therefore, this PR prebuilds a copy of the `builder` binary using `go build` and then uses it to build all plugins.
This change can be seen in the `Makefile`.

Fixes #3347

I ran `make build-install-cli-local` and saw that all plugins werebuilt properly.

The new `Makefile` pre-builds the `builder` binary under `/tmp/builder` and then uses it.  I don't like the location too much.  Any suggestions on a better approach?


go.mod
```diff
@@ -8,6 +8,7 @@ replace (
 	github.com/vmware-tanzu/tanzu-framework/apis/run => ./apis/run
 	github.com/vmware-tanzu/tanzu-framework/cli/core => ./cli/core
 	github.com/vmware-tanzu/tanzu-framework/cli/runtime => ./cli/runtime
+	github.com/vmware-tanzu/tanzu-framework/cmd/cli/plugin-admin/builder => ./cmd/cli/plugin-admin/builder
```
anujc25, line 11: Is this really needed? I believe no module will be dependent on `builder` plugin, so this might not be needed.

anujc25 (commented):
Looks good overall. One minor comment regarding `go.mod file.

prkalle (commented):
LGTM.  Thank you @marckhouzam !
I have one  comment. We have dependency on core cli because builder plugin supports compiling the core cli. We need to think about this(may be later), as we are using the builder plugin to build core cli ( we are building cli using `go build`).

go.mod
marckhouzam: Good catch!  This was left-over from my trying to get things to work early on.
Will remove.

marckhouzam:
> LGTM. Thank you @marckhouzam ! I have one comment. We have dependency on core cli because builder plugin supports compiling the core cli. We need to think about this(may be later), as we are not using the builder plugin anymore to build core cli ( we made changes to build cli using `go build` as part of refactoring effort).

That is a good point.  If we can agree that the builder plugin will not build the CLI anymore, then we could cleanup the builder plugin and probably remove that dependancy.

I will open an issue to track that.  Thanks @prkalle.

marckhouzam:
> I will open an issue to track that.

I opened #3370

anujc25 (approved):
LGTM Thanks!

marckhouzam:
@anujc25 Do you think it is ok to use `/tmp/builder` as the location for the temporary builder plugin?

anujc25:
> @anujc25 Do you think it is ok to use `/tmp/builder` as the location for the temporary builder plugin?

Would be nice to put it under `./bin` directory within the PROJECT_ROOT.

marckhouzam:
> > @anujc25 Do you think it is ok to use `/tmp/builder` as the location for the temporary builder plugin?
> 
> Would be nice to put it under `./bin/builder` within the PROJECT_ROOT.

Done.  I like that better too.  Thanks!

anujc25 (approved):

PR #3354, opened 2022-09-14
Run cli/runtime unit tests as part of `make test` target

anujc25:
- Run cli/runtime unit tests as part of `make test` target

Fixes #

- Manually verified by running `make test`

### Release note

```release-note
None
```

### DIFF
--- a/cli/runtime/config/context_test.go
+++ b/cli/runtime/config/context_test.go
@@ -329,8 +329,9 @@ func TestSetCurrentContext(t *testing.T) {
 			}
 			currSrv, err := GetCurrentServer()
 			assert.NoError(t, err)
-			// server is updated only for K8s Ctx Type, so test-mc remains the current server
-			assert.Equal(t, "test-mc", currSrv.Name)
+			if err != nil {
+				assert.Equal(t, tc.ctxName, currSrv.Name)
+			}
 		})
 	}
 }
--- a/Makefile
+++ b/Makefile
@@ -472,6 +472,9 @@ test: generate manifests build-cli-mocks ## Run tests
 	# pinniped tanzu-auth-controller-manager
 	pinniped-components/tanzu-auth-controller-manager/hack/test.sh
 
+	#Test core cli runtime library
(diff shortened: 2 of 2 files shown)


cli/runtime/config/context_test.go
```diff
@@ -329,8 +329,9 @@ func TestSetCurrentContext(t *testing.T) {
 			}
 			currSrv, err := GetCurrentServer()
 			assert.NoError(t, err)
-			// server is updated only for K8s Ctx Type, so test-mc remains the current server
-			assert.Equal(t, "test-mc", currSrv.Name)
+			if err != nil {
```
marckhouzam, line 332: Nit: We have a `assert.NoError(t, err)` at line 331, do we still need to check for `err != nil`?

marckhouzam (approved):
Assuming the build passed, this LGTM

cli/runtime/config/context_test.go
anujc25: Yes. Because `assert.Equal(t, tc.ctxName, currSrv.Name)` test should only be executed when there is no error.
Without this, it throws error because server is not updated.

marckhouzam: Aha. I had assumed of the assert failed it would stop the test.  Thanks for the explanation.

vuil (approved):
lgtm, thanks!

cli/runtime/config/context_test.go
marckhouzam: @anujc25 I just realized we should probably have `if err == nil` in this case instead of the opposite, no?  But looking at the goDoc for `assert.NoError()` we could even do:
```
currSrv, err := GetCurrentServer()
if assert.NoError(t, err) {
	assert.Equal(t, tc.ctxName, currSrv.Name)
}
```

anujc25: @marckhouzam I have PR out to address this: https://github.com/vmware-tanzu/tanzu-framework/pull/3432

PR #3356, opened 2022-09-14
Fix 'go vet' related issues

### LINKED ISSUE #3355: Fix `go vet` issues
**Bug description**
This is to address below issues reported by `go vet `
- cmd/cli/plugin/login/main.go:367:4: unreachable code
- pkg/v1/tkg/test/tkgctl/shared/tkrresolver_cc.go:335:26: Sprintf format %s has arg mdIndex of wrong type int

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

chandrareddyp:
Fixes `go vet` related issues

Fixes #3355

After the fix ran, `go vet` successfully without any issues
### Release note

```release-note

```

### DIFF
--- a/cmd/cli/plugin/login/main.go
+++ b/cmd/cli/plugin/login/main.go
@@ -361,10 +361,6 @@ func createServerWithEndpoint() (server *configv1alpha1.Server, err error) {
 		if err != nil {
 			log.Fatalf("Error creating kubeconfig with tanzu pinniped-auth login plugin: %v", err)
 			return nil, err
-			// IsSupervisorClusterIgnorePort only returns an error if the
-			// underlying HTTP GET returned one. Non-200 responses will return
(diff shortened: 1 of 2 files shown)


marckhouzam (approved):
LGTM

marckhouzam:
I was wondering if we should add running `go vet` to the CI jobs.  I found #2673 which removed `go vet` and `go fmt` saying `golangci-lint` was doing it anyway.  We would have to see if we have an issue with `golangci-lint`

anujc25 (approved):
Looks good. Thanks for the fix.

PR #3372, opened 2022-09-15
Remove go mod download and tidy from builder

### LINKED ISSUE #3371: Inconsistent use of "go mod tidy"
**Bug description**

Framework has recently started using `go mod tidy --compat-1.17` in our `Makefile` (e.g. `make verify`).
However, the builder plugin does its own `go mod tidy` when building a plugin that has its own `go.mod` file.  
This causes an inconsistency when building:
1. for CI to pass, `make verify` must pass, so we need to use `go mod tidy --compat-1.17` to generate the `go.sum` files
2. but after that, when running `make build-install-cli-local` for example, the `builder` plugin will run `go mod tidy` and change the `go.sum` file of the plugins

If a PR is submitted with the `go.sum` of point 1, the CI will pass, but if using point 2, the CI will fail.
This is likely to cause delays and confusion.

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- (x) Plugin
- ( ) Security
- (x) Test and Release
- ( ) User Experience
- (x) Developer Experience

**Expected behavior**

Running `tanzu builder cli compile` does not change `go.sum` files that were created using the `Makefile`.

```
cd tanzu-framework
make verify
git commit -a -m "updated go.sum files"
make build-install-cli-local
git status
# go.sum files will be dirty for plugins that have their own go.mod file
```

Today's `main`: bf0fc7bf

Locally

marckhouzam:

Removes `go mod tidy` and `go mod download` from the `tanzu builder cli compile` command.

Compiling a plugin should not modify the go.mod/go.sum files. It is up to the developer or if they choose, their Makefile to do any necessary cleanup.

Fixes #3371

```
cd tanzu-framework
make verify
git commit -a -m "updated go.sum files"
make build-install-cli-local
git status
# No go.mod or go.sum file should have been changed
```

Two `go.sum` files are changed because they were not up-to-date on main.


anujc25:
Created another [PR](https://github.com/vmware-tanzu/tanzu-framework/pull/3374) to fix the build issue. Closing this for now.

PR #3380, opened 2022-09-16
Add --all-namespaces to tanzu cluster list command

zhouke1991:

The namespace from which to list workload clusters. If not provided clusters from the default namespace will be returned.
if allNamespaces parameter is present, the namespace in the current context is ignored even if specified with --namespace

### DIFF
--- a/cmd/cli/plugin/cluster/list.go
+++ b/cmd/cli/plugin/cluster/list.go
@@ -17,9 +17,10 @@ import (
 )
 
 type listClusterOptions struct {
-	namespace    string
-	includeMC    bool
-	outputFormat string
+	namespace     string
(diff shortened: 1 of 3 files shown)


cmd/cli/plugin/cluster/list.go
```diff
@@ -17,9 +17,10 @@ import (
 )
 
 type listClusterOptions struct {
-	namespace    string
-	includeMC    bool
-	outputFormat string
+	namespace     string
+	includeMC     bool
+	outputFormat  string
+	AllNamespaces bool
```
wjun, line 23: not use a capital letter?

wjun (approved):
LGTM.

PR #3391, opened 2022-09-16
Use CAPV util function to retrieve VSphere cluster credentials.

HanFa:

To retrieve the credential from a VSphereCluster, instead of just getting it from the associated secret object, we use CAPV util function to cover the case of multi-tenancy. 

Fixes #

https://github.com/vmware-tanzu/tanzu-framework/issues/3392

Addon Controller integration test with `KUBEBUILDER_ASSETS=/Users/hanfa/workspace/tanzu-framework-HanFa/hack/tools/bin/kubebuilder/bin go test ./controllers -v`

Manually testing for non-paravirt uTKG

### Release note

```release-note
package-based-lcm: Use CAPV util function to retrieve VSphere cluster credentials.
```


lubronzhan (approved):
lgtm

randomvariable:
Split the testutil changes into a separate commit, otherwise lgtm

addons/controllers/vspherecpiconfig_controller_test.go
```diff
@@ -268,6 +271,49 @@ var _ = Describe("VSphereCPIConfig Reconciler", func() {
...
+			Eventually(func() bool {
+				secretKey := client.ObjectKey{
+					Namespace: clusterNamespace,
+					Name:      fmt.Sprintf("%s-%s-data-values", clusterName, constants.CPIAddonName),
+				}
+				if err := k8sClient.Get(ctx, secretKey, secret); err != nil {
+					return false
+				}
+				secretData := string(secret.Data["values.yaml"])
+				fmt.Println(secretData) // debug dump
```
lubronzhan, line 307: Remove this

addons/controllers/vspherecpiconfig_controller_test.go
```diff
@@ -268,6 +271,49 @@ var _ = Describe("VSphereCPIConfig Reconciler", func() {
...
+				Expect(len(secretData)).Should(Not(BeZero()))
```
lubronzhan, line 308: nit: shouldnot

lubronzhan (approved):

addons/controllers/vspherecpiconfig_controller_test.go
HanFa: fixed

addons/controllers/vspherecpiconfig_controller_test.go
```diff
@@ -250,6 +253,48 @@ var _ = Describe("VSphereCPIConfig Reconciler", func() {
...
+					Namespace: clusterNamespace,
+					Name:      fmt.Sprintf("%s-%s-data-values", clusterName, constants.CPIAddonName),
+				}
+				if err := k8sClient.Get(ctx, secretKey, secret); err != nil {
+					return false
+				}
+				secretData := string(secret.Data["values.yaml"])
+				Expect(len(secretData)).ShouldNot(BeZero())
+				Expect(strings.Contains(secretData, "username: administrator@vsphere.local")).Should(BeTrue())
+				Expect(strings.Contains(secretData, "password: Admin!23")).Should(BeTrue())
```
maralavi, line 291: Maybe better not to commit these credentials here?

HanFa: thanks, just fixed

maralavi:
@HanFa Will a resource of kind VSphereClusterIdentity or an identityRef on a VsphereCluster exist for sure in a cluster?  How credentials would be fetched if they don’t?

HanFa:
> Will a resource of kind VSphereClusterIdentity or an identityRef on a VsphereCluster exist for sure in a cluster? How credentials would be fetched if they don’t?

Hi @maralavi , not necessarily. If the  VSphereClusterIdentity doesn't exist, the secretRef type should be `Secret` instead, the controller shall grab the credentials from the ref-ed secret object.

addons/controllers/cpi/vspherecpiconfig_utils.go
```diff
@@ -95,15 +97,13 @@ func (r *VSphereCPIConfigReconciler) mapCPIConfigToDataValuesNonParavirtual( //
 	d.TLSThumbprint = vsphereCluster.Spec.Thumbprint
 	d.Server = vsphereCluster.Spec.Server
 
-	// derive vSphere username and password from the <cluster name> secret
-	clusterSecret, err := r.getSecret(ctx, cluster.Namespace, cluster.Name)
+	// derive vSphere username and password using CAPV util function
+	clusterCredentials, err := capvidentity.GetCredentials(ctx, r.Client, vsphereCluster, capvmanager.DefaultPodNamespace)
```
maralavi, line 101: Can we expect that this ```DefaultPodNamespace``` always get populated correctly?

HanFa: no, the DefaultPodNamespace is always `capv-system`, we can't gurantee its existence. if it doesn't exist, the `GetCredentials` will just use the cluster's namespace

HanFa:
> > > Will a resource of kind VSphereClusterIdentity or an identityRef on a VsphereCluster exist for sure in a cluster? How credentials would be fetched if they don’t?
> > 
> > 
> > Hi @maralavi , not necessarily. If the VSphereClusterIdentity doesn't exist, the secretRef type should be `Secret` instead, the controller shall grab the credentials from the ref-ed secret object.
> 
> Thanks for the clarification @HanFa . Can we expect that `capvidentity.GetCredentials()` take care of that case too? (I ask as now `getSecret()` is removed?)

yes, that's exactly the case

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/9492c4bb01f03a0cec534311ed1d61515e94eddf/pkg/identity/identity.go#L53

maralavi (approved):

PR #3476, opened 2022-09-22
Add automation tests for admission webhook

### LINKED ISSUE #2486: TC-34: Admission Webhook test to verify CA cert gets propagated to validating/mutating webhooks matching addons label selector
Verify the following listed tasks:

1. Validate webhook server picks up generated certificate.  **(This task is meant for automated test)**
2. Validate certificate is rotated according to desired lifespan (should be updated automatically on each addon package webhook definition) **(This task is meant for manual test)**
3. Validate certificate can be accessed by user through kubectl  **(This task is meant for manual test)**
4. Delete certificate by hand, and verify it is regenerated.  **(This task is meant for manual test)**
5. User uploades their own certificate.  **(This task is meant for manual test)**
6. Upload invalid certificate and verify it is regenerated and updated in the webhook definitions  **(This task is meant for manual test)**
7. Validaton -  Create config for each addonpackage and verify they are validated by webhook  **(This task is meant for manual test)**

maralavi:

- Adds test to verify addon’s certificate manager has added the CA certificate from ```webhook-tls``` secret to the CA bundle of all validating/mutating Webhook configurations which match label ```"tkg.tanzu.vmware.com/addon-webhooks": "addon-webhooks"```
- Also did cleanup to use the same constant everywhere rather than defining in multiple files

Fixes #2486 

Tested in as AWS cluster

### Release note

```release-note
Admission Webhook test to verify CA cert gets propagated to validating/mutating webhooks matching addons label selector
```


tkg/test/tkgctl/shared/utils.go
```diff
@@ -0,0 +1,71 @@
...
+// getWebhookCACert returns the value of CA certificate "ca-cert.pem" for the "webhook-tls" secret
+func getWebhookCACert(ctx context.Context, k8sClient client.Client, secretName, secretNamespace string) (string, error) {
+	secret := corev1.Secret{}
+	if err := k8sClient.Get(ctx, types.NamespacedName{Name: secretName, Namespace: secretNamespace}, &secret); err != nil {
+		return "", err
+	}
+	if value, exists := secret.Data["ca-cert.pem"]; exists {
+		return string(value), nil
+	}
+	return "", nil
```
HL-EverGreen, line 29: Is the scenario where the `ca-cert.pem` entry does not exist in the secret expected?

haoranleo (approved):
Looks good! Just a few questions about scenario of empty CA name

tkg/test/tkgctl/shared/utils.go
maralavi: Oh no:-/ I should either return an error here, or do ```Expect(caCert).NotTo(BeEmpty())``` on the caller. Thanks for catching it

tkg/test/tkgctl/shared/utils.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	for idx := range validatingWhConfigs.Items {
+		configName := validatingWhConfigs.Items[idx].Name
+		validatingWhConfig, err := admissionRegistrationClient.ValidatingWebhookConfigurations().Get(ctx, configName, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+
+		for idx := range validatingWhConfig.Webhooks {
+			if string(validatingWhConfig.Webhooks[idx].ClientConfig.CABundle) != caCert {
```
HL-EverGreen, line 46: Same as the question above - empty caCert comparison scenario. Also same for line 64

HL-EverGreen: Seems we avoid the empty string here :D, mark as resolved.

maralavi: I changed to check no emptiness before calling this function

PR #3500, opened 2022-09-27
Upload new coverage files from refactoring for codecov

codegold79:

Since refactoring, there are new tests and coverage files being generated. However, they are not being uploaded to codecov. This PR uploads those coverage files.

We need it so we can see code coverage in components that have unit tests.

Fixes #

### Release note

```release-note

```


rajathagasthya (approved):

rajathagasthya:
Why is the `codecov/project` job failing though?

codegold79:
> Why is the `codecov/project` job failing though?

It's because there's a rule where even a 1% decrease in coverage change will fail the pipeline, and we have a -8.14% decrease in test coverage.

The rule is `threshold: 1% # allow for 1% reduction without failing`: https://github.com/vmware-tanzu/tanzu-framework/blob/main/.codecov.yaml#L11

I'll try to bypass this check because I'm not really decreasing test coverage with this change; it merely shows the actual test coverage.

PR #3507, opened 2022-09-29
Fix to create CC cluster on vSphere

ydp:

Latest code in main branch cannot create CC cluster successfully, there are 2 issues identified:

* AntreaConfig is set in ClusterBootstrap when no customization is provided; in this case, the name in CB is `<cluster-name>` format, but the actual config is `<cluster-name>-antrea-config` format, which causes error to apply the CB. 
* template in vSphereMachine is empty due to there is no vSphere template resolver deployed in the cluster, because #2748 is partially done. 

Fixes #

* Make the cni part only exist when antrea config exists
* Thanks to @tenczar 's suggestion, no code change needed for the second empty template issue, it can be workaround by exporting environment 
```
_ADDITIONAL_MANAGEMENT_COMPONENT_CONFIGURATION_FILE=https://gist.githubusercontent.com/tenczar/b214012420c3e20582f2f61f0611bdf1/raw/cb46bacbca6b7aee61011c7f750cbb0ea6006e9f/tkr-hack.yml
```
> It defines a custom TKR with a template value of /dc0/photon-3-kube-v1.23.8+vmware.2-tkg.2-81d1a7892ad39f017fbaf59f9907cbe7. The benefit of this approach is that we won't need to revert it and people can create custom TKRs to use whichever template version they want.

Note that: you also need this variable when deleting cluster, otherwise it reports "spec is immutable" error when moving resources from mgmt cluster to cleanup cluster.
### Release note

```release-note

```

I have validated the creation/deletion process locally.

### DIFF
--- a/pkg/v1/providers/yttcb/clusterbootstrap.yaml
+++ b/pkg/v1/providers/yttcb/clusterbootstrap.yaml
@@ -181,7 +181,7 @@ metadata:
   annotations:
     tkg.tanzu.vmware.com/add-missing-fields-from-tkr: #@ data.values.KUBERNETES_RELEASE
 spec:
-  #@ if data.values.CNI == "antrea":
+  #@ if data.values.CNI == "antrea" and antrea_configs_exist():
   cni:
     refName: antrea*
     valuesFrom:


packages/tkg-clusterclass-vsphere/bundle/config/upstream/base.yaml
```diff
@@ -598,7 +598,7 @@ spec:
       - op: replace
         path: /spec/template/spec/template
         valueFrom:
-          template: '{{ (index .TKR_DATA .builtin.machineDeployment.version).osImageRef.template }}'
+          template: '/dc0/vm/photon-3-kube- {{- (index .TKR_DATA .builtin.machineDeployment.version).osImageRef.version }}'
```
tenczar, line 601: Using this env variable when creating a management cluster should avoid needing to make this change export _ADDITIONAL_MANAGEMENT_COMPONENT_CONFIGURATION_FILE=https://gist.githubusercontent.com/tenczar/b214012420c3e20582f2f61f0611bdf1/raw/cb46bacbca6b7aee61011c7f750cbb0ea6006e9f/tkr-hack.yml

It defines a custom TKR with a template value of /dc0/photon-3-kube-v1.23.8+vmware.2-tkg.2-81d1a7892ad39f017fbaf59f9907cbe7. The benefit of this approach is that we won't need to revert it and people can create custom TKRs to use whichever template version they want.

pkg/v1/providers/yttcb/clusterbootstrap.yaml
```diff
@@ -181,7 +181,7 @@ metadata:
   annotations:
     tkg.tanzu.vmware.com/add-missing-fields-from-tkr: #@ data.values.KUBERNETES_RELEASE
 spec:
-  #@ if data.values.CNI == "antrea":
+  #@ if data.values.CNI == "antrea" and antrea_configs_exist():
```
tenczar, line 184: this looks good.

packages/tkg-clusterclass-vsphere/bundle/config/upstream/base.yaml
ydp: Thank you @tenczar I'll validate this variable, and update PR later.

ydp: Thank you @tenczar again, the variable validated, PR commit and description updated.

wjun (approved):

ydp:
Thank you @wjun , good news is that the post-merge pipeline passed  capa/capd cc pipelines! 
https://github.com/vmware-tanzu/tanzu-framework/actions/runs/3157108587/jobs/5137982873, or check the latest commit status: https://github.com/vmware-tanzu/tanzu-framework

PR #3517, opened 2022-09-30
Decouple Feature-Flags from `cli/runtime` library and move to it to plugin specific code

### LINKED ISSUE #3447: Plugin specific Feature-Flags should not be defined in plugin runtime library (cli/runtime)
**Bug description**

Currently, we are defining all available feature-flag as part of core-cli integration library which includes plugin specific feature-flags. Reference: https://github.com/vmware-tanzu/tanzu-framework/blob/main/cli/runtime/config/featureflags.go#L24-L58

Core CLI integration library should not specify plugin specific feature-flags and it should only define core-cli specific feature-flags.
Plugins specific feature-flag should be defined in plugin implementation.

Proposal is to have each plugin define its own feature-flags and use `postInstallHook` to configure the default feature-flag values to the ClientConfig if needed.

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- (x) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**
Remove plugin specific feature-flag definition from core-cli integration library.

anujc25:

- This PR decouples feature-flag definitions from the `cli/runtime` library and move it to plugin specific code
- It also uses `postInstallHook` for a specific plugin to set the defaults for the feature-flags in the config yaml

Fixes #3447

- **Manual tests:**
  - Clean tanzu config file
  - Run `tanzu config get` command
    - It only adds tanzu cli specific default feature-flags to the config
  - Install all plugins with `tanzu plugin sync` to install all plugins
    -  (As part of plugin installation `postInstallHook` should configure the plugin specific feature-flags)
    - Run `tanzu config get` and verify plugin specific default feature-flags are configured correctly.
- **Unit tests** are written and validated as part of CI 

### Release note

```release-note
Decouple Feature-Flags from `cli/runtime` library and move to it to plugin specific code
```


prkalle (commented):
The changes LGTM.

cmd/cli/plugin/managementcluster/post_install_hook.go
```diff
@@ -3,12 +3,19 @@
...
 package main
 
+import (
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+
+	"github.com/vmware-tanzu/tanzu-framework/cli/runtime/config"
+)
+
 func postInstallHook() error {
+	// Configure default feature-flags for management-cluster plugin
```
vuil, line 13: Any thoughts on how to deal with situations where the defaults need to be repopulated for any reason after a plugin is installed?
e.g config file got nuked accidentally and though we could once use "tanzu config init" to re-create much of what is lost, this is not the case anymore.

anujc25:
Closing this over #3665

PR #3528, opened 2022-10-03
WIP: Migrate feature gate resources from ClusterResourceSet

### LINKED ISSUE #3527: Migrate feature gate resources from ClusterResourceSet to Kapp
**Bug description**

Feature gates component was previously installed using ClusterResourceSet, and now will be installed as a Carvel package. We need to gracefully handle the migration of these resources that are installed via ClusterResourceSet.

- ( ) APIs
- (x) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

rajathagasthya:

This change removes YTT customizations to install feature gates via ClusterResourceSet. Feature gates will be exclusively installed via Carvel (meta) packages. This change also adds a kapp annotation to migrate these resources to be managed by kapp according to https://carvel.dev/blog/migrate-existing-resources-to-a-new-kapp-app/

Fixes #3527 

### Release note

```release-note

```


packages/featuregates/bundle/config/upstream/default-featuregate-cr.yaml
```diff
@@ -0,0 +1,10 @@
+#@ load("@ytt:data", "data")
+
+#@ if data.values.TKG_CLUSTER_ROLE != "workload" and data.values.PROVIDER_TYPE != "tkg-service-vsphere":
```
yharish991, line 3: I think we don't need to check for cluster type here, as featuregates package is a management package and is installed only on the management cluster. Also, how are we passing the `PROVIDER_TYPE` value to featuregates package?

yharish991 (changes requested):

yharish991:
closing this pr in lieu of this pr https://github.com/vmware-tanzu/tanzu-framework/pull/3549

PR #3549, opened 2022-10-04
Migrate feature gate resources from ClusterResourceSet

### LINKED ISSUE #3527: Migrate feature gate resources from ClusterResourceSet to Kapp

yharish991:
This change removes YTT customizations to install feature gates via ClusterResourceSet. Feature gates will be exclusively installed via Carvel (meta) packages. This change also adds a kapp annotation to migrate these resources to be managed by kapp according to https://carvel.dev/blog/migrate-existing-resources-to-a-new-kapp-app/

There a couple of options to do according to this blog https://carvel.dev/blog/migrate-existing-resources-to-a-new-kapp-app/:
1.  Add [kapp.k14s.io/disable-default-label-scoping-rules](http://kapp.k14s.io/disable-default-label-scoping-rules): ""  annotation to the featuregates deployment to prevent kapp from touching the immutable fields in the deployment
2.  Add [kapp.k14s.io/update-strategy](http://kapp.k14s.io/update-strategy): "fallback-on-replace" update strategy annotation, by adding this annotation, kapp will delete the existing deployment and recreate a new one.

We decided to go with option 2, that is mainly because featuregates is previously installed as a package, so the previous installation is managed by a different kapp app and now when we try to deploy it again as part of another kapp app with the option 1 annotation added, kapp will try to remove those labels resulting in the conflict. The only consequence of going with option 2 is having brief downtime as the deployment will be deleted and re-created which we think is fine for a controller like featuregates.

Fixes https://github.com/vmware-tanzu/tanzu-framework/issues/3527

### Release note

```release-note
None
```


codegold79 (approved):

providers/ytt/02_addons/management-packages/featuregates/add_default_featuregate_cr.yaml
```diff
@@ -1,43 +0,0 @@
...
-kind: FeatureGate
-metadata:
-  name: tkg-system
-  annotations:
-    tkg.tanzu.vmware.com/addon-type: "packages/management-package"
-spec:
-  namespaceSelector: {}
-#@ end
-
-#@ if data.values.TKG_CLUSTER_ROLE != "workload" and data.values.PROVIDER_TYPE != "tkg-service-vsphere":
```
danniel1205, line 15: question: wondering who is taking care of this logic now ?

yharish991: packages in the management package repository are exclusively installed on the management cluster, so we can ignore the first condition to check for workload cluster, and regarding TKGs they make the management packages available individually with a Package CR in the cluster, so we don't need to take care of that condition as well

danniel1205 (approved):
lgtm, thanks for the PR

PR #3508, opened 2022-09-29
Golang 1.18

seemiller:
### What this PR does / why we need it

This PR updates Tanzu Framework to use Go 1.18

### Release note

```release-note
Update Tanzu Framework to use Go 1.18
```

PR #3425, opened 2022-09-20
Refactoring cluster plugin to a separate go module

sathyanarays:
Signed-off-by: Sathyanarayanan Saravanamuthu <sathyanarays@vmware.com>

### What this PR does / why we need it
Cluster plugin needs to be on its own go module.

### Which issue(s) this PR fixes

Fixes #3431

### Describe testing done for PR
Tested by running Github pipelines

### Additional changes
- Modified github workflows to upload the coverage from tkg go modules as well

### Notes to reviewers
- The cluster plugin module still depends on the main tanzu-framework go module. The following are the packages:
    - github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkr/controllers/source
    - github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkr/pkg/utils
    - github.com/vmware-tanzu/tanzu-framework/pkg/v2/tkr/util/topology
- The above dependencies will go away when tkr is made as a independent go module as part of #3475

PR #3568, opened 2022-10-06
[Tooling]Fix Go version

marckhouzam:

We moved to go version with #3508, but these two new modules got added right after but the PR was opened before, so it contained the old go version (see #3425) .

Let CI run

cc @sathyanarays @seemiller


randomvariable (approved):

anujc25 (approved):

marckhouzam:
We're going to hold on this as the change to Go 1.18 may be reverted by #3571

PR #3578, opened 2022-10-07
actions cleanup

randomvariable:
- Switch to go mod download instead go deps
- wip
- Dedupe common github workflow steps

Fixes #

### Release note

```release-note

```


vuil:
Nice! Especially with conditional turning off specific parts of the setup depending on workflow needs.

.github/actions/workflow-setup.yml
```diff
@@ -0,0 +1,139 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/SchemaStore/schemastore/master/src/schemas/json/github-action.json
```
vuil, line 1: do you need
`
on: [workflow_call]
`
?

randomvariable: I don't think so - this is an action, not a workflow, so it'll stay in the same execution path. If we did a callable workflow, you'd lose the steps that we're being carried out.

randomvariable: Callable workflows are useful for other things though, like if you wanted to trigger a command done on a github runner using the gh cli. At least that's my understanding.

vuil: ah, thanks. I see now this is using the Composite Actions feature instead of Reusable Workflows. The differences are kinda subtle since both use `uses:` to wire up different parts of the pipeline.

I was mainly checking because I didn't notice any checks running for this PR. Usually even if there are errors in the configuration of a workflow it doesn't show up in the PR's check list but there will be usually some trace of it in the actions tab.

randomvariable: I'm pretty certain i've done something wrong. I'll figure it out.

PR #3586, opened 2022-10-07
Use go mod instead of go deps

randomvariable:

Go deps hasn't been needed for many Go versions, so we can drop it.
This seems to fix occasional errors as we bump dependencies (e.g. #3157 didn't work without this change).

This is a pre-requisite of #3157 

Fixes #

### Release note

```release-note
NONE
```


vuil (approved):
lgtm, thanks!

PR #3588, opened 2022-10-07
Makefile: Add "smoke-build" target to check everything can build

randomvariable:

Makefile: Add "smoke-build" target to check everything can build
This is then added to `make verify`

This isn't a critical fix, but I don't think I would have been able to do #3157 without it. It's the most effective way to check code isn't broken when you do go mod updates.

Fixes #

### Release note

```release-note

```


vuil (approved):
real clock time for verify target went from 20s to almost 3m, so might be nice to have a faster verify target for those changes that are more localized in nature. 

That said, +1 to being conservative for now especially with the increasing number of go modules to deal with.

PR #3596, opened 2022-10-07
prevents metric-server from taking ownership of kube-system namespace.

adduarte:
fixes: 3595

prevents metric-server app on administrator cluster from taking ownership of kube-system namespace. 
The configuration for the metrics-server which is deployed to the administration server uses the default
("createNamespace: true") when it is to be deployed on the "kube-system" namespace.
Using "createNamespace: true" causes kapp-controller to take ownership of the namespace and assign it to the metrics-server app.
The result is that if a admin user makes any changes to the "kube-system" namespace, they will be reverted back when metric-server app is reconciled by kapp-controller. It also adds kapp.k14s.io annotations and labels to the namespace which can interfere with control-plane operations. For example, when the metric-server app is deleted, the system will try to delete kube-system namespace which is not allowed, resulting in a problem.

Fixes # 3595

Tested by editing the ${CLUSTERNAME}-vc-metrics-server-addon

### Release note

```release-note
prevent metric-server app from taking ownership of kube-system namespace in administrator cluster
labeled kube-system namespace and verified labels are not removed when the metrics-server app is reconciled. 
```

### DIFF
--- a/providers/ytt/02_addons/metrics-server/metrics_server_secret.yaml
+++ b/providers/ytt/02_addons/metrics-server/metrics_server_secret.yaml
@@ -8,6 +8,7 @@
 ---
 metricsServer:
   namespace: kube-system
+  createNamespace: false
   config:
     args: [] #! Add any command args here
     probe:


randomvariable:
We should merge this one. Can we rebase it @adduarte ?

shyaamsn (approved):

maralavi:
@adduarte Could you please just update this statement in the description?

 ```they will be reverted back when metric-server app is reconciled by **addons-manager**``` --> ``` they will be reverted back when metric-server app is reconciled by **kapp-controller**```

maralavi (approved):

lubronzhan (approved):

PR #3597, opened 2022-10-08
Fix linter issues in cli/core

chandrareddyp:
This PR fixes linter issues in cli/core module

Fixes https://github.com/vmware-tanzu/tanzu-framework/issues/3606

Ran `make lint` and made sure there were no lint errors for `cli/core` module

### Release note

```release-note

```

### DIFF
--- a/cli/core/pkg/cli/catalog_test.go
+++ b/cli/core/pkg/cli/catalog_test.go
@@ -4,6 +4,7 @@
 //nolint:deadcode,unused
 package cli
 
+// TODO: This is legacy code tests and will be removed soon as part of TKG-13912
 import (
 	"os"
 	"path/filepath"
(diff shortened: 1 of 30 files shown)


cli/core/pkg/cli/catalog_test.go
```diff
@@ -106,7 +107,7 @@ func testHasUpdate(t *testing.T, multi *MultiRepo, numPluginsDowngraded int) {
 		}
 	}
 	require.Equal(t, numPluginsRequiringUpdate, numPluginsDowngraded)
-}
```
vuil, line 109: Not sure I understand why this is commented out instead of removed.

chandrareddyp: The reason I commented instead of removing the code was, as part of TKG-13912, we may remove or move the code in which we might use these test cases if we don't use then we will remove it as part of TKG-13912. 
@anujc25 let me know if we can remove it right away!

chandrareddyp: I have discussed with Anuj, added `nolint's`, and un-commented the code.

anujc25 (approved):
Thanks for fixing this. LGTM

PR #3598, opened 2022-10-10
Topic/pxiao/add azurefile csi component

xiaoping8385:
As a Kubernetes 1.22+ user, I would like create PVs with the CSI driver for Azure because the in-tree provider has been deprecated.

As a Platform Operator, I would like the out-of-tree CSI drivers for Azure Virtual Disk and Azure File to be  deployed when I create a new k8s cluster on Azure.

As a platform operator, I would like the out-of-tree CSI drivers for Azure Virtual Disk and Azure File to be lifecycle managed in lockstep with the k8s cluster.

As a platform operator of existing TKG clusters, I would like my existing Azure Virtual Disk and Azure File volumes created with the in-tree driver to be migrated to the out-of-tree CSI driver when I upgrade my k8s clusters to k8s 1.23+. 

- Created azure mc and wc to verify the pod is running well in CC env and in legcy env
- Manually test the NFS feature, Snapshot feature, volume resize feature as described in https://github.com/kubernetes-sigs/azurefile-csi-driver
### Release note
Support AzureFile CSI Driver in Cluster Class as well as legacy cluster


jeffwubj (approved):
CAPZ pipeline is still failing at refreshing Azure token, the patch itself LGTM

addons/controllers/azurefilecsi/azurefilecsiconfig_utils.go
```diff
@@ -0,0 +1,79 @@
...
+			return nil, nil
+		}
+		logger.Error(err, fmt.Sprintf("Unable to fetch cluster '%s/%s'", azurefileCSIConfig.Namespace, clusterName))
+		return nil, err
+	}
+
+	return cluster, nil
+}
+
+// mapVSphereCSIConfigToDataValues maps VSphereCSIConfig CR to data values
```
wjun, line 51: Update comment. This can be fixed in a future fix.

wjun (approved):

providers/yttcb/clusterbootstrap.yaml
```diff
@@ -204,6 +210,26 @@ metadata:
...
 ---
+#@ if azurefilecsi_configs_exist():
+apiVersion: csi.tanzu.vmware.com/v1alpha1
+kind: AzureFileCSIConfig
+metadata:
+  name: #@ data.values.CLUSTER_NAME
+  namespace: #@ data.values.NAMESPACE
+spec:
+  azureFileCSIDriver:
+    namespace: kube-system
```
jeffwubj, line 221: I might have made a mistake that, we'd better to replace this line with
```
```
probably we could change them in all three drivers later, sounds if this is an issue, it should be P1~P2

PR #3611, opened 2022-10-10
Ensure generators are in sync for all modules

### LINKED ISSUE #3589: Missing generators from refactored modules
**Bug description**

Tanzu Framework cannot be built reliably at the moment because of the following:

1. Directories have been moved into their own go modules for eventual split out.
2. Go tasks that relied on `./...` no longer operate on nested modules
3. Those modules *should* have their own Makefiles that are invoked from the top-level using `$(MAKE) -C <dir> go-generate`, etc..., but the dependencies are mostly missing in the nested `hack/tools`

These makefiles need to be completed, and they need to be linked from the top-level. Right now, we cannot know if a code change somewhere requires generators to be run, and it's not clear how to run them either.

You could hack around this with a loop at the top-level using all of the top-level code generators, but this will break immediately when the packages are moved.

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- (X) Developer Experience

**Expected behavior**

vijaykatam:
* Update existing Makefile targets to generate for all modules
* Generated crds for apis/cli module are moved as a result

Fixes(partially) #3589 

1. Ran generators manually by deleting existing crds and verifying post run.
2. Deleted generated fakes and ran make targets to verify

### Release note

```release-note

```


addons/manifests/tanzu-addons-manager.yaml
```diff
@@ -1,122 +0,0 @@
-apiVersion: v1
```
vijaykatam, line 1: This file is not needed.

randomvariable (approved):

randomvariable:
LGTM.

I'll let @vuil do the honours.

Makefile
```diff
@@ -154,19 +154,30 @@ deploy: manifests tools ## Deploy controller in the configured Kubernetes cluste
...
+	$(MAKE) -C apis/core generate-manifests
+	$(MAKE) -C apis/run generate-manifests
 	$(MAKE) -C cli/runtime generate-manifests
 
 generate-go: $(COUNTERFEITER) ## Generate code via go generate.
-	PATH=$(abspath hack/tools/bin):"$(PATH)" go generate ./...
+	@for i in $(GO_MODULES); do \
+		echo "-- Running go generate ./... $$i --"; \
+		pushd $${i}; \
+		PATH=$(abspath hack/tools/bin):"$(PATH)" go generate ./...; \
```
vuil, line 169: nit: to be consistent with the other generate targets, should we convert this to invoke a generate-go target in each of the Makefile as well?

vuil:
Can you also mention what is done for testing the generate-go bit?
I suggest doing the same thing as for the CRDs, manifests, iow delete the generated files and verify that we generated the same.
you can use the following to delete the files:
```
git grep go:generate | perl -pe 's#^(.*/[^/]*/)([^/]*\.go):.* -o (\S+)\s.*$#$1$3#' | xargs rm -rf
```

When I did that
I noticed the import path ordering of the generated files are shifted because we are missing the goimports call, currently part of generate-fakes.

Makefile
vijaykatam: I tried with a make target for this in common.mk but couldn't get it to work so leaving this with fmt added.

vijaykatam:
@vuil updated testing done and also added targets to `verify`

Makefile
vuil: sgtm. we can address this as a followup.

vuil: > but just atm the "make fmt" needs fixing as well
> https://github.com/vmware-tanzu/tanzu-framework/actions/runs/3222939850/jobs/5272513892#step:13:348

Maybe not, looks like the missing fmt target lines are harmless, because for now every module that needs to generate fakes also has a fmt target to go with it? Will monitor the verify check to see if this change kept the expected behavior.

vuil (approved):
lgtm. checks came back good as well.

PR #3613, opened 2022-10-10
 Redesign Feature and FeatureGate API

yharish991:

This PR 
1. Adds a new FeatureGate and Feature API in the core.tanzu.vmware.com API group and deprecates the APIs in config.tanzu.vmware.com
2. Removes namespaces concept in FeatureGate and Feature API v1alpha2 version in core.tanzu.vmware.com
3. Adds new controller to reconcile Feature resources in core.tanzu.vmware.com
4. Moves strings util functions to util module
5. Adds validating webhook for FeatureGate v1alpha2 version in core.tanzu.vmware.com
6. Deprecates existing featuregates utility functions
7. Adds new utility functions to support new API in core.tanzu.vmware.com
8. Updates featuregates package

Fixes #

Ran the controller and tested the controller and webhook functionality manually.

### Release note

```release-note
Deprecated Feature and FeatureGate API in config.tanzu.vmware.com, added new APIs Feature and FeatureGate in core.tanzu.vmware.com and a controller to reconcile feature.core.tanzu.vmware.com resources
```


apis/config/v1alpha1/feature_types.go
```diff
@@ -40,6 +40,7 @@ type FeatureStatus struct{}
 // +kubebuilder:printcolumn:name="Description",type=string,JSONPath=.spec.description
 // +kubebuilder:printcolumn:name="Immutable",type=boolean,JSONPath=.spec.immutable
 // +kubebuilder:printcolumn:name="Maturity",type=string,JSONPath=.spec.maturity
+// +kubebuilder:deprecatedversion:warning="Feature API in config.tanzu.vmware.com is deprecated. Use Feature API from core.tanzu.vmware.com instead"
```
codegold79, line 43: Nit: sometimes in kubebuilder markers there is a space after the `//` comment tags and other times there aren't. I think we should just choose a format and stick with it. I think it's more common not to have that space, so I say remove these spaces, but this is a nit, also, so your choice.

apis/core/config/crd/bases/core.tanzu.vmware.com_features.yaml
```diff
@@ -65,27 +65,28 @@ spec:
...
-                  deactivated - alpha: the first milestone meant for limited wider
-                  consumption, discoverable and defaults to deactivated - beta: greater
-                  visibility for proven designs, discoverable and defaults to activated
-                  - ga: intended to be part of the mainline codebase, non-optional
-                  - deprecated: destined for future removal'
+              stability:
+                description: 'Stability indicates stability level of this feature.
+                  Stability levels are Work In Progress, Experimental, Technical Preview,
+                  Stable and Deprecated. - Work In Progress: the default for new resources,
+                  represents local dev. intended to be hidden and deactivated - Experimental:
```
codegold79, line 72: Nit: I think "dev." should be spelled out here.

apis/core/v1alpha1/groupversion_info.go
```diff
@@ -2,8 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 // Package v1alpha1 contains API Schema definitions for the core v1alpha1 API group
-//+kubebuilder:object:generate=true
-//+groupName=core.tanzu.vmware.com
+// +kubebuilder:object:generate=true
```
codegold79, line 5: Nit: sometimes in kubebuilder markers there is a space after the `//` comment tags and other times there aren't. I think we should just choose a format and stick with it. I think it's more common not to have that space, so I say remove these spaces, but this is a nit, also, so your choice.

apis/core/v1alpha2/feature_types.go
```diff
@@ -0,0 +1,61 @@
...
+	// - Stable: intended to be part of the mainline codebase, non-optional
+	// - Deprecated: destined for future removal
+	Stability string `json:"stability"`
+}
+
+// FeatureStatus defines the observed state of Feature
+type FeatureStatus struct{}
+
+// Feature is the Schema for the features API
+// +kubebuilder:object:root=true
```
codegold79, line 37: Nit: sometimes in kubebuilder markers there is a space after the `//` comment tags and other times there aren't. I think we should just choose a format and stick with it. I think it's more common not to have that space, so I say remove these spaces, but this is a nit, also, so your choice.

Applies to lines 24 and 52, and probably others, as well.

apis/core/v1alpha2/featuregate_types.go
```diff
@@ -0,0 +1,72 @@
...
+package v1alpha2
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// FeatureReference refers to a Feature resource and specifies its intended activation state.
+type FeatureReference struct {
+	// Name is the name of the Feature resource, which represents a feature the system offers.
+	// +kubebuilder:validation:Required
```
codegold79, line 13: Nit: To space or not to space before `+kubebuilder`?

apis/core/v1alpha2/featuregate_webhook_test.go
```diff
@@ -0,0 +1,95 @@
...
+// Two slices are considered equal when they have the same length and same elements. The order of the elements is
+// ignored while comparing. Nil and empty slices are considered equal.
+//
+// This function is intended to be used in tests for comparing expected and actual values, and printing the diff for
+// users to debug:
+//
+//	if diff := sliceDiffIgnoreOrder(got, want); diff != "" {
+//	    t.Errorf("got: %v, want: %v, diff: %s", got, want, diff)
+//	}
+func sliceDiffIgnoreOrder(a, b []string) string {
```
codegold79, line 93: I'm glad you didn't put this function in a common `utils` or `helper` type package. I know we do it all the time, but I like Dave Cheney's recommendation not to, and instead put these functions (and it's ok to copy/duplicate them) in the packages they're used. 

Dave wrote and [explained in Singapore GopherCon 2019](https://dave.cheney.net/practical-go/presentations/gophercon-singapore-2019.html#_avoid_package_names_like_base_common_or_util),
> ...move the relevant functions into their caller’s package. Even if this involves duplicating some helper code this is better than introducing an import dependency between two packages.

apis/core/v1alpha2/groupversion_info.go
```diff
@@ -0,0 +1,23 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package v1alpha2 contains API Schema definitions for the core v1alpha2 API group
+// +kubebuilder:object:generate=true
```
codegold79, line 5: Nit: To space or not to space before `+kubebuilder:`?

config/crd/patches/cainjection_in_features.yaml
```diff
@@ -0,0 +1,8 @@
+# The following patch adds a directive for certmanager to inject CA into the CRD
```
codegold79, line 1: Nit: end with full stop (period).

config/crd/patches/webhook_in_features.yaml
```diff
@@ -0,0 +1,17 @@
+# The following patch enables conversion webhook for CRD
```
codegold79, line 1: Nit: end with full stop.

featuregates/client/pkg/util/featuregate.go
```diff
@@ -82,7 +86,68 @@ func ComputeFeatureStates(featureGateSpec configv1alpha1.FeatureGateSpec, featur
 	return activated, deactivated, unavailable
 }
 
+// ComputeFeatureStatesForCoreAPIGroupResources takes a corev1alpha2 FeatureGate spec and computes the actual state (activated, deactivated or unavailable)
```
codegold79, line 89: Nit: what is our repo's style guide for line width? Whatever it is, these lines here (89-90) and the ones above (30-33) might exceed it.

featuregates/client/pkg/util/featuregate.go
```diff
@@ -82,7 +86,68 @@ func ComputeFeatureStates(featureGateSpec configv1alpha1.FeatureGateSpec, featur
...
+		if f.Activate {
+			toActivate.Insert(f.Name)
+		} else {
+			toDeactivate.Insert(f.Name)
+		}
+	}
+
+	// discovered is set a set of available features that are discoverable.
+	discovered := sets.String{}
+	// discoveredDefaultActivated is a set of available features that are discoverable and activated by default.
```
codegold79, line 105: Nit: 105 and 106 also seem too long as well as comments in the rest of this doc.

packages/featuregates/bundle/config/upstream/crds/core/core.tanzu.vmware.com_features.yaml
```diff
@@ -0,0 +1,103 @@
...
+      name: Activated
+      type: boolean
+    - jsonPath: .spec.description
+      name: Description
+      type: string
+    - jsonPath: .spec.immutable
+      name: Immutable
+      type: boolean
+    - jsonPath: .spec.maturity
+      name: Maturity
```
codegold79, line 29: Would this be "Stability" now in core?

util/cmp/compare.go
```diff
@@ -0,0 +1,101 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
```
codegold79, line 1: Nit: year? Or maybe this is just copied from 2021 file?

util/cmp/compare_test.go
```diff
@@ -0,0 +1,91 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
```
codegold79, line 1: nit: year check

util/cmp/strings/slices.go
```diff
@@ -0,0 +1,23 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
```
codegold79, line 1: nit: year check

util/cmp/strings/slices_test.go
```diff
@@ -0,0 +1,60 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
```
codegold79, line 1: nit: year check

codegold79 (approved):
I can't really see anything wrong with the new webhook, conversion from v1alpha1 -> v1alpha2, nor the new featuregate reconciler. Only left nitpicks.

For the nitpicks, you don't even need to comment on or fix them. Click "resolve" as you wish. They're really too small to even discuss.

apis/core/v1alpha2/featuregate_webhook_test.go
codegold79: Maybe I'm mistaken. I see you did follow the habit of Tanzu Framework/VMware/Kubernetes and put the function in util/cmp/strings/slices.go. In that case, does this `sliceDiffIgnoreOrder` function still need to be here?

codegold79:
I found something that should be updated as part of this PR: https://github.com/vmware-tanzu/tanzu-framework/blob/main/docs/api-machinery/features-and-featuregates.md

yharish991:
> I found something that should be updated as part of this PR: https://github.com/vmware-tanzu/tanzu-framework/blob/main/docs/api-machinery/features-and-featuregates.md

ya, this is still in WIP, i will update the docs in my last commit

config/crd/patches/cainjection_in_features.yaml
yharish991: this is generated file

featuregates/client/pkg/util/featuregate.go
```diff
@@ -131,3 +140,71 @@ func FeaturesActivatedInNamespacesMatchingSelector(ctx context.Context, c client
...
+		Name: featureName,
+	}, feature); err != nil {
+		return false, fmt.Errorf("couldn't retrieve feature %s :%w", featureName, err)
+	}
+
+	return feature.Status.Activated, nil
+}
+
+// GetFeatureGateForFeature returns FeatureGate resource that is gating the feature
+func GetFeatureGateForFeature(ctx context.Context, c client.Client, featureName string) (*corev1alpha2.FeatureGate, bool, error) {
```
rajathagasthya, line 157: Not important right away, but can we [setup FieldIndexers](https://github.com/kubernetes-sigs/controller-runtime/blob/4c9c9564e4652bbdec14a602d6196d8622500b51/pkg/client/example_test.go#L249-L268) to answer these types of "contains" queries with less cost? These functions are used the in reconcile loop meaning there will be a lot of GET API calls.

rajathagasthya (commented):
Round 1: API review

Still need to look closely at the controller code.

apis/core/v1alpha2/featuregate_types.go
```diff
@@ -0,0 +1,83 @@
...
+// FeatureGateStatus defines the observed state of FeatureGate
+type FeatureGateStatus struct {
+	// Results represents the results of all the features specified in the FeatureGate spec.
+	// +listType=map
+	// +listMapKey=name
+	Results []Result `json:"results"`
+}
+
+// Result represents the result of Feature.
+type Result struct {
```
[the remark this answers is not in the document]
yharish991: i agree `Result` being too generic, but i think changing it to `FeatureState` can imply that this is actual feature state ie either `activated` or `deactivated`. I'm open to other options

featuregates/client/pkg/util/featuregate.go
yharish991: ah right.. i totally forgot about FieldIndexers, i think this makes sense.. i will do this in a follow up pr

apis/core/v1alpha2/featuregate_types.go
```diff
@@ -0,0 +1,83 @@
...
+// FeatureReference refers to a Feature resource and specifies its intended activation state.
+type FeatureReference struct {
+	// Name is the name of the Feature resource, which represents a feature the system offers.
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+	// Activate indicates the activation intent for the feature.
+	Activate bool `json:"activate,omitempty"`
+	// SkipStabilityPolicyValidation lets you skip the policy validation performed when toggling the feature.
+	// Once set to true, cannot be set back to false
+	SkipStabilityPolicyValidation bool `json:"skipStabilityPolicyValidation,omitempty"`
```
zjs, line 19: I'd name this something more explicit; I think someone could set this without understanding the impact.

How about `PermanentlyVoidAllSupportGuaranteesForThisEnvironment`?

apis/core/v1alpha2/featuregate_webhook.go
```diff
@@ -0,0 +1,270 @@
...
+	for _, featureRef := range r.Spec.Features {
+		stabilityLevel, found := getFeatureStabilityLevel(features, featureRef.Name)
+		if !found {
+			// Feature doesn't exist and is validated in validateFeatureExistence method
+			continue
+		}
+		policy := GetPolicyForStabilityLevel(stabilityLevel)
+		// immutability validation
+		if policy.Immutable && !featureRef.SkipStabilityPolicyValidation && policy.DefaultActivation != featureRef.Activate {
+			invalidFeatures.Insert(featureRef.Name)
```
zjs, line 147: I don't think I would include `!featureRef.SkipStabilityPolicyValidation` in this case. Allowing users to deactivate stable features seems complicated. If that's necessary, we could add it later, when we better understand the use case.

apis/core/v1alpha2/featuregate_webhook.go
```diff
@@ -0,0 +1,270 @@
...
+		}
+	}
+
+	if len(invalidFeatures) != 0 {
+		allErrors = append(allErrors, field.Invalid(field.NewPath("spec").Child("features"), r.Spec.Features,
+			fmt.Sprintf("cannot toggle features due to policy violation: %v."+
+				"Set feature.skipStabilityPolicyValidation to true to override policy."+
+				"Once set to true, cannot be set back to false.", invalidFeatures.List())))
+	}
```
zjs, line 158: I think my suggestion above might require you to maintain separate lists for the two error cases in this method: immutability violations and warranty violations. I doing so would also help with the clarity of the overall error message?

capabilities/controller/main.go
```diff
@@ -55,7 +55,7 @@ func main() {
 
 	if err = (&run.CapabilityReconciler{
 		Client: mgr.GetClient(),
-		Log:    ctrl.Log.WithName("controllers").WithName("Capability"),
+		Log:    ctrl.Log.WithName("controllers").WithName("Capability").WithName("apigroup").WithName("run"),
```
zjs, line 58: **nitpick**: As above, I'd use WithValues for the API Group.

featuregates/controller/pkg/feature/feature_controller.go
```diff
@@ -0,0 +1,242 @@
...
+		// Remove feature from FeatureGate status
+		featureGate.Status.Results = computeFeatureGateStatusResults(featureGate.Status, corev1alpha2.Result{
+			Name: feature.Name,
+		}, false)
+		if err := c.Status().Update(ctx, featureGate); err != nil {
+			return fmt.Errorf("couldn't update %s FeatureGate status :%w", featureGate.Name, err)
+		}
+	}
+	// Update Feature status to set feature as deactivated
+	feature.Status.Activated = false
```
zjs, line 141: If the feature is no longer referenced from any feature gate, shouldn't we set the feature to its default activation state? (Which might be true in some cases!)

featuregates/client/pkg/util/featuregate.go
```diff
@@ -131,3 +140,71 @@ func FeaturesActivatedInNamespacesMatchingSelector(ctx context.Context, c client
...
+	}
+
+	for i, fg := range featureGateList.Items {
+		for _, f := range featureGateList.Items[i].Spec.Features {
+			if f.Name == featureName {
+				featureGate := &corev1alpha2.FeatureGate{}
+				if err := c.Get(ctx, types.NamespacedName{
+					Name: fg.Name,
+				}, featureGate); err != nil {
+					return nil, false, fmt.Errorf("couldn't get FeatureGate resource %s: %w", featureGate.Name, err)
```
zjs, line 170: I feel like I'm missing something obvious, but why not just `return fg, true, nil`?

featuregates/client/pkg/util/featuregate.go
```diff
@@ -131,3 +140,71 @@ func FeaturesActivatedInNamespacesMatchingSelector(ctx context.Context, c client
...
+	}
+
+	for i, fg := range featureGateList.Items {
+		for _, result := range featureGateList.Items[i].Status.Results {
+			if result.Name == featureName {
+				featureGate := &corev1alpha2.FeatureGate{}
+				if err := c.Get(ctx, types.NamespacedName{
+					Name: fg.Name,
+				}, featureGate); err != nil {
+					return nil, false, fmt.Errorf("couldn't get FeatureGate resource %s: %w", featureGate.Name, err)
```
zjs, line 193: As above, probably missing the same obvious thing, but why not just `return fg, true, nil`?

featuregates/controller/main.go
```diff
@@ -53,17 +56,31 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err = (&featuregate.FeatureGateReconciler{
+	if err = (&configFeatureGateController.FeatureGateReconciler{
 		Client: mgr.GetClient(),
-		Log:    ctrl.Log.WithName("controllers").WithName("FeatureGate"),
+		Log:    ctrl.Log.WithName("controllers").WithName("FeatureGate").WithName("apigroup").WithName("config"),
```
zjs, line 61: **nitpick**: As above, I'd use `WithValues` to make a key-value pair.

featuregates/controller/pkg/feature/doc.go
```diff
@@ -0,0 +1,5 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package feature has the controller for Feature in core API group.
```
zjs, line 4: **nitpick**: Maybe explain that this is responsible for determining the feature's activation state based on its stability level and any feature gates?

featuregates/controller/pkg/feature/feature_controller.go
```diff
@@ -0,0 +1,242 @@
...
+func applyPolicyToComputeFeatureResultAndActivation(policy corev1alpha2.Policy, featureRef corev1alpha2.FeatureReference, stabilityLevel string) (corev1alpha2.Result, bool) {
+	activated := policy.DefaultActivation
+	result := corev1alpha2.Result{Name: featureRef.Name}
+	// Check for immutability and change in intent of feature status
+	if policy.Immutable && !featureRef.SkipStabilityPolicyValidation && policy.DefaultActivation != featureRef.Activate {
+		result.Status = constants.InvalidFeatureReferenceStatus
+		result.Message = "Feature couldn't be toggled because it is immutable"
+	} else if policy.VoidsWarranty && !featureRef.SkipStabilityPolicyValidation {
+		result.Status = constants.InvalidFeatureReferenceStatus
+		result.Message = "Toggling this feature voids the warranty"
```
zjs, line 158: **nitpick**: Could this message be clearer? (E.g., "The stability level of this feature indicates that it should not be activated in production environments. To activate the feature, you must agree to permanently void all support guarantees for this environment by setting PermanentlyVoidAllSupportGuaranteesForThisEnvironment = true.")

apis/core/v1alpha2/feature_types.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// FeatureSpec defines the desired state of Feature
+type FeatureSpec struct {
+	// Description of the feature.
+	Description string `json:"description"`
+	// Stability indicates stability level of the feature.
+	// Stability levels are Work In Progress, Experimental, Technical Preview, Stable and Deprecated.
+	// +kubebuilder:validation:Enum=Work In Progress;Experimental;Technical Preview;Stable;Deprecated
+	// - Work In Progress: Feature is not ready, it is still being developed
```
zjs, line 17: How about:

> Feature is still under development. It is not ready to be used, except by the team working on it. Activating this feature is not recommended under any circumstances.

apis/core/v1alpha2/feature_types.go
```diff
@@ -0,0 +1,56 @@
...
+	// - Experimental: Feature is not ready, but customers can try it out in pre-production clusters. It's unsupported and voids warranty.
```
zjs, line 18: How about:

> Feature is not ready, but may be used in pre-production environments. However, if an experimental feature has ever been used in an environment, that environment will not be supported. Activating an experimental feature requires you to permanently, irrevocably void all support guarantees for this environment by setting PermanentlyVoidAllSupportGuaranteesForThisEnvironment = true. You will need to recreate the environment to return to a supported state.

apis/core/v1alpha2/feature_types.go
```diff
@@ -0,0 +1,56 @@
...
+	// - Technical Preview: Almost ready, but customers are encouraged to try it out in pre-production clusters. It's unsupported, but doesn't void warranty
```
zjs, line 19: How about:

> Feature is not ready, but is not believed to be dangerous. The feature itself is unsupported, but activating a technical preview feature does not affect the support status of the rest of the environment.

apis/core/v1alpha2/feature_types.go
```diff
@@ -0,0 +1,56 @@
...
+	// - Stable: Feature is ready and fully supported
+	// - Deprecated: Feature is destined for removal, usage is discouraged
```
zjs, line 21: How about:

> Feature is destined for removal, usage is discouraged. Deactivate this feature prior to upgrading to a release which has removed it to validate that you are not still using it and to prevent users from introducing new usage of it.

(Disclaimer: I feel like this still isn't as clear as it could be.)

zjs (dismissed):

featuregates/client/pkg/util/featuregate.go
yharish991: I remember why I did this now, I was thinking from a perspective that these functions could be helpful on the client side as well. Will revisit this comment later

apis/core/v1alpha2/feature_stability_policies.go
```diff
@@ -0,0 +1,45 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package v1alpha2
+
+// Policy represents Stability level policy
+type Policy struct {
```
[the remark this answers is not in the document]
yharish991: what do you think about `Discoverable bool`?

apis/core/v1alpha2/featuregate_webhook.go
```diff
@@ -0,0 +1,270 @@
...
+	return allErrors
+}
+
+// validateStabilityPolicyOverride validates stability policy override
```
codegold79, line 162: nit: I think rewriting to the following might make this comment easier to understand:
```
// validateStabilityPolicyOverride determines if a policy override for stability in an old FeatureGate resource object is valid.

File: apis/core/v1alpha2/featuregate_webhook.go [hunk 35]
```diff
@@ -0,0 +1,270 @@
...
+		skipStabilityPolicyValidationFlagFromOldObj, found := getSkipStabilityPolicyValidationFieldForFeature(oldObject, featureRef.Name)
+		if !found {
+			// Feature doesn't exist in old object and need not be validated
+			continue
+		}
+		if !featureRef.SkipStabilityPolicyValidation && skipStabilityPolicyValidationFlagFromOldObj {
+			invalidFeatures.Insert(featureRef.Name)
+		}
+	}
+	if len(invalidFeatures) != 0 {
```
codegold79, line 177: I know it's impossible for length here to be negative, but it would be better to be more explicit about your intention.
```suggestion
	if len(invalidFeatures) > 0 {
```

apis/core/v1alpha2/featuregate_webhook.go
```diff
@@ -0,0 +1,270 @@
...
+		allErrors = append(allErrors, field.Invalid(field.NewPath("spec").Child("features"), r.Spec.Features,
+			fmt.Sprintf("cannot toggle features due to policy violation: %v."+
+				"Set feature.skipStabilityPolicyValidation to true to override policy."+
+				"Once set to true, cannot be set back to false.", invalidFeatures.List())))
+	}
+	return allErrors
+}
+
+// validateStabilityPolicyOverride validates stability policy override
+func (r *FeatureGate) validateStabilityPolicyOverride(oldObject *FeatureGate) field.ErrorList {
```
codegold79, line 163: Why are only 'oldObjects' validated? I'm assuming that an 'oldObject' is a config v1alpha1 CRD version FeatureGate. If my assumption is correct, why aren't 'new' objects using core v1alpha2 also checked for stability policy override validation? Or maybe this function is only for checking old objects? If so, maybe mention in the comment where the new objects are validated? I'm probably misunderstanding something.

apis/core/v1alpha2/featuregate_webhook.go
codegold79: Even if it requires code duplication, I think the slight reduction in performance is worth increased readability in this case. Therefore +1 to @zjs to maintain separate lists for the two error cases.

apis/core/v1alpha2/featuregate_webhook.go
```diff
@@ -0,0 +1,270 @@
...
+
+	invalidFeatures := sets.String{}
+	for _, featureRef := range r.Spec.Features {
+		stabilityLevel, found := getFeatureStabilityLevel(features, featureRef.Name)
+		if !found {
+			// Feature doesn't exist and is validated in validateFeatureExistence method
+			continue
+		}
+		policy := GetPolicyForStabilityLevel(stabilityLevel)
+		// immutability validation
```
codegold79, line 145: Optional: I don't think this comment is very helpful. I can tell from line 146 this is some sort of validation on `policy.Immutable`. What would be helpful is a comment that says something like, "Verify that a feature that is immutable due to the stability policy has sufficient overrides to be (de)activated."

apis/core/v1alpha2/featuregate_webhook.go
```diff
@@ -0,0 +1,270 @@
...
+	if len(invalidFeatures) != 0 {
+		allErrors = append(allErrors, field.Invalid(field.NewPath("spec").Child("features"), r.Spec.Features,
+			fmt.Sprintf("cannot set the feature.skipStabilityPolicyValidation back to false after the stability "+
+				"policy has been overridden: %v", invalidFeatures.List())))
+	}
+	return allErrors
+}
+
+// validateFeatureExistence validates feature existence in the cluster
+func (r *FeatureGate) validateFeatureExistence(ctx context.Context, c client.Client) field.ErrorList {
```
codegold79, line 186: optional: I like `validateFeatureExists` better. Comment can then be changed to

```
// validateFeatureExists checks that a feature can be found in the cluster.
```

apis/core/v1alpha2/featuregate_webhook.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	allFeaturesInCluster := sets.String{}
+	for _, feature := range features.Items {
+		allFeaturesInCluster.Insert(feature.Name)
+	}
+
+	invalidFeatures := allFeaturesInSpec.Difference(allFeaturesInCluster)
+
+	if len(invalidFeatures) != 0 {
```
codegold79, line 207:
```suggestion
	if len(invalidFeatures) > 0 {
```

apis/core/v1alpha2/featuregate_webhook.go
```diff
@@ -0,0 +1,270 @@
...
+		allErrors = append(allErrors, field.Invalid(field.NewPath("spec").Child("features"), r.Spec.Features,
+			fmt.Sprintf("some features in the FeatureGate spec do not exist in cluster: %v", invalidFeatures.List())))
+	}
+	return allErrors
+}
+
+// validateFeatureConflicts validates that the feature is only being gated by one featuregate
+func (r *FeatureGate) validateFeatureConflicts(ctx context.Context, c client.Client) field.ErrorList {
```
codegold79, line 215: Thanks to the helpful comment, I think the function's name would be better described as `validateFeatureGateConflict`. Because if you have more than one FeatureGate for one feature, it's not the feature that conflicts, is it? It's the FeatureGate that will conflict.

codegold79 (changes requested):
I've only reviewed what I needed for list and activate feature so far which is mostly webhook validation code and the feature, featuregate types. If there's time, I can review more.

apis/core/v1alpha2/feature_stability_policies.go
```diff
@@ -0,0 +1,45 @@
...
+	DefaultActivation bool
+	Immutable         bool
+	VoidsWarranty     bool
+}
+
+// StabilityPolicies is map that holds the policies for different stability levels
+var StabilityPolicies = map[string]Policy{
```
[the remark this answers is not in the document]
yharish991: we cannot have map as a const, marking this as resolved

apis/core/v1alpha2/featuregate_types.go
yharish991: I think `PermanentlyVoidAllSupportGuaranteesForThisEnvironment` is too wordy, so went with `PermanentlyVoidAllSupportGuarantees`. Let me know what you think.

apis/core/v1alpha2/featuregate_webhook.go
yharish991: i think it shouldn't matter, but will update it

apis/core/v1alpha2/feature_types.go
yharish991: i think this is clear

apis/core/v1alpha2/featuregate_webhook.go
yharish991: i updated the method to `validateConflictingFeaturesInFeatureGate` and i think  it makes more sense

apis/core/v1alpha2/featuregate_webhook.go
yharish991: this webhook is for validating only core v1alpha2 resources.. here oldObject means the object before the update of the featuregate resource

yharish991:
@rajathagasthya @zjs @codegold79 addressed some of the review comments and added a reply to other review comments. Let me know what you think..

featuregates/controller/pkg/feature/feature_controller.go
```diff
@@ -0,0 +1,244 @@
...
+func (r *FeatureReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	ctxCancel, cancel := context.WithTimeout(ctx, contextTimeout)
+	defer cancel()
+
+	log := r.Log.WithValues("feature", req.NamespacedName)
+	log.Info("Starting reconcile")
+
+	feature := &corev1alpha2.Feature{}
+	if err := r.Client.Get(ctxCancel, req.NamespacedName, feature); err != nil {
+		// Reconcile deleted feature
```
rajathagasthya, line 52: Could you expand these comments? I don't think things are obvious in this controller, so the more detailed comments the better.

featuregates/controller/pkg/feature/feature_controller.go
```diff
@@ -0,0 +1,244 @@
...
+
+// reconcileFeatureInFeatureGateSpec reconciles feature that is present in FeatureGate spec
+func reconcileFeatureInFeatureGateSpec(ctx context.Context, c client.Client, featureGate *corev1alpha2.FeatureGate, feature *corev1alpha2.Feature) error {
+	policy := corev1alpha2.GetPolicyForStabilityLevel(feature.Spec.Stability)
+	featureReference, _ := util.GetFeatureReferenceFromFeatureGate(featureGate, feature.Name)
+	featureResult, activate := applyPolicyToComputeFeatureResultAndActivation(policy, featureReference)
+
+	// Update FeatureGate status
+	featureGate.Status.Results = computeFeatureGateStatusResults(featureGate.Status, featureResult, true)
+	if err := c.Status().Update(ctx, featureGate); err != nil {
```
rajathagasthya, line 90: Does this update result yet another reconcile event for this controller because of the event handler? Or should we add some predicates to filter those events? We will end up with duplicate events otherwise.

featuregates/controller/pkg/feature/feature_controller.go
```diff
@@ -0,0 +1,244 @@
...
+
+	// Update Feature status
+	feature.Status.Activated = activate
+	if err := c.Update(ctx, feature); err != nil {
+		return fmt.Errorf("could not update %s Feature status :%w", feature.Name, err)
+	}
+	return nil
+}
+
+// reconcileDeletedFeature reconciles feature that has been deleted
```
rajathagasthya, line 102: Is this comment referring to the Feature CR being deleted or the FeatureReference being removed from FeatureGate?

rajathagasthya (dismissed):

apis/core/v1alpha2/featuregate_webhook.go
codegold79: Ok, got it. Thanks.

featuregates/client/pkg/util/featuregate_test.go
```diff
@@ -305,3 +307,300 @@ func TestFeaturesActivatedInNamespacesMatchingSelector(t *testing.T) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	defer cancel()
+	testCases := []struct {
+		description string
+		featureName string
+		want        bool
+		returnErr   bool
+	}{
+		{
+			description: "should return true",
```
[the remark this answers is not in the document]
codegold79: +1

featuregates/controller/pkg/feature/feature_controller.go
```diff
@@ -0,0 +1,244 @@
...
+// +kubebuilder:rbac:groups=core.tanzu.vmware.com,resources=features,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=core.tanzu.vmware.com,resources=features/status,verbs=get;update;patch
+
+// Reconcile reconciles the FeatureGate spec by computing activated, deactivated and unavailable features.
+func (r *FeatureReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	ctxCancel, cancel := context.WithTimeout(ctx, contextTimeout)
+	defer cancel()
+
+	log := r.Log.WithValues("feature", req.NamespacedName)
+	log.Info("Starting reconcile")
```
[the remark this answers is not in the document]
codegold79: I don't know...if you're looking at pod logs, don't you want to see how many times reconcile was triggered and when? You only really look at pod logs to debug and this is I think the minimum information you want to see when your pod crashes and something goes wrong.

featuregates/controller/pkg/feature/feature_controller.go
codegold79: +1

This controller is doing some things that are not obvious to wrap my head around. So comments that quickly tell us what the goal of "reconcile deleted feature" is would be very helpful.

apis/core/v1alpha2/featuregate_types.go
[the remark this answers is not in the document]
yharish991: how about FeatureReferenceResult?

featuregates/controller/pkg/feature/feature_controller.go
yharish991: yes, this is a known issue. I need to think a little more about this on how to avoid this. but reconciliation is idempotent, so not a blocker for now IMO.

featuregates/controller/pkg/feature/feature_controller.go
yharish991: reconcileDeletedFeature is for reconciling deleted Feature CR, i can expand the comments a little more to make it easy to understand

featuregates/controller/pkg/feature/feature_controller.go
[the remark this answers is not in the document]
yharish991: i think this message would be useful when debugging the controller and i don't think this would pollute our logs too mych. Also, we do the same thing for [old featuregate controller](https://github.com/vmware-tanzu/tanzu-framework/blob/main/featuregates/controller/pkg/featuregate/featuregate_controller.go#L42) and for [capabilities controller](https://github.com/vmware-tanzu/tanzu-framework/blob/main/capabilities/controller/pkg/capabilities/run/capability_controller.go#L38). If we end up changing this here, we should also make changes in other places for consistency purposes

featuregates/controller/pkg/feature/feature_controller.go
yharish991: not sure if this [comment](https://github.com/vmware-tanzu/tanzu-framework/pull/3613/files#diff-2f55534c92afa48c0492aa91850834888c7c5fce2af6cf6a661137cf84d1f631R102) is helpful, but, yes i can expand more

apis/core/v1alpha2/featuregate_types.go
```diff
@@ -0,0 +1,91 @@
...
+// FeatureGateStatus defines the observed state of FeatureGate
+type FeatureGateStatus struct {
+	// Results represents the results of all the features specified in the FeatureGate spec.
+	// +listType=map
+	// +listMapKey=name
+	Results []Result `json:"results"`
+}
+
+// FeatureReferenceStatus represents the status of the feature reference in the FeatureGate spec
+type FeatureReferenceStatus string
```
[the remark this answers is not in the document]
codegold79: > sound too generic

I agree. I like `FeatureReferenceStatus.

featuregates/controller/pkg/feature/feature_controller.go
codegold79: > Check if feature part of any FeatureGate spec and update status in Feature result to Invalid

That comment is confusing. Do you mean,

Check if the features listed in the FeatureGate spec FeatureReference slice is invalid and update the Feature result appropriately.

Might explain for line 52, "Reconcile deleted feature" what it means to be invalid due to deletion?

apis/core/v1alpha2/featuregate_webhook.go
```diff
@@ -0,0 +1,318 @@
...
+	if len(invalidFeatures) > 0 {
+		allErrors = append(allErrors, field.Invalid(field.NewPath("spec").Child("features"), r.Spec.Features,
+			fmt.Sprintf("cannot toggle features due to policy violation: %v."+
+				"Set feature.skipStabilityPolicyValidation to true to override policy."+
+				"Once set to true, cannot be set back to false.", invalidFeatures)))
+	}
+	return allErrors
+}
+
+// computeInvalidStabilityPolicyOverridedFeatures computes and returns features that have overrided the stability policy override
```
codegold79, line 205:
```suggestion
// computeInvalidStabilityPolicyOverridedFeatures computes and returns features
// that have overridden the stability policy.
```

apis/core/v1alpha2/featuregate_webhook.go
```diff
@@ -0,0 +1,318 @@
...
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+//+kubebuilder:webhook:verbs=create;update,path=/validate-core-tanzu-vmware-com-v1alpha2-featuregate,mutating=false,failurePolicy=fail,groups=core.tanzu.vmware.com,resources=featuregates,versions=v1alpha2,name=vfeaturegate.kb.io
+
+var _ webhook.Validator = &FeatureGate{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
```
codegold79, line 68: For this and all the other comments, I'd like to suggest following the official Go style guide: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

> Comments documenting declarations should be full sentences...Comments should begin with the name of the thing being described and end in a period...

codegold79 (changes requested):

apis/core/v1alpha2/featuregate_webhook.go
codegold79: I see you are much more comfortable with longer line lengths (maybe due to your huge monitor?). I try to stick to 80 characters, but make them longer if I don't break anything up that would be better together. I wish we had a line-length number/range for TF, but we don't, so go ahead an make my above suggestion a longer line. I like to follow the guidelines at https://github.com/golang/go/wiki/CodeReviewComments#line-length

yharish991: restricting the line length to 80 characters is too constraining and outdated, I prefer to limit the line length to 120 characters as we have wide monitors these days and 120 characters doesn't seem constraining. I do see a few lines in my code that exceeded 120 character length, will fix them.

apis/core/v1alpha2/featuregate_webhook.go
```diff
@@ -0,0 +1,318 @@
...
+		}
+		policy := GetPolicyForStabilityLevel(stabilityLevel)
+		if policy.VoidsWarranty && !featureRef.PermanentlyVoidAllSupportGuarantees && policy.DefaultActivation != featureRef.Activate {
+			invalidFeatures.Insert(featureRef.Name)
+		}
+	}
+	return invalidFeatures.List()
+}
+
+// computeImmutableFeatures computes and returns features which are immutable
```
codegold79, line 173:
```suggestion
// computeImmutableFeatures computes and returns features that are immutable.
```

apis/core/v1alpha2/featuregate_webhook.go
```diff
@@ -0,0 +1,318 @@
...
+
+	if len(immutableFeatures) > 0 {
+		allErrors = append(allErrors, field.Invalid(field.NewPath("spec").Child("features"), r.Spec.Features,
+			fmt.Sprintf("cannot toggle immutable features: %v", immutableFeatures)))
+	}
+
+	return allErrors
+}
+
+// computeFeaturesThatVoidSupportWarranty computes and returns features which voids the support warranty
```
codegold79, line 156:
```suggestion
// computeFeaturesThatVoidSupportWarranty computes and returns features
// that void the support warranty.
```

apis/core/v1alpha2/featuregate_webhook.go
codegold79: Sounds good. I don't like 120 as I think it's too long, but as I said, we have no guidelines, so 120 is all right.

apis/core/v1alpha2/featuregate_webhook.go
```diff
@@ -0,0 +1,318 @@
...
+
+	if len(conflicts) > 0 {
+		allErrors = append(allErrors, field.Invalid(field.NewPath("spec").Child("features"),
+			r.Spec.Features, fmt.Sprintf("features %v cannot be gated by multiple featuregates", conflicts)))
+	}
+	return allErrors
+}
+
+// computeConflictingFeatures computes and returns features that conflict with existing feature gates.
+func computeConflictingFeatures(featureGate *FeatureGate, featureGates *FeatureGateList) []string {
```
codegold79, line 276: I probably should mention this here even though we talked about it. Even if as a reminder.

I like your suggestion that instead of writing my own validation logic that does the same thing as this webhook, if you could export these computeXXXXX functions, I can use them for my feature plugin. That would be helpful.

codegold79 (changes requested):

featuregates/controller/pkg/feature/feature_controller.go
yharish991: The controller is reconciling on the Feature resource, so it is basically saying when reconciling the deleted feature, check if it is part of any FeatureGate spec and update the Feature result to invalid in FeatureGate status

apis/core/v1alpha2/featuregate_webhook.go
yharish991: resolving this based on the discussion here https://github.com/vmware-tanzu/tanzu-framework/pull/3613#discussion_r1012342860

apis/core/v1alpha2/feature_stability_policies.go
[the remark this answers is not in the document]
zjs: Yeah, that's better!

apis/core/v1alpha2/featuregate_types.go
zjs: Works for me!

apis/core/v1alpha2/featuregate_webhook.go
yharish991: resolving this as you are going to write your own validation functions for cli plugin

apis/core/v1alpha2/feature_stability_policies.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package v1alpha2
+
+// Policy represents Stability level policy
+type Policy struct {
```
danniel1205, line 7: nit:  could we add some inline comments that briefly describe what the following four policies mean? I guess some of them are not self-explanation, e.g, `VoidWarranty`. And I guess not all people have read the design doc.

apis/core/v1alpha2/featuregate_types.go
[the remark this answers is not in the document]
danniel1205: Naming is hard, but `FeatureReferenceResult` sounds better than `Result`. I guess we could choose one and move forward. We could always iterate in the future.

apis/core/v1alpha2/featuregate_webhook.go
```diff
@@ -0,0 +1,327 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if err := k8sscheme.AddToScheme(s); err != nil {
+		return nil, err
+	}
+	return s, nil
+}
+
+// Get a cached client.
```
danniel1205, line 39: nit: `Get a cached client if it has been initiated`

apis/core/v1alpha2/featuregate_webhook.go
```diff
@@ -0,0 +1,327 @@
...
+	invalidFeatures := sets.String{}
+	for _, featureRef := range spec.Features {
+		stabilityLevel, found := getFeatureStabilityLevel(features, featureRef.Name)
+		if !found {
+			// Feature doesn't exist and is validated in validateFeatureExistence method
+			continue
+		}
+		policy := GetPolicyForStabilityLevel(stabilityLevel)
+		if policy.VoidsWarranty && !featureRef.PermanentlyVoidAllSupportGuarantees &&
+			policy.DefaultActivation != featureRef.Activate {
```
danniel1205, line 169: question: trying to understand the logic of `policy.VoidsWarranty && !featureRef.PermanentlyVoidAllSupportGuarantees`, is it equivalent to `policy.VoidsWarranty != featureRef.PermanentlyVoidAllSupportGuarantees` ?

And another question is why we need `policy.DefaultActivation != featureRef.Activate` ?

apis/core/v1alpha2/feature_stability_policies.go
yharish991: i have a minor cleanup commit coming, i already added these changes to that commit

apis/core/v1alpha2/featuregate_webhook.go
yharish991: ya we could write that as `policy.VoidsWarranty != featureRef.PermanentlyVoidAllSupportGuarantees`, the outcome is going to be the same, but i think is more readable and easy to understand `policy.VoidsWarranty && !featureRef.PermanentlyVoidAllSupportGuarantees`, it basically says if the policy voids warranty and if PermanentlyVoidAllSupportGuarantees is not set to true, its not a valid feature ref in the featuregate spec.

regarding `policy.DefaultActivation != featureRef.Activate`, we want to indicate that a feature is invalid only when toggling the feature, for example, we don't want to say the feature is invalid in the below case 

```
apiVersion: core.tanzu.vmware.com/v1alpha2
kind: FeatureGate
metadata:
  name: tkg-system
spec:
  features:
    - name: big-cache
      activate: false
---
apiVersion: core.tanzu.vmware.com/v1alpha2
kind: Feature
metadata:
  name: big-cache
spec:
  description: "A sample big cache Feature"
  stability: "Experimental"
status:
  activated: false
```

apis/core/v1alpha2/featuregate_types.go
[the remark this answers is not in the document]
yharish991: yes, lets move with `FeatureReferenceResult` as multiple people are okay with that.

> We could always iterate in the future

changing the name of the field in the CRD would be a breaking change, but we are still in alpha versions, so it is okay.

apis/core/v1alpha2/featuregate_webhook.go
danniel1205: could you help add above explanation as inline comment around L169 ? It would help future maintainers to understand the logic better. thanks

config/crd/kustomization.yaml
```diff
@@ -25,7 +27,12 @@ patchesStrategicMerge:
 #- patches/webhook_in_clusterbootstraps.yaml
 #- patches/webhook_in_clusterbootstraptemplates.yaml
 #- patches/webhook_in_awsebscsiconfigs.yaml
+<<<<<<< HEAD
 #- patches/webhook_in_azurefilecsiconfigs.yaml
+=======
+#- patches/webhook_in_featuregates.yaml
+#- patches/webhook_in_features.yaml
+>>>>>>> e8bc12f1 (Adds Featuregate,Feature API to core.tanzu.vmware.com and FeatureGate controller)
```
danniel1205, line 35: change required: Let's clean this up.

config/crd/kustomization.yaml
```diff
@@ -39,7 +46,12 @@ patchesStrategicMerge:
 #- patches/cainjection_in_clusterbootstraps.yaml
 #- patches/cainjection_in_clusterbootstraptemplates.yaml
 #- patches/cainjection_in_awsebscsiconfigs.yaml
+<<<<<<< HEAD
 #- patches/cainjection_in_azurefilecsiconfigs.yaml
+=======
+#- patches/cainjection_in_featuregates.yaml
+#- patches/cainjection_in_features.yaml
+>>>>>>> e8bc12f1 (Adds Featuregate,Feature API to core.tanzu.vmware.com and FeatureGate controller)
```
danniel1205, line 54: ditto.

featuregates/client/pkg/util/featuregate.go
danniel1205: @yharish991 , do we have an GH issue or some other tracking system for this ? If not, let's create one. Thanks

util/cmp/compare.go
danniel1205: Lets update to 2022

util/cmp/compare_test.go
danniel1205: change to 2022 ?

util/cmp/strings/doc.go
```diff
@@ -0,0 +1,5 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
```
danniel1205, line 1: change to 2022 ?

featuregates/client/pkg/util/featuregate.go
yharish991: i created an issue for this in our internal tracking system

apis/core/v1alpha2/featuregate_webhook.go
```diff
@@ -0,0 +1,330 @@
...
+	k8sscheme "k8s.io/client-go/kubernetes/scheme"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+// log is for logging in this package.
+var featuregatelog = logf.Log.WithName("featuregate-resource").WithValues("apigroup", "core")
```
danniel1205, line 24: nit: not a blocked, but I am curious on what is the naming convention for `WithName` ?

featuregates/controller/pkg/feature/feature_controller.go
danniel1205: Could we file a GH issue or JIRA for this ?

tkg/tkgconfigpaths/zz_bundled_default_bom_files_configdata.go
```diff
@@ -12,5 +12,5 @@ package tkgconfigpaths
 var (
 	TKGDefaultImageRepo               string = "projects-stg.registry.vmware.com/tkg"
 	TKGDefaultCompatibilityImagePath  string = "framework-zshippable/tkg-compatibility"
-	TKGManagementClusterPluginVersion string = "v0.28.0-dev"
+	TKGManagementClusterPluginVersion string = "v0.7.0-dev"
```
danniel1205, line 15: question: why this change ?

yharish991: this one somehow slipped through, i reverted this change

danniel1205 (approved):
Adding my approval because current change looks good to me. However, @yharish991 still needs to fix the linter failure and build failure before doing the squash merge.

yharish991:
> Adding my approval because current change looks good to me. However, @yharish991 still needs to fix the linter failure and build failure before doing the squash merge.

the linter failure is not in our package, I will just suppress it now using `//nolint` and let the appropriate team know to make the actual fix

cmd/cli/plugin-admin/codegen/generators/feature/fakeData/mykind_types.go
```diff
@@ -14,7 +14,7 @@ type MyKindSpec struct {
 type MyKindStatus struct {
 }
 
-//+tanzu:feature:name=foo,immutable=false,activated=false,discoverable=false,maturity=dev
+//+tanzu:feature:name=foo,stability=Technical Preview
```
danniel1205, line 17: question: do we need a quote for `Technical Preview` ?

pkg/v1/tkr/pkg/registry/client.go
```diff
@@ -77,7 +77,7 @@ func getFileContentFromImage(image regv1.Image, filename string) ([]byte, error)
 			return nil, err
 		}
 
-		defer layerStream.Close()
+		defer layerStream.Close() //nolint:gocritic
```
danniel1205, line 80: question: why do we want to `nolint:gocritic` here ? If there is really something that is "won't fix", could you also add some comments to provide some details ?

cmd/cli/plugin-admin/codegen/generators/feature/fakeData/mykind_types.go
yharish991: shouldn't matter, works same with or without quote

pkg/v1/tkr/pkg/registry/client.go
yharish991: i did not want to fix something that is not part of this pr and break things, also, it's not our code, but the linter was complaining.. so i just added `nolint:gocricitc` to suppress the linter. I was planning to let the tkr team know about this so that they can make appropriate fix

yharish991: add comment and created a gh issue to track this https://github.com/vmware-tanzu/tanzu-framework/issues/3917

yharish991:
//usebom:sandbox/v1.7.0-zshippable/436568906048479804/tkg-compatibility:v12

codegold79 (approved):

PR #3637, opened 2022-10-12
Add Debian package build script

marckhouzam:

This PR adds a script to build a Debian package to install the Tanzu CLI.

Part of #3 (wow third issue!)
Other package managers have to be included before we can completely resolve this issue.

```
$ cd tanzu-framework
$ cd cli/core
$ VERSION=v0.26.0 make apt-package
```

I then uploaded the built `_output/apt` directory to our `tanzu-cli-os-packages` bucket on GCloud.
Then:
```
$ docker run --rm -it ubuntu

$ tanzu
bash: tanzu: command not found

# Notice tanzu is not installed

$ apt update
[...]
$ apt install -y ca-certificates
[...]
$ echo "deb https://storage.googleapis.com/tanzu-cli-os-packages/apt jessie main" | tee /etc/apt/sources.list.d/tanzu.list
[...]
$ apt update --allow-insecure-repositories
[...]
$ apt install -y tanzu-cli --allow-unauthenticated
[...]

$ tanzu
tanzu
Tanzu CLI

Usage:
  tanzu [command]
[...]

# Notice tanzu is now installed
```
### Additional information

This PR allows to build the Debian packages and repository.  Publishing it to GCloud is currently done manually.

In future PRs we will need to:
1- Signing the repo
2- Figuring out if and how to deal with different versions of Debian/Ubuntu, currently we hard-code "Jessie"
3- when available, use an ARM64 build of the CLI for the ARM64 package
4- setup shell completion (if possible)

All changes are within the `cli/core` directory in preparation to the move to the new Github repo.


cli/core/hack/apt/build_package.sh
```diff
@@ -0,0 +1,41 @@
...
+BASE_DIR=$(cd $(dirname "${BASH_SOURCE[0]}")/..; pwd)
+OUTPUT_DIR=${BASE_DIR}/apt/_output
+
+# Install build dependencies
+apt-get update
+apt-get install -y dpkg-dev curl
+
+# Copy the tanzu binary to the package directory
+mkdir -p ${OUTPUT_DIR}/tanzu-cli_${VERSION}_linux_amd64/usr/bin
+cp ${BASE_DIR}/../../../artifacts/linux/amd64/cli/core/v${VERSION}/tanzu-core-linux_amd64 ${OUTPUT_DIR}/tanzu-cli_${VERSION}_linux_amd64/usr/bin/tanzu
```
randomvariable, line 28: How hard would it be to make this multi-arch?
Eventually we will want Graviton support for example.

chandrareddyp:
LGTM

cli/core/hack/apt/build_package.sh
```diff
@@ -0,0 +1,41 @@
...
+echo "Package: tanzu-cli
+Version: ${VERSION}
+Maintainer: Tanzu CLI project team
+Architecture: all
+Homepage: https://github.com/vmware-tanzu/tanzu-cli/
+Description: The tanzu CLI" \
+      > ${OUTPUT_DIR}/tanzu-cli_${VERSION}_linux_amd64/DEBIAN/control
+
+# Create the .deb package
+dpkg --build ${OUTPUT_DIR}/tanzu-cli_${VERSION}_linux_amd64
```
anujc25, line 41: Currently, we only support `amd64` arch. But in future we might have a requirement to support `arm` as well. 
Should we make this script more generic to support different architectures?

anujc25 (commented):
Overall looks good. Just one comment.

cli/core/hack/apt/build_package.sh
marckhouzam: Good point.  I'll see what I can do.

marckhouzam:
I think I made a mistake in building the CLI and then include it in the package.  To follow VMware's secure build process, we should use the artifacts that are published on github.  I will update the PR accordingly.

anujc25:
> I think I made a mistake in building the CLI and then include it in the package. To follow VMware's secure build process, we should use the artifacts that are published on github. I will update the PR accordingly.

Considering as part of the release artifacts we would like to publish SRP compliance core binary as well as packages containing that binary, should we still be using something what you have at the moment But decouple the build of binary and inject the binary (which can be a binary build with SRP compliant build) and use the existing script to generate the package?

anujc25:
> I think I made a mistake in building the CLI and then include it in the package. To follow VMware's secure build process, we should use the artifacts that are published on github. I will update the PR accordingly.

Considering as part of the release artifacts we would like to publish SRP compliance core binary as well as packages containing that binary, should we still be using something what you have at the moment But decouple the build of binary and inject the binary (which can be a binary build with SRP compliant build) and use the existing script to generate the package?

marckhouzam:
> > I think I made a mistake in building the CLI and then include it in the package. To follow VMware's secure build process, we should use the artifacts that are published on github. I will update the PR accordingly.
> 
> Considering as part of the release artifacts we would like to publish SRP compliance core binary as well as packages containing that binary, should we still be using something what you have at the moment But decouple the build of binary and inject the binary (which can be a binary build with SRP compliant build) and use the existing script to generate the package?

The latest forced push downloads the asset from Github and uses that in the debian package.
Notice that in the Makefile changes (`apt-package` target) I have hard-coded the version to `0.26.0`.  I will be able to improve this once we add some automation to building the OS packages.

I still have to improve things to address the multi-architecture suggested by Anuj.

marckhouzam:
I've converted this PR to a draft as it has expanded in scope.
The latest version builds a full repository that can be uploaded to a GCloud bucket to allow for public installation.

Things missing:
1- Signing the repo
2- Figuring out if and how to deal with different versions of Debian/Ubuntu, currently we hard-code "Jessie"
3- when available, use an ARM64 build of the CLI for the ARM64 package

marckhouzam:
> LGTM nit: I believe we should document the installation doc; I don't see any CLI installation instructions under https://github.com/vmware-tanzu/tanzu-framework/tree/main/cli/core/docs; it would be great if we add them as part of this story or better create a new story for the same.

I suggest that the official documentation for installation be done once we have worked out some final improvements, like signing the repo.

marckhouzam:
This is ready for review.  It provides a way to build the debian packages and repository; the repository can then be uploaded manually to GCloud, which in turn allows to install the CLI using APT.  IMO this is good enough for an MVP.

> Things missing:
> 1- Signing the repo
> 2- Figuring out if and how to deal with different versions of Debian/Ubuntu, currently we hard-code "Jessie"
> 3- when available, use an ARM64 build of the CLI for the ARM64 package
> 4- setup shell completion (if possible)

This will be done in following PRs.

cli/core/hack/apt/README.md
```diff
@@ -0,0 +1,59 @@
+# Using APT to install the Tanzu CLI
```
vuil, line 1: Nice docs!

cli/core/hack/apt/README.md
```diff
@@ -0,0 +1,59 @@
...
+Currently, the repo is not signed but will be in the future; you may get warnings.
+To install from an insecure repo:
+
+```bash
+$ docker run --rm -it ubuntu
+apt update
+apt install -y ca-certificates
+echo "deb https://storage.googleapis.com/tanzu-cli-os-packages/apt jessie main" | tee /etc/apt/sources.list.d/tanzu.list
+apt update --allow-insecure-repositories
+apt install -y tanzu-cli --allow-unauthenticated
```
anujc25, line 58: nit: What is our long term goal around using unauthenticated registries? Are we planning to use some public signing key approach similar to `kubectl` and other tools?

anujc25 (approved):
Overall changes looks good. Nice explanation and documentation.

cli/core/hack/apt/README.md
marckhouzam: Absolutely. It's important. That's one of the next improvements that have to be made.

cli/core/Makefile
```diff
@@ -158,6 +158,20 @@ release-%: ## Create release for a platform
...
+.PHONY: apt-package
+apt-package: ## Build a debian package to use with APT
+	@if [ "$$(command -v docker)" == "" ]; then \
+		echo "Docker required to build apt package" ;\
+		exit 1 ;\
+	fi
+
+	@# Until we have things automated, hard-code the version which must be available as a release on github
+	@# This can be overridden by specifying the version of the command-line. E.g., VERSION=v0.25.1 make apt-package
+	docker run --rm -e VERSION=$${VERSION-v0.26.0} -v $(ROOT_DIR):$(ROOT_DIR) ubuntu $(MODULE_ROOT_DIR)/hack/apt/build_package.sh
```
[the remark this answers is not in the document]
marckhouzam: > question: would the automation involve building this target with as part of producing a github release?

Not sure yet to be honest. We need an SRP-compliant build which we currently (IIUC) manually upload to GitHub; that's why this script downloads the version from GitHub instead of building it itself.


> nit: is defaulting to v0.26.0 necessary? Even as a temporary step it seems rather arbitrary.

You mean to instead require the version to be specified explicitly when calling the script?  Sure, I can do that.  The script will eventually be called by an automation which will specify the version, so you're right we don't need a default.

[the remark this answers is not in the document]
vuil: >that's why this script downloads the version from GitHub instead of building it itself
yeah the question is definitely tied to this point, but something to consider in the future. Seems like if some process is producing the binaries needed for the debian package build, it would be useful to extend said process to include the packaging work too.

>You mean to instead require the version to be specified explicitly when calling the script? 
Yep that's what mean.

cli/core/hack/apt/README.md
anujc25: Thanks. 👍

cli/core/Makefile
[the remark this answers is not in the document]
marckhouzam: > something to consider in the future. Seems like if some process is producing the binaries needed for the debian package build, it would be useful to extend said process to include the packaging work too.

Agreed this would be a nice improvement.

PR #3639, opened 2022-10-12
Remove `BOMRepo` and `CompatibilityFilePath` dependency from ClientConfig

anujc25:

For `BomRepo` and `CompatibilityFilePath` fields in ClientConfig, the `cluster` and `management-cluster` plugins are not reading on this information from ClientConfig. However, plugins are configuring this fields when invoking the commands.

Reference: As part of PR [#3264](https://github.com/vmware-tanzu/tanzu-framework/pull/3264/files#diff-e5b439d2272edfba1b423c302d8f167b9ec6462a706f3ab57e43db84b11ec165L142-R199) we removed the dependency of plugins on ClientConfig.

Also, core CLI is always making sure that this fields are configured in the ClientConfig to support older plugins.

So, This PR removes the `SetCompatibilityFileBasedOnEdition` invocation as part of plugin code and marks fields as Deprecated. 
 
Related to #3149

### Release note

```release-note
Remove `BOMRepo` and `CompatibilityFilePath` dependency from ClientConfig
```


cli/runtime/apis/config/v1alpha1/clientconfig_types.go
```diff
@@ -160,6 +160,7 @@ type CLIOptions struct {
 	DiscoverySources []PluginDiscovery `json:"discoverySources,omitempty" yaml:"discoverySources"`
 	// UnstableVersionSelector determined which version tags are allowed
 	UnstableVersionSelector VersionSelectorLevel `json:"unstableVersionSelector,omitempty" yaml:"unstableVersionSelector"`
+	// Deprecated: BOMRepo has been deprecated and will be removed from future version
```
tenczar, line 163: I don't think this belongs here. It is duplicated on line 166 where it actually belongs

tenczar (approved):
One small comment, but sounds good to me.

cli/runtime/apis/config/v1alpha1/clientconfig_types.go
anujc25: Thanks for catching this. Updated the PR.

PR #3652, opened 2022-10-13
Removes requirement adddon configs match cluster name

adduarte:
allowing the name of addons config to be anything

controllers changed in this patch

- Calico CNI
- Antrea CNI
- CSI
- CPI
- Kapp Controller

scenario logic:
- no customization
  user does nothing
  **addonsNS/config-template**   is cloned to **clusterNS/{CLUSTER_NAME}-{pacakge-short-name}-package**
  **CLUSTER_NAME** will be added  the owner references of **clusterNS/{CLUSTER_NAME}-{pacakge-short-name}-package**
   data value secrets is created

- customize based on template:
   create clusterNS/addonconfig withname **clusterNS/{CLUSTER_NAMEW}-{package-short-name}-pacakge**
   then
   **clusterNS/{CLUSTER_NAME}-{pacakge-short-name}-package += addonsNS/config-template** 

    The resulting addon config is a merged of template with user defined config addon
    The values from user defined addon config take precedence over the values from template config
    and **CLUSTER_NAME** will be added  the owner references of **clusterNS/{CLUSTER_NAME}-{pacakge-short-name}- 
    package**
    data value secrets is created

-  customized  based custom clusterboostrap and custom addonconfig:
   create clusterNS/addonconfig withname **clusterNS/{Any-name-for-addon-config}**
   webhook adds all undefined necessary fields.
   create clusterbootsrap **clusterNS/{CLUSTER_NAME}** pointing to **clusterNS/{Any-name-for-addon-config}**
   **CLUSTER_NAME** will be added  the owner references of **clusterNS/{Any-name-for-addon-config}**
   data value secrets is created

-  customized  based custom clusterboostrap and custom addonconfig with wildcard:
   create clusterNS/addonconfig withname **clusterNS/{Any-name-for-addon-config}**
   webhook adds all undefined necessary fields.
   create clusterbootsrap **clusterNS/{CLUSTER_NAME}** pointing to **clusterNS/{Any-name-for-addon-config}**
   **CLUSTER_NAME** will be added  the owner references of **clusterNS/{Any-name-for-addon-config}**
   data value secrets is created

Fixes #

scenarios described have been added as tests using eventest library
unit tests defined

### Release note

```release-note
Remove requirements that user defined addon config match cluster name. 
Addon configs can be based on addonConfig templates plus user defined values by using predefined name for addon config
User defined addon configs must be used in conjunction with user defined clusterboostrap

```

### DIFF
--- a/addons/controllers/antrea/antreaconfig_controller.go
+++ b/addons/controllers/antrea/antreaconfig_controller.go
@@ -24,6 +24,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	cutil "github.com/vmware-tanzu/tanzu-framework/addons/controllers/utils"
 	addonconfig "github.com/vmware-tanzu/tanzu-framework/addons/pkg/config"
 	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/constants"
 	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/util"
@@ -64,28 +65,24 @@ func (r *AntreaConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, err
 	}
 
+	annotations := antreaConfig.GetAnnotations()
+	if _, ok := annotations[constants.TKGAnnotationTemplateConfig]; ok {
+		log.Info(fmt.Sprintf("resource '%v' is a config template. Skipping reconciling", req.NamespacedName))
+		return ctrl.Result{}, nil
+	}
+
 	// deep copy AntreaConfig to avoid issues if in the future other controllers where interacting with the same copy
 	antreaConfig = antreaConfig.DeepCopy()
 
-	// get the parent cluster name from owner reference
-	// if the owner reference doesn't exist, use the same name as config CRD
-	clusterNamespacedName := req.NamespacedName
-	cluster := &clusterapiv1beta1.Cluster{}
-	for _, owner := range antreaConfig.OwnerReferences {
-		if owner.Kind == constants.ClusterKind {
-			clusterNamespacedName.Name = owner.Name
-			break
-		}
-	}
+	cluster, err := cutil.GetOwnerCluster(ctx, r.Client, antreaConfig, req.Namespace, constants.AntreaDefaultRefName)
 
-	// verify that the cluster related to config is present
-	if err := r.Client.Get(ctx, clusterNamespacedName, cluster); err != nil {
-		if apierrors.IsNotFound(err) {
-			log.Info(fmt.Sprintf("Cluster '%s'not found in '%s'", clusterNamespacedName.Name, clusterNamespacedName.Namespace))
+	if err != nil {
+		if apierrors.IsNotFound(err) && cluster != nil {
+			log.Info(fmt.Sprintf("'%s/%s' is listed as owner reference but could not be found",
+				cluster.Namespace, cluster.Name))
 			return ctrl.Result{}, nil
 		}
-
-		log.Error(err, fmt.Sprintf("unable to fetch cluster '%s' in '%s'", clusterNamespacedName.Name, clusterNamespacedName.Namespace))
+		log.Error(err, "could not determine owner cluster")
 		return ctrl.Result{}, err
 	}
 
--- a/addons/controllers/antreaconfig_controller_test.go
+++ b/addons/controllers/antreaconfig_controller_test.go
@@ -4,30 +4,44 @@
 package controllers
 
 import (
+	"fmt"
 	"os"
 	"strings"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clusterapiv1beta1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	cutil "github.com/vmware-tanzu/tanzu-framework/addons/controllers/utils"
 	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/constants"
 	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/util"
 	"github.com/vmware-tanzu/tanzu-framework/addons/test/testutil"
 	cniv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/addonconfigs/cni/v1alpha1"
+	runtanzuv1alpha3 "github.com/vmware-tanzu/tanzu-framework/apis/run/v1alpha3"
 )
 
 var _ = Describe("AntreaConfig Reconciler and Webhooks", func() {
 	var (
+		clusterName             string
+		clusterNamespace        string
 		configCRName            string
+		configName              string
 		clusterResourceFilePath string
 		err                     error
 		f                       *os.File
+		tkrString               string
+		newDefaultMTU           string
+		clusterBootstrap        *runtanzuv1alpha3.ClusterBootstrap
+		clusterBoostrapFile     string
+		clusterInfraName        string
 	)
 
 	const (
+		waitTimeout                            = waitTimeout //use this to change test speed when debugging
 		antreaManifestsTestFile1               = "testdata/antrea-test-1.yaml"
 		antreaTemplateConfigManifestsTestFile1 = "testdata/antrea-test-template-config-1.yaml"
 		antreaTestCluster1                     = "test-cluster-4"
@@ -74,26 +88,32 @@ var _ = Describe("AntreaConfig Reconciler and Webhooks", func() {
 	Context("Reconcile AntreaConfig for management cluster", func() {
 
 		BeforeEach(func() {
-			configCRName = antreaTestCluster1
+			clusterName = antreaTestCluster1
+			clusterNamespace = defaultString
+			configCRName = util.GeneratePackageSecretName(clusterName, constants.AntreaDefaultRefName)
 			clusterResourceFilePath = antreaManifestsTestFile1
 		})
 
 		It("Should reconcile AntreaConfig and create data value secret on management cluster", func() {
 
-			key := client.ObjectKey{
-				Namespace: "default",
+			clusterKey := client.ObjectKey{
+				Namespace: clusterNamespace,
+				Name:      clusterName,
+			}
+			configKey := client.ObjectKey{
+				Namespace: clusterNamespace,
 				Name:      configCRName,
 			}
 
 			cluster := &clusterapiv1beta1.Cluster{}
 			Eventually(func() bool {
-				err := k8sClient.Get(ctx, key, cluster)
+				err := k8sClient.Get(ctx, clusterKey, cluster)
 				return err == nil
 			}, waitTimeout, pollingInterval).Should(BeTrue())
 
 			config := &cniv1alpha1.AntreaConfig{}
 			Eventually(func() bool {
-				err := k8sClient.Get(ctx, key, config)
+				err := k8sClient.Get(ctx, configKey, config)
 				if err != nil {
 					return false
 				}
@@ -104,7 +124,7 @@ var _ = Describe("AntreaConfig Reconciler and Webhooks", func() {
 				}
 
 				Expect(len(config.OwnerReferences)).Should(Equal(1))
-				Expect(config.OwnerReferences[0].Name).Should(Equal(configCRName))
+				Expect(config.OwnerReferences[0].Name).Should(Equal(clusterName))
 
 				Expect(config.Spec.Antrea.AntreaConfigDataValue.TrafficEncapMode).Should(Equal("encap"))
 				Expect(config.Spec.Antrea.AntreaConfigDataValue.FeatureGates.AntreaTraceflow).Should(Equal(false))
@@ -119,7 +139,7 @@ var _ = Describe("AntreaConfig Reconciler and Webhooks", func() {
 
 			Eventually(func() bool {
 				cluster := &clusterapiv1beta1.Cluster{}
-				err := k8sClient.Get(ctx, key, cluster)
+				err := k8sClient.Get(ctx, clusterKey, cluster)
 				if err != nil {
 					return false
 				}
@@ -145,8 +165,8 @@ var _ = Describe("AntreaConfig Reconciler and Webhooks", func() {
 
 			Eventually(func() bool {
 				secretKey := client.ObjectKey{
-					Namespace: "default",
-					Name:      util.GenerateDataValueSecretName(configCRName, constants.AntreaAddonName),
+					Namespace: clusterNamespace,
+					Name:      util.GenerateDataValueSecretName(clusterName, constants.AntreaAddonName),
 				}
 				secret := &v1.Secret{}
 				err := k8sClient.Get(ctx, secretKey, secret)
@@ -173,12 +193,11 @@ var _ = Describe("AntreaConfig Reconciler and Webhooks", func() {
 			Eventually(func() bool {
 				// Check status.secretRef after reconciliation
 				config := &cniv1alpha1.AntreaConfig{}
-				err := k8sClient.Get(ctx, key, config)
+				err := k8sClient.Get(ctx, configKey, config)
 				if err != nil {
 					return false
 				}
-				Expect(config.Status.SecretRef).Should(Equal(util.GenerateDataValueSecretName(configCRName, constants.AntreaAddonName)))
-				return true
+				return config.Status.SecretRef == util.GenerateDataValueSecretName(clusterName, constants.AntreaAddonName)
 			}, waitTimeout, pollingInterval).Should(BeTrue())
 
 		})
@@ -188,7 +207,8 @@ var _ = Describe("AntreaConfig Reconciler and Webhooks", func() {
 	Context("Reconcile AntreaConfig used as template", func() {
 
 		BeforeEach(func() {
-			configCRName = antreaTestCluster1
+			clusterName = antreaTestCluster1
+			configCRName = util.GeneratePackageSecretName(clusterName, constants.AntreaDefaultRefName)
 			clusterResourceFilePath = antreaTemplateConfigManifestsTestFile1
 		})
 
@@ -209,14 +229,16 @@ var _ = Describe("AntreaConfig Reconciler and Webhooks", func() {
 	Context("Mutating webhooks for AntreaConfig", func() {
 
 		BeforeEach(func() {
-			configCRName = antreaTestCluster1
+			clusterName = antreaTestCluster1
+			clusterNamespace = defaultString
+			configCRName = util.GeneratePackageSecretName(clusterName, constants.AntreaDefaultRefName)
 			clusterResourceFilePath = antreaManifestsTestFile1
 		})
 
 		It("Should fail mutating webhooks for immutable field for AntreaConfig", func() {
 
 			key := client.ObjectKey{
-				Namespace: "default",
+				Namespace: clusterNamespace,
 				Name:      configCRName,
 			}
 			config := &cniv1alpha1.AntreaConfig{}
@@ -227,4 +249,211 @@ var _ = Describe("AntreaConfig Reconciler and Webhooks", func() {
 			Expect(k8sClient.Update(ctx, config)).ToNot(Succeed())
 		})
 	})
+
+	When("antrea config is precreated with name {CLUSTER_NAME}-{package-short-name}-package", func() {
+
+		BeforeEach(func() {
+			clusterName = "antrea-custom-cb-cluster"
+			clusterNamespace = "antrea-custom-cb-ns"
+			configName = util.GeneratePackageSecretName(clusterName, constants.AntreaDefaultRefName)
+			tkrString = "v1.23.1"
+			clusterResourceFilePath = "testdata/antrea-custom-tkg-system.yaml"
+			clusterInfraName = "custom-cb-docker-cluster"
+			newDefaultMTU = "8901"
+			ns := &v1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: clusterNamespace,
+				},
+			}
+			err := k8sClient.Create(ctx, ns)
+			if err != nil {
+				Expect(apierrors.IsAlreadyExists(err)).To(BeTrue())
+			}
+
+		})
+
+		It("should add cluster as owner reference to antrea config", func() {
+			// define custom resources in cluster namespace
+			f, err := os.Open("testdata/antrea-custom-cb-ns.yaml")
+			Expect(err).ToNot(HaveOccurred())
+			err = testutil.CreateResources(f, cfg, dynamicClient)
+			Expect(err).ToNot(HaveOccurred())
+			f.Close()
+
+			By("Create antrea config in the cluster's namespace with expected name pattern", func() {
+				datavalues := &cniv1alpha1.AntreaConfigDataValue{DefaultMTU: newDefaultMTU}
+				antreaConfig := genearateAntreaConfig(configName, clusterNamespace, datavalues)
+				err := k8sClient.Create(ctx, antreaConfig)
+				Expect(err).ToNot(HaveOccurred())
+				err = k8sClient.Get(ctx, client.ObjectKeyFromObject(antreaConfig), antreaConfig)
+				Expect(err).ToNot(HaveOccurred())
+			})
+
+			By("Create the cluster", func() {
+				cluster := generateDockerCluster(clusterName, clusterNamespace, tkrString, clusterInfraName)
+				err := k8sClient.Create(ctx, cluster)
+				Expect(err).ToNot(HaveOccurred())
+				err = k8sClient.Get(ctx, client.ObjectKeyFromObject(cluster), cluster)
+				Expect(err).ToNot(HaveOccurred())
+
+			})
+
+			By("Verify after reconciliation clusterBootstrap ProviderRef.Name points to pre-created antrea config", func() {
+				clusterBootstrap = &runtanzuv1alpha3.ClusterBootstrap{}
+				Eventually(func() bool {
+					err := k8sClient.Get(ctx, client.ObjectKey{Namespace: clusterNamespace, Name: clusterName}, clusterBootstrap)
+					return err == nil
+				}, waitTimeout, pollingInterval).Should(BeTrue())
+
+				Eventually(func() bool {
+					return clusterBootstrap.Spec.CNI.ValuesFrom.ProviderRef.Name == configName
+
+				}, waitTimeout, pollingInterval).Should(BeTrue())
+			})
+
+			By("Verify contents of resulting antrea config", func() {
+				antreaConfig := &cniv1alpha1.AntreaConfig{}
+				key := client.ObjectKey{Name: clusterBootstrap.Spec.CNI.ValuesFrom.ProviderRef.Name, Namespace: clusterNamespace}
+				err := k8sClient.Get(ctx, key, antreaConfig)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(cutil.VerifyOwnerRef(antreaConfig, clusterName, constants.ClusterKind)).To(BeTrue())
+				Expect(antreaConfig.Spec.Antrea.AntreaConfigDataValue.DefaultMTU).Should(Equal(newDefaultMTU))
+			})
+
+		})
+
+	})
+
+	When("custom clusterbootstrap points to precreated antrea config", func() {
+		BeforeEach(func() {
+			clusterName = "custom-cb-2-cluster"
+			clusterNamespace = "custom-cb-2-namespace"
+			configName = "antrea-config-custom-2"
+			clusterBoostrapFile = "testdata/antrea-custom-cb.yaml"
+			clusterInfraName = "custom-cb-docker-cluster-2"
+			tkrString = "v1.23.2"
+			clusterResourceFilePath = "testdata/antrea-custom-cb-2-resources.yaml"
+			newDefaultMTU = "8902"
+			ns := &v1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: clusterNamespace,
+				},
+			}
+			err := k8sClient.Create(ctx, ns)
+			if err != nil {
+				Expect(apierrors.IsAlreadyExists(err)).To(BeTrue())
+			}
+
+		})
+		It("should add cluster as owner reference to precreated antrea config", func() {
+			By("check resources like the kapp-controller config have been created", func() {
+				kappConfig := &runtanzuv1alpha3.KappControllerConfig{}
+				kappConfigKey := client.ObjectKey{Name: "test-cluster-custom-cb-2-kapp-controller-config",
+					Namespace: addonNamespace}
+				err := k8sClient.Get(ctx, kappConfigKey, kappConfig)
+				Expect(err).ToNot(HaveOccurred())
+
+			})
+
+			By("Create antrea config in the cluster's namespace with random name", func() {
+				datavalues := &cniv1alpha1.AntreaConfigDataValue{DefaultMTU: newDefaultMTU}
+				antreaConfig := genearateAntreaConfig(configName, clusterNamespace, datavalues)
+				err := k8sClient.Create(ctx, antreaConfig)
+				Expect(err).ToNot(HaveOccurred())
+				err = k8sClient.Get(ctx, client.ObjectKeyFromObject(antreaConfig), antreaConfig)
+				Expect(err).ToNot(HaveOccurred())
+			})
+
+			By("Create custom clusterbootstrap in clusters namespace", func() {
+				f, err := os.Open(clusterBoostrapFile)
+				Expect(err).ToNot(HaveOccurred())
+				defer f.Close()
+				Expect(testutil.CreateResources(f, cfg, dynamicClient)).To(Succeed())
+
+			})
+
+			By("Create the cluster", func() {
+				cluster := generateDockerCluster(clusterName, clusterNamespace, tkrString, clusterInfraName)
+				err := k8sClient.Create(ctx, cluster)
+				Expect(err).ToNot(HaveOccurred())
+				err = k8sClient.Get(ctx, client.ObjectKeyFromObject(cluster), cluster)
+				Expect(err).ToNot(HaveOccurred())
+
+			})
+
+			By("Verify contents of resulting antrea config", func() {
+				// eventually the secret ref to the data values should be updated
+				antreaConfig := &cniv1alpha1.AntreaConfig{}
+				Eventually(func() error {
+					configKey := client.ObjectKey{
+						Namespace: clusterNamespace,
+						Name:      configName,
+					}
+					if err := k8sClient.Get(ctx, configKey, antreaConfig); err != nil {
+						return fmt.Errorf("failed to get antrea config '%v': '%v'", configKey, err)
+					}
+					if antreaConfig.Status.SecretRef == "" {
+						return fmt.Errorf("antrea config status not yet updated: %v", configKey)
+					}
+					Expect(antreaConfig.Status.SecretRef).To(Equal(fmt.Sprintf("%s-%s-data-values", clusterName, constants.AntreaAddonName)))
+					return nil
+				}, waitTimeout, pollingInterval).Should(Succeed())
+
+				antreaConfig = &cniv1alpha1.AntreaConfig{}
+				key := client.ObjectKey{Name: configName, Namespace: clusterNamespace}
+				err := k8sClient.Get(ctx, key, antreaConfig)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(cutil.VerifyOwnerRef(antreaConfig, clusterName, constants.ClusterKind)).To(BeTrue())
+				Expect(antreaConfig.Spec.Antrea.AntreaConfigDataValue.DefaultMTU).Should(Equal(newDefaultMTU))
+			})
+
+		})
+	})
 })
+
+func genearateAntreaConfig(name, namespace string, datavalues *cniv1alpha1.AntreaConfigDataValue) *cniv1alpha1.AntreaConfig {
+	config := &cniv1alpha1.AntreaConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		Spec: cniv1alpha1.AntreaConfigSpec{
+			Antrea: cniv1alpha1.Antrea{
+				AntreaConfigDataValue: *datavalues,
+			},
+		},
+	}
+	return config
+}
+
+func generateDockerCluster(name, namespace, tkr, infraName string) *clusterapiv1beta1.Cluster {
+	labels := map[string]string{}
+	labels["tkg.tanzu.vmware.com/cluster-name"] = name
+	labels["run.tanzu.vmware.com/tkr"] = tkr
+
+	pods := &clusterapiv1beta1.NetworkRanges{CIDRBlocks: []string{"192.168.0.0/16"}}
+	services := &clusterapiv1beta1.NetworkRanges{CIDRBlocks: []string{"192.168.0.0/16", "fd00:100:96::/48"}}
+	clusterNetworks := &clusterapiv1beta1.ClusterNetwork{
+		Services: services,
+		Pods:     pods,
+	}
+	clusterSpecs := clusterapiv1beta1.ClusterSpec{
+		ClusterNetwork: clusterNetworks,
+		InfrastructureRef: &v1.ObjectReference{
+			Kind:       "DockerCluster",
+			Name:       infraName,
+			APIVersion: "infrastructure.cluster.x-k8s.io/v1beta1",
+		}}
+
+	cluster := &clusterapiv1beta1.Cluster{
+		TypeMeta: metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels:    labels,
+		},
+		Spec:   clusterSpecs,
+		Status: clusterapiv1beta1.ClusterStatus{},
+	}
+	return cluster
+}
--- a/addons/controllers/calico/calicoconfig_controller.go
+++ b/addons/controllers/calico/calicoconfig_controller.go
@@ -25,6 +25,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	cutil "github.com/vmware-tanzu/tanzu-framework/addons/controllers/utils"
 	addonconfig "github.com/vmware-tanzu/tanzu-framework/addons/pkg/config"
 	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/constants"
 	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/util"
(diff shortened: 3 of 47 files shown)


addons/controllers/antrea/antreaconfig_controller.go
```diff
@@ -50,47 +51,42 @@ type AntreaConfigReconciler struct {
...
 
-		log.Error(err, fmt.Sprintf("unable to fetch cluster '%s' in '%s'", clusterNamespacedName.Name, clusterNamespacedName.Namespace))
-		return ctrl.Result{}, err
+	cluster, err := cutil.GetOwnerCluster(ctx, r.Client, addonConfig, req.Namespace, constants.AntreaDefaultRefName)
+
+	if err != nil {
+		if apierrors.IsNotFound(err) && cluster != nil {
+			log.Info(fmt.Sprintf("'%s/%s' is listed as owner reference but could not be found",
+				cluster.Name, cluster.Namespace))
+			return ctrl.Result{}, err // (TODO) adduarte check if return  ctrl.Result{}, err ot  ctrl.Result{}, nil is best
```
maralavi, line 82: typo in the comment ```ot```. For the comment itself, I think you should return ```nil``` for the NotFound case & ```err``` for all other errors

addons/controllers/antrea/antreaconfig_controller.go
```diff
@@ -50,47 +51,42 @@ type AntreaConfigReconciler struct {
...
+		}
+		log.Info(fmt.Sprintf("could not determine owner cluster"))
+		return ctrl.Result{}, nil
```
maralavi, line 85: I think you should return ```err``` for all other errors. We were doing that in the previous logic too

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -47,42 +48,38 @@ type CalicoConfigReconciler struct {
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
 func (r *CalicoConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	r.Log = r.Log.WithValues("CalicoConfig", req.NamespacedName)
+	log := r.Log.WithValues("CalicoConfig", req.NamespacedName)
```
maralavi, line 51: Any reason to make this change? Cause as far as I know, having this assignment to ```r.Log``` would add those values to the logger instance of the reconciler, for better tracking the log messages.

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -47,42 +48,38 @@ type CalicoConfigReconciler struct {
...
+	if err != nil {
+		if apierrors.IsNotFound(err) && cluster != nil {
+			log.Info(fmt.Sprintf("'%s/%s' is listed as owner reference but could not be found",
+				cluster.Name, cluster.Namespace))
+			return ctrl.Result{}, err //TODO(adduarte) test using nil instead
 		}
-		r.Log.Error(err, fmt.Sprintf("Unable to fetch cluster '%s/%s'", clusterNamespacedName.Namespace, clusterNamespacedName.Name))
-		return ctrl.Result{}, err
+		log.Info(fmt.Sprintf("could not determine owner cluster"))
+		return ctrl.Result{}, nil
```
maralavi, line 82: Same comment as above, generally would return ```nil``` for ```NotFound``` and ```err``` otherwise.

addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go
```diff
@@ -684,8 +684,23 @@ func (h *Helper) cloneProviderRef(
...
+		// Did not find provider in sourceNamespace so we look in cluster namespace
+		h.Logger.Info(fmt.Sprintf("%s/%s not found. looking for %s/%s instead", sourceNamespace, cbPkg.ValuesFrom.ProviderRef.Name, cluster.Namespace, cbPkg.ValuesFrom.ProviderRef.Name))
+		provider, err = h.DynamicClient.Resource(*gvr).Namespace(cluster.Namespace).Get(h.Ctx, cbPkg.ValuesFrom.ProviderRef.Name, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			h.Logger.Info(fmt.Sprintf("%s/%s not found.", cluster.Namespace, cbPkg.ValuesFrom.ProviderRef.Name))
+		}
+		if err == nil {
+			// found provider in cluster namespace thus we don't alter name listed in clusterbootstrap
+			newProviderName = cbPkg.ValuesFrom.ProviderRef.Name
+		}
```
maralavi, line 700: How errors other than ```NotFound``` for ```h.DynamicClient.Resource.Get()```? Those are not handled here

addons/controllers/antrea/antreaconfig_controller.go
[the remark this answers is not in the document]
adduarte: yes. I have been going back and forth. It works both ways. 
my last way of thinking was that if the  cluster is not listed in the ownerReference and does not match clustername-package , the config will have to be modified to match either one of those, which would trigger a reconciliation no? 

Lets keep  this open. I'll submit the next patch with return err.. see if it is better or creates any trouble.

addons/controllers/antrea/antreaconfig_controller.go
adduarte: will remove comment and return nil for not found and err for all others.

addons/controllers/antrea/antreaconfig_controller.go
adduarte: ack !

addons/controllers/calico/calicoconfig_controller.go
adduarte: you are right this is unecesary change. will revert

addons/controllers/calico/calicoconfig_controller.go
adduarte: ack

addons/controllers/antrea/antreaconfig_controller.go
adduarte: Set this to return nil on isNOtFound and err for all others.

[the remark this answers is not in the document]
adduarte: resolution:  return nil for isNotFound and return err for all others

addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go
adduarte: yes. It seems that way, but the logic actually drops an err != nil && apierros.iSNotFound(err) to line 702 
if err != nil, but I'll add an "else" to make it clear.

adduarte: hmm doing so just results in duplicate code. it looks something like 
```
	provider, err := h.DynamicClient.Resource....
	if apierrors.IsNotFound(err) && sourceNamespace ... {
		log message-1 here
		provider, err = h.DynamicClient.Resource...
			log message-2 here
		}
		if err != nil {.  <----- this is caught by the logic out of this loop below. 
			log message-3 here
			return nil, err
		}
		// found provider in cluster namespace, which means we should not mutate provider name.
		newProviderName = cbPkg.ValuesFrom.ProviderRef.Name
	}

       //at this point we only go on if err == nil 
	if err != nil {
		log message-3 here (again)
	}

Comment (adduarte) [reply addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go:700]: we'll go with this just to avoid more discussion. anyways the compiler will probably optimize this to the same resulting code

Comment (adduarte) [reply addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go:700]: fixed.

Comment (adduarte) [reply addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go:700]: turns out goimports does not like the fix. running golint we get 
```
pkg/util/clusterbootstrapclone/clusterbootstrapclone.go:670: Function 'cloneProviderRef' has too many statements (51 > 50) (funlen)
func (h *Helper) cloneProviderRef(
```

Looks like we have to go for compactnes over clarity in this case.

File: addons/controllers/antreaconfig_controller_test.go [hunk 6]
```diff
@@ -227,4 +254,212 @@ var _ = Describe("AntreaConfig Reconciler and Webhooks", func() {
 			Expect(k8sClient.Update(ctx, config)).ToNot(Succeed())
 		})
 	})
+
+	When("antrea config is precreated with {CLUSTER_NAME}-{package-short-name}-package", func() {
```
adduarte, line 258: "antrea config is precreated with name {CLUSTER_NAME}-{package-short-name}-package

addons/controllers/antreaconfig_controller_test.go
```diff
@@ -227,4 +254,212 @@ var _ = Describe("AntreaConfig Reconciler and Webhooks", func() {
...
+			err := k8sClient.Create(ctx, ns)
+			if err != nil {
+				Expect(apierrors.IsAlreadyExists(err)).To(BeTrue())
+			}
+
+		})
+
+		It("should add cluster as owner reference to antrea config", func() {
+			// define custom resources in cluster namespace
+			f, err := os.Open("testdata/antrea-custom-cb-ns.yaml")
```
adduarte, line 282: This file could be split into two, one with the current resources and the other file could be the section sited in the readme.md file. Thus we can test the example and make sure we don't break it

addons/controllers/antreaconfig_controller_test.go
```diff
@@ -227,4 +254,212 @@ var _ = Describe("AntreaConfig Reconciler and Webhooks", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			err = testutil.CreateResources(f, cfg, dynamicClient)
+			Expect(err).ToNot(HaveOccurred())
+			f.Close()
+
+			By("Create antrea config in the cluster's namespace with expected pattern", func() {
```
adduarte, line 288: expected name patter

addons/controllers/antreaconfig_controller_test.go
```diff
@@ -227,4 +254,212 @@ var _ = Describe("AntreaConfig Reconciler and Webhooks", func() {
...
+			By("Create the cluster", func() {
+				cluster := generateDockerCluster(clusterName, clusterNamespace, tkrString, clusterInfraName)
+				err := k8sClient.Create(ctx, cluster)
+				Expect(err).ToNot(HaveOccurred())
+				err = k8sClient.Get(ctx, client.ObjectKeyFromObject(cluster), cluster)
+				Expect(err).ToNot(HaveOccurred())
+
+			})
+
+			By("Verify clusterBootstrap points to pre-created antrea config", func() {
```
adduarte, line 306: Verify after reconciliation clusterBootstrap ProviderRef.Name  points to pre-created antrea config

addons/controllers/antreaconfig_controller_test.go
```diff
@@ -227,4 +254,212 @@ var _ = Describe("AntreaConfig Reconciler and Webhooks", func() {
...
+				},
+			}
+			err := k8sClient.Create(ctx, ns)
+			if err != nil {
+				Expect(apierrors.IsAlreadyExists(err)).To(BeTrue())
+			}
+
+		})
+		It("should add cluster as owner reference to precreated antrea config", func() {
+			By("lets make sure resources are available", func() {
```
adduarte, line 355: "check  resources like the kapp-controller config have been created.

addons/controllers/antreaconfig_controller_test.go
```diff
@@ -227,4 +254,212 @@ var _ = Describe("AntreaConfig Reconciler and Webhooks", func() {
...
+						Namespace: clusterNamespace,
+						Name:      configName,
+					}
+					if err := k8sClient.Get(ctx, configKey, antreaConfig); err != nil {
+						return fmt.Errorf("failed to get antrea config '%v': '%v'", configKey, err)
+					}
+					if antreaConfig.Status.SecretRef == "" {
+						return fmt.Errorf("antrea config status not yet updated: %v", configKey)
+					}
+					Expect(antreaConfig.Status.SecretRef).To(Equal(fmt.Sprintf("%s-%s-data-values", clusterName, constants.AntreaAddonName)))
```
adduarte, line 404: There should be no Expect inside an Eventually node. It shortcircuits the evenually.
also change the way the clusters  data value secret name is deduced.  It should use the util.Generate.... fucntions

addons/controllers/antreaconfig_controller_test.go
```diff
@@ -227,4 +254,212 @@ var _ = Describe("AntreaConfig Reconciler and Webhooks", func() {
...
+				Expect(cutil.VerifyOwnerRef(antreaConfig, clusterName, constants.ClusterKind)).To(BeTrue())
+				Expect(antreaConfig.Spec.Antrea.AntreaConfigDataValue.DefaultMTU).Should(Equal(newDefaultMTU))
+			})
+
+		})
+	})
 })
+
+func genearateAntreaConfig(name, namespace string, datavalues *cniv1alpha1.AntreaConfigDataValue) *cniv1alpha1.AntreaConfig {
+	config := &cniv1alpha1.AntreaConfig{
```
adduarte, line 421: This might be too confusing. Although it is easier to programatically control the tests, it would seem best to use yaml files and just load them appriately, and use a directory structure to organize

addons/controllers/kapp-controller/kappcontrollerconfig_controller.go
```diff
@@ -8,6 +8,8 @@ import (
 	"context"
 	"fmt"
 
+	cutil "github.com/vmware-tanzu/tanzu-framework/addons/controllers/utils"
+
```
adduarte, line 12: gratiuitous space

addons/controllers/testdata/test-tkg-system-ns-resources.yaml
```diff
@@ -174,6 +178,8 @@ kind: KappControllerConfig
 metadata:
   name: test-cluster-tcbt-kapp-controller-config
   namespace: tkg-system
+  annotations:
+    tkg.tanzu.vmware.com/template-config: "true"
```
adduarte, line 182: why were this not here before? was it just an omittion or are this supposed to be non-template configs. 
From the rest of the code it seem they are supposed to be so they need to have the  correct label

addons/controllers/utils/utils.go
```diff
@@ -134,3 +135,57 @@ func TryParseClusterVariableString(ctx context.Context, cluster *clusterapiv1bet
 	}
 	return res
 }
+
+// GetOwnerCluster returns cluster owner for the object.
+// Returns cluster=nil and err if no cluster can be determined as owner.
+// Returns cluster object with only name and namespace  and err,
+// if details for the cluster determined to be an woner can not be found.
```
adduarte, line 142: spelling. This commet will show oup in go docs. so fix it!!

addons/controllers/utils/utils.go
```diff
@@ -134,3 +135,57 @@ func TryParseClusterVariableString(ctx context.Context, cluster *clusterapiv1bet
...
+			clusterName = ownerRef.Name
+			break
+		}
+	}
+	if clusterName != "" {
+		cluster.Name = clusterName
+		cluster.Namespace = namespace
+	} else {
+		// Means no cluster listed
+		// look at object name and match against patern of {clustern-name}-{packagerefname}-package
```
adduarte, line 159: This commet (booooo ) is wrong. We don't check against the patter shown. We can check against whatever pattern GenratePackage will return which in this case is {clustern-name}-{packagerefname}-package

addons/controllers/utils/utils.go
```diff
@@ -134,3 +135,57 @@ func TryParseClusterVariableString(ctx context.Context, cluster *clusterapiv1bet
...
+		if possibleClusterName != "" {
+			cluster.Name = possibleClusterName
+			cluster.Namespace = k8sObject.GetNamespace()
+		} else {
+			errorString := "no owner cluster could be determined"
+			logger.Info(errorString)
+			err := errors.New(errorString)
+			return nil, err
+		}
+	}
```
adduarte, line 170: Do we also need to match against an existing clustername?  it could expand the options. 
the case it would handle is one where, cluster does not have owner refs, does not have a name matching pattern.... 

I think it might cause problems in the case it tries to reconcile a custom addons-config which has been genrated before the matching custom clusterbootstrap has been created.  
It could end up being a bug because if by coincidence a cluster with matching name exists, it would incorrectly assuem the new config belongs to it.

addons/controllers/vspherecpiconfig_controller_test.go
```diff
@@ -31,7 +31,7 @@ import (
 )
 
 const (
-	clusterNamespace            = "default"
+	defaultString               = "default"
```
adduarte, line 34: This has to be done because this package is in the controllers package scope, and lint complains about it being the string "default" floating around and a constant which is = "default". Also clusterNamespace is a variable used everywhere in this package, probably don't want to create a constant with that name at package level.

addons/controllers/utils/utils.go
adduarte: we only need to match on owner reference

addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go
```diff
@@ -405,15 +425,42 @@ func (h *Helper) HandleExistingClusterBootstrap(clusterBootstrap *runtanzuv1alph
 		return nil, err
 	}
 
-	// ensure ownerRef of clusterBootstrap on created secrets and providers, this can only be done after
-	// clusterBootstrap is created
+	providers = append(providers, providerRefs...)
```
shyaamsn, line 428: Wouldnt appending add duplicates ? Shouldnt we be using providerRefs instead ?

adduarte: yes! you are right... at this point provider is a subset of providerRef

adduarte: but I guess we don't care about providers with valuesfrom.inline, so I think you are correct we can get away with just providerRef

alfredthenarwhal:
Build failed! Build no: 3097

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -47,47 +48,45 @@ type CalicoConfigReconciler struct {
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
 func (r *CalicoConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	r.Log = r.Log.WithValues("CalicoConfig", req.NamespacedName)
+	log := r.Log.WithValues("CalicoConfig", req.NamespacedName)
```
shyaamsn, line 51: In other places r.Log is used. Since this is changed to log the other places where r.Log is used it is going to not log calicoconfig req.NamespacedName. It is important to know the name and namespace of calicoconfig object in the logs.

If this is unnecessary please keep it as r.Log

addons/controllers/calico/calicoconfig_controller.go
```diff
@@ -47,47 +48,45 @@ type CalicoConfigReconciler struct {
...
-		if apierrors.IsNotFound(err) {
-			r.Log.Info(fmt.Sprintf("Cluster resource '%s/%s' not found", clusterNamespacedName.Namespace, clusterNamespacedName.Name))
+	if err != nil {
+		if apierrors.IsNotFound(err) && cluster != nil {
+			log.Info(fmt.Sprintf("'%s/%s' is listed as owner reference but could not be found",
+				cluster.Namespace, cluster.Name))
 			return ctrl.Result{}, nil
 		}
-		r.Log.Error(err, fmt.Sprintf("Unable to fetch cluster '%s/%s'", clusterNamespacedName.Namespace, clusterNamespacedName.Name))
+		log.Info("could not determine owner cluster")
```
shyaamsn, line 83: Same comment as above. Please also log error object so it is clear what the error is

addons/controllers/csi/vspherecsiconfig_controller.go
```diff
@@ -160,10 +159,12 @@ func (r *VSphereCSIConfigReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	// deep copy VSphereCSIConfig to avoid issues if in the future other controllers where interacting with the same copy
 	vcsiConfig = vcsiConfig.DeepCopy()
-
-	cluster, err := r.getOwnerCluster(ctx, vcsiConfig)
+	cluster, err := cutil.GetOwnerCluster(ctx, r.Client, vcsiConfig, req.Namespace, constants.CSIDefaultRefName)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return ctrl.Result{}, err
+	}
 	if cluster == nil {
```
shyaamsn, line 166: What about the case where cluster is not nil and apierrors.IsNotFound ?

addons/controllers/kapp-controller/kappcontrollerconfig_controller.go
```diff
@@ -57,28 +58,24 @@ func (r *KappControllerConfigReconciler) Reconcile(ctx context.Context, req ctrl
...
+	cluster, err := cutil.GetOwnerCluster(ctx, r.Client, kappControllerConfig, req.Namespace, constants.KappControllerDefaultRefName)
+	if err != nil {
+		if apierrors.IsNotFound(err) && cluster != nil {
+			r.Log.Info(fmt.Sprintf("'%s/%s' is listed as owner reference but could not be found",
+				cluster.Namespace, cluster.Name))
 			return ctrl.Result{}, nil
 		}
-
-		log.Error(err, fmt.Sprintf("unable to fetch cluster %s/%s", clusterNamespacedName.Namespace, clusterNamespacedName.Name))
+		r.Log.Info("could not determine owner cluster")
```
shyaamsn, line 78: Please log the error message as well so its clear what error is

addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go
```diff
@@ -827,6 +872,51 @@ func (h *Helper) cloneEmbeddedLocalObjectRef(cluster *clusterapiv1beta1.Cluster,
...
+		err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			// We need to get and update, otherwise there could have concurrency issue: ["the object has been modified; please
+			// apply your changes to the latest version and try again"]
+			newChild, errGetProvider := h.DynamicClient.Resource(*gvr).Namespace(child.GetNamespace()).Get(h.Ctx, child.GetName(), metav1.GetOptions{})
+			if errGetProvider != nil {
+				h.Logger.Error(errGetProvider, fmt.Sprintf("unable to get %s %s/%s", child.GetKind(), child.GetNamespace(), child.GetName()))
+				return errGetProvider
+			}
+			newChild = newChild.DeepCopy()
+			newChild.SetOwnerReferences(clusterapiutil.EnsureOwnerRef(newChild.GetOwnerReferences(), *ownerRef))
```
shyaamsn, line 904: Before setting ownerreference can we make sure the provider doesnt have any other ownerreference belonging to a different cluster. If it has throw error. I am thinking instead of webhook we can live with this check for now.

addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go
```diff
@@ -379,14 +379,33 @@ func (h *Helper) HandleExistingClusterBootstrap(clusterBootstrap *runtanzuv1alph
 			packagesToBeCloned = append(packagesToBeCloned, nonEmptyInlinePackages...)
 			packages = packagesToBeCloned
 		}
+	} else {
+		// packages with values from inline need to be cloned.
```
shyaamsn, line 383: A comment with more details here explaining why this is done would be useful.

e.g. This else is hit in two cases:
1. When no Clusterbootstrap is given by user and clusterbootstraptemplate is cloned as clusterbootstrap but resolved tkr is not set - In this case the providers are already cloned so providers have to be cloned again
2. When full clusterbootstrap without tkr annotation is given by user - In this case we only want to handle inline packages for creating secret. The providers are expected to be present in the clusternamespace i.e. created by user before clusterbootstrap is created.

shyaamsn (changes requested):

addons/controllers/calico/calicoconfig_controller.go
adduarte: this causes a race condition Test failure. (reasigning r.log =r.log) if you note other  providers use  log = r.log (use log). will try again, but the build fails with race

addons/controllers/calico/calicoconfig_controller.go
adduarte: will do, but again, using r.log=r.log causes race condition that fails the tests.

addons/controllers/cpi/vspherecpiconfig_controller.go
```diff
@@ -98,7 +98,7 @@ func (r *VSphereCPIConfigReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	// deep copy VSphereCPIConfig to avoid issues if in the future other controllers where interacting with the same copy
 	cpiConfig = cpiConfig.DeepCopy()
 
-	cluster, err := r.getOwnerCluster(ctx, cpiConfig)
+	cluster, err := cutil.GetOwnerCluster(ctx, r.Client, cpiConfig, req.Namespace, constants.CPIDefaultRefName)
```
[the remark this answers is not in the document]
adduarte: fixed. Now check for error

addons/controllers/antrea/antreaconfig_controller.go
```diff
@@ -64,28 +65,24 @@ func (r *AntreaConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-			log.Info(fmt.Sprintf("Cluster '%s'not found in '%s'", clusterNamespacedName.Name, clusterNamespacedName.Namespace))
+	if err != nil {
+		if apierrors.IsNotFound(err) && cluster != nil {
+			log.Info(fmt.Sprintf("'%s/%s' is listed as owner reference but could not be found",
+				cluster.Namespace, cluster.Name))
 			return ctrl.Result{}, nil
 		}
-
-		log.Error(err, fmt.Sprintf("unable to fetch cluster '%s' in '%s'", clusterNamespacedName.Name, clusterNamespacedName.Namespace))
+		log.Info("could not determine owner cluster")
```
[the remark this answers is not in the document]
adduarte: logged error

addons/controllers/calico/calicoconfig_controller.go
adduarte: currently testing to see if we can get away with using r.log=r.log

addons/controllers/calico/calicoconfig_controller.go
adduarte: logged error

addons/controllers/calico/calicoconfig_controller.go
adduarte: yes this fails if we use r.Log = r.Log. 
see https://github.com/vmware-tanzu/tanzu-framework/actions/runs/3388165981/jobs/5629815922#step:15:1305
WARNING: DATA RACE
Write at 0x00c0012c8290 by goroutine 158:
  github.com/vmware-tanzu/tanzu-framework/addons/controllers/calico.(*CalicoConfigReconciler).Reconcile()
      /home/runner/work/tanzu-framework/tanzu-framework/addons/controllers/calico/calicoconfig_controller.go:51 +0x1cd
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile()
      /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:121 +0x16a
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler()
      /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:320 +0x488
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem()
      /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:273 +0x324
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2()
      /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:234 +0xc8

Previous read at 0x00c0012c8290 by goroutine 359:
  runtime.racereadrange()
      <autogenerated>:1 +0x1b
  github.com/vmware-tanzu/tanzu-framework/addons/controllers/calico.(*CalicoConfigReconciler).ClusterToCalicoConfig-fm()
      <autogenerated>:1 +0x4d
  sigs.k8s.io/controller-runtime/pkg/handler.(*enqueueRequestsFromMapFunc).mapAndEnqueue()
      /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/handler/enqueue_mapped.go:80 +0x84
  sigs.k8s.io/controller-runtime/pkg/handler.(*enqueueRequestsFromMapFunc).Create()
      /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/handler/enqueue_mapped.go:57 +0xd1
  sigs.k8s.io/controller-runtime/pkg/source/internal.EventHandler.OnAdd()
      /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/source/internal/eventsource.go:63 +0x311
  sigs.k8s.io/controller-runtime/pkg/source/internal.(*EventHandler).OnAdd()
      <autogenerated>:1 +0xd5
  k8s.io/client-go/tools/cache.(*processorListener).run.func1()
      /home/runner/go/pkg/mod/k8s.io/client-go@v0.24.6/tools/cache/shared_informer.go:818 +0x151
  k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1()
      /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.24.6/pkg/util/wait/wait.go:155 +0x48
  k8s.io/apimachinery/pkg/util/wait.BackoffUntil()
      /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.24.6/pkg/util/wait/wait.go:156 +0xce
  k8s.io/apimachinery/pkg/util/wait.JitterUntil()
      /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.24.6/pkg/util/wait/wait.go:133 +0x104
  k8s.io/apimachinery/pkg/util/wait.Until()
      /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.24.6/pkg/util/wait/wait.go:90 +0x78
  k8s.io/client-go/tools/cache.(*processorListener).run()
      /home/runner/go/pkg/mod/k8s.io/client-go@v0.24.6/tools/cache/shared_informer.go:812 +0x18
  k8s.io/client-go/tools/cache.(*processorListener).run-fm()
      <autogenerated>:1 +0x39
  k8s.io/apimachinery/pkg/util/wait.(*Group).Start.func1()
      /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.24.6/pkg/util/wait/wait.go:73 +0x73

Goroutine 158 (running) created at:
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2()
      /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:230 +0x4a4
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start()
      /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:241 +0x2a8
  sigs.k8s.io/controller-runtime/pkg/manager.(*runnableGroup).reconcile.func1()
      /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/manager/runnable_group.go:219 +0x1b8
  sigs.k8s.io/controller-runtime/pkg/manager.(*runnableGroup).reconcile.func2()
      /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/manager/runnable_group.go:222 +0x47

Goroutine 359 (running) created at:
  k8s.io/apimachinery/pkg/util/wait.(*Group).Start()
      /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.24.6/pkg/util/wait/wait.go:71 +0xdc
  k8s.io/client-go/tools/cache.(*sharedProcessor).addListener()
      /home/runner/go/pkg/mod/k8s.io/client-go@v0.24.6/tools/cache/shared_informer.go:634 +0x379
  k8s.io/client-go/tools/cache.(*sharedIndexInformer).AddEventHandlerWithResyncPeriod()
      /home/runner/go/pkg/mod/k8s.io/client-go@v0.24.6/tools/cache/shared_informer.go:560 +0x63d
  k8s.io/client-go/tools/cache.(*sharedIndexInformer).AddEventHandler()
      /home/runner/go/pkg/mod/k8s.io/client-go@v0.24.6/tools/cache/shared_informer.go:496 +0x54
  sigs.k8s.io/controller-runtime/pkg/source.(*Kind).Start.func1()
      /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/source/source.go:158 +0x310
==================

shyaamsn: Got it. Thanks for checking out

addons/controllers/antrea/antreaconfig_controller.go
```diff
@@ -64,28 +65,24 @@ func (r *AntreaConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
-	// verify that the cluster related to config is present
-	if err := r.Client.Get(ctx, clusterNamespacedName, cluster); err != nil {
-		if apierrors.IsNotFound(err) {
-			log.Info(fmt.Sprintf("Cluster '%s'not found in '%s'", clusterNamespacedName.Name, clusterNamespacedName.Namespace))
+	if err != nil {
+		if apierrors.IsNotFound(err) && cluster != nil {
+			log.Info(fmt.Sprintf("'%s/%s' is listed as owner reference but could not be found",
+				cluster.Namespace, cluster.Name))
 			return ctrl.Result{}, nil
```
adduarte, line 83: This should return ctr.Result{}., err to use backoff algorithm for requeueing. Check all other addonconfigs for concictency. and only then resolve this comment.

addons/controllers/cpi/vspherecpiconfig_controller.go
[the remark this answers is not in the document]
adduarte: return on all errors, and on cluster == nil. 
Only way forward is to have cluster !=nil and err == nil

addons/controllers/csi/vspherecsiconfig_controller.go
adduarte: added logic to handle

addons/controllers/kapp-controller/kappcontrollerconfig_controller.go
adduarte: log as error now with error message

addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go
adduarte: recreated list of providers with latest information from clusterbootstrap

addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go
adduarte: done

addons/controllers/antrea/antreaconfig_controller.go
adduarte: done

addons/controllers/antrea/antreaconfig_controller.go
```diff
@@ -64,31 +65,31 @@ func (r *AntreaConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			return ctrl.Result{}, err
 		}
-
-		log.Error(err, fmt.Sprintf("unable to fetch cluster '%s' in '%s'", clusterNamespacedName.Name, clusterNamespacedName.Namespace))
+		log.Error(err, "could not determine owner cluster")
 		return ctrl.Result{}, err
 	}
 
+	if cluster == nil {
+		return ctrl.Result{}, err // retry until corresponding cluster is found
```
shyaamsn, line 90: If cluster is nil will err be nil ? If err is nil anyways it will return from above if err != nil.

addons/controllers/cpi/vspherecpiconfig_controller.go
```diff
@@ -80,34 +80,45 @@ var vsphereCPIProviderServiceAccountAggregatedClusterRole = &rbacv1.ClusterRole{
...
+				cluster.Namespace, cluster.Name))
+			return ctrl.Result{}, err
+		}
+		logger.Error(err, "could not determine owner cluster")
+		return ctrl.Result{}, err
+	}
+
 	if cluster == nil {
-		return ctrl.Result{}, err // no need to requeue if cluster is not found
+		return ctrl.Result{}, err // retry until corresponding cluster is found
```
shyaamsn, line 113: Will error be there if cluster is nil ?

addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go
```diff
@@ -379,14 +379,42 @@ func (h *Helper) HandleExistingClusterBootstrap(clusterBootstrap *runtanzuv1alph
 			packagesToBeCloned = append(packagesToBeCloned, nonEmptyInlinePackages...)
 			packages = packagesToBeCloned
 		}
+	} else {
+		// packages with values from inline need to be cloned for two cases:
+		//
+		// 1. When no Clusterbootstrap is given by user and clusterbootstraptemplate
+		//    is cloned as clusterbootstrap but resolved tkr is not set - In this
+		//   case the providers are already cloned so providers have to be cloned again
```
shyaamsn, line 387: providers **dont** have to be cloned again.

addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go
```diff
@@ -379,14 +379,42 @@ func (h *Helper) HandleExistingClusterBootstrap(clusterBootstrap *runtanzuv1alph
...
 	}
 
-	secrets, providers, err := h.CloneReferencedObjectsFromCBPackages(cluster, packages, systemNamespace)
+	secrets, _, err := h.CloneReferencedObjectsFromCBPackages(cluster, packages, systemNamespace)
 	if err != nil {
 		h.Logger.Error(err, "unable to clone secrets or providers")
 		return nil, err
 	}
 
+	providers, err := h.getListOfProvidersFromRef(clusterBootstrap)
```
shyaamsn, line 410: I thought about the case where in the prior release what if any object didnt have owner reference ? Then the new config controller wouldnt work. i.e. I think we are breaking backwards compatibility with this change.

To be safe I think we need to set ownerreference for all providers outside this function in the reconcileNormal function as well. i.e. In any case set the ownerreference for all providers.

shyaamsn (changes requested):

addons/controllers/antrea/antreaconfig_controller.go
```diff
@@ -64,31 +65,31 @@ func (r *AntreaConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-	// verify that the cluster related to config is present
-	if err := r.Client.Get(ctx, clusterNamespacedName, cluster); err != nil {
-		if apierrors.IsNotFound(err) {
-			log.Info(fmt.Sprintf("Cluster '%s'not found in '%s'", clusterNamespacedName.Name, clusterNamespacedName.Namespace))
-			return ctrl.Result{}, nil
+	if err != nil {
+		if apierrors.IsNotFound(err) && cluster != nil {
+			log.Info(fmt.Sprintf("'%s/%s' is listed as owner reference but could not be found",
+				cluster.Namespace, cluster.Name))
+			return ctrl.Result{}, err
```
lubronzhan, line 83: The return err won't be necessary since we have enqueue function here 
https://github.com/vmware-tanzu/tanzu-framework/blob/main/addons/controllers/antrea/antreaconfig_util.go#L93
So it can also help avoid error spam in addon controller

adduarte: we want to return err, because if we return nil it won't be requed. i.e. returning ctrl.Result{}, nil means the request was reconciled and it will not be requed. returning error here ensured req goes back in the queue. If not it could take too long for the config to be requed again.

addons/controllers/cpi/vspherecpiconfig_controller.go
adduarte: yes. cutil.GetOwner aways returns an error with cluster=nil, but even without knowing the internals of cutil.GetOwnerCluster,  if cluster is nil && and err = nil (which is hard to think of a case that would happe), it would just flag the request as done which is what we want if err == nil .

addons/controllers/antrea/antreaconfig_controller.go
adduarte: There is a logical chance that we have cluster=nil and err=nil, can't think of one but it would happen if the function ```err := k8sClient.Get(ctx, client.ObjectKeyFromObject(cluster), cluster)`` returns cluster=nil and err = nil.  
Basically GetOwnerCluster has no way of determining what it means cluster=nil and err=nil so it just kicks it up (to here) and that situation is probably an error. Maybe we should handle it differently then (since err=nil).

addons/controllers/cpi/vspherecpiconfig_controller.go
adduarte: actully after looking at GetOwner description, and code, it will never return cluster=nil & err = nil 
the only time cluster =nil is when err != nil, so we can skip this altogehter.

addons/controllers/antrea/antreaconfig_controller.go
adduarte: GetOwnerCluster will never return cluster==nil and err == nil, thus this is dead code, will remove

addons/controllers/antrea/antreaconfig_controller.go
adduarte: so the reason for returning err here it to take advantage of backoff reque algorithm of an error. If we "return ctrl.Result{}, nil" won't that flag the request as reconciled (which is not the case here, we want to keep trying until we find a cluster owner.

adduarte: yes you are correct. will fix

adduarte: will change to return nil

addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go
adduarte: added check to see if any of the providers already have the cluster as owner

addons/controllers/calico/calicoconfig_controller.go
lubronzhan: Also run into this issue in other util func

addons/controllers/antrea/antreaconfig_controller.go
```diff
@@ -64,28 +65,24 @@ func (r *AntreaConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, err
 	}
 
+	annotations := antreaConfig.GetAnnotations()
+	if _, ok := annotations[constants.TKGAnnotationTemplateConfig]; ok {
+		log.Info(fmt.Sprintf("resource '%v' is a config template. Skipping reconciling", req.NamespacedName))
```
knabben, line 70: Warning should be more appropriate?

addons/controllers/antrea/antreaconfig_controller.go
```diff
@@ -64,28 +65,24 @@ func (r *AntreaConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-	}
+	cluster, err := cutil.GetOwnerCluster(ctx, r.Client, antreaConfig, req.Namespace, constants.AntreaDefaultRefName)
 
-	// verify that the cluster related to config is present
-	if err := r.Client.Get(ctx, clusterNamespacedName, cluster); err != nil {
-		if apierrors.IsNotFound(err) {
-			log.Info(fmt.Sprintf("Cluster '%s'not found in '%s'", clusterNamespacedName.Name, clusterNamespacedName.Namespace))
+	if err != nil {
+		if apierrors.IsNotFound(err) && cluster != nil {
+			log.Info(fmt.Sprintf("'%s/%s' is listed as owner reference but could not be found",
```
knabben, line 81: log.Error should be more appropriate?

knabben, line 81: log.Error could be more appropriate.

addons/controllers/antrea/antreaconfig_controller.go
adduarte: ? don't think so. this is pretty common. Why  a warning?

addons/controllers/antrea/antreaconfig_controller.go
adduarte: Is not an error though. Is just a state of things.  logging as an error might suggest to the user something is not correct. and it will make the logs very chatty

addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go
adduarte: added logic to add owner reference to any existing object listed under clusterbootstrap on reconcilenormal of the clusterbootstrap controller

addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go
```diff
@@ -379,14 +379,44 @@ func (h *Helper) HandleExistingClusterBootstrap(clusterBootstrap *runtanzuv1alph
...
+		// 2. When full clusterbootstrap without tkr annotation is given by user.
+		//    In this case we only want to handle inline packages for creating
+		//    secret. The providers are expected to be present in the
+		//   clusternamespace, created by user before clusterbootstrap is created.
+		var nonEmptyInlinePackages []*runtanzuv1alpha3.ClusterBootstrapPackage
+		for _, cbPackage := range packages {
+			if cbPackage != nil && cbPackage.ValuesFrom != nil && cbPackage.ValuesFrom.Inline != nil {
+				nonEmptyInlinePackages = append(nonEmptyInlinePackages, cbPackage)
+			}
+		}
```
shyaamsn, line 398: Not sure why codecov is reporting that no test is covering this. I thought you have added a test case for this ? Could you please ensure this is covered as part of the test.

adduarte: will do

addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go
```diff
@@ -321,72 +321,100 @@ func (h *Helper) HandleExistingClusterBootstrap(clusterBootstrap *runtanzuv1alph
...
-	if clusterBootstrap.Annotations != nil {
-		if annotationValue, annotationExist := clusterBootstrap.Annotations[constants.AddCBMissingFieldsAnnotationKey]; annotationExist {
-			tkrName = annotationValue
-			clusterBootstrapTemplateName := annotationValue // TanzuKubernetesRelease and ClusterBootstrapTemplate share the same name
-			clusterBootstrapTemplate := &runtanzuv1alpha3.ClusterBootstrapTemplate{}
-			// Get the ClusterBootstrapTemplate mentioned in the annotation under system namespace
-			if err := h.K8sClient.Get(h.Ctx, client.ObjectKey{Namespace: systemNamespace, Name: clusterBootstrapTemplateName}, clusterBootstrapTemplate); err != nil {
-				return nil, err
-			}
+	if annotationValue, annotationExist := clusterBootstrap.Annotations[constants.AddCBMissingFieldsAnnotationKey]; annotationExist {
```
shyaamsn, line 324: **Question**: Would Annotations be nil anytime ? Should we handle it ?

adduarte: The statment 
```
if annotationValue, annotationExist := clusterBootstrap.Annotations[constants.AddCBMissingFieldsAnnotationKey]; annotationExist
``
is special in golang and if annotations = nil, then annotationExist == ... returns nil. It won't error out. 
you can run this code to verify: 

```type ClusterBootstrap struct {
	Annotations map[string]string
}

func main() {
	clusterBootstrap := ClusterBootstrap{}
	if annotationValue, annotationExist := clusterBootstrap.Annotations["somefield"]; annotationExist {
		println(annotationValue)
	} else {
		println("Annotatinons == nil but we! good it works")
	}
}
```

shyaamsn: Sounds good. Just wanted to confirm.

shyaamsn (approved):
LGTM

PR #3663, opened 2022-10-13
Skip adding last applied annotation to kapp-controller on workload cluster

haoranleo:
- Let the logic only works on management cluster when the feature flag is off.

Fixes #
- Fix the bug observed in downstream pipelines that when feature flag is off, workload cluster fails to be upgraded due to error in adding last-applied annotation to `kapp-controller`, which actually should **not** run on workload cluster.

Verified on TKG cluster on AWS using the tanzu CLI built from this branch, and `kapp-controller` is upgraded successfully.

### Release note

```release-note
Fix kapp-controller upgrade issue on workload cluster when package-based-lcm feature flag is off
```

### DIFF
--- a/tkg/client/upgrade_addon.go
+++ b/tkg/client/upgrade_addon.go
@@ -248,8 +248,8 @@ func (c *TkgClient) DoUpgradeAddon(regionalClusterClient clusterclient.Client, /
 			return errors.Wrap(err, "unable to get cluster configuration")
 		}
 
-		// ensure current kapp-controller deployment has last-applied annotation, which is required for future apply operations
-		if addonName == "addons-management/kapp-controller" {
+		// ensure current kapp-controller deployment on the management cluster has last-applied annotation, which is required for future apply operations
+		if addonName == "addons-management/kapp-controller" && options.IsRegionalCluster {
 			if err := clusterClient.PatchKappControllerLastAppliedAnnotation(options.Namespace); err != nil {
 				return errors.Wrap(err, "unable to add last-applied annotation on kapp-controller")
 			}


tkg/managementcomponents/management_component_install.go
```diff
@@ -290,14 +290,18 @@ func InstallManagementComponents(mcip *ManagementComponentsInstallOptions) error
 
 // InstallKappController installs kapp-controller to the cluster
 func InstallKappController(clusterClient clusterclient.Client, kappControllerOptions KappControllerOptions) error {
+	// Ensure last-applied annotation is present before kapp-controller upgrade
+	if err := clusterClient.WaitForDeployment(constants.KappControllerDeploymentName, kappControllerOptions.KappControllerInstallNamespace); err == nil { // only attempt to add annotation when kapp-controller exists on the cluster
```
shyaamsn, line 294: Does this function only get called for management cluster ? Should we skip workload cluster here too ?

shyaamsn, line 294: **Question**: Do we want to use WaitForDeployment or just check if Deployment is there. WaitForDeployment actually makes sure the replicas are available. Do we need to check that here ?

HL-EverGreen: Hi Shyaam, thanks for pointing this out. Yeah we only want to use it to check if Deployment is there. I just reuse this function to differentiate the install/upgrade case

HL-EverGreen: I use `GetResource` instead to check existence of `kapp-controller`

HL-EverGreen: Yes, this function and also function `InstallOrUpgradeKappController` ([here](https://github.com/vmware-tanzu/tanzu-framework/blob/main/tkg/client/management_components.go#L25)) is only called for management clusters.

tkg/client/upgrade_addon.go
```diff
@@ -248,8 +248,8 @@ func (c *TkgClient) DoUpgradeAddon(regionalClusterClient clusterclient.Client, /
 			return errors.Wrap(err, "unable to get cluster configuration")
 		}
 
-		// ensure current kapp-controller deployment has last-applied annotation, which is required for future apply operations
-		if addonName == "addons-management/kapp-controller" {
+		// ensure current kapp-controller deployment on the management cluster has last-applied annotation, which is required for future apply operations
+		if addonName == "addons-management/kapp-controller" && options.IsRegionalCluster {
```
yharish991, line 252: curious: what does a `RegionalCluster` mean?

HL-EverGreen: `RegionalCluster` stands for management cluster.

yharish991 (approved):

tkg/managementcomponents/management_component_install.go
```diff
@@ -290,14 +291,18 @@ func InstallManagementComponents(mcip *ManagementComponentsInstallOptions) error
 
 // InstallKappController installs kapp-controller to the cluster
 func InstallKappController(clusterClient clusterclient.Client, kappControllerOptions KappControllerOptions) error {
+	// Ensure last-applied annotation is present before kapp-controller upgrade
+	if err := clusterClient.GetResource(&appsv1.Deployment{}, constants.KappControllerDeploymentName, kappControllerOptions.KappControllerInstallNamespace, nil, nil); err == nil { // only attempt to add annotation when kapp-controller exists on the cluster
```
maralavi, line 295: What if an error happens here?

shyaamsn: Should we return error in case its other than NotFound and only call PatchKappControllerLastAppliedAnnotation if deployment is found ?

HL-EverGreen: If error happens, then we skip adding last-applied annotation on the `kapp-controller`. Because it might be `NotFound` error in the install phase, so this add annotation logic is not required

HL-EverGreen: > Should we return error in case its other than NotFound
I don't think it's necessary. If error other than `NotFound` is returned, during install phase, other unexpected error returned by `GetResource` will stop installing `kapp-controller`. We should continue to install `kapp-controller` regardless to these errors because they are not relevant to kapp-controller install.

maralavi: Let's ask @vijaykatam opinion on whether we can leave other unexpected errors unhandled. Otherwise the changes look good to me

maralavi (approved):

alfredthenarwhal:
Build failed! Build no: 2941

shyaamsn (approved):

HL-EverGreen:
The fix when the `package-based-lcm` feature flag is on has been moved to another PR https://github.com/vmware-tanzu/tanzu-framework/pull/3700

PR #3665, opened 2022-10-14
Decouple Feature-Flags from cli/runtime library and move to it to plugin specific code

### LINKED ISSUE #3447: Plugin specific Feature-Flags should not be defined in plugin runtime library (cli/runtime)

anujc25:

- This PR decouples feature-flag definitions from the `cli/runtime` library and move it to plugin specific code
- This change uses `PluginDescriptor` for a specific plugin to set the defaults for the feature-flags and core CLI configures the default values during plugin installation or during plugin upgrade.
- The same default feature-flag values can also be configured with `tanzu config init` command if config file is updated manually and defaults needs to be set again.

Fixes #3447

- **Manual tests:**
  - Clean tanzu config file
  - Run `tanzu config get` command
    - It only adds tanzu cli specific default feature-flags to the config
  - Install all plugins with `tanzu plugin sync` to install all plugins
    - Run `tanzu config get` and verify plugin specific default feature-flags are configured correctly.
  - Manually update the config.yaml to remove certain default feature-flags for installed plugin and then run `tanzu config init`
    - Verify that all default feature-flags for installed plugin are configured as expected.

- **Unit tests** are written and validated as part of CI 

### Release note

```release-note
Decouple Feature-Flags from `cli/runtime` library and move to it to plugin specific code
```


marckhouzam (commented):
I haven't looked at the code yet, but I noticed that I still see the `tkr-version-v1alpha3-beta` feature flag under the `global` list.  Is this correct?

```
$ rm ~/.config/tanzu/config.yaml
$ tanzu config init
✔  successfully initialized the config
$ tanzu config get |grep global: -A5
    global:
      context-target: "false"
      tkr-version-v1alpha3-beta: "false"
    management-cluster:
```

marckhouzam:
> I haven't looked at the code yet, but I noticed that I still see the `tkr-version-v1alpha3-beta` feature flag under the `global` list. Is this correct?

The constant used for this option is `features.global.tkr-version-v1alpha3-beta` which uses `global` instead of the plugin name.  Maybe we cannot change it because of backwards compatibility?

cli/core/pkg/command/config.go
```diff
@@ -216,39 +216,29 @@ var initConfigCmd = &cobra.Command{
...
-			if !exists {
-				finalRepos = append(finalRepos, repo)
-			}
+
+		serverName := ""
+		server, err := cfg.GetCurrentServer()
+		if err == nil && server != nil {
+			serverName = server.Name
 		}
-		cfg.ClientOptions.CLI.Repositories = finalRepos
```
marckhouzam, line 233: We no longer need this I gather?
Is it ok if the user turns off context-aware discovery, or maybe we don't support that at all?

cli/core/pkg/command/config.go
```diff
@@ -216,39 +216,29 @@ var initConfigCmd = &cobra.Command{
...
+		}
+
+		err = configlib.StoreClientConfig(cfg)
 		if err != nil {
 			return err
 		}
 
-		errList := []error{}
-		for _, desc := range descriptors {
-			if err := cli.InitializePlugin(desc.Name); err != nil {
```
marckhouzam, line 247: This seems to be a change in behaviour where `tanzu config init` will no longer initialize all the plugins.
I guess that is ok?

marckhouzam (commented):
This is a vey nice decoupling.  Some minor comments in line and one question below.

When we delete a plugin, we don't delete its feature flags from the config file.  That is probably on purpose so that if the user installs the plugin back, they keep their configuration? I think this is a good approach, I just wanted to confirm in case the question comes up later.

cmd/cli/plugin/managementcluster/featureflags.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"github.com/vmware-tanzu/tanzu-framework/tkg/constants"
+)
+
+// DefaultFeatureFlagsForManagementClusterPlugin is used to populate default feature-flags for the management-cluster plugin
+var (
+	DefaultFeatureFlagsForManagementClusterPlugin = map[string]bool{
+		"features.management-cluster.import":                       false,
+		"features.management-cluster.export-from-confirm":          true,
+		"features.management-cluster.standalone-cluster-mode":      false,
```
marckhouzam, line 15: I've been thinking further about these constants.  I think we have an opportunity here to tighten the CLIs API with its plugins.

I don't think it should be plugins that define the entire feature-flag path for their flags.  It gives them too much control over where flags will be stored.  They should only be able to specify a subpath under "features.\<pluginName\>", and it should be the CLI that prepends the fixed "features.\<pluginName\>" prefix.

For example, the feature-flag `features.management-cluster.standalone-cluster-mode` should only be `standalone-cluster-mode` in this file and passed that way to the CLI in the PluginDescriptor.  The CLI will then add the `features.management-cluster` prefix before storing it in the config file.

We can make this change without breaking backwards-compatibility, except for ""features.global.tkr-version-v1alpha3-beta" which is named wrong I believe.

We don't need to do this in this PR actually, but I think we should eventually do it.

What do you think?

What do you think?

marckhouzam (commented):
Also, we should update the `config-feature.md` file with the new approach.

https://github.com/vmware-tanzu/tanzu-framework/blob/a6760d37ec59410154daf732331be7061d5a3279/cli/core/docs/cli/config-features.md?plain=1#L34-L50

tkg/constants/featureflags.go
```diff
@@ -0,0 +1,38 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package constants
+
```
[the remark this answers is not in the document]
anujc25: `tanzu-framework/tkg` because `cluster` and `management-cluster` plugin share most of the codebase. This package is an implementation of what both of the plugins does. As both plugins imports this dependency and both plugin uses this featureflags, this constants were added to the common library instead of plugin specific code.
Also, this library heavily uses this constants.

cli/runtime/config/featureflags.go
```diff
@@ -1,96 +0,0 @@
...
-	// support of ARM should be included when discovering available AWS instance types. Setting feature flag to true
-	// filters out ARM supporting instance types; false allows ARM instance types to be included in results.
-	FeatureFlagAwsInstanceTypesExcludeArm = "features.management-cluster.aws-instance-types-exclude-arm"
-	// PackageBasedLCM feature flag determines whether to use package based lifecycle management of management component
-	// or legacy way of managing management components. This is also used for clusterclass based management and workload
-	// cluster provisioning
-	FeatureFlagPackageBasedLCM = "features.global.package-based-lcm-beta"
-	// TKR version v1alpha3 feature flag determines whether to use Tanzu Kubernetes Release API version v1alpha3. Setting
-	// feature flag to true will allow to use the TKR version v1alpha3; false allows to use legacy TKR version v1alpha1
-	FeatureFlagTKRVersionV1Alpha3 = "features.global.tkr-version-v1alpha3-beta"
```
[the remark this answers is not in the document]
anujc25: Yeah. I believe this should have been used as a plugin specific constants but it was used as `global` assuming multiple plugin might rely on them.

Considering, these feature-flags are always `false` by default at the moment, we might be able to change it if required. But that means all testing scripts need to be updated as well.

cli/core/pkg/config/init.go
```diff
@@ -21,7 +21,7 @@ func init() {
 	}
 
 	addedDefaultDiscovery := populateDefaultStandaloneDiscovery(c)
-	addedFeatureFlags := addDefaultFeatureFlagsIfMissing(c, config.DefaultCliFeatureFlags)
+	addedFeatureFlags := AddDefaultFeatureFlagsIfMissing(c, DefaultCliFeatureFlags)
```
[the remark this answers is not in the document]
anujc25: We have integrated it with `tanzu config init` command. So, if user manually deletes feature-flag. The default value can be brought by by running `tanzu config init` command.

cli/core/pkg/command/config.go
anujc25: We don't support it at all now. And this is kind of a dead code that I removed to improve the readability of that function.

cli/core/pkg/command/config.go
anujc25: It looks like a change in behavior but it is actually not considering the bug in implementation here where we always return empty `descriptors` array because of missing implementation of `context-aware-cli-for-plugins` featureflag enabled.

cli/core/pkg/config/features.go
```diff
@@ -6,9 +6,35 @@ package config
...
+	configlib "github.com/vmware-tanzu/tanzu-framework/cli/runtime/config"
 )
 
+// ConfigureDefaultFeatureFlagsIfMissing configures default feature-flags to ClientConfig if missing
+func ConfigureDefaultFeatureFlagsIfMissing(defaultFeatureFlags map[string]bool) error {
+	// Acquire tanzu config lock
+	configlib.AcquireTanzuConfigLock()
+	defer configlib.ReleaseTanzuConfigLock()
+
+	c, err := configlib.GetClientConfigNoLock()
```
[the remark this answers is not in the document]
anujc25: There are some fundamental differences between 
```
AcquireTanzuConfigLock()
defer ReleaseTanzuConfigLock()

c = GetClientConfigNoLock()
// Update c in memory
StoreClientConfig(c)
```
and
```
c = GetClientConfig()
// Update c in memory
StoreClientConfig(c)
```

The 1st approach locks the file while reading/updating-in-memory/writing the configuration file. 
While 2nd approach locks the file while reading only.

To just use `GetClientConfig` we might need to do some changes in how we are locking to read and write configuration.

tkg/constants/featureflags.go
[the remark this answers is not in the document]
marckhouzam: Sounds good

cli/core/pkg/command/config.go
marckhouzam: Great

cli/core/pkg/command/config.go
marckhouzam: Aha 😁.  Then nice cleanup 👍

cli/core/pkg/command/config.go
```diff
@@ -216,39 +216,29 @@ var initConfigCmd = &cobra.Command{
...
 
-		err = configlib.StoreClientConfig(cfg)
+		serverPluginDescriptors, standalonePluginDescriptors, err := pluginmanager.InstalledPlugins(serverName)
 		if err != nil {
 			return err
 		}
-		// TODO: cli.ListPlugins is deprecated: Use pluginmanager.AvailablePluginsFromLocalSource or pluginmanager.AvailablePlugins instead
-		descriptors, err := cli.ListPlugins()
+
+		for _, desc := range append(serverPluginDescriptors, standalonePluginDescriptors...) {
```
vuil, line 231: The (newly added) context-specific aspect of `config init` is not very intuitive (especially for folks who have used it before without having to think about this aspect). To properly reconstruct all the defaults in the config file, one may now have to switch contexts so that all plugins which offers default values are covered. If the user does not learn of this new behavior, he could think that a single `config init` would `fix/initialize` the config file, without realizing that it only partially initializes the file.

Would iterating through all installed plugins be an option?

If we choose to stick with the current implementation, I would still suggest
- putting a comment to note this behavior
- see if we can provide some doc and/or UX feedback to the user about the same (maybe in the help text of the config init command)

cli/core/pkg/command/config.go
```diff
@@ -216,39 +216,29 @@ var initConfigCmd = &cobra.Command{
...
+			config.AddDefaultFeatureFlagsIfMissing(cfg, desc.DefaultFeatureFlags)
```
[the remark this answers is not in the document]
anujc25: Older plugins can take care of updating this DefaultFeatureFlag by themselves because of the way they are importing the tanzu-framework as a dependency. There is a `func init` that gets run every-time which configures this feature-flags if they are missing with each plugin invocation.

cli/core/pkg/command/config.go
anujc25: I think iterating through all installed plugins might be overkill. Considering user might have substantial amount of plugins installed and some of them are not used anymore.

I think adding a more details in the `tanzu config init` help text and providing comment in the code should be sufficient I feel.

cli/runtime/config/featureflags.go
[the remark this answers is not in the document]
marckhouzam: What testing scripts are you referring to? I don't see that string used in TF anywhere else.

cli/core/pkg/config/init.go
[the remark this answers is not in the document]
marckhouzam: You're right.  I like that better as well. 👍

cmd/cli/plugin/managementcluster/featureflags.go
marckhouzam: I agree we can do this improvement separately, but I also think we should do it before we separate the CLI code from TF as we could do it without dealing with compatibility.  I'll open an issue so we don't forget about it.

cli/core/pkg/config/features.go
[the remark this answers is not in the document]
marckhouzam: Oh my, I completely missed that difference!  It's the behaviour of `defer` being in a different function that I missed.
Thanks for clarifying.

marckhouzam (approved):
LGTM
Sorry for the delay

cmd/cli/plugin/managementcluster/featureflags.go
marckhouzam: I have opened #3824 to track this

cli/core/pkg/command/config.go
vuil: I think that sounds reasonable. We should make sure any docs we produce capture the details as well. 
I suggest we tag the issue and this PR with docs-triage-needed to raise some awareness.

cli/core/pkg/command/config.go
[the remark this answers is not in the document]
vuil: ah, thanks.

vuil (approved):
lgtm

cli/runtime/config/featureflags.go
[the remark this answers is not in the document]
anujc25: I was more talking about product level testing and not in TF testing.

PR #3674, opened 2022-10-14
Enable Single Node Cluster deployment via tanzu CLI

### LINKED ISSUE #3621: Single Node Cluster Deployment using Tanzu CLI
Currently, tanzu framework has restrictions of creating workload clusters with minimum 1 control plane and 1 one worker node. This feature would enable users to create single node workload clusters which can be used for the edge locations with limited resources.

Changes that would be required in Tanzu Framework are as proposed below:
- Introduce a feature toggle to enable deployment of Single Node Clusters. 
- Introduce a variable in [default](https://github.com/vmware-tanzu/tanzu-framework/blob/main/pkg/v1/providers/infrastructure-vsphere/v1.3.1/cconly/base.yaml#L189) vSphere ClusterClass to represent control plane nodes having no master taint with the inline patches for initConfiguration and joinConfiguration.
- If the feature gate is enabled and the infrastructure is vSphere. 
  - Tanzu CLI has restrictions of [minimum 1 control plane and 1 worker node](https://github.com/vmware-tanzu/tanzu-framework/blob/f7ef2a839571031a133130fd077e62a685a028a9/tkg/client/machine_deployment_cc.go#L23) in a TKGm cluster. We need to add the tolerance to support 0 worker nodes (allow cluster.Spec.Topology.Workers.MachineDeployments as 0 or cluster.Spec.Topology.Workers be nil) in the cluster. 
  - During classy upgrades(ClusterClass based upgrades), the worker node count is [hardcoded as 1](https://github.com/vmware-tanzu/tanzu-framework/blob/c1137d344dd30f1f811b2613bc22c26922fc0f88/pkg/v1/tkg/clusterclient/clusterclient.go#L705). Tanzu CLI must remove the hardcoding and allow for 0 replica value for the workers in order to support upgrades of SNC.  
  - Tanzu CLI should introduce a validation that SNC have 1 control plane node with no master taints enabled, otherwise the cluster creation should fail. 
  - Tanzu CLI keeps track of the ready replicas of control plane and worker nodes to display the cluster status, we need to add tolerance for single node cluster, such that `tanzu cluster list` output doesn’t show cluster status as creating/createStalled even after the cluster is provisioned successfully.

- ( ) APIs
- ( ) Addons
- (X) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- (X) User Experience
- ( ) Developer Experience

**Additional context**

Ankitasw:
This PR enabled deployment of Single Node Clusters (SNC) using Tanzu CLI.

Fixes #3621 

### Describe testing done for PR:

With feature gate enabled:

- Created single node workload cluster with standard kubernetes runtime on vSphere.
- Created standard kubernetes workload cluster on vSphere.
- Checked that SNC creation is blocked when `controlPlaneTaint` is set to true.

With feature gate disabled:

- Checked that creation of SNC is blocked.
- Created standard kubernetes workload cluster on vSphere.

### Release note

```release-note
 Enable Single Node Cluster deployment via tanzu CLI
```

### Additional information

- Added a new behavior of not allowing to create single node clusters without the feature gate. Previously, there was no failure in case of creating a single node cluster.
- `controlPlaneTaint` variable is introduced in default vSphere ClusterClass, which is set to true by default. If users want to deploy SNC, they should set the variable as false, such that the control plane node taints would be removed and workloads could be scheduled onto SNC. Otherwise, the default behavior would be preserved.


vuil:
@Ankitasw please add @wjun @imikushin @tenczar to the reviewer list when the PR is ready for review

imikushin (commented):
LGTM. But since it's still WIP, I'd like to take another look when it's ready.

Ankitasw:
@imikushin @tenczar @wjun this PR is ready for review.

tkg/client/cluster.go
```diff
@@ -862,3 +874,55 @@ func (c *TkgClient) ValidateManagementClusterVersionWithCLI(regionalClusterClien
 
 	return nil
 }
+
+// validateControlPlaneTaint validates if controlPlaneTaint CC variable is not set for the single node workload cluster, otherwise returns error
+func validateControlPlaneTaint(stream []byte, options *CreateClusterOptions, tkgClient *TkgClient, regionalClient clusterclient.Client) error {
```
dharmjit, line 879: suggestion: It seems like the function scope is more than validating the taint. Do you think it's better to rename it for e.g. `ValidateSNCCreation`?

tkg/client/machine_deployment.go
```diff
@@ -103,12 +105,20 @@ func (c *TkgClient) SetMachineDeployment(options *SetMachineDeploymentOptions) e
...
 			return errors.Wrap(err, "Unable to retrieve cluster resource")
 		}
-		return DoSetMachineDeploymentCC(clusterClient, &cluster, options)
+		skip, err := skipMDCreation(clusterClient, c, cluster, options)
+		if err != nil {
+			return err
+		}
+		if skip {
+			return nil
+		}
```
dharmjit, line 120: Question: Is this required because the clusterclass will have MD? I am asking because the cluster manifest will not have workers spec.. no? Also, could the skip logic be derived from some annotation on the cluster object?

dharmjit (commented):
some nits only

tkg/client/machine_deployment.go
Ankitasw: In case of single node cluster, we don't need MD creation as there are no worker nodes.
But in case of standard cluster, MD creation is needed, so this func just checks if the MD creation is to be skipped or not.

tkg/client/cluster.go
```diff
@@ -862,3 +874,55 @@ func (c *TkgClient) ValidateManagementClusterVersionWithCLI(regionalClusterClien
...
+	infra, err := regionalClient.GetClusterInfrastructure()
+	if err != nil {
+		return errors.New("failed to get current management cluster infrastructure")
+	}
+
+	if topology.IsSingleNodeCluster(cluster) {
+		// Adding a new behavior of not allowing to create single node clusters without the feature gate. Previously, there was no failure in case of creating a single node cluster.
+		if !tkgClient.IsFeatureActivated(config.FeatureFlagSingleNodeClusters) {
+			return errors.New("Worker count cannot be 0, minimum worker count required is 1")
+		} else if controlPlaneTaint && infra == constants.InfrastructureRefVSphere {
```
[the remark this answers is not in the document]
Ankitasw: Right now we are enabling creation of SNC only for vSphere and not for other infras.

tkg/client/cluster.go
```diff
@@ -862,3 +874,55 @@ func (c *TkgClient) ValidateManagementClusterVersionWithCLI(regionalClusterClien
...
+
+// validateControlPlaneTaint validates if controlPlaneTaint CC variable is not set for the single node workload cluster, otherwise returns error
+func validateControlPlaneTaint(stream []byte, options *CreateClusterOptions, tkgClient *TkgClient, regionalClient clusterclient.Client) error {
+	cluster, err := getClusterObjectFromYaml(stream)
+	if err != nil {
+		return err
+	}
+
+	controlPlaneTaint := true
+	err = topology.GetVariable(cluster, "controlPlaneTaint", &controlPlaneTaint)
```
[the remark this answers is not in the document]
Ankitasw: We dont have to set it, defining here by default as true means that if it is not set, it's going to be true, but if it's set, it is read from topology.

tkg/client/cluster.go
Ankitasw: good point, but I am still thinking of a better name 😄

tkg/client/machine_deployment.go
Ankitasw: > Also, could the skip logic be derived from some annotation on the cluster object?

We are not introducing any annotations for single node cluster, and we have access to machinedeployments only, at this point, so I guess we dont have any other alternative to derive skip logic, but to depend on topology.

tkg/client/cluster.go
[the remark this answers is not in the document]
Ankitasw: We decided to move away from infra check and provide disclaimer that this has been tested only for vSphere infra.

tkg/client/cluster.go
Ankitasw: Renamed to `validateConfigForSingleNodeCluster`

providers/infrastructure-vsphere/v1.4.1/cconly/base.yaml
```diff
@@ -468,6 +468,12 @@ spec:
       openAPIV3Schema:
         type: object
         properties: {}
+  - name: controlPlaneTaint
+    required: false
+    schema:
+      openAPIV3Schema:
+        type: boolean
+        default: true
```
tenczar, line 476: why not wrap this under the controlPlane object and have it set node taints more generally? this is almost certainly going to be a requested feature.

providers/infrastructure-vsphere/v1.4.1/cconly/base.yaml
```diff
@@ -1421,6 +1427,21 @@ spec:
...
+    definitions:
+    - selector:
+        apiVersion: controlplane.cluster.x-k8s.io/v1beta1
+        kind: KubeadmControlPlaneTemplate
+        matchResources:
+          controlPlane: true
+      jsonPatches:
+      - op: add
+        path: /spec/template/spec/kubeadmConfigSpec/initConfiguration/nodeRegistration/taints
+        value: []
```
tenczar, line 1441: this just needs to be an empty value? can you help explain why?

tkg/client/cluster.go
```diff
@@ -862,3 +874,63 @@ func (c *TkgClient) ValidateManagementClusterVersionWithCLI(regionalClusterClien
...
+	}
+
+	if controlPlaneTaint {
+		return errors.New(fmt.Sprintf("unable to create single node cluster %s as control plane node has taint", options.ClusterName))
+	}
+
+	return nil
+}
+
+func getClusterObjectFromYaml(stream []byte) (*capi.Cluster, error) {
```
tenczar, line 907: we don't have a method that does this already?

tkg/client/cluster.go
```diff
@@ -862,3 +874,63 @@ func (c *TkgClient) ValidateManagementClusterVersionWithCLI(regionalClusterClien
...
+	clusterYaml, err := findClusterDefinitionIn(stream)
+	if err != nil {
+		return nil, err
+	}
+
+	cluster := &capi.Cluster{}
+	err = runtime.DefaultUnstructuredConverter.FromUnstructured(clusterYaml, cluster)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to convert yaml to unstructured format")
```
tenczar, line 916:
```suggestion
		return nil, errors.Wrap(err, "unable to convert yaml to structured format")
```

tkg/client/machine_deployment.go
tenczar: Is there any reason a single node cluster should be disallowed from adding node pools?

tkg/client/machine_deployment_test.go
```diff
@@ -1681,4 +1685,198 @@ var _ = Describe("Machine Deployment", func() {
 			})
 		})
 	})
+	Context("SetMachineDeployment", func() {
```
tenczar, line 1688: This probably belongs in machine_deployment_cc_test.go

tenczar (commented):
just a couple questions

providers/infrastructure-vsphere/v1.4.1/cconly/base.yaml
Ankitasw: Taints has to be empty as the control plane will have a master schedule taint, so we are setting it to empty such that the single node in cluster would be able to schedule the workloads.

tkg/client/cluster.go
Ankitasw: Didnt find one, thats why used own helper.

tkg/client/machine_deployment_test.go
Ankitasw: We made changes to machine_deployment.go  and testing the whole flow, hence no test required in machine_deployment_cc_test.go

providers/infrastructure-vsphere/v1.4.1/cconly/base.yaml
Ankitasw: We wanted to make it flexible by the user, this is conditional patch and would remove taint only if the user sets this variable as false. We are setting this variable to true by default.

jamiemonserrate: > have it set node taints more generally
By this, did you mean accepting a list of taints from the user? Assuming that's what you meant - we decided to keep it a bool to make the experience of creating a single node cluster easier. 

That is fair feedback though. When we went through our TEP review, we opted for the boolean to limit scope. We could always enhance this in a future iteration to make this more generic. 

Does that make sense @tenczar?

tenczar: Yeah, that makes sense.

providers/infrastructure-vsphere/v1.4.1/cconly/base.yaml
tenczar: Ok, that makes sense

tkg/client/machine_deployment_test.go
tenczar: My comment was more that this just has a context for the clusterclass version, so if the test were moved into `machine_deployment_cc_test.go` then you could remove one level of nesting from the test hierarchy.

tenczar (approved):
LGTM

tkg/client/machine_deployment_test.go
Ankitasw: Makes sense, I will make the change.

Ankitasw:
The only change after approval was testcase movement from one file to another.

providers/infrastructure-vsphere/v1.4.1/cconly/base.yaml
```diff
@@ -454,6 +454,12 @@ spec:
       openAPIV3Schema:
         type: object
         properties: {}
+  - name: controlPlaneTaint
```
tenczar, line 457: Because we are changing the variable defintions we should probably bump the version number in the annotation on line 174 `identifier.cc.vmware.com: tanzu-vsphere-default-0.0.1`. Then we should make sure that Shimon on the UX team is aware of this change. Their UI uses this version number to help the UI keep track if it is compatible with this clusterclass.

tenczar (approved):
minor nit about versions

providers/infrastructure-vsphere/v1.4.1/cconly/base.yaml
Ankitasw: Updated, thanks.

Ankitasw:
Somehow CAPZ tests are failing after rebasing the changes, all tests were passing before the rebase.

util/topology/cluster.go
```diff
@@ -76,6 +77,18 @@ func GetMDVariable(cluster *clusterv1.Cluster, mdIndex int, name string, value i
...
 
+// IsSingleNodeCluster checks if the cluster topology is single node cluster(with CP count as 1 and worker count as 0).
+// Pre-reqs: cluster.Spec.Topology != nil
+func IsSingleNodeCluster(cluster *clusterv1.Cluster) bool {
+	return *cluster.Spec.Topology.ControlPlane.Replicas == *pointer.Int32(1) && HasWorkerNodes(cluster)
+}
+
+// HasWorkerNodes checks if the cluster topology has worker nodes.
+// Pre-reqs: cluster.Spec.Topology != nil
+func HasWorkerNodes(cluster *clusterv1.Cluster) bool {
```
wjun, line 88: HasWorkerNodes->NoWorkerNodes

Ankitasw: since its a bool returning func, doesn't HasWorkerNodes makes more sense?

tkg/client/machine_deployment.go
```diff
@@ -762,3 +771,13 @@ func updateAzureSecret(kcTemplate *v1beta1.KubeadmConfigTemplate, machineTemplat
 		}
 	}
 }
+
+func skipMDCreation(clusterClient clusterclient.Client, c *TkgClient, cluster *capi.Cluster, options *SetMachineDeploymentOptions) (bool, error) {
+	if topology.IsSingleNodeCluster(cluster) && c.IsFeatureActivated(constants.FeatureFlagSingleNodeClusters) {
+		return true, nil
+	} else if topology.HasWorkerNodes(cluster) {
```
wjun, line 778: HasWorkerNodes->NoWorkerNodes

wjun (approved):
LGTM. Some minor comments can be fixed in the next pr.

util/topology/cluster.go
wjun: Then it is better to use !HasWorkerNodes in the references.

PR #3557, opened 2022-10-05
[WIP] Oracle Cloud Support

randomvariable:
### Which issue(s) this PR fixes

Fixes #

### Release note

```release-note
Added support for provisioning TKG on Oracle Cloud. Refer to product documentation for more information.
```

PR #3681, opened 2022-10-17
tkgconfigbom: Revise help messages when downloading BOM

randomvariable:

When downloading the BOM, print extra debug information for developers.
Also cleans up the language when returning errors to be more specific.
Adds some code comments to explain what is happening to newcomers.

Was used in development of #3557 .

### Release note

```release-note
NONE
```

### DIFF
--- a/tkg/tkgconfigbom/bom.go
+++ b/tkg/tkgconfigbom/bom.go
@@ -373,7 +373,7 @@ func GetTKRBOMImageTagNameFromTKRVersion(tkrVersion string) string {
 }
 
 var errorDownloadingDefaultBOMFiles = `failed to download the BOM file from image name '%s':%v
-If this is an internet-restricted environment please refer to the documentation to set TKG_CUSTOM_IMAGE_REPOSITORY and related configuration variables in %s 
+If this is an internet-restricted environment please refer to the documentation to set TKG_CUSTOM_IMAGE_REPOSITORY and related configuration variables in %s
 `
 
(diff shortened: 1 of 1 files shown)


tkg/tkgconfigbom/bom.go
```diff
@@ -471,12 +471,17 @@ func (c *client) DownloadTKGCompatibilityFileFromRegistry(repo, resource string,
 
 	// begin download of compatibility file
 	tkgCompatibilityImagePath := fmt.Sprintf("%s/%s", repo, resource)
-	log.Infof("Downloading TKG compatibility file from '%s'", tkgCompatibilityImagePath)
+	log.Infof("Listing TKG compatibility tags from '%s'", tkgCompatibilityImagePath)
```
marckhouzam, line 474: I wonder if printing "listing compatibility tags" would be confusing to a user since that list will not be printed to the screen (I think).  The listing part seems to be an implementation detail that the user would not be aware of.

But just my inexperienced 2-cents

randomvariable: Stick it under a V(6) or something?

marckhouzam: I was thinking that the original message of "Downloading TKG compatibility" made sense to the user as they understand that there is a download going on.  The listing was an "obscure" step however.

How about reverting the original messages and adding a V(6) (or whatever level you feel appropriate) for the implementation-specific steps?

vuil: +1 to put this behind a log level. L505 probably suffice in happy path.
Rest of changes lgtm.

PR #3686, opened 2022-10-17
Fix Capabilities schema definition.

seemiller:

Had incorrectly added a second type to the schema as a part of 

Run the following task `make generate-package-secret PACKAGE=capabilities`
and script `cd capabilities/hack && generate-package-secret.test.sh`

### Release note

```release-note

```

### DIFF
--- a/capabilities/hack/ytt/schema.yaml
+++ b/capabilities/hack/ytt/schema.yaml
@@ -8,13 +8,7 @@ deployment:
   nodeSelector: {}
   tolerations:
     - effect: NoSchedule
-      key: node-role.kubernetes.io/master
-      value: ""
-      operator: ""
-      #@schema/nullable
-      tolerationSeconds: 0
-    - effect: NoSchedule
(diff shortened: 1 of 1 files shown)


codegold79:
I'm a bit confused. The `capabilities/hack/ytt/schema.yaml` file had the old toleration (with the word `master`) a couple months ago. But now, you'll get rid of both the old and new (`control-plane`)? Isn't there supposed to be at least one toleration item? Or was it not needed two months ago because this schema was for a `Kind: Secret`, and tolerations are only for `Kind:Pod`?

seemiller:
@codegold79 The schema is defining the shape of the "objects" in the array. Originally, it was defined as only allowing objects with a `key` that has a value of `master`. Now it should allow objects that have a key of a string type (assuming that I'm understanding schemas correctly ;) )

codegold79 (approved):

yharish991 (approved):

danniel1205 (approved):

PR #3724, opened 2022-10-20
TKR webhook should not reject a paused cluster.

HanFa:

A fix cherrypicked from https://github.com/vmware-tanzu/tanzu-framework/pull/3557

Generally, the webhook should always admit a Cluster resource that is paused.

_**Oracle MC deletion is blocked without this change**_

When deleting a management cluster, the CAPI Cluster resource will be first paused an moved to the bootstrap kind cluster. During this process, we can not guarantee that the cluster's metadata is resolve-able, because its  corresponding TKR might not be there yet.

In fact, the webhook is rejecting the Cluster CR when deleting the MC, if that MC created was using `--additional-manifests <some-tkr-url>`. This TKR from manifests won't be presented in the Bootstrap Kind cluster during deletion.

Fixes #

Webhook unit test

### Release note

```release-note
TKR webhook should not reject a paused cluster.
```

### DIFF
--- a/tkr/webhook/cluster/tkr-resolver/cluster/cluster.go
+++ b/tkr/webhook/cluster/tkr-resolver/cluster/cluster.go
@@ -65,6 +65,10 @@ func (cw *Webhook) Handle(ctx context.Context, req admission.Request) admission.
 }
 
 func (cw *Webhook) getClusterClass(ctx context.Context, cluster *clusterv1.Cluster) (*clusterv1.ClusterClass, *admission.Response) {
+	if cluster.Spec.Paused {
+		return nil, respPtr(admission.Allowed("Doing nothing. Cluster is paused."))
+	}
+
 	if !cluster.GetDeletionTimestamp().IsZero() {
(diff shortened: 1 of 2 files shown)


HanFa:
@imikushin @randomvariable any thoughts on this proposed fix?

imikushin (approved):
Looks good! Thanks @HanFa

tkr/webhook/cluster/tkr-resolver/cluster/cluster.go
```diff
@@ -65,6 +65,10 @@ func (cw *Webhook) Handle(ctx context.Context, req admission.Request) admission.
 }
 
 func (cw *Webhook) getClusterClass(ctx context.Context, cluster *clusterv1.Cluster) (*clusterv1.ClusterClass, *admission.Response) {
+	if cluster.Spec.Paused {
```
HanFa, line 68: Suggested by @imikushin 

During management cluster deletion, the CLI should replicate TKRs and OSimages (specified by the cluster object label) from MC to the cleanup cluster.

PR #3750, opened 2022-10-25
Update nodepool cc to default top level variable

tenczar:
This change adds a top level empty array for the nodePoolLabels variable so that any nodePoolLabels override in machineDeployments will not fail with a 'no top-level variable found' error.

Fixes a regression where node pool cli feature would fail for clusterclass based clusters if the cluster topology did not define a top level variable for `nodePoolLabels`

Fixes #

### Release note

```release-note

```

### DIFF
--- a/tkg/client/machine_deployment_cc.go
+++ b/tkg/client/machine_deployment_cc.go
@@ -11,6 +11,7 @@ import (
 	capi "sigs.k8s.io/cluster-api/api/v1beta1"
 
 	"github.com/pkg/errors"
+
 	"github.com/vmware-tanzu/tanzu-framework/tkg/clusterclient"
 	"github.com/vmware-tanzu/tanzu-framework/util/topology"
 )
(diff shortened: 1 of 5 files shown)


tkg/client/machine_deployment_cc.go
```diff
@@ -24,6 +24,23 @@ func DoSetMachineDeploymentCC(clusterClient clusterclient.Client, cluster *capi.
 		return errors.New("cluster topology workers are not set. please repair your cluster before trying again")
 	}
 
+	var nodePoolVarFound bool
+	for i := range cluster.Spec.Topology.Variables {
+		if cluster.Spec.Topology.Variables[i].Name == "nodePoolLabels" {
+			nodePoolVarFound = true
+		}
+	}
```
imikushin, line 32: We helpers to make it easier to read and write cluster variables: see "github.com/vmware-tanzu/tanzu-framework/util/topology" package

tenczar: The util function in that package is for ClusterClass. I'm operating directly on a Cluster, should I update the utils to include a similar function for Clusters?

imikushin (approved):
LGTM

PR #3755, opened 2022-10-26
Replicate TKR and OSImage to cleanup cluster during management cluster deletion.

HanFa:

A follow-up of https://github.com/vmware-tanzu/tanzu-framework/pull/3724, cherrypicked from https://github.com/vmware-tanzu/tanzu-framework/pull/3557

During the management cluster deletion, if the user has brought their own TKR and OSImages, then in the cleanup cluster, the moved Cluster CR will be rejected by the TKR cluster webhook. A sample error looks like,

```
error: clusters.cluster.x-k8s.io "test-lb-cluster" could not be patched: admission webhook "tkr-resolver-cluster-webhook.tanzu.vmware.com" denied the request: could not resolve TKR/OSImage for controlPlane, machineDeployments: [md-0], q
uery: {controlPlane: {k8sVersionPrefix: 'v1.23.8+vmware.2', tkrSelector: '', osImageSelector: 'oci-region=us-phoenix-1,os-name=ubuntu'}, machineDeployments: [{k8sVersionPrefix: 'v1.23.8+vmware.2', tkrSelector: '', osImageSelector: 'oci-
region=us-phoenix-1,os-name=ubuntu'}]}, result: {controlPlane: {k8sVersion: '', tkrName: '', osImagesByTKR: map[]}, machineDeployments: [{k8sVersion: '', tkrName: '', osImagesByTKR: map[]}]}
You can run `kubectl replace -f /tmp/kubectl-edit-1292888427.yaml` to try this update again.
``` 

This is because the TKR Cluster webhook rejects the moved Cluster CR because of lacking the self-brought TKR and OSImages in the cleanup cluster.

Therefore, when performing region deletion, the tanzu CLI should be responsible for replicating the TKR and OSImages that is associated with the moved management cluster. This PR adds this extra step.

Fixes #

With this fix presented at https://github.com/vmware-tanzu/tanzu-framework/pull/3557, the Oracle management cluster is able to be deleted successfully and the webhook no longer rejects with moved Cluster.

### Release note

```release-note
package-based-lcm: Replicate TKR and OSImage to cleanup cluster during management cluster deletion.
```

### DIFF
--- a/tkg/client/delete_region.go
+++ b/tkg/client/delete_region.go
@@ -168,7 +168,7 @@ func (c *TkgClient) DeleteRegion(options DeleteRegionOptions) error { //nolint:f
 
 		isStartedRegionalClusterDeletion = true
 
-		log.Info("Moving Cluster API objects from management cluster to cleanup cluster...")
+		log.Info("Moving TKR and Cluster API objects from management cluster to cleanup cluster...")
 		regionalClusterKubeConfigPath, err := regionalClusterClient.ExportCurrentKubeconfigToFile()
 		if err != nil {
 			return errors.Wrap(err, "unable to export management cluster's kubeconfig")
@@ -176,6 +176,11 @@ func (c *TkgClient) DeleteRegion(options DeleteRegionOptions) error { //nolint:f
 		defer func() {
 			_ = utils.DeleteFile(regionalClusterKubeConfigPath)
 		}()
+
+		// Copy coresponding TKR and OSImages from management cluster to cleanup cluster for all namespaces
+		if err = c.CopyNeededTKRAndOSImages(regionalClusterClient, cleanupClusterClient); err != nil {
+			return errors.Wrap(err, "unable to copy coresponding TKR and OSImages from management cluster to cleanup cluster")
+		}
 
 		// Move all Cluster API objects from management cluster to cleanup cluster for all namespaces
 		if err = c.MoveObjects(regionalClusterKubeConfigPath, cleanupClusterKubeconfigPath, regionalClusterNamespace); err != nil {
--- a/tkg/client/init.go
+++ b/tkg/client/init.go
@@ -14,6 +14,7 @@ import (
 	"github.com/go-openapi/swag"
 	"github.com/juju/fslock"
 	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	capav1beta1 "sigs.k8s.io/cluster-api-provider-aws/api/v1beta1"
@@ -26,6 +27,7 @@ import (
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	crtclient "sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/vmware-tanzu/tanzu-framework/apis/run/v1alpha3"
 	"github.com/vmware-tanzu/tanzu-framework/cli/runtime/config"
 	"github.com/vmware-tanzu/tanzu-framework/tkg/clusterclient"
 	"github.com/vmware-tanzu/tanzu-framework/tkg/constants"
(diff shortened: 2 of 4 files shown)


tkg/client/init.go
```diff
@@ -456,6 +456,30 @@ func (c *TkgClient) MoveObjects(fromKubeconfigPath, toKubeconfigPath, namespace
...
+func (c *TkgClient) MoveTKRAndOSImages(fromClusterClient, toClusterClient clusterclient.Client) error {
+	tkr, err := fromClusterClient.GetClusterResolvedTanzuKubernetesRelease()
+	if tkr == nil {
+		return err // error occurs or management cluster does not have resolved TKR
+	}
+	osImages, err := fromClusterClient.GetClusterResolvedOSImages()
+	if err != nil {
+		return err
+	}
+	tkr.SetResourceVersion("")
```
XudongLiuHarold, line 470: It is better to recreate tkr object with only copy name/namespace in metadata & whole tkr.spec?

Since it may have creation timestamp, annotation, labels and etc in metadata.

tkg/client/init.go
```diff
@@ -456,6 +456,30 @@ func (c *TkgClient) MoveObjects(fromKubeconfigPath, toKubeconfigPath, namespace
...
+	if err := toClusterClient.CreateResource(tkr, tkr.Name, tkr.Namespace); err != nil {
+		return err
+	}
+	for _, osImage := range osImages {
+		osImage.SetResourceVersion("")
```
XudongLiuHarold, line 475: Same here

tkg/client/init.go
HanFa: this should be fine. checkout how CAPI performs the move [here](https://github.com/kubernetes-sigs/cluster-api/blob/1bf70d5c24004977844b9da7570853c32e29c6b3/cmd/clusterctl/client/cluster/mover.go#L865). it is getting the object from cluster API server like what we do here (instead of constructing one)

tkg/client/delete_region.go
```diff
@@ -177,6 +177,11 @@ func (c *TkgClient) DeleteRegion(options DeleteRegionOptions) error { //nolint:f
 			_ = utils.DeleteFile(regionalClusterKubeConfigPath)
 		}()
 
+		// Move coresponding TKR and OSImages from management cluster to cleanup cluster for all namespaces
+		if err = c.MoveTKRAndOSImages(regionalClusterClient, cleanupClusterClient); err != nil {
```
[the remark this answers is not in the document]
HanFa: it's intended for use cases other than OCI, for example, when user applied their own TKR with OSImages, they will hit the same issue when deleting it.

tkg/client/init.go
```diff
@@ -456,6 +456,30 @@ func (c *TkgClient) MoveObjects(fromKubeconfigPath, toKubeconfigPath, namespace
 	return c.clusterctlClient.Move(moveOptions)
 }
 
+// MoveTKRAndOSImages moves the resolved TKR and associated OSImage CR for a Cluster resource
+// from namespaceto a target management
+func (c *TkgClient) MoveTKRAndOSImages(fromClusterClient, toClusterClient clusterclient.Client) error {
```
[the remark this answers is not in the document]
HanFa: yea, I would probaly consider renaming it since it only performs replication and not deleting anything from MC

tkg/clusterclient/clusterclient.go
```diff
@@ -1206,19 +1211,61 @@ func (c *client) GetBomConfigMap(tkrNameLabel string) (corev1.ConfigMap, error)
...
+	}
+	return &tkr, nil
+}
+
+// GetClusterResolvedOSImages get a list of OSImage resource from the resolved TKR
+func (c *client) GetClusterResolvedOSImages() ([]runv1alpha3.OSImage, error) {
+	tkr, err := c.GetClusterResolvedTanzuKubernetesRelease()
+	if err != nil {
+		return []v1alpha3.OSImage{}, nil
+	}
```
imikushin, line 1259: No need to call `c.GetClusterResolvedTanzuKubernetesRelease()`: just pass `tkr` as an argument. 

Also, return `[]*runv1alpha3.OSImage` instead of `[]runv1alpha3.OSImage`: to avoid unnecessary copying (e.g. when constructing the list) and possible bugs when iterating over the slice.

```suggestion
func (c *client) GetClusterResolvedOSImages(tkr *runv1alpha3.TanzuKubernetesRelease) ([]*runv1alpha3.OSImage, error) {
```

tkg/client/init.go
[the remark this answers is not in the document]
HanFa: will change to `CopyNeededTKRAndOSImages`.

XudongLiuHarold (approved):
LGTM

imikushin (approved):
LGTM

randomvariable (approved):

PR #3756, opened 2022-10-26
Fix makefile comments

mpanchajanya:

Fix extra space caused by inline comment

Fixes #

Run and Pass the Pipeline tests for any change in Makefile


marckhouzam (approved):
Assuming CI passes this LGTM
Thanks for the quick fix @mpanchajanya

codegold79 (approved):
Thank you for the fix

anujc25 (approved):

PR #3762, opened 2022-10-26
Remove extra space in no_proxy definition

lubronzhan:
Creating management cluster with proxy will fail. Since there are extra space in the no_proxy settings, making the kubeadm command to error out, thus kube-apiserver is not getting correct NO_PROXY value

Fixes #https://github.com/vmware-tanzu/tanzu-framework/issues/3763

### Release note

```release-note
Remove extra space in no_proxy in clusterclass template
```


lubronzhan:
Wrong target branch

PR #3823, opened 2022-11-02
Correct the label selector of MachineDeployment in CLI to support backup/restore management cluster with Velero

zhouke1991:
Currently, the CLI sets the label selector of MachineDeployment as follows, the selectors are not unique for each MachineDeployment:
```
apiVersion: cluster.x-k8s.io/v1alpha3
kind: MachineDeployment
metadata: 
  labels: 
    cluster.x-k8s.io/cluster-name: '${ CLUSTER_NAME }'
  name: '${ CLUSTER_NAME }-md-0'
  namespace: '${ NAMESPACE }'
spec: 
  clusterName: '${ CLUSTER_NAME }'
  replicas: ${ WORKER_MACHINE_COUNT }
  selector: 
    matchLabels: 
      cluster.x-k8s.io/cluster-name: '${ CLUSTER_NAME }'
```
As Velero removes the OwnerReferences when restoring, this triggers MachineDeployment to adopt MachineSets which not belong to it and causes the actual MachineSets count doesn't match the desired replicas and causes the restoration failure.

### DIFF
--- a/providers/infrastructure-vsphere/v1.4.1/ytt/base-template.yaml
+++ b/providers/infrastructure-vsphere/v1.4.1/ytt/base-template.yaml
@@ -249,14 +249,12 @@ spec:
   clusterName: '${ CLUSTER_NAME }'
   replicas: ${ WORKER_MACHINE_COUNT }
   selector:
-    matchLabels:
-      cluster.x-k8s.io/cluster-name: '${ CLUSTER_NAME }'
+    matchLabels: {}
   strategy:
     type: RollingUpdate
   template:
     metadata:
       labels:
-        cluster.x-k8s.io/cluster-name: '${ CLUSTER_NAME }'
         node-pool: "${CLUSTER_NAME}-worker-pool"
     spec:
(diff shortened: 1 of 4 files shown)


zhouke1991:
//usebom:sandbox/v1.7.0-zshippable/433578160030751737/tkg-compatibility:v12

wjun (approved):
LGTM

providers/infrastructure-vsphere/v1.4.1/ytt/base-template.yaml
```diff
@@ -249,14 +249,12 @@ spec:
   clusterName: '${ CLUSTER_NAME }'
   replicas: ${ WORKER_MACHINE_COUNT }
   selector:
-    matchLabels:
-      cluster.x-k8s.io/cluster-name: '${ CLUSTER_NAME }'
+    matchLabels: {}
```
jayunit100, line 252: how does this save us? maybe bc now nothing matches?

PR #3830, opened 2022-11-03
[WIP] seperate cc packages from meta package

hongtaiwang:
remove tkg-clusterclass package

move object-propagation out of tkg-clusterclass and create a new package for it

Fixes #

### Release note

```release-note

```


packages/tkg-clusterclass/bundle/config/packageinstalls/tkg-infra-clusterclass-pi.yaml
```diff
@@ -1,74 +0,0 @@
-#@ load("@ytt:overlay", "overlay")
```
chrischdi, line 1: By removing this: 

Would there still be a possibility to write a single overlay which would get applied for all clusterclasses (without copying it to every clusterclass?)

We make use of that functionality in #3862 🤔

PR #3833, opened 2022-11-03
Fix vCenter connection validation error on classy cluster manifest

DanielXiao:
Tanzu cli fails to validate vcenter connection when the config file is a classy cluster manifest. See the following error: Error: workload cluster configuration validation failed: vSphere config validation failed: failed to get VC client: failed to get VSPHERE_USERNAME

There is no cluster variable which can be mapped to them so we need to retrieve them from VSphereCluster's identityRef.

Fixes #

### Release note

```release-note

```

### DIFF
--- a/tkg/tkgctl/create_cluster.go
+++ b/tkg/tkgctl/create_cluster.go
@@ -160,7 +160,7 @@ func (t *tkgctl) processManagementClusterInputFile(ir *InitRegionOptions) (bool,
 			return isInputFileClusterClassBased, err
 		}
 		if isInputFileClusterClassBased {
-			err = t.processClusterObjectForConfigurationVariables(clusterobj)
+			err = t.processClusterObjectForConfigurationVariables(clusterobj, ir.ClusterConfigFile)
 			if err != nil {
 				return isInputFileClusterClassBased, err
 			}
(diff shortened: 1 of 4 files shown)


ydp:
LGTM

tkg/tkgctl/helpers.go
```diff
@@ -232,6 +275,13 @@ func (t *tkgctl) processClusterObjectForConfigurationVariables(clusterObj unstru
 	}
 	legacyVarMap[constants.ConfigVariableInfraProvider] = providerName
 
+	// VSPHERE_USERNAME and VSPHERE_PASSWORD do not have mapping to any Cluster variables and should be retrieved from VSphereCluster's identityRef
```
wjun, line 278: Is that possible to add them into cluster variables? Just try to understand which approach is better.

DanielXiao: We use Secret as vSphereCluster's identityRef which is not a Cluster-API template so Cluster variable can not patch it.

tkg/tkgctl/create_cluster.go
```diff
@@ -160,7 +160,7 @@ func (t *tkgctl) processManagementClusterInputFile(ir *InitRegionOptions) (bool,
 			return isInputFileClusterClassBased, err
 		}
 		if isInputFileClusterClassBased {
-			err = t.processClusterObjectForConfigurationVariables(clusterobj)
+			err = t.processClusterObjectForConfigurationVariables(clusterobj, ir.ClusterConfigFile)
```
wjun, line 163: Since this is only invoked by mc creation, how to pass it to workload cluster creation?

DanielXiao: When users create a workload cluster with a legacy configuration file, cli terminate creation and generate a classy cluster manifest. Users need to run `tanzu cluster create --file <generated manifest file>`. So this is mainly for workload cluster creation.

wjun: But why is this api only invoked in a mc creation api?

DanielXiao: It is invoked in both mc and wc creation. Please refer to line 186

wjun (approved):

PR #3850, opened 2022-11-07
Make feature plugin work for Feature and FeatureGate APIs in core.tanzu.vmware.com

codegold79:

This PR updates the Tanzu CLI plugin for managing Features on a cluster. It required updating the FeatureGate client in order to make the plugin work. Unit tests for the plugin and client have been updated.

Fixes #

Unit tests are passing, but the plugin code test coverage is less than 33% while the client test coverage rounds up to 70%. Increased unit test coverage for the plugin will happen in a future PR.

I have done some manual testing by setting up the new Feature controller and Feature(Gate) APIs. I have documented the commands I ran in a gist: https://gist.github.com/codegold79/0b4e76d437e58173dc17fbb2047e2e69

Please ignore the FeatureGate status section as the logic for the controller is not in the scope of this Tanzu CLI Feature plugin PR.

### Release note

```release-note
Update the Tanzu Feature CLI plugin and FeatureGate client to work with the new (core.tanzu.vmware.com v1alpha2) Feature and FeatureGate APIs. This change is a breaking change as the Tanzu Feature CLI will no longer work with config.tanzu.vmware.com v1alpha1. Please upgrade Feature and FeatureGate resources to use the new v1alpha2 CRDs.
```

### DIFF
--- a/cmd/cli/plugin/feature/README.md
+++ b/cmd/cli/plugin/feature/README.md
@@ -1,20 +1,20 @@
-# feature
+# Feature
 
-Feature plugin lets you operate on Features and FeatureGates
+Feature plugin gives access to features using featuregates.
 
 ## Usage
 
-Feature plugin has 3 commands
+Feature plugin has three commands:
 
 1. list - allows to list the features that are gated by a particular
    FeatureGate.
 2. activate - allows to activate a feature.
 3. deactivate - allows to deactivate a feature.
 
-By default, Feature plugin operates on Features that are gated by `tkg-system`
-FeatureGate, but that can be changed by specifying `featuregate` flag.
+Feature plugin is able to list all discoverable features on the cluster.
+Optionally, a FeatureGate may be specified by using the `featuregate` flag.
 
-Ex:
+Example:
 
 ```sh
 # list Features associated with tkg-system FeatureGate
@@ -45,26 +45,25 @@ Use "tanzu feature [command] --help" for more information about a command.
 
 ```sh
 >>> tanzu feature list --help
-List Features
+List features
 
 Usage:
   tanzu feature list [flags]
 
 Examples:
   
-    # List a clusters Features
+    # List feature(s) in the cluster.
     tanzu feature list --activated
-    tanzu feature list --unavailable
     tanzu feature list --deactivated
 
 Flags:
-  -a, --activated            List only activated Features
-  -d, --deactivated          List only deactivated Features
-  -e, --extended             Include extended output. Higher latency as it requires more API calls.
-  -f, --featuregate string   List Features gated by a particular FeatureGate (default "tkg-system")
-  -h, --help                 help for list
-  -o, --output string        Output format (yaml|json|table)
-  -u, --unavailable          List only Features specified in the gate but missing from cluster
+  -a, --activated                             List only activated features
+  -d, --deactivated                           List only deactivated features
+  -e, --extended                              Include extended output
+  -f, --featuregate string                    List features gated by a particular FeatureGate
+  -h, --help                                  Help for list
+  -x, --include-experimental                  Allows displaying experimental features
+  -o, --output string                         Output format (yaml|json|table)
 ```
 
 ### activate command
--- a/cmd/cli/plugin/feature/activate.go
+++ b/cmd/cli/plugin/feature/activate.go
@@ -1,4 +1,4 @@
-// Copyright 2021 VMware, Inc. All Rights Reserved.
+// Copyright 2022 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package main
@@ -6,16 +6,22 @@ package main
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/spf13/cobra"
 
+	corev1alpha2 "github.com/vmware-tanzu/tanzu-framework/apis/core/v1alpha2"
 	"github.com/vmware-tanzu/tanzu-framework/featuregates/client/pkg/featuregateclient"
+)
+
+var (
+	userAllowsVoidingWarranty bool
 )
 
 // FeatureActivateCmd is for activating Features
 var FeatureActivateCmd = &cobra.Command{
 	Use:   "activate <feature>",
-	Short: "Activate Features",
+	Short: "Activate features",
 	Args:  cobra.ExactArgs(1),
 	Example: `
 	# Activate a cluster Feature
@@ -24,22 +30,112 @@ var FeatureActivateCmd = &cobra.Command{
 }
 
 func init() {
-	FeatureActivateCmd.Flags().StringVarP(&featuregate, "featuregate", "f", "tkg-system", "Activate a Feature gated by a particular FeatureGate")
+	FeatureActivateCmd.Flags().BoolVar(&userAllowsVoidingWarranty, "permanentlyVoidAllSupportGuarantees", false, "Allow for the permanent voiding of all support guarantees for this environment. For some features, e.g. experimental features, if a user sets the activation status to one that does not match the default activation, all support guarantees for this environment will be permanently voided.")
 }
 
 func featureActivate(cmd *cobra.Command, args []string) error {
 	featureName := args[0]
-	featureGateClient, err := featuregateclient.NewFeatureGateClient()
+
+	fgClient, err := featuregateclient.NewFeatureGateClient()
 	if err != nil {
-		return fmt.Errorf("couldn't get featureGateRunner: %w", err)
+		return fmt.Errorf("could not get FeatureGate client: %w", err)
 	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), contextTimeout)
 	defer cancel()
 
-	if err := featureGateClient.ActivateFeature(ctx, featureName, featuregate); err != nil {
-		return fmt.Errorf("couldn't activate feature %s: %w", featureName, err)
+	feature, err := fgClient.GetFeature(ctx, featureName)
+	if err != nil {
+		return fmt.Errorf("could not get Feature %s: %w", featureName, err)
 	}
-	cmd.Printf("Feature %s Activated", featureName)
+
+	gates, err := fgClient.GetFeatureGateList(ctx)
+	if err != nil {
+		return fmt.Errorf("could not get FeatureGate List: %w", err)
+	}
+
+	gateName, featRef := featuregateclient.FeatureRefFromGateList(gates, featureName)
+
+	var proceedWithVoidingWarranty bool
+	if willWarrantyBeVoided(featRef, feature) {
+		// The warranty will be voided with the request, so check that user allows it.
+		var userAllows *bool
+		if cmd.Flags().Changed("permanentlyVoidAllSupportGuarantees") {
+			userAllows = &userAllowsVoidingWarranty
+		}
+		proceedWithVoidingWarranty, err = userGivesPermissionToVoidWarranty(feature, userAllows)
+		if err != nil {
+			return fmt.Errorf("could not get user permission to void warranty for Feature %s: %w", featureName, err)
+		}
+	}
+
+	err = fgClient.ActivateFeature(ctx, featureName, proceedWithVoidingWarranty)
+	if err != nil {
+		return fmt.Errorf("could not activate Feature %s gated by FeatureGate %s: %w", featureName, gateName, err)
+	}
+
+	displayActivationWarnings(feature)
+
+	cmd.Printf("Feature %s gated by FeatureGate %s is activated.\n", featureName, gateName)
 	return nil
+}
+
+// displayActivationWarnings warns the user that technical preview features are
+// unstable and lack support.
+func displayActivationWarnings(feature *corev1alpha2.Feature) {
+	if feature.Spec.Stability == corev1alpha2.TechnicalPreview {
+		fmt.Printf("Warning: Technical preview features are not ready, but are not believed to be dangerous. The feature itself is unsupported, but activating technical preview features does not affect the support status of the environment. Use at your own risk.\n\n")
+	}
+}
+
+// willWarrantyBeVoided checks that activating the Feature will cause warranty to be voided.
+// Warranty will be voided only if all the following conditions are met:
+//   - The stability policy's VoidsWarranty field value is true. This means changing the activation
+//     set point away from default activation (in this case, default activate will have to be false)
+//     will void the warranty.
+//   - The warranty is not already void. If it's void, then it cannot be voided as it's already void.
+//   - The current Feature activation set point in the FeatureGate reference is false. If the opposite is
+//     true--that the Feature is already set to be activated--then there is nothing to change and
+//     the warranty will not be voided.
+//   - Activating the Feature will set it different than its default activation setting. This means the
+//     default activation of a Feature must be false for the warranty to be voided. On the other hand,
+//     if the default activation is true, then activating the Feature will not void the warranty.
+//
+// If all of the above are true, then the warranty will be voided and the function returns true.
+// Otherwise, if any are false, the warranty will not be voided and the function returns false.
+func willWarrantyBeVoided(ref corev1alpha2.FeatureReference, feature *corev1alpha2.Feature) bool {
+	stability := feature.Spec.Stability
+	policy := corev1alpha2.GetPolicyForStabilityLevel(stability)
+	return policy.VoidsWarranty && !ref.PermanentlyVoidAllSupportGuarantees && !ref.Activate && !policy.DefaultActivation
+}
+
+func userGivesPermissionToVoidWarranty(feature *corev1alpha2.Feature, userAllowsByFlag *bool) (bool, error) {
+	if userAllowsByFlag == nil {
+		// Request user permission interactively to void the warranty if not already set by flag.
+		fmt.Printf("Warning: activating %q, a %s Feature will irrevocably void all support guarantees for this environment. You will need to recreate the environment to return to a supported state.\nWould you like to continue [y/N]?", feature.Name, feature.Spec.Stability)
+		return userAllowsByInteractiveCLI()
+	}
+	// If the user did pass a value to the flag, use the user's input. No interactive user input is needed.
+	return *userAllowsByFlag, nil
+}
+
+func userAllowsByInteractiveCLI() (bool, error) {
+	var userAnswer string
+	if n, err := fmt.Scanln(&userAnswer); err != nil {
+		if n == 0 {
+			return false, nil
+		}
+		return false, fmt.Errorf("could not read user input: %w", err)
+	}
+
+	userAnswer = strings.TrimSpace(strings.ToLower(userAnswer))
+	switch {
+	case userAnswer == "yes" || userAnswer == "y":
+		return true, nil
+	case userAnswer == "" || userAnswer == "no" || userAnswer == "n":
+		return false, nil
+	default:
+		fmt.Printf("Please respond with %q or %q\n", "y", "n")
+		return userAllowsByInteractiveCLI()
+	}
 }
--- a/cmd/cli/plugin/feature/deactivate.go
+++ b/cmd/cli/plugin/feature/deactivate.go
@@ -22,22 +22,21 @@ var FeatureDeactivateCmd = &cobra.Command{
 	tanzu feature deactivate myfeature`,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		featureName := args[0]
-		featureGateClient, err := featuregateclient.NewFeatureGateClient()
+
+		fgClient, err := featuregateclient.NewFeatureGateClient()
 		if err != nil {
-			return fmt.Errorf("couldn't get a featureGateRunner: %w", err)
+			return fmt.Errorf("could not get FeatureGate client: %w", err)
 		}
 
 		ctx, cancel := context.WithTimeout(context.Background(), contextTimeout)
 		defer cancel()
 
-		if err := featureGateClient.DeactivateFeature(ctx, featureName, featuregate); err != nil {
-			return fmt.Errorf("couldn't deactivate feature %s: %w", featureName, err)
+		gateName, err := fgClient.DeactivateFeature(ctx, featureName)
+		if err != nil {
+			return fmt.Errorf("could not deactivate Feature %s gated by FeatureGate %s: %w", featureName, gateName, err)
 		}
-		cmd.Printf("Feature %s Deactivated", featureName)
+
+		cmd.Printf("Feature %s gated by FeatureGate %s is deactivated.\n", featureName, gateName)
 		return nil
 	},
-}
-
-func init() {
-	FeatureDeactivateCmd.Flags().StringVarP(&featuregate, "featuregate", "f", "tkg-system", "Deactivate Feature gated by a particular FeatureGate")
 }
--- a/cmd/cli/plugin/feature/list.go
+++ b/cmd/cli/plugin/feature/list.go
@@ -1,4 +1,4 @@
-// Copyright 2021 VMware, Inc. All Rights Reserved.
+// Copyright 2022 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package main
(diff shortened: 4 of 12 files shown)


featuregates/client/pkg/featuregateclient/client.go
```diff
@@ -8,33 +8,34 @@ import (
...
 // FeatureGateClient defines methods to interact with FeatureGate resources
 type FeatureGateClient struct {
-	c client.Client
+	crClient client.Client
 }
 
-// NewFeatureGateClient returns an instance of FeatureGateClient.
-func NewFeatureGateClient(options ...Option) (*FeatureGateClient, error) {
+// New returns an instance of FeatureGateClient.
+func New(options ...Option) (*FeatureGateClient, error) {
```
yharish991, line 27: we have to be careful when changing the names of public functions. We first need to deprecate and remove them later

featuregates/client/pkg/featuregateclient/client.go
```diff
@@ -45,133 +46,217 @@ type Option func(*FeatureGateClient) *FeatureGateClient
...
-		return fmt.Errorf("couldn't get feature %s: %w", featureName, err)
+		return fmt.Errorf("get Feature %s: %w", featureName, err)
+	}
+
+	gates, err := f.GetFeatureGateList(ctx)
+	if err != nil {
+		return fmt.Errorf("get FeatureGateList: %w", err)
+	}
+
+	gateName, featRef := FeatureRefFromGateList(gates, featureName)
```
yharish991, line 129: same comment as above you can use the existing functions `GetFeatureGateForFeature` and `GetFeatureReferenceFromFeatureGate` from util package

featuregates/client/pkg/featuregateclient/client.go
```diff
@@ -45,133 +46,217 @@ type Option func(*FeatureGateClient) *FeatureGateClient
...
+
+	if featRef.Activate {
```
yharish991, line 131: featureref just holds the intent, you should check for `feature.status.activated`

featuregates/client/pkg/featuregateclient/validation.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// featureExistsInOneFeaturegate checks that the Feature exists in one and only one FeatureGate.
+func featureExistsInOneAndOnlyOneFeaturegate(gates *core.FeatureGateList, featureName string) error {
+	var n int
+
+	for i := range gates.Items {
+		for _, ref := range gates.Items[i].Spec.Features {
+			if ref.Name == featureName {
+				n++
```
yharish991, line 32: can we simply check for `n`'s value here and return?

yharish991 (changes requested):
This is a Round 1 review, need to review this more thoroughly

cmd/cli/plugin/feature/activate.go
```diff
@@ -24,22 +30,89 @@ var FeatureActivateCmd = &cobra.Command{
...
+	feature, err := fgClient.GetFeature(ctx, featureName)
+	if err != nil {
+		return fmt.Errorf("get Feature %s: %w", featureName, err)
 	}
-	cmd.Printf("Feature %s Activated", featureName)
+
+	displayActivationWarnings(feature)
+
+	gates, err := fgClient.GetFeatureGateList(ctx)
+	_, featRef := featuregateclient.FeatureRefFromGateList(gates, featureName)
```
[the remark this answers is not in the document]
codegold79: But I don't want FeatureGate. I want the FeatureReference from FeatureGate. Isn't that different?

cmd/cli/plugin/feature/activate.go
```diff
@@ -24,22 +30,89 @@ var FeatureActivateCmd = &cobra.Command{
...
+
+	var userAllows *bool
+	if cmd.Flag("userAllowsVoidingWarranty").Changed {
+		userAllows = &userAllowsVoidingWarranty
+	}
+	warrantyVoidAllowed := isVoidingWarrantyAllowed(featRef, feature, userAllows)
```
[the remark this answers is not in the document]
codegold79: Nope, because `userAllowsVoidingWarranty` is boolean and `userAllows` is pointer to boolean. I need that pointer to see if user set it or not. I don't want to assume `false` if user didn't set it. Why not? Because I need to know if I should ask the user if ze wants to allow voiding warranty. If I assume it was false because of response, I don't ask the user if the user would like to void the warranty.

cmd/cli/plugin/feature/activate.go
```diff
@@ -24,22 +30,89 @@ var FeatureActivateCmd = &cobra.Command{
...
+// isVoidingWarrantyAllowed checks if a warranty will be voided and if so, if it
+// is allowed by the user. Request for permission to void warranty only happens
+// if toggling the Feature activation will void the warranty. Otherwise, it is
+// considered not allowed.
+func isVoidingWarrantyAllowed(ref core.FeatureReference, feature *core.Feature, userAllowsByFlag *bool) bool {
+	stability := feature.Spec.Stability
+	policy := core.GetPolicyForStabilityLevel(stability)
+
+	// Check first if warranty will be voided if Feature is activated.
+	if policy.DefaultActivation == feature.Status.Activated && !ref.PermanentlyVoidAllSupportGuarantees && policy.VoidsWarranty {
```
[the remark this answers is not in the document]
codegold79: I agree with your intention, but not the suggested change in code. 

The `featureRef.Activate` value is the value before the toggle was made. If it matches the default before user toggles activation, that means the activation will deviate from default. So, when it's equal, it means the change will be away from default.

cmd/cli/plugin/feature/list.go
```diff
@@ -5,148 +5,158 @@ package main
...
+	return selectFeatures, nil
+}
 
-		// Many API calls to gather and join on Features
-		return listExtended(cmd, features)
+// isFeatureListable will prevent non-discoverable features from being listed. It will
+// also require a special flag to be able to display experimental features.
+func isFeatureListable(feat *core.Feature, inclExperimental bool) bool {
+	if feat == nil {
+		// Since the stability level information is only in the Feature resource,
```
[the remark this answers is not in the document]
codegold79: Yup, the nil check is to check for features that do not exist in the cluster.

cmd/cli/plugin/feature/list.go
```diff
@@ -5,148 +5,158 @@ package main
...
+			} else {
+				info.Available = true
+				info.Activated = feature.Status.Activated
+				info.Stability = feature.Spec.Stability
+				info.Description = feature.Spec.Description
+			}
+			info.Immutable = core.GetPolicyForStabilityLevel(info.Stability).Immutable
+
+			if isFeatureListable(feature, includeExperimental) {
+				switch {
```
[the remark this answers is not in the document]
codegold79: Yup. So all this is in a function called `filterFeatures`. I'll write in a comment explaining what it does. Something like,

```
// filterFeatures will determine, based on which flags the user provided, which
// features can be listed and which will be filtered out (not listed).
```

What the switch does is it looks at three different flags you can send into the `feature list` command.

```
switch {
				case activated && info.Activated:
					selectFeatures = append(selectFeatures, info)
				case deactivated && !info.Activated:
					selectFeatures = append(selectFeatures, info)
				case unavailable && !info.Available:
					selectFeatures = append(selectFeatures, info)
				}

```

The three flags are `--actiaved` (first switch case), `--deactivated` (second switch case), and `--unavailable` (third swich case). 

If flag is --activated and the FeatureRef shows it's supposed to be activated, we'll append to my collection of Features only the activated ones. The rest of the features will not show up. And so forth for the rest of the flags.

cmd/cli/plugin/feature/list.go
```diff
@@ -5,148 +5,158 @@ package main
...
+		return true
 	}
 
-	return listBasic(cmd, features)
+	policy := core.GetPolicyForStabilityLevel(feat.Spec.Stability)
+	if !policy.Discoverable {
+		return false
+	}
+
+	if feat.Spec.Stability == core.Experimental && !inclExperimental {
```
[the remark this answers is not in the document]
codegold79: Yup, it's discoverable, so if people want to see it, they can. But, by default, it should not be listed. By default, it cannot be listed. The user has to use a special flag, in this case, it is `include-experimental` and the value of it is saved in `inclExperimental`. If that variable is false, don't list the experimental feature. However, if the user explicitly makes it true, then we can list it.

featuregates/client/pkg/featuregateclient/client.go
```diff
@@ -8,33 +8,34 @@ import (
...
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	configv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/config/v1alpha1"
+	core "github.com/vmware-tanzu/tanzu-framework/apis/core/v1alpha2"
```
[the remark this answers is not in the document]
codegold79: I don't think it helps readability to add version, but I guess it's consistent with how we name these for metav1, corev1, etc. But I thought just doing `core` without version would distinguish it from tanzu core vs k8s core. Anyway, if you think it's better to add version, let me know. I'll go ahead and replace all `core` with `corev1alpha`. Seems like a lot of letters that hinders reading and doesn't really help with reading, but maybe you have a different perspective.

featuregates/client/pkg/featuregateclient/client.go
codegold79: Aren't we going with a whole new API? But I see your point. I'll revert to the longer name.

featuregates/client/pkg/featuregateclient/client.go
codegold79: And again. I don't want to get FeatureGate using `GetFeatureGateForFeature`. I want to find the FeatureRef from a list of Feature Gates. It's backwards.

featuregates/client/pkg/featuregateclient/client.go
codegold79: Can't status change by other things? I want to make sure that the intent is already set to activated. If you do `ActivateFeature`, your goal is to set that you intend this Feature to be active or not. When I do a check on its current intent, and see it's already what the user wants, I'm not going to change the intent. Whether or not the status changes to the intent is up to the controller.

featuregates/client/pkg/featuregateclient/validation.go
codegold79: Well, yes, but here's why I didn't do that:
- if I return n, I'm forced to go through and count how many Feature Refs there are. I don't know that if I see more than 1, I can return.
- if I return n, I'll need to check in another function if n is 0 or more than 1. I guess it's separate, but I think it's easier to follow my intention all in one place. But maybe not.

So, the question is, do you think I should check if n is not 1 elsewhere and have a count of a particular Feature in a separate function? I can do that.

cmd/cli/plugin/feature/list.go
```diff
@@ -5,148 +5,158 @@ package main
...
 	tanzu feature list --activated
 	tanzu feature list --unavailable
 	tanzu feature list --deactivated`,
-	RunE: featureList,
+	RunE: printFeatures,
 }
 
 func init() {
-	FeatureListCmd.Flags().BoolVarP(&extended, "extended", "e", false, "Include extended output. Higher latency as it requires more API calls.")
-	FeatureListCmd.Flags().StringVarP(&featuregate, "featuregate", "f", "tkg-system", "List Features gated by a particular FeatureGate")
```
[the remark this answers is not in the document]
codegold79: Oops.  I didn't mean to delete that line. I even stilll use `featuregate` in the `filterFeatures` function.

cmd/cli/plugin/feature/activate.go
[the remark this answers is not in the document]
yharish991: void support warranty happens when activating a feature that is unstable, and the unstable features are always deactivated by default, so we just need to check the intent ie if the feature that voids warranty support is being activated or not

cmd/cli/plugin/feature/activate.go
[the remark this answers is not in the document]
codegold79: After speaking more on this with @yharish991 , we found that my `FeatureRefFromGateList` reduces the number of network calls to the API, which is a significant time cost that should not be incurred in this application. It reduces the need for getting the feature gate list every time the GetFeatureGateForFeature function is called. See below:

```
func GetFeatureGateForFeature(ctx context.Context, c client.Client, featureName string) (*corev1alpha2.FeatureGate, bool, error) {
	featureGateList := &corev1alpha2.FeatureGateList{}
	if err := c.List(ctx, featureGateList); err != nil {
		return nil, false, fmt.Errorf("could not list FeatureGate resources: %w", err)
	}

	for i, fg := range featureGateList.Items {
		for _, f := range featureGateList.Items[i].Spec.Features {
			if f.Name == featureName {
				return &fg, true, nil
			}
		}
	}
	return nil, false, nil
}
```

featuregates/client/pkg/featuregateclient/client.go
codegold79: Refer to previous mention of this comment for decision and resolution.

cmd/cli/plugin/feature/activate.go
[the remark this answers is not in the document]
codegold79: Thanks for explaining the logic here. Here's what we decided to do:
```
if policy.VoidsWarranty && !ref.PermanentlyVoidAllSupportGuarantees && !policy.DefaultActivation {
<warranty will be voided, check user permission>
```

cmd/cli/plugin/feature/list.go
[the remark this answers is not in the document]
codegold79: We discussed this in person further and I showed @yharish991 our design document where it shows experimental features should not be listed by default.

featuregates/client/pkg/featuregateclient/client.go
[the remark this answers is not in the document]
codegold79: After speaking with @yharish991, I decided it's better to have versions because what if the plugin had to support more than one version of the same group? In that case, we need to have the version in alias name.

cmd/cli/plugin/feature/activate.go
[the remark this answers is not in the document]
codegold79: I thought about this some more, and I have yet another solution. I left a very long comment to explain why. Here's what I ended up with: 

```
// willWarrantyBeVoided checks that activating the Feature will cause warranty to be voided.
// Warranty will be voided only if all the following conditions are met:
//   - The stability policy's VoidsWarranty field value is true. This means changing the activation
//     set point away from default activation (in this case, default activate will have to be false)
//     will void the warranty.
//   - The warranty is not already void. If it's void, then it cannot be voided as it's already void.
//   - The Feature activation set point in the FeatureGate reference is false. If the opposite is
//     true--that the Feature is already set to be activated--then there is nothing to change and
//     the warranty will not be voided.
//   - Activating the Feature will set it different than its default activation setting. This means the
//     default activation of a Feature must be false for the warranty to be voided. On the other hand,
//     if the Feature set point is false, and the default activation is true, then the request to activate
//     the Feature will not void warranty.
//
// If all of the above are true, then the warranty will be voided and the function returns true.
// Otherwise, if any are false, the warranty will not be voided and the function returns false.
func willWarrantyBeVoided(ref core.FeatureReference, feature *core.Feature) bool {
	stability := feature.Spec.Stability
	policy := core.GetPolicyForStabilityLevel(stability)
	return policy.VoidsWarranty && !ref.PermanentlyVoidAllSupportGuarantees && !ref.Activate && !policy.DefaultActivation
}
```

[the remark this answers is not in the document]
yharish991: is this `policy.VoidsWarranty && !ref.PermanentlyVoidAllSupportGuarantees && !ref.Activate && !policy.DefaultActivation` equivalent to ` policy.VoidsWarranty && !ref.PermanentlyVoidAllSupportGuarantees && policy.DefaultActivation != featureRef.Activate`

cmd/cli/plugin/feature/activate.go
```diff
@@ -24,22 +30,112 @@ var FeatureActivateCmd = &cobra.Command{
...
+		}
 	}
-	cmd.Printf("Feature %s Activated", featureName)
+
+	err = fgClient.ActivateFeature(ctx, featureName, proceedWithVoidingWarranty)
+	if err != nil {
+		return fmt.Errorf("could not activate Feature %s gated by FeatureGate %s: %w", featureName, gateName, err)
+	}
+
+	displayActivationWarnings(feature)
```
yharish991, line 77: can we display this message before activating?

cmd/cli/plugin/feature/activate.go
```diff
@@ -24,22 +30,112 @@ var FeatureActivateCmd = &cobra.Command{
...
+
+	gates, err := fgClient.GetFeatureGateList(ctx)
+	if err != nil {
+		return fmt.Errorf("could not get FeatureGate List: %w", err)
+	}
+
+	gateName, featRef := featuregateclient.FeatureRefFromGateList(gates, featureName)
+
+	var proceedWithVoidingWarranty bool
+	if willWarrantyBeVoided(featRef, feature) {
```
yharish991, line 60: where is the immutability check being done when activating the feature?

cmd/cli/plugin/feature/deactivate.go
```diff
@@ -22,22 +22,25 @@ var FeatureDeactivateCmd = &cobra.Command{
...
-		cmd.Printf("Feature %s Deactivated", featureName)
+
+		cmd.Printf("Feature %s gated by FeatureGate %s is deactivated.\n", featureName, gateName)
 		return nil
 	},
 }
 
 func init() {
-	FeatureDeactivateCmd.Flags().StringVarP(&featuregate, "featuregate", "f", "tkg-system", "Deactivate Feature gated by a particular FeatureGate")
+	FeatureDeactivateCmd.Flags().StringVarP(&featuregate, "featuregate", "f", "", "Deactivate Feature gated by a particular FeatureGate")
```
yharish991, line 45: should this flag be removed?

cmd/cli/plugin/feature/list.go
```diff
@@ -7,187 +7,206 @@ import (
...
-// FeatureListCmd is for activating Features
+// FeatureListCmd is for listing features in (a) featuregate(s).
 var FeatureListCmd = &cobra.Command{
 	Use:   "list",
-	Short: "List Features",
+	Short: "List features",
 	Args:  cobra.NoArgs,
 	Example: `
-	# List a clusters Features
+	# List feature(s) referenced in featuregate(s)
```
yharish991, line 29: we would also list features that are not part of featuregate(s), so i this can be reworded

cmd/cli/plugin/feature/list.go
```diff
@@ -7,187 +7,206 @@ import (
...
-	Available   bool
-	Immutable   bool
+	Name         string
+	Description  string
+	Stability    corev1alpha2.StabilityLevel
+	Immutable    bool
+	Discoverable bool
+	FeatureGate  string
+	Activated    bool
+	Listable     bool
```
yharish991, line 53: what is Listable for?

cmd/cli/plugin/feature/activate.go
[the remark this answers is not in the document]
codegold79: It's so complicated, I can't tell. Are both correct? If so, I think mine is easier to understand and if you don't understand, I have a very long comment to explain each part.

cmd/cli/plugin/feature/README.md
```diff
@@ -1,20 +1,20 @@
-# feature
+# Feature
 
-Feature plugin lets you operate on Features and FeatureGates
+Feature plugin lets you operate on features and featuregates.
```
rajathagasthya, line 3: `Feature plugin gives access to features using featuregates` or something like that?

cmd/cli/plugin/feature/README.md
```diff
@@ -45,26 +45,28 @@ Use "tanzu feature [command] --help" for more information about a command.
...
-  -f, --featuregate string   List Features gated by a particular FeatureGate (default "tkg-system")
-  -h, --help                 help for list
-  -o, --output string        Output format (yaml|json|table)
-  -u, --unavailable          List only Features specified in the gate but missing from cluster
+  -a, --activated                             List only activated features
+  -d, --deactivated                           List only deactivated features
+  -e, --extended                              Include extended output
+  -x, --include-experimental                  Allows displaying experimental features
+  -f, --featuregate string                    List features gated by a particular FeatureGate
+      --permanentlyVoidAllSupportGuarantees   Allow activating a non-stable Feature that will permanently void warranty support. Use with caution.
```
rajathagasthya, line 66: Why do we need this option for listing features? Or this is a documentation error?

cmd/cli/plugin/feature/activate.go
codegold79: When I did manual testing, having this message before activation did not work out, but after it did, worked great. The reason is because the line that does the activation might error out and cause the problem of having conflicting messages. As Jeff said, "Is that a warning or an error?" 

I think you should either have a warning or an error and not both and by having the activation warning after the activation ensures that the warning only shows up if there's no err. Here's what it looks like currently (shown in [a Gist)](https://gist.github.com/codegold79/0b4e76d437e58173dc17fbb2047e2e69),

```
$ ./build/feature activate small-cache
Warning: Technical preview features are not ready, but are not believed to be dangerous. The feature itself is unsupported, but activating technical preview features does not affect the support status of the environment. Use at your own risk.
Feature small-cache gated by FeatureGate tkg-system-b is activated.
```

cmd/cli/plugin/feature/activate.go
codegold79: It's done by the featuregate client in featuregates/client/pkg/featuregateclient/validation.go lines 51-61.

```bash
// featureActivationToggleAllowed checks if a Feature is considered immutable by its stability
// level and associated policy. Immutable means a Feature's activation setting cannot be toggled.
func featureActivationToggleAllowed(feature *corev1alpha2.Feature) error {
	stability := feature.Spec.Stability
	policy := corev1alpha2.GetPolicyForStabilityLevel(stability)

	if policy.Immutable {
		return fmt.Errorf("activation setting for Feature %s cannot be toggled as its stability level is %s: %w", feature.Name, stability, ErrTypeForbidden)
	}
	return nil
}
```

cmd/cli/plugin/feature/list.go
codegold79: I should probably rename it since it's confusing. I'm thinking `showInList` or something.

cmd/cli/plugin/feature/deactivate.go
rajathagasthya: Wouldn't this be a breaking change? Or are we okay with breaking it?

codegold79: Everything in this PR is a breaking change. This new iteration of Feature plugin has been changed to work with the new Feature(Gates) core.tanzu.vmware.com v1alpha2 API. So I guess we're ok with breaking it.

cmd/cli/plugin/feature/list.go
codegold79: I've changed it to `ShowInList` (exported like all other fields). I'll resolve this for now and you can suggest something else in another review comment.

cmd/cli/plugin/feature/README.md
codegold79: I changed it to `Feature plugin gives access to feature(s) that are gated by featuregate(s).`. I'll resolve for now and you can provide another suggestion in a separate review comment.

cmd/cli/plugin/feature/README.md
codegold79: `--permanentlyVoidAllSupportGuarantees` is required per Feature Graduation Process design.

codegold79: Oh, you're right. I shouldn't have this in the `feature list` command documentation. It's only applicable for `feature activate` command.

cmd/cli/plugin/feature/activate.go
codegold79: Per @yharish991's request, I have added a blank line in between the warning and Feature activation notice. It now looks like
```
$ ./build/feature activate small-cache
Warning: Technical preview features are not ready, but are not believed to be dangerous. The feature itself is unsupported, but activating technical preview features does not affect the support status of the environment. Use at your own risk.

Feature small-cache gated by FeatureGate tkg-system-b is activated.
```

cmd/cli/plugin/feature/activate.go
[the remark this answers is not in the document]
yharish991: i think this logic can be made more concise, but we can revisit this later

cmd/cli/plugin/feature/README.md
yharish991: > Feature plugin gives access to feature(s) that are gated by featuregate(s)
This is not always true, when can list features that are not part of any featuregate

cmd/cli/plugin/feature/activate.go
yharish991: I would like to display the warning message first and any errors later for a consistent user experience. kubectl does the same, for example, when you try to get a resource that doesn't exist of an API type that is deprecated, it first displays the warning message and error later.

Also, as a user when the warning message is displayed first, it gives me an idea if i want to further perform the operation if any error. Otherwise, if i see the error first, i would figure out how to fix it the error first to perform the operation and then later know that there is a warning when performing the operation. If i had known about the warning earlier, i would have changed my mind about not performing the operation at all.

yharish991: this is a non-blocker though

codegold79: Since it's a non-blocker, I'm going to resolve this comment and the warning/error message issue can be addressed another time.

cmd/cli/plugin/feature/README.md
codegold79: @yharish991, what do you think of @rajathagasthya 's suggestion, `Feature plugin gives access to features using featuregates`? If you don't like that, I'll revert the changes to the orginal wording, `Feature plugin lets you operate on features and featuregates.`

rajathagasthya (approved):
Please make sure release note is updated with breaking changes and what users can use as an alternative.

codegold79:
> Please make sure release note is updated with breaking changes and what users can use as an alternative.

@rajathagasthya. I updated the release note to contain your suggestions,

```
Update the Tanzu Feature CLI plugin and FeatureGate client to work with the new (core.tanzu.vmware.com v1alpha2) Feature and FeatureGate APIs. This change is a breaking change as the Tanzu Feature CLI will no longer work with config.tanzu.vmware.com v1alpha1. Please upgrade Feature and FeatureGate resources to use the new v1alpha2 CRDs.
```

Let me know if it can be improved.

yharish991 (approved):

PR #3851, opened 2022-11-07
Removing not used import header

knabben:
It adds a not required dependency when building the output YAML with the YTT CLI, removing this not used header can make easier manual tests.

Fixes #

### Release note

```release-note

```

### DIFF
--- a/providers/yttcb/clusterbootstrap.yaml
+++ b/providers/yttcb/clusterbootstrap.yaml
@@ -1,7 +1,6 @@
 #@ load("@ytt:assert", "assert")
 #@ load("@ytt:data", "data")
 #@ load("@ytt:ip", "ip")
-#@ load("lib/helpers.star", "get_tkr_version_from_tkr_name")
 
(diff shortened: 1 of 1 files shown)


providers/yttcb/clusterbootstrap.yaml
```diff
@@ -1,7 +1,6 @@
 #@ load("@ytt:assert", "assert")
 #@ load("@ytt:data", "data")
 #@ load("@ytt:ip", "ip")
-#@ load("lib/helpers.star", "get_tkr_version_from_tkr_name")
```
benjaminapetersen, line 4: Wondering how this is currently invoked such that `lib/helpers.star` is loaded with `get_tkr_version_from_tkr_name` and not currently erroring. 🤔 

This path must be meaningful and this function must be loaded right now, even though this file is not using it.

benjaminapetersen:
@knabben looks like tests pass.

knabben:
this was only a CI test.

PR #3849, opened 2022-11-07
Separate vsphere template resolver package

hongtaiwang:
### What this PR does / why we need it
remove vsphere template resolver from vsphere clusterclass package since we are going to separate clusterclass package with tanzu-framework.

### Which issue(s) this PR fixes

Fixes #

### Release note

```release-note

```

### Additional information

tested locally. pi for vsphere-resolver will be installed and the component works fine.
pi will not install in aws. Only package resources exist. working as expected

PR #3683, opened 2022-10-17
[TKG-13275] package tkr-vsphere-template webhook for vsphere-cc

hongtaiwang:
package tkr-vsphere-template webhook in tkg-clusterclass-vsphere pkg

add corresponding rules for tanzu-infra-clusterclass-package-cluster-role

update makefile for building tkr-vsphere-template image

### Which issue(s) this PR fixes

Fixes https://github.com/vmware-tanzu/tanzu-framework/issues/3726

### Describe testing done for PR

tested on local env.
vsphere resolver manager works
```
kubo@UWF9unoTO8zSw:~/kubeconfigs$ k -n tkg-system logs tkr-vsphere-resolver-webhook-manager-5f58c6b8d7-bbbp7
1.666017839321263e+09	INFO	setup	Version	{"version": "v0.28.0-dev", "buildDate": "2022-10-17", "sha": "eeb44d68-dirty"}
1.6660178393213103e+09	INFO	setup	setting up manager
I1017 14:44:00.374250       1 request.go:601] Waited for 1.037620136s due to client-side throttling, not priority and fairness, request: GET:https://100.64.0.1:443/apis/controlplane.cluster.x-k8s.io/v1alpha3?timeout=32s
1.6660178406260228e+09	INFO	setup	setting up webhook server
1.6660178406260667e+09	INFO	setup	registering webhooks to the webhook server
1.6660178406284528e+09	INFO	controller-runtime.webhook	Registering webhook	{"path": "/resolve-template"}
1.666017840628638e+09	INFO	setup	registering webhooks to the webhook server
1.6660178406286545e+09	INFO	setup	starting manager
1.6660178406289191e+09	INFO	controller-runtime.webhook.webhooks	Starting webhook server
1.6660178406291919e+09	INFO	controller-runtime.certwatcher	Updated current TLS certificate
1.6660178406296694e+09	INFO	controller-runtime.webhook	Serving webhook server	{"host": "", "port": 9443}
1.666017840629843e+09	INFO	controller-runtime.certwatcher	Starting certificate watcher
1.6660179125250592e+09	DEBUG	controller-runtime.webhook.webhooks	received request	{"webhook": "/resolve-template", "UID": "ba6fc9b3-63ea-478c-9d6d-b6865c2cd031", "kind": "cluster.x-k8s.io/v1beta1, Kind=Cluster", "resource": {"group":"cluster.x-k8s.io","version":"v1beta1","resource":"clusters"}}
[...]

PR #3875, opened 2022-11-09
Allow using VSphereClusterIdentity based identityRef

DanielXiao:
API consumers like TMC can not save any vCenter credentials, so it can not create a Secret based identityRef for each Cluster. It relies on TKG admin to create a VSphereClusterIdentity shared by multiple Clusters.

In this patch a Cluster variable is introduced to allow users to specify an alternative identityRef. Nevertheless, Tanzu cli still use Secret as the default identity type to keep backward UX compatible.

This patch also enhances VM template resolver to get vCenter credentials from any types of identityRef.

Fixes #

### Release note

```release-note

```


DanielXiao:
Wait for https://github.com/vmware-tanzu/tanzu-framework/pull/3849 to be merged.

tenczar (approved):
This looks good to me. Have we done any tests with a VSphereClusterIdentity that specifies allowedNamespaces? I want to make sure that the cli can't leak these credentials to namespaces that aren't allowed to use them. In particular a workload cluster should not be created if it references a VSphereClusterIdentity that doesn't have the namespace of the workload cluster in its allowedNamespaces list. If we have validated that, then this looks good to me.

DanielXiao:
Good point. I will add checks if the cluster meet allowedNamespaces.

packages/tkg-clusterclass/bundle/config/packageinstalls/tkg-infra-clusterclass-service-account.yaml
```diff
@@ -29,41 +29,6 @@ rules:
       - get
       - list
       - delete
-      - watch
```
wjun, line 32: Why these rules are not required?

DanielXiao: It was introduced by https://github.com/vmware-tanzu/tanzu-framework/pull/3683 and should be cleaned up by https://github.com/vmware-tanzu/tanzu-framework/pull/3849.

wjun (approved):

PR #3880, opened 2022-11-09
Enable error on linter check

### LINKED ISSUE #3442: Lint job should fail on linter errors
**Bug description**

The Lint job was not failing on linter errors, leading to issues that required 
https://github.com/vmware-tanzu/tanzu-framework/pull/3433 and retroactive fixing to fully address.

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- (X) Test and Release
- ( ) User Experience
- (X) Developer Experience

**Expected behavior**

Lint job should fail on linter errors

See for instance any Lint jobs last week. They all incorrected passed.

vuil:

There are about a dozen modules that still have linter errors. Linting
    will still be reported on these modules, but they are temporarily
    excluded from failing the Lint check.

To prevent regression on other modules, the lint check will now fail if
    any of them is not lint-free.

Note: Once a module from the excluded list is declared lint-free, update
    the Makefile to exclude it from MODULES_NEEDING_LINT_FIX

Also: addressed lints in cli/runtime, cli/core

Fixes #3442

Verified CI Lint check
Also manual testing.

showing module that still needs fixin'
```
INFO File cache stats: 1 entries of total size 579B
INFO Memory: 9 samples, avg is 58.7MB, max is 87.3MB
INFO Execution took 712.182501ms

****** WARNING: module ./cli/runtime needs lint fixes

-- Linting ./cli/runtime/hack/tools --
```

showing fixed module that can be excluded from the MODULES_NEEDING_LINT_FIX list
```
INFO Memory: 8 samples, avg is 57.1MB, max is 95.1MB
INFO Execution took 689.524048ms
****** NOTE: ./cli/runtime is now lint-free. Recommend it be excluded from MODULES_NEEDING_LINT_FIX in Makefile
```

### Release note

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -574,12 +574,26 @@ misspell:
 yamllint:
 	hack/check/check-yaml.sh
 
+
+# These are the modules that still contain issues reported by golangci-lint.
+# Once a module is updated to be lint-free, remove from list to catch lint regressions.
+MODULES_NEEDING_LINT_FIX=./cmd/cli/plugin/cluster ./cmd/cli/plugin/login ./cmd/cli/plugin/package ./cmd/cli/plugin-admin/builder . ./apis/core ./apis/addonconfigs ./capabilities/client ./addons ./hack/packages/package-tools ./packageclients ./tkg ./tkr ./pinniped-components/post-deploy ./pkg/v1/tkr
+
 go-lint: tools ## Run linting of go source
 	@for i in $(GO_MODULES); do \
 		echo "-- Linting $$i --"; \
-		pushd $${i}; \
-		$(GOLANGCI_LINT) run -v --timeout=10m; \
-		popd; \
+		pushd $${i} > /dev/null; \
+		if echo ${MODULES_NEEDING_LINT_FIX} | tr ' ' '\n' | grep -q -x $$i; then \
+			$(GOLANGCI_LINT) run -v --timeout=10m; \
+			if [ $$? -eq 0 ]; then \
+				echo "****** NOTE: $$i is now lint-free. Recommend it be excluded from MODULES_NEEDING_LINT_FIX in Makefile"; \
+			else \
+				echo; echo "****** WARNING: module $$i needs lint fixes"; echo; \
+			fi; \
+		else \
+			$(GOLANGCI_LINT) run -v --timeout=10m || exit 1; \
+		fi; \
+		popd > /dev/null; \
 	done
 
 	# Prevent use of deprecated ioutils module
--- a/cli/core/pkg/command/config_test.go
+++ b/cli/core/pkg/command/config_test.go
@@ -155,7 +155,7 @@ func TestConfigEditionCommunity(t *testing.T) {
 		t.Errorf("Unexpected error returned for cli.edition argument: %s", err.Error())
 	}
 
-	if cfg.ClientOptions.CLI.Edition != configapi.EditionCommunity {
+	if cfg.ClientOptions.CLI.Edition != configapi.EditionCommunity { //nolint:staticcheck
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -574,12 +574,21 @@ misspell:
 yamllint:
 	hack/check/check-yaml.sh
 
+
+# These are the modules that still contain issues reported by golangci-lint.
+# Once a module is updated to be lint-free, remove from list to catch lint regressions.
+MODULES_NEEDING_LINT_FIX=./cmd/cli/plugin/login ./cmd/cli/plugin/package ./cmd/cli/plugin-admin/builder . ./apis/core ./apis/addonconfigs ./capabilities/client ./addons ./hack/packages/package-tools ./packageclients ./tkg ./tkr ./pinniped-components/post-deploy ./pkg/v1/tkr
```
navidshaikh, line 580: can we use `skip-dirs` in `.golangci.yaml` instead ?

https://github.com/vmware-tanzu/tanzu-framework/blob/c84836a4c5a1d9c7b5f3f9b244537f9e71d3c888/.golangci.yaml#L155

ref: https://golangci-lint.run/usage/configuration/#run-configuration

navidshaikh: @vuil : once this is in, we can mob on taking the one dir at a time and fixing the lint issues. WDYT ?

anujc25: My understanding is, we would like to run the linter jobs on all the modules so that the errors are shown as part of the CI logs, however only block the PR for the modules that has all the linter errors fixed.

Using `skip-dirs` will completely skip the job run on those directories.

anujc25 (approved):
LGTM

Makefile
```diff
@@ -574,12 +574,21 @@ misspell:
...
+
 go-lint: tools ## Run linting of go source
 	@for i in $(GO_MODULES); do \
 		echo "-- Linting $$i --"; \
-		pushd $${i}; \
-		$(GOLANGCI_LINT) run -v --timeout=10m; \
-		popd; \
+		pushd $${i} > /dev/null; \
+	    if echo ${MODULES_NEEDING_LINT_FIX} | tr ' ' '\n' | grep -q -x $$i; then \
```
marckhouzam, line 586: Nit: replace spaces with tab to match indentation

Nice use of `-x` to handle the lone `.` path.  That's now a new grep flag in my arsenal 👍

Makefile
```diff
@@ -574,12 +574,21 @@ misspell:
...
+			$(GOLANGCI_LINT) run -v --timeout=10m || echo "module $$i needs lint fixes"; \
```
marckhouzam, line 587: What do you think about adding a prefix to the printout that would make it more noticeable?  Maybe something people would search for?  Maybe "ERROR: module $$i needs lint fixes".  Or if you don't like using "ERROR", we could use "WARNING", although I don't think people search for that string when looking at those giant CI logs.

marckhouzam: Oh, and how about adding a different printout if the linter succeeds here, to say that this module should be removed from the `MODULES_NEEDING_LINT_FIX` variable?

marckhouzam (commented):
This is a great idea

Makefile
vuil: Good point, that thought came across my mind too, but I reasoned that fixing the module requires a conscious effort so it is quite likely the need to update MODULES_NEEDING_LINT_FIX is not lost on the person making the fix. 
But turns out it is not complicated to introduce another conditional check, so let me fix it.

vuil: updated

Makefile
vuil: updated

PR #3888, opened 2022-11-10
Hide the default completion command of plugins

### LINKED ISSUE #3160: Plugins should hide or disable their default 'completion' command
**Bug description**

Cobra automatically creates a `completion` sub-command if one does not already exist in a binary.  Since Cobra supports shell completions out-of-the-box, this commands is fully functional and aims to make it easier for CLI developers to provide shell completion to their users.

In our Tanzu CLI plugin model, each plugin is a Cobra binary and therefore gets its own `completion` command.  This is not useful as the core CLI already provides such a command.

- (x) CLI
- (x) Plugin
- (x) User Experience

**Expected behavior**

There should not be any visible `completion` command for plugins.
We could completely disable that command using `rootCmd.CompletionOptions.DisableDefaultCmd = true` or we could choose to simply hide it using `rootCmd.CompletionOptions.DisableDefaultCmd = true`; either way, it must be done for each plugin.

The idea of hiding it instead of disabling it completely is that it would allow to use shell completion if the plugin is used directly, without being used through the tanzu cli. But that would require the user to know the `completion` command was available.

One can try the help of any plugin to see it has a `completion` command.  For example:
```
$ tanzu builder -h
Build Tanzu components

Usage:
  tanzu builder [command]

Available Commands:
  cli           Build CLIs
  completion    Generate the autocompletion script for the specified shell
  init          Initialize a new plugin repository
  publish       Publish operations
[...]
```
It is also suggested when using shell completion itself, which is useless:
```
$ tanzu builder c<TAB>
cli         -- Build CLIs
completion  -- Generate the autocompletion script for the specified shell
```

`main` branch: 482ee91c

Locally

marckhouzam:

Plugins automatically get a default `completion` command from Cobra.  For example:
```
tanzu cluster completion
tanzu package completion
```
This commit hides this default command for any plugin through the plugin runtime library.

Notice that users enable shell completion using `tanzu completion`, and that this allows completion to work for plugins as well.  Therefore plugins do not need to also have their own `completion` command.  Having such a command for a plugin (e.g., `tanzu cluster completion`) is just confusing for users.

Fixes #3160

I built the CLI and the plugins and then ran
``` 
$ tanzu cluster -h
$ tanzu management-cluster -h
$ tanzu package -h
```
and confirmed there was no `completion` command listed.

I then ran
```
$ tanzu cluster completion
$ tanzu management-cluster completion
$ tanzu package completion
```
and confirmed the command still existed but is simply hidden.
The reasons for keeping it hidden instead of removing it is explained in the section below.

### Release note
```release-note
The completion command of any plugin is removed since it is not needed.
```

### Additional information

We don't completely disable a plugin's default completion command for two reasons:
 1. backwards-compatibility, as the command used to be available for some plugins
 2. to allow to use shell completion when using the plugin as a native program (mostly for testing)

Note that a plugin can completely disable this command itself using: 
```
plugin.Cmd.CompletionOptions.DisableDefaultCmd = true
```

This change impacts **all** plugins that will be compiled with this new version of the runtime library.

### DIFF
--- a/cli/runtime/plugin/root.go
+++ b/cli/runtime/plugin/root.go
@@ -14,6 +14,17 @@ func newRootCmd(descriptor *cliapi.PluginDescriptor) *cobra.Command {
 		Use:     descriptor.Name,
 		Short:   descriptor.Description,
 		Aliases: descriptor.Aliases,
+		// Hide the default completion command of the plugin.
+		// Shell completion is enabled using the Tanzu CLI's `completion` command so a plugin
+		// does not need its own `completion` command.  Having such a command is just
+		// confusing for users. However, we don't disable it completely for two reasons:
+		//   1. backwards-compatibility, as the command used to be available for some plugins
+		//   2. to allow shell completion when using the plugin as a native program (mostly for testing)
+		// Note that a plugin can completely disable this command itself using:
+		//  plugin.Cmd.CompletionOptions.DisableDefaultCmd = true
+		CompletionOptions: cobra.CompletionOptions{
+			HiddenDefaultCmd: true,
+		},
 	}
 	cobra.AddTemplateFuncs(TemplateFuncs)
 	cmd.SetUsageTemplate(CmdTemplate)


marckhouzam:
Although the code is ready I've left this PR as a draft because is like to discuss if plugin owners need to be consulted

cli/runtime/plugin/root.go
```diff
@@ -14,6 +14,17 @@ func newRootCmd(descriptor *cliapi.PluginDescriptor) *cobra.Command {
 		Use:     descriptor.Name,
 		Short:   descriptor.Description,
 		Aliases: descriptor.Aliases,
+		// Hide the default completion command of the plugin.
```
vuil, line 17: Nice comment. 
I think the approach here looks reasonable.

marckhouzam:
This is ready for review

anujc25 (approved):
This looks good.

PR #3905, opened 2022-11-12
Tanzu CLI Configuration - Replace Strategy

### LINKED ISSUE #2652: CLI: Make the config read/write libraries forwards compatible
The Tanzu CLI supports multiple plugins. Different plugins can have different versions of the framework. Reading or writing the Tanzu CLI config should be forwards compatible. For example, an plugin using an older version of the framework should not lose configuration added by a newer version of the CLI.

- [ ] APIs
- [ ] Addons
- [x] CLI
- [ ] Docs
- [ ] IAM
- [ ] Installation
- [x] Plugin
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

**Additional context**
This is needed to support newer server/context types (like workload cluster) in the Tanzu CLI.

mpanchajanya:

Introduce new config-metadata.yaml to configure the metadata related info for config handling.
Example :- 
 configMetadata:
 patchStrategy:
  "contexts.clusterOpts.annotation": "replace"

Replace Strategy:- To the above metadata the Merge on setting the contexts will replace/delete the clusterOpts.annotation field if not specified in the change object.

By Default Set APIs will use Merge strategy on all the yaml nodes.

New API's are added to Read/Write config metadata.
Existing API's are updated to handle Replace Nodes configuration.

Other Minor Fixes:
- Changing the default-local discovery source label to default.
- `default` and `default-local` will co exist in the config.yaml i.e. If local discovery source is used and is now assigned the default name, the discovery source named default-local will still exist.
- And recommend that it be manually removed from the config file.

Fixes #2652 

Unit Testing and Integration Testing


Makefile
```diff
@@ -25,7 +25,7 @@ ADDONS_DIR := addons
 YTT_TESTS_DIR := providers/tests
 PACKAGES_SCRIPTS_DIR := $(abspath hack/packages/scripts)
 UI_DIR := tkg/web
-GO_MODULES=$(shell find . -path "*/go.mod" | grep -v "^./pinniped/" | xargs -I _ dirname _)
+GO_MODULES=$(shell find . -path "*/go.mod" | grep -v "^./pinniped/" | grep runtime | xargs -I _ dirname _)
```
mpanchajanya, line 28: will revert this before Merging

cli/runtime/config/discovery_sources_node.go
```diff
@@ -106,3 +150,15 @@ func getDiscoverySourceTypeAndName(discoverySource configapi.PluginDiscovery) (s
 	}
 	return "", ""
 }
+
+// Find the matching discovery source type and index from accepted discovery sources
+func findDiscoverySourceTypeAndIndexByWeakMatch(currentNodes []*yaml.Node) (string, int) {
```
vuil, line 155: nit: could use a more specific variable name, e.g discoverySourceContentNodes, so it does not look like this function is used on any generic node tree

vuil (commented):
Looking good. Have a few minor comments. 
Also on the PR description, where Merge is used (when describing the effect of the patch strategy), we should say when Setting the context...

On "default and default-local will co exist in the config.yaml"
we should instead mention that even though if local discovery source is used and is now assigned the default name, the ds named default-local will still exist.
And recommend that it be manually removed from the config file.

cli/runtime/config/discovery_sources_node.go
mpanchajanya: updated the variable name as suggested

cli/runtime/config/patchstrategy_it_test.go
```diff
@@ -0,0 +1,240 @@
...
+`
+	return tanzuConfigBytes, expectedConfig
+}
+func setupConfigMetadata() string {
+	metadata := `configMetadata:
+  patchStrategy:
+    contexts.group: replace
+    contexts.clusterOpts.endpoint: replace
+    contexts.clusterOpts.annotation: replace
+    contexts.discoverySources.gcp.required: replace
```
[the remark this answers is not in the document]
mpanchajanya: updated the test data as suggested

cli/core/pkg/config/defaults.go
```diff
@@ -22,7 +22,7 @@ var (
 	DefaultStandaloneDiscoveryImagePath  = ""
 	DefaultStandaloneDiscoveryImageTag   = ""
 	DefaultStandaloneDiscoveryName       = "default"
-	DefaultStandaloneDiscoveryNameLocal  = "default-local"
+	DefaultStandaloneDiscoveryNameLocal  = "default"
```
vuil, line 25: nit: A comment explaining the reason for the rename could be useful.

mpanchajanya: Added the doc

cli/runtime/config/nodeutils/replace_nodes.go
```diff
@@ -0,0 +1,88 @@
...
+				if strings.EqualFold(patchStrategies[key], "replace") {
+					dst.Content = append(dst.Content[:i], dst.Content[i+1:]...)
+					dst.Content = append(dst.Content[:i], dst.Content[i+1:]...)
+					i--
+					i--
+					break
+				}
+
+				if err := replaceNodes(src.Content[j+1], dst.Content[i+1], key, patchStrategies); err != nil {
+					return errors.New("replace at key " + src.Content[i].Value + ": " + err.Error())
```
vuil, line 61: nit: when errors are passed back up, it is more idiomatic to use 
errors.Wrap or in this case errors.Wrapf. Same for case in L82 as well as a few others in merge_nodes.go
It's not blocking. And since there are also preexisting instances, I suggest taking note and addressing in a followup.

vuil (commented):
I just left a tiny nit that is not blocking.

@mpanchajanya I might have missed it, but can you point me to the test that exhibited the replace behavior when a discovery source type is altered as part of the change?

vuil:
@mpanchajanya : please update the commits (squashing to a list of logical commits/ be consistent with the commit subject, / provide some more description if helpful, etc)

mpanchajanya:
> I just left a tiny nit that is not blocking.
> 
> @mpanchajanya I might have missed it, but can you point me to the test that exhibited the replace behavior when a discovery source type is altered as part of the change?

The test are in cli/runtime/config/cli_discovery_sources.go

vuil:
> > I just left a tiny nit that is not blocking.
> > @mpanchajanya I might have missed it, but can you point me to the test that exhibited the replace behavior when a discovery source type is altered as part of the change?
> 
> @vuil The test are in cli/runtime/config/cli_discovery_sources.go and cli/runtime/config/patchstrategy_it_test.go

Thanks for updating patchstrategy_it_test.go. I see it now.

vuil (approved):
Thanks for the updates. LGTM.

vuil (commented):
Few additional comments for consideration in the followup. @mpanchajanya

mpanchajanya:
> Few additional comments for consideration in the followup. @mpanchajanya

Yes made the changes into multi File PR https://github.com/vmware-tanzu/tanzu-framework/pull/3954/

cli/runtime/config/nodeutils/replace_nodes.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+)
+
+// ReplaceNodes replace nodes in dst as per patchStrategy prior performing merge
+func ReplaceNodes(src, dst *yaml.Node, opts ...PatchStrategyOpts) (bool, error) {
```
vuil, line 15: @mpanchajanya thinking about it more. As implemented the replace logic is done in two phases, first deleting all the nodes marked for 'replacement' by the strategy, then performing the merge from source to destination node, creating the nodes again. 

If the merge logic is not going to be part of this function or replaceNodes that it calls, would it be clearer to call implementation in this file DeleteNodes, deleteNodes?

cli/runtime/config/nodeutils/replace_nodes_test.go
```diff
@@ -0,0 +1,280 @@
...
+		dst      string
+		src      string
+		metadata map[string]string
+		output   string
+	}{
+		{
+			name:     "success replace nodes with empty patch strategy",
+			metadata: make(map[string]string),
+			dst: `name: test-mc
+type: k8s
```
vuil, line 25: nit: the test data is unnecessarily complex, containing too much stuff that wasn't germane to demonstrating the node replacing (actual deleting) effect. It is difficult to manually diff 3 sets of yaml content when one cannot keep most of the data on screen simultaneously. I suggest we pare it down some without sacrificing coverage.

vuil (commented):
@mpanchajanya couple more suggestions.

cli/runtime/config/nodeutils/replace_nodes.go
mpanchajanya: yes makes sense. updated in the multi file PR

cli/runtime/config/nodeutils/replace_nodes_test.go
mpanchajanya: refactored tests

PR #3908, opened 2022-11-14
Add version suffix v1.0.0 to all default ClusterClass names

DanielXiao:
According to ClusterClass best practice, when we make any changes on it or its reference templates, we 'd better make a copy of all and rename them. Given multiple versions of the default ClusterClasses will coexist in the same management cluster, we add a version suffix in their names to identify them and avoid package conflicts.

There is no time to separate ClusterClass packages from the TKG meta package in Glasgow, we decided to hard code version v1.0.0 for now and will start to package them individually from Halifax.

Fixes #

### Release note

```release-note

```


providers/infrastructure-aws/v2.0.0-beta.1/cconly/base.yaml
```diff
@@ -1015,22 +1019,22 @@ spec:
...
 metadata:
-  name: aws-cluster-template
+  name: #@ "tkg-aws-default-{}-cluster".format(data.values.AWS_CLUSTER_CLASS_VERSION)
   namespace: #@ data.values.NAMESPACE
 spec:
   template:
     spec:
-      region: "${AWS_REGION}"
-      sshKeyName: "${AWS_SSH_KEY_NAME}"
+      region: #@ data.values.AWS_REGION
```
ydp, line 1027: I had a confusion for a long time about the difference between environment variable (by export) and config variable (in config file), when should we use environment variable and when should use config variable, or they are equivalent through some cli internal Golang code (uncertain...)? Thus, does this kind of change break user behavior?

ydp: I remember a (uncertain) time ago when deleting AWS management cluster, AWS_REGION is necessary to be present as environment, not sure what's the current status.

DanielXiao: Environment variable can not be interpolated by our ytt processor, you will see them still as ${AWS_REGION} after it is rendered

zhouke1991:
LGTM

zhouke1991 (approved):

packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml
```diff
@@ -5,22 +5,23 @@
 apiVersion: cluster.x-k8s.io/v1beta1
 kind: ClusterClass
 metadata:
-  name: tkg-aws-default
```
jeffwubj, line 8: modify it in overlay.yaml should be an option as well ?

12345lcr (approved):
LGTM

packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml
DanielXiao: Yes but overlay is usually for more complicated cases

wjun (approved):
LGTM.

PR #3911, opened 2022-11-14
Fix upgrade logic for additional packages in ClusterBoostrap but not in CBT

### LINKED ISSUE #3937: Packages are not upgraded if they are in clusterboostrap but not in clusterbootstrap template
**Bug description**
Packages are not upgraded if they are in clusterboostrap but not in clusterbootstrap template. There can be optional packages in clusterbootstrap but not present in clusterbootstrap template. They need to be upgraded if they are specified in TKr's spec.bootstrappackages

- ( ) APIs
- (x) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

shyaamsn:
There might be additional packages in clusterbootstrap but not in clusterboostraptemplate. During upgrade these packages refName to be updated. The logic is to update the package refName if its present in tkr spec.bootstrappackages. If its not present then its the user's responsibility to fix them

Fixes #3937 

Unit test

### Release note

```release-note
Fix upgrade logic for additional packages in ClusterBoostrap but not in CBT
```

### DIFF
--- a/addons/controllers/clusterbootstrap_controller.go
+++ b/addons/controllers/clusterbootstrap_controller.go
@@ -544,9 +544,7 @@ func (r *ClusterBootstrapReconciler) mergeClusterBootstrapPackagesWithTemplate(
 	// Meaning the users will not be able to customize the packageRefName
 	// Find all the corresponding pairs in ClusterBootstrap and new ClusterBootstrapTemplate to update
 	// Add the additional package if it's only present in the new ClusterBootstrapTemplate
-	// Leave the package as it is if it's only present in ClusterBootstrap but not in the new Template
 	additionalPackageMap := map[string]*runtanzuv1alpha3.ClusterBootstrapPackage{}
-
 	for _, pkg := range updatedClusterBootstrap.Spec.AdditionalPackages {
 		packageRefName, _, err := util.GetPackageMetadata(r.context, r.aggregatedAPIResourcesClient, pkg.RefName, cluster.Namespace)
 		if err != nil || packageRefName == "" {
@@ -557,6 +555,7 @@ func (r *ClusterBootstrapReconciler) mergeClusterBootstrapPackagesWithTemplate(
 		additionalPackageMap[packageRefName] = pkg
 	}
 
+	additionalPackageTemplateMap := map[string]*runtanzuv1alpha3.ClusterBootstrapPackage{}
 	for _, templatePkg := range clusterBootstrapTemplate.Spec.AdditionalPackages {
 		// use the refName in package CR, since the package CR hasn't been cloned at this point, use SystemNamespace to fetch packageCR
 		packageRefName, _, err := util.GetPackageMetadata(r.context, r.aggregatedAPIResourcesClient, templatePkg.RefName, r.Config.SystemNamespace)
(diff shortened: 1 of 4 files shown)


alfredthenarwhal:
Commit: 609a855

Build failed! Build no: 3316

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -575,6 +574,21 @@ func (r *ClusterBootstrapReconciler) mergeClusterBootstrapPackagesWithTemplate(
...
+	}
+
+	// There might be packages in clusterbootstrap but not in clusterboostraptemplate.
+	// The below loop tries to find those packages and updates the RefName if it is found in tkr's spec.bootstrapPackages.
+	// If its not there in tkr's spec.bootstrappackages then user has to update the version of it in clusterbootstrap.
+	for additionalPkgRefName, additionalPkg := range additionalPackageMap {
+		if _, ok := additionalPackageTemplateMap[additionalPkgRefName]; !ok {
+			additionalPkgFromTKr, additionalPkgNamePrefix, err := util.GetBootstrapPackageNameFromTKR(r.context, r.Client, additionalPkg.RefName, cluster)
+			if err != nil {
+				log.Info(fmt.Sprintf("unable to find any additional bootstrap package prefixed with '%s' for ClusterBootstrap %s/%s in TKR", additionalPkgNamePrefix, cluster.Name, cluster.Namespace))
```
maralavi, line 587: Seems like when error is returned on ```GetBootstrapPackageNameFromTKR```, it does not return prefix information any more. Not related to this PR but a bug with existing logic. Could you please modify the ```additionalPkgNamePrefix``` to return prefix even on error so that the log above can be useful?

maralavi: Also I noticed in most places we log with this format & ordering ```<NAMESPACE>/<NAME>```. Could you please fix this in the log message above and in the other existing usage of ```GetBootstrapPackageNameFromTKR``` in the code to stay consistent?

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -575,6 +574,21 @@ func (r *ClusterBootstrapReconciler) mergeClusterBootstrapPackagesWithTemplate(
 			updatedClusterBootstrap.Spec.AdditionalPackages = append(updatedClusterBootstrap.Spec.AdditionalPackages, newPkg)
 			packages = append(packages, newPkg)
 		}
+		additionalPackageTemplateMap[packageRefName] = templatePkg
+	}
+
+	// There might be packages in clusterbootstrap but not in clusterboostraptemplate.
+	// The below loop tries to find those packages and updates the RefName if it is found in tkr's spec.bootstrapPackages.
+	// If its not there in tkr's spec.bootstrappackages then user has to update the version of it in clusterbootstrap.
```
maralavi, line 582: If its not in tkr's ```bootstrappackages```, does it work if the user just update the version in ```clusterbootstrap```? Should we provide instructions on how? Wouldn't it immutable?

addons/controllers/clusterbootstrap_controller.go
shyaamsn: Fixed

addons/controllers/clusterbootstrap_controller.go
shyaamsn: I have added a testcase to update the version in clusterbootstrap and ensured it can work. I am not sure there is anything special to provide in documentation. It's just a matter of updating clusterbootstrap.spec.AdditionalPackages[index].RefName to updated version

lubronzhan (approved):

maralavi (approved):

PR #3915, opened 2022-11-15
Fix few lint issues

mpanchajanya:

Fix few lint issues reported by `make go-lint`

Fixes #


vuil (commented):
@mpanchajanya Thanks for doing this! lgtm for the most part.
There are a few //nolint directives (the ones with no accompanying explanation), which I will leave to the codeowners of the code unless they give them the thumbs up.
Either way, whichever modules are not linting cleanly after the change should be remove from the MODULES_NEEDING_LINT_FIX variable so that CI will enforce they stay lint free.

PR #3916, opened 2022-11-15
Plumb legacy PinnipedConfigs to ClusterBootstrap

### LINKED ISSUE #2373: Plumb Pinniped addon configs to PinnipedConfig API
CLI has legacy addon configs like ANTREA_TRAFFIC_ENCAP_MODE that is used to create an addon secret. In new Addons API since we are using ClusterBootStrap and AddonConfig API we need to plumb the legacy addon configs to new APIs.

Create ClusterBootstrap and AddonConfig API overalys.

benjaminapetersen:

Plumb legacy PinnipedConfigs to ClusterBootstrap to make the Pinniped addon work on classy management and workload clusters.

- [x] The pinniped addon Secret is generated by `providers/yttcb` at the default `name: <cluster-name>-pinniped-package`, with the `type: clusterbootstrap-secret`, with the `move` label, and with the rest of the expected content, but only for management clusters and only when the user has provided the pinniped configuration settings.
- [x] For classy workload clusters, never generate a pinniped addon secret in the `providers/yttcb` code. The default Secret (from the ClusterBootstrapTemplate?) will be created. It will initially contain the values which cause Pinniped to be disabled, but the `pinniped-comonents/tanzu-auth-controller-manager/controllers/v3_cascade_controller.go` controller will automatically repopulate the Secret to have the correct contents based on the management cluster's Pinniped config.
- [x] In the case where the user has not enable pinniped in their management cluster's config during `tanzu management-cluster create` then do not create a pinniped addon Secret in the `providers/yttcb` code. The default Secret (from the ClusterBootstrapTemplate?) will be created. It will contain the appropriate settings to disable pinniped on the management cluster. - Note: we decided to change this and always create a secret for management clusters to enable `--dry-run` use case for setting pinniped back to the defaults, in this follow-up PR https://github.com/vmware-tanzu/tanzu-framework/pull/4154
- [x] The `pinniped-comonents/tanzu-auth-controller-manager/controllers/v3_cascade_controller.go` is enhanced ignore any Secret which is associated with a `Cluster` which has the label indicating that it is a management cluster (see https://github.com/vmware-tanzu/tanzu-framework/blob/main/addons/controllers/clusterbootstrap_controller.go#L299 for an example of another controller which is checking for that label). This is also unit tested.
- [x] Enhance `tanzu cluster kubeconfig get` to add the Cluster resources' UID to the pinniped audience name for classy clusters. It was already doing this for TKGs (Pacific) workload clusters, but it is also needed for other classy workload clusters.
- [x] Confirm that these changes will not negatively impact TKGs, which was already using these same code paths for creating and configuring workload clusters (but not for management clusters). TKGs does not use the same mechanisms to create management (aka supervisor) clusters. Care was taken to preserve the preexisting behavior of the code for TKGs workload clusters.
- [x] Integration tests for configuring Pinniped on classy clusters live in a different repo and we will create a follow-up PR in that repo to make the pre-existing tests work with classy clusters. This is not included in this PR.

Known issues related to making the Pinniped addon work for classy clusters which are not fixed by this PR:
- [x] The Pinniped addon Secret is not copied into the management cluster during `tanzu management-cluster-create`, preventing Pinniped from ever being installed on classy management clusters. 
   - This was resolved by PR #3960.
- [x] #4038 the `--dry-run` flag may not work for existing management-clusters. There needs to be a way for a user to use `tanzu management-cluster create --dry-run` to output only the Pinniped addon secret. 
    - Addressed by https://github.com/vmware-tanzu/tanzu-framework/pull/4060
- [x] the stdout output of `tanzu management-cluster create  --yes --deploy-tkg-on-vSphere7 -f tkg-mgmt-vc.yaml --dry-run` ends with:
    ```
    Checking for required plugins...
    All required plugins are already installed and up-to-date
    ```
    which makes it impossible to redirect the stdout to a file and get a valid yaml file as the result, since that line will always appear at the bottom of the file
    - Addressed by https://github.com/vmware-tanzu/tanzu-framework/pull/4106
- [x] After the above dry run issues are fixed, we need also to document how to perform day 2 configuration of Pinniped on classy clusters. The user needs to know how to filter out the Pinniped addon Secret from the dry run output (e.g. by using yq) and how to `kubectl apply` the pinniped addon Secret to the management cluster. This doc should make it clear that the other resources returned by the dry run should not be applied to the cluster. Unfortunately, the tanzu CLI feature allowing the use of the `FILTER_BY_ADDON_TYPE` env var with the `tanzu management-cluster create` command seems to be broken for classy clusters. The [documented steps for TKG 1.6](https://docs.vmware.com/en/VMware-Tanzu-Kubernetes-Grid/1.6/vmware-tanzu-kubernetes-grid-16/GUID-iam-configure-id-mgmt.html#generate-the-pinniped-addon-secret-for-the-management-cluster-7) no longer work because the `FILTER_BY_ADDON_TYPE` feature is broken. This PR adds the `FILTER_BY_ADDON_TYPE` env var feature back to the CLI for TKGm classy clusters.
- [x] For management clusters configured to use LDAP, the `pinniped-post-deploy-job` consistently fails its maximum 7 tries. This prevents the pinniped configuration from being finished on the management cluster. One way that a user might notice this is when they run `tanzu mc kubeconfig get` it will error with a message `Error: failed to get pinniped-info from cluster`. Running `kubectl describe job pinniped-post-deploy-job -n pinniped-supervisor` will show 7 failures. The workaround is to wait a little while and then run `kubectl delete job pinniped-post-deploy-job -n pinniped-supervisor`. kapp-controller will recreate the job within a couple of minutes, and the job should succeed. At this point, Pinniped should be ready for use on the management cluster. This `communtiy-edition` PR attempts to address this problem by extending the retry limit: https://github.com/vmware-tanzu/community-edition/pull/5594. Also, there might be a bug in Kubernetes which is causing the Job to retry to quickly: https://github.com/kubernetes/kubernetes/issues/114391
- [x] There is no easy way to set the Pinniped config back to the default of identity management type "none". See below for more information and for a workaround. There are two separate problems here:
  - [x] When the user deletes the Pinniped addon secret, nothing happens. The default secret is not automatically created. No reconciliation happens. (See below for workaround.)
  - [X] When the Pinniped Supervisor is uninstalled by setting the addon secret back to its default values, nothing deletes the `pinniped-info` configmap. This is create by the pinniped-post-deploy-job. To fix this in this PR, we made the post-deploy-job add an ownerRef during creation such that the ConfigMap would be auto-deleted whenever the Pinniped Supervisor is deleted.
- [X] For management clusters without Pinniped configuration, we do not create an addon Secret in the ytt templates to allow the default Secret to be generated instead. This Secret contains `cluster_type: workload` (even on management clusters) because the Tanzu Community Edition only allows `identity_management_type: none` for `workload` clusters.  [See here](https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/pinniped/0.12.1/bundle/config/values.star#L17).  It might be a little confusing for the Secret to say that it is a workload cluster when it is actually for a management cluster, but it is currently required and it works. Maybe we could fix this some day to make it more clear, but it doesn't seem important right now. We decided not to worry about this for now.

Fixes #2373

Manual testing performed:

1. Used a bolt MR to create a build of this PR: https://gitlab.eng.vmware.com/TKG/bolt/bolt-release-yamls/-/merge_requests/3486#note_13435102
2. Used Jenkins to create a VC7 testbed with the changes from this PR by referring to the buildinfo file built by bolt: https://kscom.svc.eng.vmware.com/job/tkg-release/job/install-vc7/20553/
3. The Jenkins job automatically creates a management cluster and a workload cluster without configuring Pinniped. Manually inspect them to check if Pinniped is in the correct default state: each Cluster resource on the management cluster should have a corresponding `<cluster-name>-pinniped-package` Secret with the expected default values, and Pinniped should not be installed on either cluster (no Pinniped-related namespaces).
3. Test the use case of configuring Pinniped on existing clusters where it was not previously configured. Use OIDC for this step. Make sure that Pinniped gets installed on both the management cluster and the workload cluster. Test logging in via Pinniped to both clusters.
   1. `cp tkg-mgmt-vc.yaml tkg-mgmt-vc-oidc.yaml`
   2. Edit `tkg-mgmt-vc-oidc.yaml` and add all the required OIDC settings:
      ```
      IDENTITY_MANAGEMENT_TYPE: oidc
      OIDC_IDENTITY_PROVIDER_CLIENT_ID: <redacted>
      OIDC_IDENTITY_PROVIDER_CLIENT_SECRET: <redacted, but note that this should NOT be base64 encoded>
      OIDC_IDENTITY_PROVIDER_ISSUER_URL: https://gitlab.eng.vmware.com
      OIDC_IDENTITY_PROVIDER_SCOPES: email
      OIDC_IDENTITY_PROVIDER_USERNAME_CLAIM: email
      # OIDC_IDENTITY_PROVIDER_GROUPS_CLAIM: # unnecessary
      ```
   3. `FILTER_BY_ADDON_TYPE="authentication/pinniped" tanzu management-cluster create  --yes -v 9 --deploy-tkg-on-vSphere7 -f tkg-mgmt-vc-oidc.yaml --dry-run > dry-run-mgmt-auth-oidc-filtered-by-addon-type.yaml`
   4. `kubectl config use-context tkg-mgmt-vc-admin@tkg-mgmt-vc`
   5. `kubectl apply -f dry-run-mgmt-auth-oidc-filtered-by-addon-type.yaml` to configure the pinniped addon for the management cluster, and the tanzu-auth controller should automatically reconcile it into the workload cluster as well.
   6. Wait for the install of Pinniped to finish by waiting for this job to succeed once: `watch kubectl describe job pinniped-post-deploy-job -n pinniped-supervisor`
   7. Confirm that the Pinniped Supervisor and Concierge were both installed on the management cluster, and the Pinniped Concierge was installed on the workload cluster.
   8. Get the issuer string: `kubectl get cm pinniped-info -n kube-public -o jsonpath='{.data.issuer}'`. Append `/callback` to the end of that issuer string and add the url to your OIDC client's allowed redirect list.
   9. Confirm that getting the pinniped-compatible kubeconfigs using the tanzu CLI allows you to successfully auth into both clusters using an identity from the OIDC provider.
      1. `tanzu mc kubeconfig get --export-file kubeconfig-tkg-mgmt-vc-oidc.yaml`
      2. `kubectl get pods --kubeconfig kubeconfig-tkg-mgmt-vc-oidc.yaml`
      3. `tanzu cluster kubeconfig get tkg-vc-antrea --export-file kubeconfig-tkg-vc-antrea-oidc.yaml`
      4. `kubectl get pods --kubeconfig kubeconfig-tkg-vc-antrea-oidc.yaml`
5. Delete both clusters and create new clusters with Pinniped configured during cluster creation of the management cluster. Use LDAP for this step, so that we will have tested both LDAP and OIDC by the end of our testing.
   1.  `cp tkg-mgmt-vc.yaml tkg-mgmt-vc-ldap.yaml`
   2. Edit `tkg-mgmt-vc-ldap.yaml` and add all the required LDAP settings (not shown here).
   2. Create management cluster by running `tanzu mc create -f tkg-mgmt-vc-ldap.yaml --yes -v 9 --deploy-tkg-on-vSphere7`
   3. Create the workload cluster by running `tanzu cluster create tkg-vc-antrea -v 9 -f tkg-vc-antrea.yaml`.
   4. Make sure that Pinniped gets installed on both the management cluster and the workload cluster.
   5. Test logging in via Pinniped to both clusters.
6. Unfortunately, deleting the addon secret on the management cluster using `kubectl delete secret tkg-mgmt-vc-pinniped-package -n tkg-system` does not seem to do anything. We need a way to configure Pinniped back to the default of identity management type "none" on the management cluster, and to have the v3_cascade_controller automatically be triggered to set the addon secret for all workload clusters back to the default contents. (See below for workaround.)

A possible workaround to configure Pinniped back to the default of identity management type "none" is to follow this one-step procedure:
1. On the management cluster, manually edit the pinniped addon secret to look like the following. This will cause the Pinniped Supervisor and Concierge to be deleted from the management cluster.
   ```yaml
   apiVersion: v1
   kind: Secret
   metadata:
     annotations:
       tkg.tanzu.vmware.com/addon-type: authentication/pinniped
     labels:
       clusterctl.cluster.x-k8s.io/move: ""
       tkg.tanzu.vmware.com/addon-name: pinniped
       tkg.tanzu.vmware.com/cluster-name: tkg-mgmt-vc # must be actual name of your mgmt cluster
     name: tkg-mgmt-vc-pinniped-package # must start with actual name of your mgmt cluster
     namespace: tkg-system
   stringData:
     values.yaml: |
       infrastructure_provider: vsphere # use the actual infra type for your mgmt cluster
       tkg_cluster_role: workload # use workload, even though it is a mgmt cluster
       identity_management_type: none # use none to turn off pinniped
   type: clusterbootstrap-secret
   ```

### Release note

```release-note

```


providers/yttcb/pinniped_addon_secret.lib.yaml
```diff
@@ -0,0 +1,204 @@
...
+    azure:
+      dnsNames:
+        - tkg-dex.com
+      DEX_SVC_LB_HOSTNAME: dex.example.com #! <DEX_SVC_LB_HOSTNAME> is required for azure
+#@ end
+
+#@ def getValuesForWC():
+#@ bomData = get_bom_data_for_tkr_name()
+#@ pinnipedImage = bomData.components["pinniped"][0].images.pinnipedImage
+#@ postDeployPinnipedImage = bomData.components["tanzu-framework-addons"][0].images.tkgPinnipedPostDeployImage
```
benjaminapetersen, line 170: It appears, from our testing, that the shape of the `tkr-bom` has changed since this bit of code was originally written. This works in `tkg 1.6.0` but in the `1.7.0 classy` cluster I am experimenting with, the following change is required:

```yaml
# old bad
#! postDeployPinnipedImage = bomData.components["tanzu-framework-addons"][0].images.tkgPinnipedPostDeployImage
# new good
#@ postDeployPinnipedImage = bomData.components["pinniped-post-deploy"][0].images.tkgPinnipedPostDeployImage
```

tenczar (approved):
LGTM

providers/yttcb/clusterbootstrap.yaml
```diff
@@ -315,7 +362,11 @@ spec:
   additionalPackages:
     - refName: metrics-server*
     - refName: secretgen-controller*
+    #@ if pinniped_configs_exist():
```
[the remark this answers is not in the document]
benjaminapetersen: @shyaamsn

providers/yttcb/clusterbootstrap.yaml
```diff
@@ -16,6 +19,33 @@
 #@   end
 #@ end
 
+#@ def pinniped_configs_exist():
+#@   return (data.values.IDENTITY_MANAGEMENT_TYPE != "none" and
```
benjaminapetersen, line 23: There are 2 other CERT duration types of variables that Pinniped understands, but alone do not contribute meaningfully to Pinniped package, so I left them out of this function.

providers/yttcb/pinniped_addon_secret.lib.yaml
benjaminapetersen: Pushing an update for this.

providers/yttcb/clusterbootstrap.yaml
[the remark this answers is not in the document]
shyaamsn: I think this is coming from default secret thats published in the TKr package. Checkout the TKr package and it should have default pinniped secret.

If you dont want this default secret then in Clusterboostrap  you should provide the secret in Clusterboostrap.

providers/yttcb/clusterbootstrap.yaml
```diff
@@ -315,7 +362,11 @@ spec:
   additionalPackages:
     - refName: metrics-server*
     - refName: secretgen-controller*
+    #@ if pinniped_configs_exist():
     - refName: pinniped*
+      valuesFrom:
+        secretRef: #@ "{}-pinniped-addon".format(data.values.CLUSTER_NAME)
```
benjaminapetersen, line 368: We tried creating a `management-cluster` using:

```bash
DEPLOY_TKG_ON_VSPHERE7=true tanzu management-cluster create -f tkg-mgmt-vc-with-pinniped.yaml --plan dev
```
with `tkg-mgmt-vc-with-pinniped.yaml` being a basic config with some pinniped env vars:

```yaml
# default mgmt config vals:
CLUSTER_NAME: tkg-mgmt-vc-with-pinniped
CLUSTER_PLAN: dev
INFRASTRUCTURE_PROVIDER: vsphere
... trimmed for brevity

# pinniped config
# --------------------
IDENTITY_MANAGEMENT_TYPE: ldap

# LDAP
# using VMware LDAP config:
# LDAP_BIND_DN:                                                 # skip this field
# LDAP_BIND_PASSWORD:                                           # skip this field
LDAP_HOST: ldaps.eng.vmware.com:636
LDAP_USER_SEARCH_BASE_DN: 'ou=people,dc=vmware,dc=com'
LDAP_USER_SEARCH_FILTER: '(objectClass=posixAccount)'
LDAP_USER_SEARCH_USERNAME: uid            # i guess?
LDAP_USER_SEARCH_ID_ATTRIBUTE: uid
LDAP_USER_SEARCH_EMAIL_ATTRIBUTE: uid
LDAP_USER_SEARCH_NAME_ATTRIBUTE: uid
LDAP_GROUP_SEARCH_BASE_DN: dc=vmware,dc=com
LDAP_GROUP_SEARCH_FILTER: (objectClass=posixGroup)
LDAP_GROUP_SEARCH_USER_ATTRIBUTE: uid
LDAP_GROUP_SEARCH_GROUP_ATTRIBUTE: memberUid
LDAP_GROUP_SEARCH_NAME_ATTRIBUTE: cn
# LDAP_ROOT_CA_DATA_B64:                                        # not sure2

```

which created the expected `pinniped secret` as well as this cluster bootstrap file:

```yaml
# cat .config/tanzu/tkg/clusterconfigs/tkg-mgmt-vc-with-pinniped.yaml
# generated when we ran tanzu cluster create above...
...
apiVersion: run.tanzu.vmware.com/v1alpha3
kind: ClusterBootstrap
metadata:
  annotations:
    tkg.tanzu.vmware.com/add-missing-fields-from-tkr: v1.23.8---vmware.2-tkg.2-zshippable
  name: tkg-mgmt-vc-with-pinniped
  namespace: tkg-system
spec:
  additionalPackages:
  - refName: metrics-server*
  - refName: secretgen-controller*
  # this is what we expressed in our template logic.
  # the *-pinniped-addon is the correct name.
  # but does not match in the next block below
  - refName: pinniped*
    valuesFrom:
      secretRef: tkg-mgmt-vc-with-pinniped-pinniped-addon  # pinniped-addon, not pinniped-package!
...
```
However, the actual cluster boostrap file on the cluster does not look like the configuration file that we generated and then used to create the cluster.  It has defaulted back to the default pinniped package secret:

```yaml
# kubectl get clusterbootstrap tkg-mgmt-vc-with-pinniped -n tkg-system -o yaml
apiVersion: run.tanzu.vmware.com/v1alpha3
kind: ClusterBootstrap
metadata:
  creationTimestamp: "2022-11-16T19:59:20Z"
  generation: 1
  name: tkg-mgmt-vc-with-pinniped
  namespace: tkg-system
  ownerReferences:
  - apiVersion: cluster.x-k8s.io/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: Cluster
    name: tkg-mgmt-vc-with-pinniped
    uid: 91a8faba-5b3b-47cb-a047-4ed88cb49a13
  resourceVersion: "6456"
  uid: 3f7daf58-131a-4e63-b649-f8f2b03ee776
spec:
  additionalPackages:
  - refName: metrics-server.tanzu.vmware.com.0.6.1+vmware.1-tkg.3-zshippable
  - refName: secretgen-controller.tanzu.vmware.com.0.11.0+vmware.2-tkg.1-zshippable
  - refName: tkg-storageclass.tanzu.vmware.com.0.28.0-dev-115-gd7751adc+vmware.1
  - refName: pinniped.tanzu.vmware.com.0.12.1+vmware.2-tkg.3-zshippable
   # this is for the default valuesFrom,
   # not the updated valuesFrom *-pinniped-addon that we 
   # changed.  Something is incorrect here.
    valuesFrom:
      secretRef: tkg-mgmt-vc-with-pinniped-pinniped-package # pinniped-package? this is wrong.
...
```
However, it did create the `pinniped-addon secret`, as we expected it to:

```bash
kubectl get secret tkg-mgmt-vc-with-pinniped-pinniped-addon -n tkg-system
# NAME                                       TYPE                         DATA   AGE
# tkg-mgmt-vc-with-pinniped-pinniped-addon   tkg.tanzu.vmware.com/addon   1      11m
```

It appears that we declared what we want, but something overrode our declaration with the default?

providers/yttcb/clusterbootstrap.yaml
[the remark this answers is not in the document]
ydp: I agree, you can provide an `else` unless you don't want the default.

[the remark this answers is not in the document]
benjaminapetersen: Adding the else then.

providers/yttcb/clusterbootstrap.yaml
```diff
@@ -315,7 +362,13 @@ spec:
...
+    #@ else:
```
benjaminapetersen, line 369: This actually seems CONCERNING to us.
Given that a user might not configure pinniped with LDAP or OIDC initially, but may configure it later, I think we want a consistent output here of a certain specific secret name so a user can find it easily.

providers/yttcb/clusterbootstrap.yaml
```diff
@@ -61,9 +91,26 @@
...
+metadata:
+  name: #@ "{}-pinniped-addon".format(data.values.CLUSTER_NAME)
+  namespace: #@ data.values.NAMESPACE
+  labels:
+    tkg.tanzu.vmware.com/addon-name: pinniped
+    tkg.tanzu.vmware.com/cluster-name: #@ data.values.CLUSTER_NAME
+    clusterctl.cluster.x-k8s.io/move: ""
+  annotations:
+    tkg.tanzu.vmware.com/addon-type: "authentication/pinniped"
+type: tkg.tanzu.vmware.com/addon
```
benjaminapetersen, line 110: Theorizing that this might be the wrong type, given I see:

```bash
kubectl get secret -n tkg-system | grep pinn
tkg-mgmt-vc-customized-pinn-pinniped-addon                              tkg.tanzu.vmware.com/addon            1      8m41s
tkg-mgmt-vc-customized-pinn-pinniped-data-values                        Opaque                                1      15m
tkg-mgmt-vc-customized-pinn-pinniped-fetch-0                            kubernetes.io/dockerconfigjson        1      15m
tkg-mgmt-vc-customized-pinn-pinniped-package                            clusterbootstrap-secret               1      8m37s
```
I am speculating that the type `clusterbootstrap-secret` should be on our secret here, not `tkg.tanzu.vmware.com/addon` and that the ClusterBoostrapController may not be seeing what its looking for and then is dumping out the default generic secret instead.

providers/yttcb/clusterbootstrap.yaml
cfryanr: Josh had a goo question... what is the default value of data.values.IDENTITY_MANAGEMENT_TYPE in this yttcb context? Is it "none" or "" or something else? There is no values.yaml file in the providers/yttcb directory.

providers/yttcb/clusterbootstrap.yaml
```diff
@@ -61,9 +91,26 @@
...
 
+#@ if pinniped_configs_exist():
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: #@ "{}-pinniped-addon".format(data.values.CLUSTER_NAME)
+  namespace: #@ data.values.NAMESPACE
+  labels:
+    tkg.tanzu.vmware.com/addon-name: pinniped
```
cfryanr, line 105: It seems like only the auth v1_cascade_controller cares about this label, not the v3 controller. Should we remove it?

providers/yttcb/clusterbootstrap.yaml
benjaminapetersen: This is indeed a great question!

providers/yttcb/pinniped_addon_secret.lib.yaml
```diff
@@ -0,0 +1,204 @@
...
+---
+infrastructure_provider: #@ data.values.PROVIDER_TYPE
+tkg_cluster_role: #@ data.values.TKG_CLUSTER_ROLE
+identity_management_type: #@ data.values.IDENTITY_MANAGEMENT_TYPE
+pinniped:
+  cert_duration: #@ data.values.CERT_DURATION
+  cert_renew_before: #@ data.values.CERT_RENEW_BEFORE
+  supervisor_svc_endpoint: #@ data.values.SUPERVISOR_ISSUER_URL
+  supervisor_ca_bundle_data: #@ data.values.SUPERVISOR_ISSUER_CA_BUNDLE_DATA_B64
+  image: #! For backwards compatibility. i.e. C.1 mgmt cluster creates a C.0 wlc with older addon templates.
```
benjaminapetersen, line 180: @shyaamsn asked if we can remove the `image` stanza here. Seems backwards compat for `C.1` (Calgary) is likely unnecessary.

providers/yttcb/pinniped_addon_secret.lib.yaml
```diff
@@ -0,0 +1,204 @@
...
+    name: #@ pinnipedImage.imagePath
+    tag: #@ pinnipedImage.tag
+    repository: #@ get_image_repo_for_component(pinnipedImage)
+    pull_policy: IfNotPresent
+  post_deploy_job_image: #! For backwards compatibility. i.e. C.1 mgmt cluster creates a C.0 wlc with older addon templates.
```
benjaminapetersen, line 185: @shyaamsn also asked if we can remove `post_deploy_job_image`.  Calgary is quite old.

pinniped-components/tanzu-auth-controller-manager/controllers/v3_cascade_controller.go
```diff
@@ -122,6 +135,14 @@ func (c *PinnipedV3Controller) reconcileSecret(ctx context.Context, secret *core
 		return nil
 	}
 
+	// Check the Cluster's labels to determine if it is a management cluster. Do not update the pinniped
```
benjaminapetersen, line 138: We could further add, it is the user's job to provide mgmt cluster configuration, our controllers ensure workload clusters remain in sync.

pinniped-components/tanzu-auth-controller-manager/controllers/v3_cascade_controller_test.go
```diff
@@ -51,9 +62,24 @@ var _ = Describe("Controller", func() {
...
+					tkgClusterNameLabel: managementCluster.Name,
+				},
+			},
+			Data: map[string][]byte{tkgDataValueFieldName: []byte("should-not-be-edited")},
+			Type: clusterBootstrapManagedSecret,
+		}
 	})
 
-	Context("ClusterBootstrap Secret", func() {
+	Context("pinniped-info configmap does not exist", func() {
```
benjaminapetersen, line 82: The `pinniped-info` configmap  is a contract for several use cases:
- the WCP (TKGs) Supervisor cluster provides it, to provide consistency with our usual Management (TKGm) clusters. 
- the tanzu cli knows to look for this configmap

pinniped-components/tanzu-auth-controller-manager/controllers/v3_cascade_controller_test.go
```diff
@@ -435,4 +437,71 @@ var _ = Describe("Controller", func() {
...
+		})
+
+		AfterEach(func() {
+			deleteObject(ctx, cluster)
+			deleteObject(ctx, managementCluster)
+			deleteObject(ctx, pinnipedCBSecret)
+			deleteObject(ctx, managementClusterPinnipedCBSecret)
+		})
+
+		When("the pinniped cluster bootstrap secret gets created", func() {
```
benjaminapetersen, line 456: Nice, this is pretty critical coverage.

providers/yttcb/pinniped_addon_secret.lib.yaml
```diff
@@ -0,0 +1,204 @@
+#@ load("@ytt:data", "data")
+#@ load("/lib/helpers.star", "get_no_proxy")
+#@ load("/lib/helpers.star", "get_bom_data_for_tkr_name", "get_image_repo_for_component")
```
benjaminapetersen, line 3: @shyaamsn per discussion (and related below) this dependency may be able to removed.

benjaminapetersen:
Not this replaces #3350

providers/yttcb/clusterbootstrap.yaml
cfryanr: This was resolved by https://github.com/vmware-tanzu/tanzu-framework/pull/3960.

providers/yttcb/pinniped_addon_secret.lib.yaml
cfryanr: It was removed.

providers/yttcb/pinniped_addon_secret.lib.yaml
cfryanr: Removed.

providers/yttcb/pinniped_addon_secret.lib.yaml
cfryanr: It was removed.

providers/yttcb/clusterbootstrap.yaml
cfryanr: It looks like the addons controller might also use the label, so I suppose we should leave it.

providers/yttcb/clusterbootstrap.yaml
cfryanr: We decided to let the `secretRef` use its default value for this PR.

providers/yttcb/clusterbootstrap.yaml
cfryanr: We ended up using clusterbootstrap-secret as the type.

pinniped-components/tanzu-auth-controller-manager/controllers/v3_cascade_controller_test.go
cfryanr: 👍

providers/yttcb/clusterbootstrap.yaml
cfryanr: We ended up writing the code to specifically check for "ldap" or "oidc". Any other value will disable Pinniped.

We also decided to only check for the data.values.IDENTITY_MANAGEMENT_TYPE value. If it is either of the supported values, then assume that the user is trying to configure Pinniped.

benjaminapetersen:
Testing done via https://github.com/vmware-tanzu/tanzu-framework/pull/4115, now closed.

cfryanr (approved):

PR #3919, opened 2022-11-16
Comsume AKOO as a composite package

lubronzhan:
Comsume AKOO as a composite package.
The package name now is `ako-operator-v2` to avoid conflict with old legacy package from core-repository
The package version is same as the tf version. 
To take care of upgrade, old AKOO packageInstall and secret is gracefully deleted once new pkgi is installed

Fixes #

### Release note

```release-note

```


tkg/managementcomponents/management_component_install.go
```diff
@@ -167,10 +167,28 @@ func NoopDeletePackageInstall(clusterClient clusterclient.Client, addonName, nam
...
+
+	if err := pauseAddonSecretReconciliation(clusterClient, akoOperatorAddonName, constants.TkgNamespace); err != nil {
+		return err
+	}
+
+	if err := pausePackageInstallReconciliation(clusterClient, akoOperatorPkgiName, constants.TkgNamespace); err != nil {
+		return err
+	}
+
+	if err := NoopDeletePackageInstall(clusterClient, akoOperatorPkgiName, constants.TkgNamespace); err != nil {
```
12345lcr, line 182: May I ask what is the package version of akoo after integrated into tkg-pkg? I am asking this because if the akoo's version becomes same as t-f, then the problem below would happen:
* After delete the old pkgi and let tkg-pkg to install the new pkgi, it may still point to the old akoo package, which is from core package repo. Becase t-f version is 0.2x.x and akoo version was 1.x.x, kapp will treat the old akoo package as the default newer version.
* Core package repo will be removed after tkg-pkg deployed, so the old akoo package will be gone in the end. But the pkgi won't become to point to the new akoo package, because kapp does not allow package downgrade. There will be a ReconcileFailure in the package status.
I suggest to
* follow the addon-manager's behavior to retrieve the version from bom to make sure the new pkgi points to the right version, and expose a dev env param to overwrite the version in case the tkg-pkg is not from bom for developing purpose.
* OR, remain the akoo version instead of following t-f version

lubronzhan: Do you mean this will still happens even if we changed the packageInstall name and packageName?
The previous package name is ako-operator.tanzu.vmware.com and now it's ako-operator-v2.tanzu.vmware.com.

12345lcr: Oh, with the name changed it should be fine, thanks!

alfredthenarwhal:
Commit: 366911d

Tests passed! Build no: 3323

packages/ako-operator-v2/kbld-config.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kbld.k14s.io/v1alpha1
+kind: Config
+overrides:
+  - image: ako-operator:latest
+    newImage: "projects-stg.registry.vmware.com/tkg/sandbox/ako-operator:v1.7.0_vmware.1"
```
wjun, line 5: What is this line for?

tkg/constants/files.go
```diff
@@ -39,3 +39,67 @@ const (
 
 	TKGPackageValuesFile = "tkgpackagevalues.yaml"
 )
+
+// Config content
+const (
```
wjun, line 44: Why do we need a const here?

tkg/client/init.go
```diff
@@ -231,6 +231,17 @@ func (c *TkgClient) InitRegion(options *InitRegionOptions) error { //nolint:funl
...
 		}
+
+		akoRequired, err := c.isAKORequiredInBootstrapCluster()
+		if err != nil {
+			return errors.Wrap(err, "unable to check whether avi ha is enabled")
+		}
+		if akoRequired {
+			log.Info("Installing AKO...")
+			if err = c.InstallAKO(bootStrapClusterClient); err != nil {
+				return errors.Wrap(err, "unable to upgrade ako")
```
12345lcr, line 242: The error message here should be `unable to install ako on bootstrap cluster`, right?
And I also think since we are installing ako on bootstrap cluster, maybe it is better to change the log info `"Installing AKO..."` to `Installing AKO on bootstrap cluster...` to be the same log patterns of other bootstrap cluster components. This can also reduce some confusions.

alfredthenarwhal:
Commit: a275e59

Build failed! Build no: 3330

tkg/constants/files.go
lubronzhan: Separate const with different usage

packages/ako-operator-v2/kbld-config.yaml
lubronzhan: We don't build ako-operator image inside tanzu-framework, it's a different repo, so it doesn't have the step to replace the `kbld-config` with real image after docker image is generated. For example, when building tanzu-addon-manager package, it will first build the image, then replace the kbld-config with that newly built image. 
If we don't add this hardcoded image, when running the `make build` target, it will fail because `newImage` is empty.
Downstream is different, bolt will just replace the image with ako-operator component's reference. But to get it work on upstream, I need to hardcode something here

tkg/client/init.go
lubronzhan: Let me fix it

tkg/constants/files.go
lubronzhan:
```
// ConfigFilePermissions defines the permissions of the config file
const (
	ConfigFilePermissions       = 0o600
	DefaultDirectoryPermissions = 0o700
)

// File name related constants
const (

```

lubronzhan: Or you want me to put the content in a real fie

lubronzhan: Ok I moved the content to providers folder

wjun (approved):
LGTM.

12345lcr (approved):

12345lcr:
LGTM

PR #3921, opened 2022-11-16
fix harded coded overlay k8s version for docker clusterclass

hongtaiwang:
use data.values.KUBERNETES_VERSION as topology.version

the latest k8s version is upgraded to `v1.24.6+vmware.1-tkg.1-zshippable` and the hardcoded logic will break all docker provider clusters

Fixes #

### Release note

```release-note

```

### DIFF
--- a/packages/tkg-clusterclass-docker/bundle/config/upstream/base.yaml
+++ b/packages/tkg-clusterclass-docker/bundle/config/upstream/base.yaml
@@ -148,10 +148,10 @@ spec:
           jsonPatches:
             - op: add
               path: "/spec/template/spec/kubeadmConfigSpec/initConfiguration/nodeRegistration/kubeletExtraArgs/cgroup-driver"
-              value: cgroupfs
+              value: systemd
             - op: add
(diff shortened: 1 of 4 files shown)


packages/tkg-clusterclass-docker/bundle/config/upstream/base.yaml
```diff
@@ -148,10 +148,10 @@ spec:
           jsonPatches:
             - op: add
               path: "/spec/template/spec/kubeadmConfigSpec/initConfiguration/nodeRegistration/kubeletExtraArgs/cgroup-driver"
-              value: cgroupfs
+              value: systemd
```
vuil, line 151: What's the reason for this? Seems like this is already conditionally setting the right value for the cgroup-driver based on the kubernetes version.

hongtaiwang: This mr is just making some simple change to test whether it works for the pipelines :)

I saw you posted #3933 and it's much more well formed. Will close this pr and let's track it in #3933

hongtaiwang:
closing since #3933 posted

PR #3924, opened 2022-11-16
[WIP] add retry for WaitForClusterInitialized when vcenter connection slow 

hongtaiwang:
cluster creation will failed due to `cluster secret not found` in dimension sddc on top of nimbus if the env is slow.

cluster secret will be created afterwards. Add retry for this condition

Fixes #

### Release note

```release-note

```

### DIFF
--- a/tkg/clusterclient/clusterclient.go
+++ b/tkg/clusterclient/clusterclient.go
@@ -587,12 +587,21 @@ func (c *client) WaitForClusterInitialized(clusterName, namespace string) error 
 		err = currentClusterInfo.RetrievalError
 
 		if err == nil {
-			// If cluster's ReadyCondition is False and severity is Error, it implies non-retriable error, so return error
+			// should retry in this condition
+			// For dimension env, sometimes it will take some time to connect to vcenter to create resources like cluster secret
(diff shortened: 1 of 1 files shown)


tkg/clusterclient/clusterclient.go
```diff
@@ -587,12 +587,21 @@ func (c *client) WaitForClusterInitialized(clusterName, namespace string) error
 		err = currentClusterInfo.RetrievalError
 
 		if err == nil {
-			// If cluster's ReadyCondition is False and severity is Error, it implies non-retriable error, so return error
+			// should retry in this condition
+			// For dimension env, sometimes it will take some time to connect to vcenter to create resources like cluster secret
```
randomvariable, line 591: Can we rewrite this comment? I don't think it makes sense, or at least I don't understand it.
vCenter isn't Kubernetes, so we're not creating a cluster secret. And this doesn't feel like it would be limited only to Dell Dimension, so what is the real thing we're fixing here? I suspect the real answer is disk I/O causing etcd to be slower...


```suggestion
			// Retry waiting for cluster ready status for slow I/O infrastructure or resource constrained environments.
```

PR #3925, opened 2022-11-16
Add a seperate controller for cluster-metadata

zhouke1991:
The cluster-metadata existed as a configmap in a legacy cluster, both mgmt and workload cluster.
It was used to install by ytt on the client side, but in utkg2.0 we will deprecate the way of installing resources by client side.

### DIFF
--- a/addons/controllers/cluster_metadata_controller.go
+++ b/addons/controllers/cluster_metadata_controller.go
@@ -9,9 +9,6 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
-	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/constants"
-	"github.com/vmware-tanzu/tanzu-framework/addons/pkg/util"
-	"github.com/vmware-tanzu/tanzu-framework/addons/predicates"
 	"gopkg.in/yaml.v3"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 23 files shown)


zhouke1991:
The bootstrap controller UT failed in Main pipeline, but it passed locally.
image

addons/controllers/cluster_metadata_controller.go
```diff
@@ -0,0 +1,372 @@
...
+	// if tkr is not found, should not requeue for the reconciliation
+	if tkr == nil {
+		log.Info("TKR object not found", "name", tkrName)
+		return ctrl.Result{}, nil
+	}
+
+	//if _, labelFound := tkr.Labels[constants.TKRLabelLegacyClusters]; labelFound {
+	//	log.Info("Skipping reconciling due to tkr label", "name", tkrName, "label", constants.TKRLabelLegacyClusters)
+	//	return ctrl.Result{}, nil
+	//}
```
12345lcr, line 118: This seems not necessary

12345lcr (approved):
LGTM

wjun (approved):
LGTM after removing the useless codes.

shyaamsn:
@zhouke1991 @wjun I am concerned adding more and more things to addons-manager is going to cause memory issues. I am not comfortable with this change. Lets hold on to this until we have @zjs and Steven, Yang approve this

shyaamsn:
Thanks for the PR. Since this is a small controller the impact on memory should be less. If there are issues we can revisit it. Thanks

zhouke1991:
@shyaamsn Thanks!

PR #3927, opened 2022-11-16
Kickstart UI: Containerize Angular UI build

miclettej:
Moves Angular UI build into a docker container

Includes:
 - Dockerfile with node v16 base image
 - Copies all UI src files into image, installs npm dependencies and generates REST API models/services from swagger doc
 - Modifies Makefile UI build targets to generate the image and then start a container from which we can copy the compiled 'dist' folder from the Angular build to the host machine
 - Cleans up docker container and image once copy is completed
 - Updates Github CI workflows to remove node v16 install and npm cache
 
Fixes #

### Release note

```release-note
none
```


vuil (approved):
lgtm, thanks!

PR #3932, opened 2022-11-16
Add logic for clusterboostrap webhook to handle deleting cluster

### LINKED ISSUE #3931: clusterbootstap webhook does not handle cluster deletion correctly
**Bug description**
When a cluster is deleted it triggers a clusterbootstrap (update) reconciliation. 
The clusterbootstrap webhook fails because some of the objects it tries to verify 
have already been deleted. 

- ( ) APIs
- (x ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

adduarte:
When a cluster is being deleted is possible that some of the objects the clusterbootstrap webhook is looking for might be deleted by the time the webhook checks for them.
In that case the validation should not fail because the cluster is being deleted.

fixes: #3931 

Fixes # 3931

### Release note

```release-note
Clusterbootstrap validation is handled different when cluster is marked by deletion. 
Deleted resources do not fail validation
```

### DIFF
--- a/addons/webhooks/clusterbootstrap_webhook.go
+++ b/addons/webhooks/clusterbootstrap_webhook.go
@@ -20,6 +20,7 @@ import (
 	cacheddiscovery "k8s.io/client-go/discovery/cached/memory"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
+	clusterv1beta1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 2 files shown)


maralavi:
@adduarte Thanks for the PR. Please add release note:)

maralavi (approved):

shyaamsn:
@adduarte Could you please check if its possible to add unit test for this ? Thanks

PR #3933, opened 2022-11-16
Test CAPD cc and non-cc integ test fixes

vuil:

(This combines relevant changes in 3921 and 3928)

Test  CAPD cc and non-cc integ test fixes
Fixes the CAPD integration tests which fail to create v1.24 clusters because the templates always set the cgroup driver to cgroupfs.
Kind images >= v1.24 only support the systemd cgroup driver.

Fix hardcode topology.version of docker CC overlay 

Fixes N/A

monitoring CI

### Release Notes
```release-note
None
```


tenczar (approved):
LGTM

vuil:
It looks like both commits accomplishes what they set out to do. Even though there are lingering issues with the CAPD cc tests, let's get this in.

PR #3934, opened 2022-11-17
add retry for WaitForClusterInitialized when VCenterUnreachable

hongtaiwang:
cluster creation will failed due to `cluster secret not found` in dimension sddc on top of nimbus if the env is slow.

cluster secret will be created afterwards. Add retry for this condition

will return error if exceed timeout

Fixes #

### Release note

```release-note

```

### DIFF
--- a/pkg/v1/tkg/clusterclient/clusterclient.go
+++ b/pkg/v1/tkg/clusterclient/clusterclient.go
@@ -542,12 +542,21 @@ func (c *client) WaitForClusterInitialized(clusterName, namespace string) error 
 		err = currentClusterInfo.RetrievalError
 
 		if err == nil {
-			// If cluster's ReadyCondition is False and severity is Error, it implies non-retriable error, so return error
+			// should retry in this condition
+			// For dimension env, sometimes it will take some time to connect to vcenter to create resources like cluster secret
 			if conditions.IsFalse(currentClusterInfo.ClusterObject, capi.ReadyCondition) &&
 				(*conditions.GetSeverity(currentClusterInfo.ClusterObject, capi.ReadyCondition) == capi.ConditionSeverityError) {
-				return true, errors.Errorf("cluster creation failed, reason:'%s', message:'%s'",
-					conditions.GetReason(currentClusterInfo.ClusterObject, capi.ReadyCondition),
-					conditions.GetMessage(currentClusterInfo.ClusterObject, capi.ReadyCondition))
+				reason := conditions.GetReason(currentClusterInfo.ClusterObject, capi.ReadyCondition)
+				message := conditions.GetMessage(currentClusterInfo.ClusterObject, capi.ReadyCondition)
+
+				maxTimeoutCounter++
+				// if exceeds maxTimeout, return error
+				if interval*time.Duration(maxTimeoutCounter) > maxTimeout {
+					return true, errors.Errorf("timed out waiting for cluster creation, reason:'%s', message:'%s'",
+						reason,
+						message)
+				}
+				return false, errors.Errorf("cluster not ready, reason:'%s', message:'%s'", reason, message)
 			}
 			// Could have checked cluster's ReadyCondition is True which is currently aggregation of ControlPlaneReadyCondition
 			// and InfrastructureReadyCondition, however in future if capi adds WorkersReadyCondition into aggregation, it would


pkg/v1/tkg/clusterclient/clusterclient.go
```diff
@@ -542,12 +542,21 @@ func (c *client) WaitForClusterInitialized(clusterName, namespace string) error
...
+			// For dimension env, sometimes it will take some time to connect to vcenter to create resources like cluster scret
 			if conditions.IsFalse(currentClusterInfo.ClusterObject, capi.ReadyCondition) &&
 				(*conditions.GetSeverity(currentClusterInfo.ClusterObject, capi.ReadyCondition) == capi.ConditionSeverityError) {
-				return true, errors.Errorf("cluster creation failed, reason:'%s', message:'%s'",
-					conditions.GetReason(currentClusterInfo.ClusterObject, capi.ReadyCondition),
-					conditions.GetMessage(currentClusterInfo.ClusterObject, capi.ReadyCondition))
+				reason := conditions.GetReason(currentClusterInfo.ClusterObject, capi.ReadyCondition)
+				message := conditions.GetMessage(currentClusterInfo.ClusterObject, capi.ReadyCondition)
+
+				maxTimeoutCounter++
```
wjun, line 552: In case there is a real error, will this change make the creation failure after 3*c.operationTimeout? How about retry 4 times which is 1 minute?

pkg/v1/tkg/clusterclient/clusterclient.go
```diff
@@ -542,12 +542,21 @@ func (c *client) WaitForClusterInitialized(clusterName, namespace string) error
 		err = currentClusterInfo.RetrievalError
 
 		if err == nil {
-			// If cluster's ReadyCondition is False and severity is Error, it implies non-retriable error, so return error
+			// should retry in this condition
+			// For dimension env, sometimes it will take some time to connect to vcenter to create resources like cluster scret
```
jeffwubj, line 546: typo `scret`

pkg/v1/tkg/clusterclient/clusterclient.go
```diff
@@ -542,12 +542,21 @@ func (c *client) WaitForClusterInitialized(clusterName, namespace string) error
...
-				return true, errors.Errorf("cluster creation failed, reason:'%s', message:'%s'",
-					conditions.GetReason(currentClusterInfo.ClusterObject, capi.ReadyCondition),
-					conditions.GetMessage(currentClusterInfo.ClusterObject, capi.ReadyCondition))
+				reason := conditions.GetReason(currentClusterInfo.ClusterObject, capi.ReadyCondition)
+				message := conditions.GetMessage(currentClusterInfo.ClusterObject, capi.ReadyCondition)
+
+				maxTimeoutCounter++
+				// if exceeds maxTimeout, return error
+				if interval*time.Duration(maxTimeoutCounter) > maxTimeout {
+					return true, errors.Errorf("timed out waiting for cluster createion, reason:'%s', message:'%s'",
```
jeffwubj, line 555: typo `createion`

pkg/v1/tkg/clusterclient/clusterclient.go
```diff
@@ -542,12 +542,21 @@ func (c *client) WaitForClusterInitialized(clusterName, namespace string) error
...
 		if err == nil {
-			// If cluster's ReadyCondition is False and severity is Error, it implies non-retriable error, so return error
+			// should retry in this condition
+			// For dimension env, sometimes it will take some time to connect to vcenter to create resources like cluster scret
 			if conditions.IsFalse(currentClusterInfo.ClusterObject, capi.ReadyCondition) &&
 				(*conditions.GetSeverity(currentClusterInfo.ClusterObject, capi.ReadyCondition) == capi.ConditionSeverityError) {
-				return true, errors.Errorf("cluster creation failed, reason:'%s', message:'%s'",
-					conditions.GetReason(currentClusterInfo.ClusterObject, capi.ReadyCondition),
-					conditions.GetMessage(currentClusterInfo.ClusterObject, capi.ReadyCondition))
+				reason := conditions.GetReason(currentClusterInfo.ClusterObject, capi.ReadyCondition)
```
jeffwubj, line 549: ditto : add a log ?

hongtaiwang: yes, got log message

pkg/v1/tkg/clusterclient/clusterclient.go
hongtaiwang: I re-use the `maxTimeoutCounter` for other conditions in `WaitForClusterInitialized`. Seems it's a reasonable timeout val.

pkg/v1/tkg/clusterclient/clusterclient.go
```diff
@@ -542,12 +542,21 @@ func (c *client) WaitForClusterInitialized(clusterName, namespace string) error
...
+				message := conditions.GetMessage(currentClusterInfo.ClusterObject, capi.ReadyCondition)
+
+				maxTimeoutCounter++
+				// if exceeds maxTimeout, return error
+				if interval*time.Duration(maxTimeoutCounter) > maxTimeout {
```
jeffwubj, line 554: two comments here:
1. so the "side-effect" is: for a real "error" level failure, it takes more time to fail ? is it possible to add this logic for vSphere if it is not yet
2. how should we determine a suitable maxTimeout value ?

hongtaiwang: 1. correct, it will retry several times before failure. Initially I added retry for `VCenterUnreachable` only. But then I found out fuji capz pipeline failed due to `reason: Failed @ Machine/github-5516-mc-control-plane-xtzgp`. Thus I changed to retry for all kinds of `reason`. For now seems like the machine failure is kind of like pipeline issue. But I'm not sure if other conditions should also be retried. I think it's better to retry than failed by some other unknown `reason`. :) But sure, if it's better to retry only when `VCenterUnreachable`, I could change it back.
2. I re-use the timeout retry counter for `WaitForClusterInitialized`. Hopefully it might be a suitable value since it's been used for a long while.

pkg/v1/tkg/clusterclient/clusterclient.go
wjun: Sorry, what I mean is exactly Jeff's comment below: "so the "side-effect" is: for a real "error" level failure, it takes more time to fail ?" How about retry 4 times which is 1 minute?

pkg/v1/tkg/clusterclient/clusterclient.go
jeffwubj: Thanks for the explanations.

As a dot release to fix must have issues, maybe just fix specific issue that has already been found, and we are confidential for the impact ? In this case, maybe just limited to vsphere unless we found similar issues on the other platforms? (do we want/need to expand tests to public cloud?) 

For the main branch, which has more test cycles, seems good to keep current design to see the impact.

Feel free to evaluate baed on your observations.

renmaosheng: Jeff, do you mean we need to find the real root-cause of why "vCenterUnreachable" happened? instead of just adding retry? thanks.

jeffwubj: @renmaosheng : retry seems good for an urgent patch. 
`vCenterUnreachable` happened for a slow "internal" env inside lab, and we want to support that. Instead of adding some logic to wait for VC to be ready before reconciling, @hongtaiwang 's fix here is lightweight, and seems good for an urgent fix, but maybe better to limit to vSphere's scenario ONLY, given test team may just verify vSphere platform for this improvement.

hongtaiwang:
closing this pr since we are not going to fix it in fuji.1

PR #3936, opened 2022-11-17
Enable AWS Garbage Collection by default

ydp:

Starting from CAPA 1.5.0, it supports to garbage collect non-capi created resources, which help cleanup orphaned resources on aws, we should enable it by default.

About this feature: https://cluster-api-aws.sigs.k8s.io/topics/external-resource-gc.html

This change will get all clusters with GC feature, users can disable it per cluster basis by patching AWSCluster object annotation like:

```
kubectl patch awscluster test-aws -n tkg-system --type='json' -p='[{"op": "add", "path": "/metadata/annotations/aws.cluster.x-k8s.io~1external-resource-gc", "value":"false"}]'
```

## how providers-sync-tools works
Take `cluster-api-provider-aws` as an example, `make all` will run
* `make clean`: remove all files in `cluster-api-provider-aws/build` directory
* `make sync`: copy files in `packages/cluster-api-provider-aws/bundle/config` to `cluster-api-provider-aws/build/upstream`
* `make render`: Use ytt to merge overlay and upstream manifests in `cluster-api-provider-aws/build/upstream`, output is in  `cluster-api-provider-aws/build/rendered`
* `make move`: copy files in `cluster-api-provider-aws/build/rendered/upstream` to `providers/infrastructure-aws/v2.0.0-beta.1`
* `make validate`: compare all files in `providers/infrastructure-aws/v2.0.0-beta.1` and `cluster-api-provider-aws/build/rendered/upstream`

Fixes #

Manually create aws cluster, and deployed NLB service, then delete clusters.

### Release note

```release-note

```

### DIFF
--- a/packages/cluster-api-provider-aws/bundle/config/overlay/set-features-gates.yaml
+++ b/packages/cluster-api-provider-aws/bundle/config/overlay/set-features-gates.yaml
@@ -1,0 +1,15 @@
+#@ load("@ytt:data", "data")
+#@ load("@ytt:overlay", "overlay")
+
+#@overlay/match by=overlay.subset({"kind":"Deployment", "metadata": {"name": "capa-controller-manager"}})
+---
+spec:
+  template:
+    spec:
+      containers:
+      #@overlay/match by=overlay.subset({"name": "manager"})
+      - args:
+        #@overlay/match by=overlay.index(1)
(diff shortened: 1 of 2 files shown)


wjun (approved):
LGTM

PR #3938, opened 2022-11-17
Temporarily disable CAPD integration tests from CI workflow

danniel1205:
Signed-off-by: Daniel Guo <danniel1205@gmail.com>

Due to the known issue from CAPD integration tests ([link](https://github.com/vmware-tanzu/tanzu-framework/pull/3933)), this PR temporarily disables/removes CAPD integration tests from CI workflow. So that we could unblock all other PRs from merging. 

**IMPORTANT**: We need to turn it back on once the CAPD issue gets fixed.

Fixes #

### Release note

```release-note

```

### DIFF
--- a/.github/workflows/tkg_integration_tests_clusterclass.yaml
+++ b/.github/workflows/tkg_integration_tests_clusterclass.yaml
@@ -194,6 +194,7 @@ jobs:
         make clean-registry
 
   integ-tests-docker-cc:
+    if: false
     name: TKG Integration Tests CAPD (ClusterClass)
     runs-on: tkg
     needs: build-and-publish-packages


.github/workflows/tkg_integration_tests.yaml
```diff
@@ -15,98 +15,6 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  integ-tests-docker:
```
yharish991, line 18: instead of deleting the job, can we use conditions to control job execution? like this https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution. All we need to do is to add this `if: false`

danniel1205: It's a good idea. Let me have a try.

vuil (approved):
It's not ideal, but looks the pragmatic thing to do.

PR #3939, opened 2022-11-17
Update unit tests to fix config file corruption issue

### LINKED ISSUE #3930: `~/.config/tanzu/config.yaml` is getting corrupted after running unit tests
**Bug description**
The CLI commands `tanzu login` and `tanzu config server list` not working after running tests in grpc_test.go and token_test.go, it seems test cases are corrupting the ~/.config/tanzu/config.yaml file.

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- (x) User Experience
- ( ) Developer Experience

**Expected behavior**
The `tanzu login`, `tanzu config server list` and other CLI commands should work fine even after running any unit test suites.
Get latest tanzu-framework code, delete existing ~/.config/tanzu/config.yaml file, run `make build-install-local` and `make test`, then run tests in grpc_test.go and token_test.go, now run commands `tanzu login` and `tanzu config server list`, you can see error.
❯ tanzu version
version: v0.28.0-dev
buildDate: 2022-11-16
sha: 1d60a188
Mac
[cli-commands-output.txt](https://github.com/vmware-tanzu/tanzu-framework/files/10026373/cli-commands-output.txt)
[config.yml.txt](https://github.com/vmware-tanzu/tanzu-framework/files/10026377/config.yml.txt)

chandrareddyp:
The `~/.config/tanzu/config.yaml` is being corrupted by the unit tests, as part of this PR the unit test cases are updated to not to update the file `~/.config/tanzu/config.yaml` and mock the calls.

Fixes #3930

Unit test cases are executed

### Release note

```release-note

```


cli/core/pkg/interfaces/config_client_interface.go
```diff
@@ -22,3 +24,13 @@ func NewConfigClient() ConfigClient {
 func (cc *configClientImpl) GetEnvConfigurations() map[string]string {
 	return config.GetEnvConfigurations()
 }
+
+func (cc *configClientImpl) StoreClientConfig(clientConfig *configapi.ClientConfig) error {
+	// Acquire tanzu config lock
+	config.AcquireTanzuConfigLock()
+	defer config.ReleaseTanzuConfigLock()
```
anujc25, line 31: Adding this means `configClientImpl.StoreClientConfig ` is not same as actual `config.StoreClientConfig` which is concerning to me. Why do we need update the logic? Can we keep it just as a wrapper and not add things like this?

If needed we could add additional functions like `AcquireTanzuConfigLock` and `ReleaseTanzuConfigLock` to `configClientImpl` but this should only be treated as wrapper.

chandrareddyp: Curious, getting the lock and releasing the lock shouldn't be part of function `config.StoreClientConfig(clientConfig)`? whats the reason we are getting and releasing the lock outside of it?

cli/core/pkg/interfaces/config_client_interface.go
```diff
@@ -5,20 +5,38 @@
...
+	config.AcquireTanzuConfigLock()
+}
+
+func (cc *configClientWrapperImpl) ReleaseTanzuConfigLock() {
+	config.ReleaseTanzuConfigLock()
+}
+
+func (cc *configClientWrapperImpl) StoreClientConfig(clientConfig *configapi.ClientConfig) error {
+	// TODO: Add Read/Write locking mechanism before updating the configuration
+	// Currently we are only acquiring the lock while updating the configuration
```
anujc25, line 40: nit: We should remove this TODO comment.

anujc25 (approved):
Just a minor comment. Otherwise looks good.

cli/core/pkg/interfaces/config_client_interface.go
chandrareddyp: it was an existing comment. That's why I kept it.

anujc25: We need to keep this comment but not here, it needs to be kept in `cli/core/pkg/auth/csp/grpc.go` and not in the wrapper.

anujc25 (approved):

PR #3940, opened 2022-11-17
Update Cluster Toplogy utils to populate empty arr

tenczar:
This change makes sure that our cluster topology utils sets array typed cluster variables to an empty array instead of nil. This does the same for map typed variables.

The SetMDVariable util function will set the override variable for a given machine deployment in a cluster topology. The cluster topology requires the corresponding cluster variable is defined for the cluster, so the SetMDVariable function would also set a null variable in the cluster level variable definition. null is an invalid value for variables of type array or type map. This change updates the topology util functions to set a proper empty array or empty map when setting a MachineDeployment override on a variable that does not currently have a top level definition. 

Fixes #

### Release note

```release-note

```

### DIFF
--- a/util/topology/cluster.go
+++ b/util/topology/cluster.go
@@ -41,21 +41,31 @@ func GetVariable(cluster *clusterv1.Cluster, name string, value interface{}) err
 // SetMDVariable sets the variable for the given machineDeployment, overriding it if necessary, to the given value.
 // Pre-reqs: cluster.Spec.Topology != nil && cluster.Spec.Topology.Workers != nil
 func SetMDVariable(cluster *clusterv1.Cluster, mdIndex int, name string, value interface{}) error {
-	jsonValue, err := jsonValue(value)
+	jsonVal, err := jsonValue(value)
 	if err != nil {
 		return err
 	}
 
+	var varDefault *apiextensionsv1.JSON
+	switch reflect.TypeOf(value).Kind() {
+	case reflect.Slice:
+		varDefault, _ = jsonValue([]interface{}{})
+	case reflect.Map:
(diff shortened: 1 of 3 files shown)


util/topology/cluster.go
```diff
@@ -41,21 +41,31 @@ func GetVariable(cluster *clusterv1.Cluster, name string, value interface{}) err
...
 func SetMDVariable(cluster *clusterv1.Cluster, mdIndex int, name string, value interface{}) error {
-	jsonValue, err := jsonValue(value)
+	jsonVal, err := jsonValue(value)
 	if err != nil {
 		return err
 	}
 
+	var varDefault *apiextensionsv1.JSON
+	switch value.(type) {
+	case []interface{}:
```
timothysc, line 51: where are the unit-tests to do input verification (fuzz) for different types?

timothysc: Just make a test array.

tenczar: I've added tests for maps and arrays.

randomvariable (dismissed):

util/topology/cluster.go
imikushin, line 51: This will not match if value is of type `[]SomeOtherType`. Looks like the only way to detect it's a slice is using reflection.

util/topology/cluster.go
```diff
@@ -41,21 +41,31 @@ func GetVariable(cluster *clusterv1.Cluster, name string, value interface{}) err
...
+		varDefault, _ = jsonValue([]interface{}{})
+	case map[interface{}]interface{}:
```
imikushin, line 53: Same as above.

imikushin (approved):
LGTM. Thanks @tenczar!

PR #3954, opened 2022-11-20
Tanzu CLI Configuration - Multi file approach

### LINKED ISSUE #2652: CLI: Make the config read/write libraries forwards compatible

### LINKED ISSUE #3810: CurrentContext map can potentially be erased by older plugins saving into the CLI config file
**Bug description**
Due to the same issue reported in https://github.com/vmware-tanzu/tanzu-framework/issues/2652,
Any data in the CurrentContext map can potentially be wiped out when an old plugin attempts to save the configuration.

This can lead to commands that rely on it (e.g. `tanzu tmc cluster create [Create workload cluster while talking with TMC endpoint])` to fail

```
1. tanzu context use tmc-unstable [sets currentContext for TMC to tmc-unstable]


   current: tmc-unstable
   currentContext:
     tmc: tmc-unstable


2. tanzu context use tkg-mgmt-cluster-1 [sets currentContext for k8s to tkg-mgmt-cluster-1]

   current: tkg-mgmt-cluster-1
   currentContext:
     tmc: tmc-unstable
     k8s: tkg-mgmt-cluster-1



3. tanzu tmc cluster create [Create workload cluster while talking with TMC endpoint]
4. tanzu kubernetes cluster create [Create workload cluster using TKG management cluster]

Assuming cluster plugin here has bug [GHI-2652](https://github.com/vmware-tanzu/tanzu-framework/issues/2652). So, running above command will wipes out currentContext for tmc target as shown below)

   current: tkg-mgmt-cluster-1
   currentContext:
     k8s: tkg-mgmt-cluster-1

5. tanzu tmc cluster --help [this command is no longer available to be invoked]
   (Running this command will fail after running an old buggy plugin which is not the expected behavior.)
```

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- (x) User Experience
- ( ) Developer Experience

**Expected behavior**
The active contexts associated with each target level should be maintained and used when needed.

See description

mpanchajanya:

- Introduce Multi Config files `config.yaml` and `config-ng.yaml` to manage client config data.

- Introduce new API's to config-metadata to manage what goes into `config-ng.yaml`.

- `contexts` and `currentContexts` will be moved to `config-ng.yaml`.

- New type`settings` has been added to `config-metadata.yaml` to store any internal settings related to config operations which are controlled by core cli.

- `useUnifiedConfig` -  Used to move all the config related operations to `config-ng.yaml`. By default this setting is disabled and will be enabled in next future versions of cli.
- Existing `getClientConfigNode` and `getClientConfigNodeNoLock ` function implementation is updated to handle multiple config files.
- New factory and filesytem files are added to perform operations on `config-ng.yaml`.
- `TANZU_CONFIG_NEXT_GEN` Env key is defined to explicitly set `config-ng.yaml` file.

##### Read from Config

1. Check for `useUnifiedConfig` settings in `config-metadata.yaml`.
- if enabled Read `config-ng.yaml` and construct yaml node.
- If disabled go to (2)
2. Construct root node with cfgItems from config.yaml and next gen nodes from config-ng.yaml.
3. Return the multi config root node.

##### Write to Config

1. Check for `useUnifiedConfig` settings in `config-metadata.yaml`.
- if enabled Write to `config-ng.yaml`.
- If disabled go to (2)
2. Create root cfg node and root next gen node.
3. Process through the change node and construct the root cfg node and  root next gen node based on cfgItems.
4. Write the root cfg node to config.yaml.
5. Write the root next gen node to config-ng.yaml.
6. Also populate the legacyClientConfig directory `.tanzu`.

#### Tanzu Command changes
`tanzu config get` cmd implementation is changed to handle multi files. now it will retrieve the combination of both `config.yaml` and `config-ng.yaml`

#### Minor Bug Fix

- Fix the legacyConfigFilePath that uses `.tanzu` directory to not rely on `TANZU_CONFIG` env variable that is conflicted with `config.yaml`.

Example:- Default behaviour
```
configMetadata:
  settings:
    useUnifiedConfig: false
```
Outcome :- Only contexts and currentContexts read and write to config-ng.yaml

Example :-
```
configMetadata:
  settings:
    useUnifiedConfig: true
```
Outcome :- All config item read and write to config-ng.yaml

Fixes #2652
Fixes #3810

Unit Testing and Integration Testing
Tested `tanzu config get`, `tanzu login`, `tanzu context create` etc..

### Release note

```release-note
contexts and current context are now stored in new config-ng.yaml
```

Review the `clientconfig_factory.go` and  `config_factory.go` first to familiarize with the Multi File Read/Write works.


cli/runtime/config/config_factory.go
```diff
@@ -8,22 +8,194 @@ import (
...
+	if node2.Content[0].Content != nil && len(node2.Content[0].Content) > 0 {
+		node2Bytes, err := yaml.Marshal(node2)
+		if err != nil {
+			return nil, err
+		}
+
+		if len(node2Bytes) != 0 {
+			multiBytes = append(multiBytes, node2Bytes...)
+		}
+	}
```
chandrareddyp, line 94: better have a helper function to avoid code duplication

cli/runtime/config/nodeutils/helpers.go
```diff
@@ -18,3 +23,21 @@ func UniqNodes(nodes []*yaml.Node) []*yaml.Node {
...
+		return left.Value == right.Value, nil
+	}
+	return false, errors.New("equals on non-scalars not implemented")
+}
+
+func checkErrors(src, dst *yaml.Node) error {
+	if src.Kind != dst.Kind {
+		return ErrDifferentArgumentsTypes
+	}
+	if dst != nil && reflect.ValueOf(dst).Kind() != reflect.Ptr {
```
chandrareddyp, line 39: we don't need to check dst as `reflect.Ptr` as a function accepts only pointers

chandrareddyp (changes requested):

cli/runtime/config/nodeutils/helpers.go
mpanchajanya: removed

cli/runtime/config/config_factory.go
```diff
@@ -8,22 +8,194 @@ import (
...
+	if err != nil {
+		return cfgNextGenNode, err
+	}
+
+	return makeMultiFileCfg(cfgNode, cfgNextGenNode)
 }
 
-type CfgOpts func(config *CfgOptions)
+// Construct multi file node from config.yaml and config-ng.yaml nodes
+func makeMultiFileCfg(node1, node2 *yaml.Node) (*yaml.Node, error) {
```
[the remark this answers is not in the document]
mpanchajanya: updated

cli/runtime/config/config_factory.go
mpanchajanya: refactored

cli/runtime/config/cli_discovery_sources_it_test.go
```diff
@@ -115,43 +122,75 @@ contexts:
...
-	f, err := os.CreateTemp("", "tanzu_config")
+	// Setup config data
+	cfg, expectedCfg, cfg2, expectedCfg2 := setupData()
+
+	f1, err := os.CreateTemp("", "tanzu_config")
 	assert.Nil(t, err)
-	err = os.WriteFile(f.Name(), []byte(tanzuConfigBytes), 0644)
+	err = os.WriteFile(f1.Name(), []byte(cfg), 0644)
 	assert.Nil(t, err)
+
```
vuil, line 158: nit: Can we refactor this config setup/cleanup code that seems to be identical or very similar across more than a dozen places? It would reduce the amount of test changes considerably. You can have the setup return a cleanup function that you can use to control when to call as appropriate in your test.

vuil (commented):
Will send additional comments shortly. 
Nit: did you mean the section headers in the PR description to be rendered differently in markdown (they are not ATM)

chandrareddyp (commented):
this is a review for below code in in node_types.go file:
```
type CfgNode struct {
	ForceCreate bool  // Set to True to create nodes of missing keys. Ex: True for Add/Update operations on yaml node, False for Get/Delete operations on yaml node
	Keys        []Key // keys of config nodes passed in hierarchical order. Ex: [ClientOptions, CLI, DiscoverySources] to get the DiscoverySources node from ClientOptions yaml node
}
```
its not part of this PR, it was there in previous PR https://github.com/vmware-tanzu/tanzu-framework/pull/3905
if the flag `ForceCreate` is set true in case of Get/Delete, then it creates new node with given key, i don't think its good approach, because assume that i am checking if a context or server is available, if not then we are creating server/context with given key and empty value, later if i want to create server/context, i can not create as its already exists.

I feel better to restrict the flag `ForceCreate` for just Add/Update operations.
its being used in api `func FindNode(node *yaml.Node, opts ...Options)`

mpanchajanya (commented):
/test install-vc7

cli/runtime/config/metadata_settings_api.go
```diff
@@ -0,0 +1,182 @@
...
+	return "", errors.New("not found")
+}
+
+func deleteSpecificSetting(node *yaml.Node, key string) (err error) {
+	// find feature flags node
+	keys := []nodeutils.Key{
+		{Name: KeyConfigMetadata, Type: yaml.MappingNode},
+		{Name: KeySettings, Type: yaml.MappingNode},
+	}
+	featureFlagsNode := nodeutils.FindNode(node.Content[0], nodeutils.WithKeys(keys))
```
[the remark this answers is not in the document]
mpanchajanya: updated the naming

cli/runtime/config/metadata_settings_api.go
```diff
@@ -0,0 +1,182 @@
...
+		return nil, err
+	}
+	if cfgMetadata != nil && cfgMetadata.ConfigMetadata != nil &&
+		cfgMetadata.ConfigMetadata.Settings != nil {
+		return cfgMetadata.ConfigMetadata.Settings, nil
+	}
+	return nil, nil
+}
+
+func getConfigMetadataSpecificSetting(node *yaml.Node, key string) (string, error) {
```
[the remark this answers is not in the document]
mpanchajanya: updated the function naming

mpanchajanya:
> Will send additional comments shortly. Nit: did you mean the section headers in the PR description to be rendered differently in markdown (they are not ATM)

yea fixed it

cli/runtime/config/config_factory.go
```diff
@@ -8,22 +8,194 @@ import (
...
+	cfgNode := node1.Content[0]
+
+	for _, nextGenItem := range nextGenCfgItems {
+		// Find the next gen item node from the config.yaml
+		nextGenItemNodeIndex := nodeutils.GetNodeIndex(cfgNode.Content, nextGenItem)
+		if nextGenItemNodeIndex == -1 {
+			continue
+		}
+		// Delete the next gen item node from the config.yaml
+		cfgNode.Content = append(cfgNode.Content[:nextGenItemNodeIndex-1], cfgNode.Content[nextGenItemNodeIndex:]...)
```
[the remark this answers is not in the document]
mpanchajanya: as discussed this is expected behaviour

cli/runtime/config/cli_discovery_sources_it_test.go
mpanchajanya: refactored the test data helper functions

cli/runtime/config/config_factory.go
```diff
@@ -8,22 +8,184 @@ import (
...
+		return err
+	}
+
+	cfgNextGenNode, err := getClientConfigNextGenNodeNoLock()
+	if err != nil {
+		return err
+	}
+
+	// next gen config items that goes to config-ng.yaml
+	nextGenItemNodes := []*yaml.Node{
```
vuil, line 138: something for consideration:

since the set keys of nodes we intent to write back to the legacy config file is not only known, but fixed _forever_, 
would it be cleaner to hard code that list instead? We can then loop through the keys of the incoming node's context and crosscheck against that list, and update the appropriate config/configng node. 

The implementation would then not need updating when new config sections are added to config-ng, and we run no risk of somehow introducing new stuff back into the legacy config.
We also avoid need to do the content resetting bit in L164

cli/runtime/config/nodeutils/delete_nodes.go
```diff
@@ -50,15 +50,13 @@ func replaceNodes(src, dst *yaml.Node, patchStrategyKey string, patchStrategies
...
-					i--
+					dst.Content = append(dst.Content[:i], dst.Content[i+2:]...)
+					i -= 2
 					break
 				}
 
-				if err := replaceNodes(src.Content[j+1], dst.Content[i+1], key, patchStrategies); err != nil {
-					return errors.New("replace at key " + src.Content[i].Value + ": " + err.Error())
+				if err := deleteNodes(src.Content[j+1], dst.Content[i+1], key, patchStrategies); err != nil {
+					return errors.Wrap(err, " replace at key "+src.Content[i].Value)
```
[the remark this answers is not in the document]
mpanchajanya: updated

cli/runtime/config/nodeutils/delete_nodes.go
```diff
@@ -67,22 +65,20 @@ func replaceNodes(src, dst *yaml.Node, patchStrategyKey string, patchStrategies
...
 	case yaml.DocumentNode:
-		err := replaceNodes(src.Content[0], dst.Content[0], patchStrategyKey, patchStrategies)
+		err := deleteNodes(src.Content[0], dst.Content[0], patchStrategyKey, patchStrategies)
 		if err != nil {
-			return errors.New("replace at key " + src.Content[0].Value + ": " + err.Error())
+			return errors.Wrap(err, "replace at key "+src.Content[0].Value)
 		}
 	default:
-		return errors.New("can only merge mapping nodes")
+		return errors.New("unknown node type: can only replace know node types")
```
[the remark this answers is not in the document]
mpanchajanya: updated

cli/runtime/config/config_factory.go
mpanchajanya: Changed the Implementation it is now much simpler

mpanchajanya:
> this is a review for below code in in node_types.go file:
> 
> ```
> type CfgNode struct {
> 	ForceCreate bool  // Set to True to create nodes of missing keys. Ex: True for Add/Update operations on yaml node, False for Get/Delete operations on yaml node
> 	Keys        []Key // keys of config nodes passed in hierarchical order. Ex: [ClientOptions, CLI, DiscoverySources] to get the DiscoverySources node from ClientOptions yaml node
> }
> ```
> 
> its not part of this PR, it was there in previous PR #3905 if the flag `ForceCreate` is set true in case of Get/Delete, then it creates new node with given key, i don't think its good approach, because assume that i am checking if a context or server is available, if not then we are creating server/context with given key and empty value, later if i want to create server/context, i can not create as its already exists.
> 
> I feel better to restrict the flag `ForceCreate` for just Add/Update operations. its being used in api `func FindNode(node *yaml.Node, opts ...Options)`

as discussed in earlier meeting closing the comment

mpanchajanya (commented):
/test install-vc7

alfredthenarwhal:
> @anujc25:  /test install-vc7
Commit: cc53f32

Build failed! Build no: 3427

cli/runtime/config/config_factory.go
```diff
@@ -8,22 +8,161 @@ import (
...
+	}
+
+	// Store the nextGenItem config data to config-ng.yaml
+	err = persistClientConfigNextGen(rootCfgNextGenNode)
+	if err != nil {
+		return err
+	}
+
+	// Store the config data to legacy client config file/location
+	err = persistLegacyClientConfig(node)
```
vuil, line 157: Thanks for updating the implementation. It is cleaner and less subject to change now.

Unless I am missing something, doesn't this line override what L145 is trying to do?

cli/runtime/config/nodeutils/helpers.go
```diff
@@ -18,3 +21,32 @@ func UniqNodes(nodes []*yaml.Node) []*yaml.Node {
...
+}
+
+func checkErrors(src, dst *yaml.Node) error {
+	if src.Kind != dst.Kind {
+		return ErrDifferentArgumentsTypes
+	}
+	return nil
+}
+
+func AppendNodeBytes(rootBytes []byte, documentNode *yaml.Node) ([]byte, error) {
```
vuil, line 40: Is this getting used anywhere?

cli/runtime/config/config_factory.go
```diff
@@ -8,22 +8,161 @@ import (
...
+	}
+
+	return makeMultiFileCfg(cfgNode, cfgNextGenNode)
+}
+
+// Construct multi file node from config.yaml and config-ng.yaml nodes
+func makeMultiFileCfg(cfgNode, cfgNextGenNode *yaml.Node) (*yaml.Node, error) {
+	// config nodes that goes to config.yaml
+	cfgItems := []string{
+		KeyAPIVersion,
```
vuil, line 53: nit: perhaps the the list itself should be a constant since it is used in more than one place?

cli/runtime/config/clientconfignextgen_filesystem.go
```diff
@@ -0,0 +1,37 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
```
vuil, line 3: question: what's the rationale for naming a set of files in this dir **_filesystem**.go? 
maybe _file.go would be more appropriate? 
(If we elect to change, and especially since there are preexisting files already committed, let's do it as a followup.)

cli/runtime/config/config_factory.go
mpanchajanya: legacy referred is .tanzu folder L145 is config.yaml.

cli/runtime/config/nodeutils/helpers.go
mpanchajanya: removing its unused

cli/runtime/config/config_factory.go
mpanchajanya: renaming

vuil (approved):
Thanks for the changes! Looks good to me. 
Couple of minor things. 
At top of the PR description, where you described a lot of what the PR does, please also add a line or two about the why.

cli/runtime/config/config_factory_test.go
```diff
@@ -0,0 +1,1245 @@
...
+	ctx, err := getContext(node, "test-mc")
+	assert.NoError(t, err)
+	assert.Equal(t, expectedCtx, ctx)
+
+	_, err = getServer(node, "test-mc")
+	assert.Equal(t, "could not find server \"test-mc\"", err.Error())
+}
+
+func TestPersistConfig(t *testing.T) {
+	// Setup config test data
```
vuil, line 441: should you need to rebase, one more suggestion:

As implemented the persistConfig functionality will actually preserve any nodes in the main config.yaml that is not on the hardcode list. It would be good to have a test communicate that behavior.

perhaps something like:
```
--- a/cli/runtime/config/config_factory_test.go
+++ b/cli/runtime/config/config_factory_test.go
@@ -25,6 +25,7 @@ clientOptions:
         local:
           name: default-local
           path: standalone
+      # extra comment my local admin source
       - local:
           name: admin-local
           path: admin
@@ -99,6 +100,9 @@ contexts:
         contextType: tmc
 currentContext:
   k8s: test-mc
+# extraneous fiels/comments are preserved on file update
+extrafield:
+    extrasubfield: 1
 `
        expectedCfg := `apiVersion: config.tanzu.vmware.com/v1alpha1
 clientOptions:
@@ -110,6 +114,7 @@ clientOptions:
               local:
                 name: default-local
                 path: standalone
+            # extra comment my local admin source
             - local:
                 name: admin-local
                 path: admin
@@ -184,6 +189,9 @@ contexts:
           contextType: tmc
 currentContext:
     k8s: test-mc
+# extraneous fiels/comments are preserved on file update
+extrafield:
+    extrasubfield: 1
 `
```

anujc25 (commented):
Looks good to me.

chandrareddyp (approved):
Great job. LGTM! 
Thanks.

anujc25:
This PR looks good to get in. If there are any other minor changes required, we can do it as a followup.

cli/runtime/config/config_factory_test.go
mpanchajanya: addressing  this in follow up PR

PR #3962, opened 2022-11-21
Remove package installing when deleting cc mgmt cluster

zhouke1991:
It is not necessary to install the kapp controller and meta-package when deleting classy mgmt cluster, so drop it.


DanielXiao (approved):

wjun (approved):
LGTM.

PR #3964, opened 2022-11-21
Skip deletion of ako packageInstall when cluster is deleted since AKO…

lubronzhan:
… needs to clean up external resources.

AKO pod needs to clean up resource when cluster is getting deleted. But currently, its packageInstall just got deleted when cluster is deleted so it can't finish clean up. This will block cluster deletion since AKO has finalizer on the cluster

Added a new annotations on ClusterBootstrap
`run.tanzu.vmware.com/skip-packageinstall-deletion`
`SkipDeletePackageInstallAnnotation` is the annotation that contains a list of comma separated packageInstalls to skip when cluster is deleted
for example
```
"run.tanzu.vmware.com/skip-packageinstall-deletion": "cpi,cni,load-balancer-and-ingress-service"
```

Fixes #

### Release note

```release-note

```

### DIFF
--- a/addons/pkg/util/addon_util.go
+++ b/addons/pkg/util/addon_util.go
@@ -87,14 +87,6 @@ func GetPackageMetadata(ctx context.Context, c client.Client, carvelPkgName, car
 	return pkg.Spec.RefName, pkg.Spec.Version, nil
 }
 
-func GetPackageAnnotations(ctx context.Context, c client.Client, carvelPkgName, carvelPkgNamespace string) (map[string]string, error) {
-	pkg := &kapppkgv1alpha1.Package{}
-	if err := c.Get(ctx, client.ObjectKey{Name: carvelPkgName, Namespace: carvelPkgNamespace}, pkg); err != nil {
-		return nil, err
-	}
-	return pkg.Annotations, nil
-}
-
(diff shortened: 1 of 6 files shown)


addons/pkg/util/addon_util.go
```diff
@@ -87,6 +87,14 @@ func GetPackageMetadata(ctx context.Context, c client.Client, carvelPkgName, car
 	return pkg.Spec.RefName, pkg.Spec.Version, nil
 }
 
+func GetPackageAnnotations(ctx context.Context, c client.Client, carvelPkgName, carvelPkgNamespace string) (map[string]string, error) {
```
wjun, line 90: Where is this func used?

wjun:
If we skip the packageinstall deletion, when/how do we delete it?

lubronzhan:
We don't delete the packageIntall itself, it will be deleted as part of cluster. Like other packageInstall, cpi/csi/cni

addons/pkg/util/addon_util.go
lubronzhan: Good catch

wjun (approved):
LGTM.

12345lcr (approved):

12345lcr:
LGTM

PR #3968, opened 2022-11-22
CLI Core - base framework for end-to-end  test cases implementation

chandrareddyp:

- Currently, CLI core doesn't have end-to-end test cases for CLI core functionalities; this PR introduces the end-to-end test basic framework and helpers to write E2E test cases.
- Makefile target `make test-e2e` has been added to execute end-to-end test cases
- GitHub action has been added to execute end-to-end test cases as part of PR raised for any changes in CLI core and CLI runtime code.
- E2E pre-setup creates separate directory `$HOME/.tanzu-cli-e2e` to create config files and use while executing E2E tests, E2E tests does not touch default config files from `$HOME/.config/tanzu`
- E2E Test cases implemented for `tanzu config`, `tanzu init` and `tanzu version`

Fixes #NA

- Makefile target `make test-e2e` added, and tested all e2e test cases using the `make test-e2e` target
- E2E tests executed locally and confirmed that it does update default config file. It creates separate directory `$HOME/.tanzu-cli-e2e` to create config files and to execute e2e tests.
- Github action `Tanzu CLI Core E2E Tests` added and its executed E2E tests as part of pull request

### Release note

```release-note

```

### DIFF
--- a/cli/core/test/e2e/config/config_test.go
+++ b/cli/core/test/e2e/config/config_test.go
@@ -1,0 +1,75 @@
+// Package config_e2e_test provides config command specific E2E test cases
+package config_e2e_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/vmware-tanzu/tanzu-framework/cli/core/test/e2e/framework"
+)
+
+var _ = framework.CLICoreDescribe("[Tests:E2E][Feature:Command-Config]", func() {
+	var (
+		tf *framework.Framework
+	)
+	BeforeEach(func() {
+		tf = framework.NewFramework()
+	})
+	Context("config feature flag operations", func() {
+		When("new config flag set with value", func() {
+			It("should set flag and unset flag successfully", func() {
+				randomFlagName := "e2e-test-" + framework.RandomString(4)
+				randomFeatureFlagPath := "features.global." + randomFlagName
+				flagVal := "true"
+				err := tf.ConfigSetFeatureFlag(randomFeatureFlagPath, flagVal)
+				Expect(err).To(BeNil())
+
+				val, err := tf.ConfigGetFeatureFlag(randomFeatureFlagPath)
+				Expect(err).To(BeNil())
+				Expect(val).Should(Equal("true"))
+
+				err = tf.ConfigUnsetFeature(randomFeatureFlagPath)
+				Expect(err).To(BeNil())
+
+				val, err = tf.ConfigGetFeatureFlag(randomFeatureFlagPath)
+				Expect(err).To(BeNil())
+				Expect(val).Should(Equal(""))
+			})
+		})
+		When("config init called when config files not exists", func() {
+			It("should initialize configuration successfully", func() {
+				// delete config files
+				err := tf.DeleteCLIConfigurationFiles()
+				Expect(err).To(BeNil())
+				// call init
+				err = tf.ConfigInit()
+				Expect(err).To(BeNil())
+				// should create config files
+				Expect(tf.IsCLIConfigurationFilesExists()).To(BeTrue())
+
+				// set feature flag
+				randomFlagName := "e2e-test-" + framework.RandomString(4)
+				randomFeatureFlagPath := "features.global." + randomFlagName
+				flagVal := "true"
+				err = tf.ConfigSetFeatureFlag(randomFeatureFlagPath, flagVal)
+				Expect(err).To(BeNil())
+
+				val, err := tf.ConfigGetFeatureFlag(randomFeatureFlagPath)
+				Expect(err).To(BeNil())
+				Expect(val).Should(Equal("true"))
+
+				// call init
+				err = tf.ConfigInit()
+				Expect(err).To(BeNil())
+				// second run of init should not remove the existing feature flag
+				val, err = tf.ConfigGetFeatureFlag(randomFeatureFlagPath)
+				Expect(err).To(BeNil())
+				Expect(val).Should(Equal("true"))
+
+				// unset the feature flag
+				err = tf.ConfigUnsetFeature(randomFeatureFlagPath)
+				Expect(err).To(BeNil())
+			})
+		})
+	})
+})
--- a/cli/core/test/e2e/framework/config_operations.go
+++ b/cli/core/test/e2e/framework/config_operations.go
@@ -1,0 +1,129 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package framework
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+
+	configapi "github.com/vmware-tanzu/tanzu-framework/cli/runtime/apis/config/v1alpha1"
+)
+
+const (
+	ConfigFileName   = "config.yaml"
+	ConfigNGFileName = "config-ng.yaml"
+	ConfigFileDir    = ".config/tanzu/"
+)
+
+// ConfigOps performs "tanzu config" command operations
+type ConfigOps interface {
+	ConfigSetFeatureFlag(path, value string) error
+	ConfigGetFeatureFlag(path string) (string, error)
+	ConfigUnsetFeature(path string) error
+	ConfigInit() error
+	ConfigServerList() error
+	ConfigServerDelete(serverName string) error
+	DeleteCLIConfigurationFiles() error
+	IsCLIConfigurationFilesExists() bool
+}
+
+// configOps is the implementation of ConfOps interface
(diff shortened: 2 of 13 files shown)


cli/core/test/e2e/config/config_test.go
```diff
@@ -0,0 +1,59 @@
...
+				Expect(err).To(BeNil())
+
+				val, err = tf.ConfigFeatureFlagNotExists(randomFlagName)
+				Expect(err).To(BeNil())
+				Expect(val).To(BeTrue())
+			})
+		})
+		When("config init called", func() {
+			It("should initialize configuration successfully", func() {
+				err := tf.ConfigInit()
```
anujc25, line 42: We need to make sure that `tanzu config init` command performed the things that it should have rather then just checking for error here.

One another way to test this would be to do tests in below sequence:
1. Clean all the config files (we might need a helper function in the framework to cleanup all the configuration that tanzu cli uses)
2. Run `tanzu config init`
3. Verify that `$HOME/.config/tanzu/config.yaml` and `$HOME/.config/tanzu/config-ng.yaml` gets created
4. Run feature-flag set tests
5. Run feature-flag get tests
6. Run `tanzu config init` again
7. Verify that second run of `tanzu config init` doesn't cleanup the existing feature-flag which was already configured
8. Run feature-flag unset tests

cli/core/test/e2e/framework/config_operations.go
```diff
@@ -0,0 +1,80 @@
...
+
+// ConfigFeatureFlagNotExists validates the existence of feature flag in tanzu config
+func (co *configOps) ConfigFeatureFlagNotExists(flagName string) (bool, error) {
+	getConfCmd := ConfigGet
+	err := co.ExecNotContainsString(getConfCmd, flagName+": ")
+	if err != nil {
+		return false, err
+	}
+	return true, err
+}
```
anujc25, line 54: I think instead of implementing functions like `ConfigFeatureFlagNotExists` and `ConfigFeatureFlagExists` here, it would be better to implement function like `ConfigGetFeatureFlag` which parses the yaml and returns the feature-flag value if it exists else returns error.

As part of the tests, we can just to `ConfigGetFeatureFlag` to get the feature-flag value and compare the result as part of the tests.

cli/core/test/e2e/config/config_test.go
```diff
@@ -0,0 +1,59 @@
...
+			})
+		})
+
+		// TODO: test list and delete use cases as part of context or login commands
+		When("config server list called", func() {
+			It("should list all available servers", func() {
+				err := tf.ConfigServerList()
+				Expect(err).To(BeNil())
+			})
+		})
```
anujc25, line 53: Let's not add a partial test for server list here. I am not seeing much meaning of this test here considering we are not adding or deleting any servers for the tests. It might be good to add tests for this along with other scenarios in a separate test file.

anujc25:
When we run this e2e test suite, does it update users existing tanzu config files under `~/.config/tanzu/config.yaml` ?

If yes, we need to make sure that when running this locally, we do not update user's existing configuration by configuring different `$HOME` environment variable for tests.

cli/core/test/e2e/config/config_test.go
chandrareddyp: I have updated the test case, to delete config files, initialize, validate config set, and again initialize and validate config.

cli/core/test/e2e/config/config_test.go
chandrareddyp: I have removed tests for now, i will add servers specific tests in separate file.

cli/core/test/e2e/framework/config_operations.go
chandrareddyp: Yes thats better API, i have created the new API `ConfigGetFeatureFlag`

cli/core/go.mod
```diff
@@ -48,6 +48,7 @@ require (
 	k8s.io/api v0.24.2
 	k8s.io/apimachinery v0.24.2
 	k8s.io/client-go v0.24.2
+	sigs.k8s.io/cluster-api v1.2.7
```
anujc25, line 51: We should avoid adding this dependency on core-cli as we are just using `util.RandomString(4)` function. We should try to find some alternative.

cli/core/test/e2e/config/config_test.go
```diff
@@ -0,0 +1,82 @@
...
+				// val, err := tf.ConfigFeatureFlagExists(randomFlagName, flagVal)
+				Expect(err).To(BeNil())
+				Expect(val).Should(Equal("true"))
+
+				// call init
+				err = tf.ConfigInit()
+				Expect(err).To(BeNil())
+				// second run of init should not remove the existing feature flag
+				val, err = tf.ConfigGetFeatureFlag(randomFeatureFlagPath)
+				// val, err = tf.ConfigFeatureFlagExists(randomFlagName, flagVal)
```
anujc25, line 68: nit: remove this

cli/core/test/e2e/framework/config_operations.go
```diff
@@ -0,0 +1,144 @@
...
+		return "", errors.Wrap(err, "failed to construct yaml node from config get output")
+	}
+	cnf, err := convertNodeToClientConfig(&node)
+	if err != nil {
+		return "", err
+	}
+	featureName := strings.Split(path, ".")[len(strings.Split(path, "."))-1]
+	if cnf != nil && cnf.ClientOptions.Features["global"] != nil {
+		return cnf.ClientOptions.Features["global"][featureName], nil
+	}
```
anujc25, line 72: `path` here can be `features.<plugin-name>.<featureName>` so we shouldn't expect that it is `features.global.<featureName>`.

cli/core/test/e2e/framework/config_operations.go
```diff
@@ -0,0 +1,144 @@
...
+
+	var node yaml.Node
+	err = yaml.Unmarshal(out.Bytes(), &node)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to construct yaml node from config get output")
+	}
+	cnf, err := convertNodeToClientConfig(&node)
+	if err != nil {
+		return "", err
+	}
```
anujc25, line 68: You can directly unmarshal the struct into `configapi.ClientConfig` object here for this tests. We do not need to unmarshal it into `Node` and convert that into `configapi.ClientConfig` object.

Something like:
```
	var cnf configapi.ClientConfig
	err = yaml.Unmarshal(out.Bytes(), &node)
	if err != nil {
		return "", errors.Wrap(err, "failed to unmarshal config get output into ClientConfig")
	}
```

cli/core/test/e2e/framework/config_operations.go
```diff
@@ -0,0 +1,144 @@
...
+	_, err1 := os.Stat(configFilePath)
+	_, err2 := os.Stat(configNGFilePath)
+	if err1 == nil && err2 == nil {
+		return true
+	}
+	return false
+}
+
+// convertNodeToClientConfig converts yaml node to client config type
+func convertNodeToClientConfig(node *yaml.Node) (obj *configapi.ClientConfig, err error) {
```
anujc25, line 135: based on the above comment, I think we can don't need this function and it can be removed.

cli/core/go.mod
chandrareddyp: Removed the cluster-api dependency and added util function.

cli/core/test/e2e/config/config_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+				// set feature flag
+				randomFlagName := RandomNameWithE2ETestString()
+				randomFeatureFlagPath := "features.global." + randomFlagName
+				flagVal := "true"
+				err = tf.ConfigSetFeatureFlag(randomFeatureFlagPath, flagVal)
+				Expect(err).To(BeNil())
+
+				val, err := tf.ConfigGetFeatureFlag(randomFeatureFlagPath)
+				// val, err := tf.ConfigFeatureFlagExists(randomFlagName, flagVal)
```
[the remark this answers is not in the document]
chandrareddyp: done

cli/core/test/e2e/config/config_test.go
chandrareddyp: done

cli/core/test/e2e/framework/config_operations.go
chandrareddyp: updated, thanks.

cli/core/test/e2e/framework/config_operations.go
chandrareddyp: removed

cli/core/test/e2e/framework/config_operations.go
chandrareddyp: Logic has updated to consider any plugin-name

anujc25 (approved):
LGTM. Thanks!

PR #3983, opened 2022-11-23
[Workaround]Schedule featuregate pod on controlplane nodes

12345lcr:
Schedule featuregate pod on controlplane nodes by adding the nodeSelector to fix the CAPA ClusterClass CI.

This may not be the final solution since it may related to aws security group and node ports.

Fixes #

### Release note

```release-note

```

### DIFF
--- a/packages/featuregates/bundle/config/values.yaml
+++ b/packages/featuregates/bundle/config/values.yaml
@@ -1,9 +1,9 @@
 #@data/values
 ---
-
 namespace: tkg-system
 deployment:
   hostNetwork: true
-  nodeSelector: null
+  nodeSelector:
+    node-role.kubernetes.io/control-plane: ""
   tolerations: []
   webhookServerPort: 9443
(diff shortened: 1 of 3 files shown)


danniel1205:
Thanks for the change. If adding NodeSelector is the fix, I think it could be added via https://github.com/vmware-tanzu/tanzu-framework/blob/main/packages/featuregates/bundle/config/values.yaml#L7

12345lcr:
Thanks @danniel1205 , just realized that this field is already defined in the overlay, I updated this part.

yharish991 (approved):
LGTM!!

danniel1205:
Thanks for the PR, changes seem reasonable but it's worth a revisit. The CAPA is still failing but it is a different issue. After looking into the log, it does fix the featuregate deployment issue we saw earlier. So, I am adding my approval on this PR. Feel free to merge it if it blocks everyone, otherwise I would suggest hold on until 28th once folks are back from Thanksgiving holiday.

danniel1205 (approved):

hongtaiwang:
should we add node selector for other packages with `hostNetwork: true` configured such as addonsManager and tanzuauth?

PR #3990, opened 2022-11-23
Fix upgrade check to allow upgrades to 2.x.x

tenczar:
This change updates are upgrade validation logic to allow users to upgrade from 1.6.x to 2.x.x. It will return an error if a user attempts to upgrade from a version less than 1.6 or if they try to upgrade to major version 3. Obviously that check will need to be updated on release of major version 3.

This fix allows us to move to TKG with major version 2. A previous check would throw an error if the major version was changed.

Fixes #

Updated unit tests to include validation that upgrades to major version 3 still throw an error. Also added test to make sure an error is thrown when upgrade to v2 from a version less than v1.6

### Release note

```release-note
Upgrades to major version 2 no longer throw an error.
```

### DIFF
--- a/tkg/client/upgrade_region.go
+++ b/tkg/client/upgrade_region.go
@@ -494,9 +494,13 @@ func (c *TkgClient) ValidateManagementClusterUpgradeVersionCompatibility(options
 	log.V(9).Infof("Management cluster SemVersion is %q and current TKG SemVersion(to be upgraded to) is %q",
 		mgmtClusterSemVersion.String(), currentTKGSemVersion.String())
 
+	if currentTKGSemVersion.Major() == uint(2) && (mgmtClusterSemVersion.Major() == uint(1) && mgmtClusterSemVersion.Minor() < uint(6)) {
+		return errors.Errorf("management cluster version must be 1.6.x to upgrade to 2.1.x")
+	}
+
 	// TODO: Update this condition when TKG major version is changed
-	if mgmtClusterSemVersion.Major() != currentTKGSemVersion.Major() {
-		return errors.Errorf("major version mismatch detected")
+	if currentTKGSemVersion.Major() > uint(2) {
+		return errors.Errorf("upgrading to beyond major version 2 is not yet supported.")
 	}
 	warningMsg := ""
 	minorGap := int(currentTKGSemVersion.Minor()) - int(mgmtClusterSemVersion.Minor())
--- a/tkg/client/upgrade_region_test.go
+++ b/tkg/client/upgrade_region_test.go
@@ -221,13 +221,26 @@ var _ = Describe("Unit tests for upgrade management cluster", func() {
 					Expect(err.Error()).To(ContainSubstring("unable to parse management cluster version InvalidTKGSemanticVersion"))
 				})
 			})
-			Context("When TKG version user trying to upgrade has different major version", func() {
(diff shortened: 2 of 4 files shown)


tkg/client/upgrade_region.go
```diff
@@ -494,9 +494,13 @@ func (c *TkgClient) ValidateManagementClusterUpgradeVersionCompatibility(options
...
 
+	if currentTKGSemVersion.Major() == uint(2) && (mgmtClusterSemVersion.Major() == uint(1) && mgmtClusterSemVersion.Minor() < uint(6)) {
+		return errors.Errorf("management cluster version must be at least 1.6.x to upgrade to 2.1.x")
+	}
+
 	// TODO: Update this condition when TKG major version is changed
-	if mgmtClusterSemVersion.Major() != currentTKGSemVersion.Major() {
-		return errors.Errorf("major version mismatch detected")
+	if currentTKGSemVersion.Major() > uint(2) {
+		return errors.Errorf("upgrading to beyond major version 2 is not uet supported.")
```
zjs, line 503:
```suggestion
		return errors.Errorf("upgrading to beyond major version 2 is not yet supported.")
```

tkg/client/upgrade_region.go
```diff
@@ -494,9 +494,13 @@ func (c *TkgClient) ValidateManagementClusterUpgradeVersionCompatibility(options
 	log.V(9).Infof("Management cluster SemVersion is %q and current TKG SemVersion(to be upgraded to) is %q",
 		mgmtClusterSemVersion.String(), currentTKGSemVersion.String())
 
+	if currentTKGSemVersion.Major() == uint(2) && (mgmtClusterSemVersion.Major() == uint(1) && mgmtClusterSemVersion.Minor() < uint(6)) {
+		return errors.Errorf("management cluster version must be at least 1.6.x to upgrade to 2.1.x")
```
zjs, line 498:
```suggestion
		return errors.Errorf("management cluster version must be 1.6.x to upgrade to 2.1.x")
```

tkg/client/upgrade_region_test.go
```diff
@@ -221,13 +221,26 @@ var _ = Describe("Unit tests for upgrade management cluster", func() {
...
+			Context("When TKG version user trying to upgrade has major version greater than 2", func() {
 				BeforeEach(func() {
-					regionalClusterClient.GetManagementClusterTKGVersionReturns("v2.0.0-rc.1", nil)
+					regionalClusterClient.GetManagementClusterTKGVersionReturns("v1.5.4", nil)
+					tkgClient, err = CreateTKGClient("../fakes/config/config.yaml", testingDir, "../fakes/config/bom/tkg-bom-v3.0.0.yaml", 2*time.Millisecond)
+					Expect(err).NotTo(HaveOccurred())
+				})
+				It("should return an error", func() {
+					Expect(err).To(HaveOccurred())
+					Expect(err.Error()).To(ContainSubstring("upgrading to beyond major version 2 is not uet supported."))
```
zjs, line 232:
```suggestion
					Expect(err.Error()).To(ContainSubstring("upgrading to beyond major version 2 is not yet supported."))
```

zjs (changes requested):

zjs (approved):
LGTM

PR #3998, opened 2022-11-24
Trim the version suffix from the default ClusterClass name

DanielXiao:
When we dry run management Cluster creation, the bootstrap Cluster is not running yet but clusterctl will try to render the ClusterClass template which referenced by the Cluster at the same time.

We can get the template file name by trimming the version suffix from the default ClusterClass name.

Fixes #

### Release note

```release-note

```


wjun (approved):
LGTM

PR #4007, opened 2022-11-25
Fix the judgment of  custom storageclass

m1zzx2:

Fixes #

### Release note

```release-note

```

### DIFF
--- a/providers/yttcb/clusterbootstrap.yaml
+++ b/providers/yttcb/clusterbootstrap.yaml
@@ -87,7 +87,9 @@
 #@ end
 
 #@ def storage_class_configs_exist():
-#@   return data.values.ENABLE_DEFAULT_STORAGE_CLASS and data.values.PROVIDER_TYPE != "tkg-service-vsphere" and data.values.PROVIDER_TYPE != "docker"
+#@   return data.values.ENABLE_DEFAULT_STORAGE_CLASS and (data.values.PROVIDER_TYPE == "vsphere" or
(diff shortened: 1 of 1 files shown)


providers/yttcb/clusterbootstrap.yaml
```diff
@@ -87,7 +87,9 @@
 #@ end
 
 #@ def storage_class_configs_exist():
-#@   return data.values.ENABLE_DEFAULT_STORAGE_CLASS and data.values.PROVIDER_TYPE != "tkg-service-vsphere" and data.values.PROVIDER_TYPE != "docker"
+#@   return data.values.ENABLE_DEFAULT_STORAGE_CLASS and (data.values.PROVIDER_TYPE == "vsphere" or
```
shyaamsn, line 90: ENABLE_DEFAULT_STORAGE_CLASS is by default true, which means that this function is going to return true always.

providers/yttcb/clusterbootstrap.yaml
```diff
@@ -87,7 +87,9 @@
...
+#@                                                      data.values.PROVIDER_TYPE == "aws" or
+#@                                                      data.values.PROVIDER_TYPE == "azure")
```
randomvariable, line 92: I think this is cleaner as per https://github.com/bazelbuild/starlark/blob/master/spec.md#membership-tests 

```suggestion
#@   return data.values.ENABLE_DEFAULT_STORAGE_CLASS and (data.values.PROVIDER_TYPE in ["vsphere", "aws", "azure"]
```

PR #4011, opened 2022-11-28
Implement automatic certificates renewal

weiwenli97:
Enable certificates automatic renewal in TKG

Fixes #TKG-14228

1. create vsphere/azure/aws legacy cluster using --dry-run
    - without variable CONTROLPLANE_CERTIFICATE_ROTATION_ENABLED and CONTROLPLANE_CERTIFICATE_ROTATION_DAYS_BEFORE
       - rolloutBefore with 90 days is enabled in KCP 
    - with variable CONTROLPLANE_CERTIFICATE_ROTATION_ENABLED false
       - rolloutBefore is disabled in KCP
    - with variable CONTROLPLANE_CERTIFICATE_ROTATION_DAYS_BEFORE 30
       - rolloutBefore with 30 days is enabled in KCP 
2. create vpshere legacy cluster and upgrade
    - with variable CONTROLPLANE_CERTIFICATE_ROTATION_ENABLED false
       - Cluster was upgraded successfully and rolloutBefore is disabled as before in KCP
    - with variable CONTROLPLANE_CERTIFICATE_ROTATION_DAYS_BEFORE 30
       - Cluster was upgraded successfully and rolloutBefore is enabled with the same value as before in KCP  

### Release note

```release-note

```


packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml
```diff
@@ -273,6 +273,11 @@ spec:
       openAPIV3Schema:
         type: object
         properties: {}
+  - name: controlPlaneCertificateRotationBefore
+    required: false
+    schema:
+      openAPIV3Schema:
+        type: integer
```
sbueringer, line 280:
```suggestion
        type: integer
        minimum: 7
```
Might make sense to set a minimum here. Inclusive minimum in Cluster API is 7 
(https://github.com/kubernetes-sigs/cluster-api/blob/650b1879a586228d9ec5d81f3dd86016f588aa5d/controlplane/kubeadm/api/v1beta1/kubeadm_control_plane_webhook.go#L339)

sbueringer: Same in other places

tkg/client/upgrade_cluster.go
```diff
@@ -1115,9 +1116,13 @@ func (c *TkgClient) PatchKubernetesVersionToKubeadmControlPlane(regionalClusterC
 	}
 
 	pollOptions := &clusterclient.PollOptions{Interval: upgradePatchInterval, Timeout: upgradePatchTimeout}
-	err = regionalClusterClient.UpdateResourceWithPolling(currentKCP, clusterUpgradeConfig.KCPObjectName, clusterUpgradeConfig.KCPObjectNamespace, pollOptions)
+	kcpBytes, err := json.Marshal(currentKCP)
 	if err != nil {
-		return errors.Wrap(err, "unable to update the kubernetes version for kubeadm control plane nodes")
+		return err
+	}
+	err = regionalClusterClient.PatchResource(&capikubeadmv1beta1.KubeadmControlPlane{}, clusterUpgradeConfig.KCPObjectName, clusterUpgradeConfig.KCPObjectNamespace, string(kcpBytes), types.MergePatchType, pollOptions)
```
sbueringer, line 1123: Should we add a comment why we specifically patch this resource now? (vs. update before)

Q: I assume this was specifically tested and it doesn't drop a pre-existing rolloutBefore field?

sbueringer: Update: looks like it works:

Hm looks like my intuition is wrong.

1 Deployed a KCP with rolloutBefore

2 `kubectl patch kcp capi-quickstart-control-plane --type merge --patch-file test.yaml --v=8`
with
```yaml
apiVersion: controlplane.cluster.x-k8s.io/v1beta1
kind: KubeadmControlPlane
metadata:
  name: capi-quickstart-control-plane
  namespace: default
spec:
  machineTemplate:
    infrastructureRef:
      apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
      kind: DockerMachineTemplate
      name: capi-quickstart-control-plane
      namespace: default
  replicas: 1
  version: v1.25.0
```
log from kubectl:
```
 Request Body: {"apiVersion":"controlplane.cluster.x-k8s.io/v1beta1","kind":"KubeadmControlPlane","metadata":{"name":"capi-quickstart-control-plane","namespace":"default"},"spec":{"machineTemplate":{"infrastructureRef":{"apiVersion":"infrastructure.cluster.x-k8s.io/v1beta1","kind":"DockerMachineTemplate","name":"capi-quickstart-control-plane","namespace":"default"}},"replicas":1,"version":"v1.25.0"}}
 ```
 
3 rolloutBefore is preserved

Comment (sbueringer) [reply tkg/client/upgrade_cluster.go:1123]: It would be probably good to cover this with an envtest-based unit/integration test though

File: packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml (hunk 1 above)
Comment (randomvariable) [reply packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml:280]: We also wanted a default of 90 as per the TEP for classy clusters, and opt-in for existing YTT based clusters.

File: tkg/client/upgrade_cluster.go (hunk 2 above)
Comment (randomvariable) [reply tkg/client/upgrade_cluster.go:1123]: Thanks for the check. +1 on having an envtest test.

**Review by randomvariable — CHANGES_REQUESTED** [review]
Mostly need the default for classy clusters, and ideally an envtest test for the patch.

File: packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml (hunk 1 above)
Comment (weiwenli97) [reply packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml:280]: Yes. But if users want to disable certificate automatic renewal, how should users do? Now, if users don't set the variable, the cluster is created without certificate automatic renewal.

File: tkg/client/upgrade_cluster.go (hunk 2 above)
Comment (weiwenli97) [reply tkg/client/upgrade_cluster.go:1123]: > Should we add a comment why we specifically patch this resource now? (vs. update before)
> 
> Q: I assume this was specifically tested and it doesn't drop a pre-existing rolloutBefore field?
> 
> (I would have guessed that a json merge patch with spec.{Version,MachineTemplate,...} would drop rolloutBefore, but I could be wrong)

Yes. rolloutBefore is immutable in CAPI. It failed to upgrade if using UpdateResource (removing rolloutBefore). So we used PatchResource instead.

**Comment by weiwenli97** [comment]
> Mostly need the default for classy clusters, and ideally an envtest test for the patch.

Yes. If we give the default value, users have to enable certificate automatic renewal. How to disable it? Thanks!

File: packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml (hunk 1 above)
Comment (sbueringer) [reply packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml:280]: That's a good point. If we have a default folks cannot just simply "unset" it. We could use 0 or -1 as a way to opt-out

File: tkg/client/upgrade_cluster.go (hunk 2 above)
Comment (sbueringer) [reply tkg/client/upgrade_cluster.go:1123]: @randomvariable Wondering if that gives us enough additional signal that we don't need an additional test?
(situation will be resolved as soon as we bump to Cluster API v1.3 anyway and until then the field is immutable anyway which would lead to broken pipelines if the CLI tries to modify the field)

File: packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml (hunk 1 above)
Comment (weiwenli97) [reply packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml:280]: Now I have updated the logic and It used default value 90.
1. Users don't set the value: they will create the cluster enabling certificate automatic renewal with default value 90
2. Users set the value with the number less than 1: they will create the cluster disabling certificate automatic renewal
3. User set the value with the integer greater than 1: they will create the cluster enabling certificate automatic renewal with the input value

Comment (sbueringer) [reply packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml:280]: Probably the best tradeoff if we want to encode this behavior in just one single basic variable.

Only issue is that every value < 7 and >= 1 will be rejected by the KubeadmControlPlane webhook as the supported minimum there is 7

Comment (weiwenli97) [reply packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml:280]: > Probably the best tradeoff if we want to encode this behavior in just one single basic variable.
> 
> Only issue is that every value < 7 and >= 1 will be rejected by the KubeadmControlPlane webhook as the supported minimum there is 7
> 
> We could validate here but that would require us setting `minimum: 7` which makes it impossible to use < 1 as a signal for disabled.

Based on it, we should throw validation error if the value is set in [1, 7).

Comment (sbueringer) [reply packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml:280]: Yup, Unfortunately that's something you cannot express with the current Open API schema validation (there's only min/max allowed values).

Comment (weiwenli97) [reply packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml:280]: Yes. I cannot use minimum annotation for we support -1/0. I will update it using function checking.

File: tkg/client/upgrade_cluster.go (hunk 2 above)
Comment (randomvariable) [reply tkg/client/upgrade_cluster.go:1124]: Yeah, I think that's fine.
Maybe add a `TODO: @weiwenli97 Would be ideal to add an envtest unit test here` 
as it's a larger refactor to move from fakeclient to envtest.

**Comment by randomvariable** [comment]
Marking as WIP as the ClassyCluster implementation has been removed pending resolution of UX.

**Comment by randomvariable** [comment]
As discussed elsewhere, in order to resolve the UX issues for ClusterClass and remain consistent for legacy YTT, change the patch to something like 

image

and we should be good to go. Thanks for all your efforts so far.

File: packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml [hunk 3]
```diff
@@ -290,6 +290,23 @@ spec:
...
+      openAPIV3Schema:
+        type: object
+        properties:
+          activate:
+            type: boolean
+            default: true
+          daysBefore:
+            type: integer
+            format: int32
+            minimum: 7
```
wjun, line 305: What is the behavior if users provide a value less than 7 here?

weiwenli97: > What is the behavior if users provide a value less than 7 here?

The Cluster "test01" is invalid: spec.topology.variables[0].daysBefore: Invalid value: 6: daysBefore in body should be greater than or equal to 7
: exit status 1

It returns errors as above during creating cluster. And it doesn't fail using --dry-run.

wjun (approved):
LGTM

DanielXiao (approved):

randomvariable (approved):

randomvariable:
Rerunning the failed test and then will merge.

PR #4033, opened 2022-11-29
Support configuring etcd flags

DanielXiao:
Add Cluster variable etcdExtraArgs to set arbitrary etcd flags. Add Cluster variable tlsCipherSuites to store the default cipher suites. When etcdExtraArgs does not contain cipher-suites, the default value is set.

Add variable ETCD_EXTRA_ARGS for UX backward compatibility.

We need to support configuring etcd flags since Nimbus testbeds have only low performance storages.

Fixes #

### Release note

```release-note
Use the variable ETCD_EXTRA_ARGS to configure etcd flags. See the following example:
ETCD_EXTRA_ARGS: "heartbeat-interval=300;election-timeout=2000"
```

### DIFF
--- a/packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml
+++ b/packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml
@@ -45,6 +45,20 @@ spec:
       name: #@ "tkg-aws-default-{}-cluster".format(data.values.AWS_CLUSTER_CLASS_VERSION)
       namespace: #@ data.values.NAMESPACE
   variables:
+  - name: etcdExtraArgs
+    required: false
+    schema:
+      openAPIV3Schema:
+        type: object
+        additionalProperties:
+          type: string
+        default: {}
+  - name: tlsCipherSuites
+    required: false
+    schema:
+      openAPIV3Schema:
+        type: string
+        default: TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
   - name: region
     required: true
     schema:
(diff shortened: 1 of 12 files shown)


packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml
```diff
@@ -45,6 +45,20 @@ spec:
...
       namespace: #@ data.values.NAMESPACE
   variables:
+  - name: etcdExtraArgs
+    required: false
+    schema:
+      openAPIV3Schema:
+        type: object
+        additionalProperties:
+          type: string
+        default: {}
```
wjun, line 55: Is that possible to set default value as {cipher-suites: TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384}, so you can remove the tlsCipherSuites variable?

providers/yttcc/lib/helpers.star
```diff
@@ -323,15 +323,25 @@ end
...
 # get_labels_map_from_string constructs a map from given string of the format "key1=label1,key2=label2"
 def get_labels_map_from_string(labelString):
-   labelMap = {}
-   for val in labelString.split(','):
+   return get_map_from_string(labelString, ',')
+end
+
+# get_extra_args_map_from_string constructs a map from given string of the format "key1=label1,key2=label2"
+def get_extra_args_map_from_string(labelString):
+   return get_map_from_string(labelString, ';')
```
wjun, line 331: Why not use ',' as the delimiter?

providers/config_default.yaml
```diff
@@ -62,6 +62,11 @@ ENABLE_TKGS_ON_VSPHERE7:
 #! Valid values for BUILD_EDITION are 'tce' and 'tkg'.
 BUILD_EDITION:
 
+#! ETCD_EXTRA_ARGS specified etcd flags.
+#! It should be of the format "key1=value1;key2=value2".
```
wjun, line 66: Why not use ',' as the delimiter?

tenczar (approved):
it seems reasonable to me. I'm becoming a little concerned with how much customization is being baked directly into every clusterclass. I think podsecuritypolicy added clusterclass customizations in an interesting way, or maybe we even adopt kustomize in the future...

packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml
DanielXiao: Exactly. Users must specify cipher-suites in the variable etcdExtraArgs if we don't append the default value. E.g: If users only specify the following args, cipher-suites will be missed.
```
    - name: etcdExtraArgs
      value:
        election-timeout: "2000"
        heartbeat-interval: "300"
```

DanielXiao: variable tlsCipherSuites will be shared as the default cipher suites by all components: apiserver, kube-controller-mananger, kubelet etc. I just don't have time to change them.

providers/yttcc/lib/helpers.star
DanielXiao: As you can see, ',' is already used as the delimiter of some flags, e.g. cipher-suites `TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384`

providers/config_default.yaml
DanielXiao: Same as above.

DanielXiao:
@tenczar @wjun As you can see jsonpatch is aweful, other folks can introduce a patch to change the same fields by accident and there is no way to write unit test to ensure the expected behavior. We should try runtime sdk next release so we can easily modify map, array and leverage unit tests to avoid regression.

wjun (approved):
LGTM

PR #4034, opened 2022-11-29
fix #3817  use latest available k8s version if not specified when upgrading cluster

### LINKED ISSUE #3817: tanzu upgrade command returns "no upgrade available" after failing verification in a prior invocation
**Bug description**
1. When trying to upgrade (without arguments), we shouldn't default to the highest version; we should default to the highest valid version. If you're on v1.X, we shouldn't try to upgrade to v1.(X+2)

2. When an upgrade fails because we've attempted to skip a minor version, it seems like we need a better way to recover; just retrying with the right version should work.

- ( ) APIs
- ( ) Addons
- (x ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

- should handle empty tkrversion/k8s version correct instead of using default k8s version: https://github.com/vmware-tanzu/tanzu-framework/blob/main/tkg/tkgctl/upgrade_cluster.go#L64 
- should restore tkr version label if upgrade failed https://github.com/vmware-tanzu/tanzu-framework/blob/main/tkg/client/upgrade_cluster.go#L181

hongtaiwang:

* restore to previous k8s version if upgrade failed for cc clusters

* restore to previous tkr version label if upgrade failed for legacy clusters

* use latest available k8s version if not specified when upgrading cluster

Fixes #3817 

tests done on local testbed:
* restore tkr label for clusters when upgrade failing
```
Upgrading workload cluster 'my-legacy-wc' to kubernetes version 'v1.23.13+vmware.1'. Are you sure? [y/N]: y
Warning: Use of ytt based cluster templates will be deprecated in favor of ClusterClass templates in a future version of TKG. Please work to move your workloads to a ClusterClass enabled cluster.
Verifying kubernetes version...
Upgrading kubernetes cluster from `v1.22.11---vmware.2-tkg.2-zshippable` to `v1.23.13+vmware.1-tkg.1-zshippable` version
Restoring kubernetes cluster to tkr `v1.22.11---vmware.2-tkg.2-zshippable` version
```
* restore k8s version for cc cluster when upgrade failing
```
Upgrading workload cluster 'my-wc1' to kubernetes version 'v1.24.6+vmware.1'. Are you sure? [y/N]: y
Upgrading kubernetes cluster from `v1.22.11+vmware.2` to `v1.24.6+vmware.1` version
Restoring kubernetes cluster to `v1.22.11+vmware.2` version
```

* use latest available version when upgrade cluster without `--tkr` specified
```
kubo@0LZ8DZiMkuKTi:~/tanzu-framework$ tanzu cluster upgrade my-legacy-wc
Upgrading workload cluster 'my-legacy-wc' to kubernetes version 'v1.23.13+vmware.1'. Are you sure? [y/N]: y
Warning: Use of ytt based cluster templates will be deprecated in favor of ClusterClass templates in a future version of TKG. Please work to move your workloads to a ClusterClass enabled cluster.
Verifying kubernetes version...
Upgrading kubernetes cluster from `v1.22.11---vmware.2-tkg.2-zshippable` to `v1.23.13+vmware.1-tkg.1-zshippable` version
Retrieving configuration for upgrade cluster...
Create InfrastructureTemplate for upgrade...
Upgrading control plane nodes...
Patching KubeadmControlPlane with the kubernetes version v1.23.13+vmware.1...
Updating the KCP object with k8s version v1.23.13+vmware.1
Waiting for kubernetes version to be updated for control plane nodes
Upgrading worker nodes...
Patching MachineDeployment with the kubernetes version v1.23.13+vmware.1...
Waiting for kubernetes version to be updated for worker nodes...
Waiting for packages to be up and running...
Cluster 'my-legacy-wc' successfully upgraded to kubernetes version 'v1.23.13+vmware.1'
```

### Release note

```release-note

```

### DIFF
--- a/cmd/cli/plugin/cluster/upgrade.go
+++ b/cmd/cli/plugin/cluster/upgrade.go
@@ -115,18 +115,15 @@ func upgradeCluster(server *configapi.Server, clusterName string) error {
 		return err
 	}
 
-	tkrVersion := ""
-	if uc.tkrName != "" {
-		clusterClientOptions := clusterclient.Options{GetClientInterval: 2 * time.Second, GetClientTimeout: 5 * time.Second}
-		clusterClient, err := clusterclient.NewClient(server.ManagementClusterOpts.Path, server.ManagementClusterOpts.Context, clusterClientOptions)
-		if err != nil {
-			return err
-		}
+	clusterClientOptions := clusterclient.Options{GetClientInterval: 2 * time.Second, GetClientTimeout: 5 * time.Second}
+	clusterClient, err := clusterclient.NewClient(server.ManagementClusterOpts.Path, server.ManagementClusterOpts.Context, clusterClientOptions)
+	if err != nil {
+		return err
+	}
 
-		tkrVersion, err = getValidTkrVersionFromTkrForUpgrade(tkgctlClient, clusterClient, clusterName)
-		if err != nil {
-			return err
-		}
+	tkrVersion, err := getValidTkrVersionFromTkrForUpgrade(tkgctlClient, clusterClient, clusterName)
+	if err != nil {
+		return err
 	}
 
 	edition, err := config.GetEdition()
@@ -352,6 +349,15 @@ func getClusterTKRNameFromClusterLabels(clusterLabels map[string]string) (string
 }
 
 func getValidTKRVersionFromClusterForUpgrade(cluster *capiv1.Cluster, tkrName string) (string, error) {
+	// if tkrname is not specified, return the latest available version
+	if tkrName == "" {
+		updates := clusters.AvailableUpgrades(cluster)
+		if len(updates) == 0 {
+			return "", errors.Errorf("no available upgrades for cluster '%s', namespace '%s'", cluster.Name, cluster.Namespace)
+		}
+		return latestTkrVersion(updates.Slice()), nil
+	}
+
 	tkrVersion := strings.ReplaceAll(tkrName, "---", "+")
 	// If the existing TKR on the cluster and the one requested are same
 	// continue the upgrade using the given TKR
--- a/cmd/cli/plugin/cluster/upgrade_test.go
+++ b/cmd/cli/plugin/cluster/upgrade_test.go
@@ -393,6 +393,19 @@ var _ = Describe("getValidTKRVersionFromClusterForUpgrade", func() {
 			Expect(err.Error()).To(ContainSubstring(fmt.Sprintf("no available upgrades for cluster '%s', namespace '%s'", cluster.Name, cluster.Namespace)))
 		})
 	})
+
+	Context("when user provided TKR prefix name is empty", func() {
+
+		BeforeEach(func() {
+			availableUpdates := fmt.Sprintf("[%s %s %s]", "v1.18.18+vmware.1-tkg.2", "v1.18.8+vmware.1-tkg.1", "v1.18.14+vmware.1-tkg.1-rc.1")
+			cluster = getFakeCluster("fake-cluster", availableUpdates)
+			tkrNamePrefix = ""
+		})
+		It("should return the highest available version", func() {
+			Expect(err).ToNot(HaveOccurred())
+			Expect(latestTKRVersion).To(Equal("v1.18.18+vmware.1-tkg.2"))
+		})
+	})
 })
 
 var _ = Describe("getClusterResource", func() {


tkg/client/upgrade_cluster.go
```diff
@@ -181,18 +181,35 @@ func (c *TkgClient) UpgradeCluster(options *UpgradeClusterOptions) error {
...
 	if err != nil {
 		return errors.Wrapf(err, "unable to patch the cluster object with TanzuKubernetesRelease label")
 	}
 
+	// restore tkr version if upgrade failed after tkr label updated
+	defer func() {
+		if err != nil {
+			log.Infof("Restoring kubernetes cluster to tkr `%v` version", currentTkrVersion)
+			options.TkrVersion = currentTkrVersion
+			if restoreErr := c.addKubernetesReleaseLabel(regionalClusterClient, options); restoreErr != nil {
```
12345lcr, line 207: Hi @hongtaiwang , what if the cluster node is successfully upgraded and all the core addons are upgraded, but one of the addon's reconciliation just got a little timeout? I think switching the tkr version back to origin no matter what the error is will result in some unexpected behavior, like the successfully reconciled core addons will be downgraded , which is either not allowed by kapp or bring compatibility issues.

hongtaiwang: does kapp refer to  `tanzuKubernetesRelease` label for legacy cluster or `topology.version` for cc cluster? 

The revert is only for those fields when upgrade failed. If upgrade failed due to some reason, it's expected that the cluster is broken, and users should be able to re-run the upgrade command to get through it.

tkg/client/upgrade_cluster.go
```diff
@@ -181,18 +181,35 @@ func (c *TkgClient) UpgradeCluster(options *UpgradeClusterOptions) error {
...
 	err = c.addKubernetesReleaseLabel(regionalClusterClient, options)
 	if err != nil {
 		return errors.Wrapf(err, "unable to patch the cluster object with TanzuKubernetesRelease label")
 	}
 
+	// restore tkr version if upgrade failed after tkr label updated
+	defer func() {
+		if err != nil {
+			log.Infof("Restoring kubernetes cluster to tkr `%v` version", currentTkrVersion)
+			options.TkrVersion = currentTkrVersion
```
jeffwubj, line 206: besides @12345lcr 's comments, curious what will be the suggested actions for users after setting tkr version back to the original version? upgrade again directly or do some "repair" then upgrade? if user need to do some manual "repair", seems set tkr version back could be part of that as well

hongtaiwang: > upgrade again directly or do some "repair" then upgrade?
either way should be good, depends on what kind of error they meet. But if the version is not set back, customer will not be able to run upgrade again since the current version is overwritten and tanzu cli does not support upgrade to "old" versions.

tkg/client/upgrade_cluster.go
12345lcr: `tanzuKubernetesRelease` took effect on tanzu cli, addons-manager, pinniped... I am not sure what is the full consequence for these components to set it back and forth.

And also `tanzuKubernetesRelease` is a legacy tkr label. In ClusterClass clusters the tkr label seems to be `run.tanzu.vmware.com/tkr`.

hongtaiwang: tanzu cli upgrade cluster read `tanzuKubernetesRelease` in legacy clusters , while `topology.version`  in cc clusters.

12345lcr: From the addon-manager's perspective, this `tanzuKubernetesRelease` decide which core package repository should be deployed to to legacy wl cluster. Once it is updated the core addon's upgrade is started.

12345lcr: So what I mean is this `tanzuKubernetesRelease` is a very high level information that multiple components are depends on it. If the cluster upgrade is failed and we want to retry, it is better to have something like a `force upgrade` to adjust the cluster into the desired state, instead of `fake` the cluster that the upgrade did not happened before.

12345lcr: And maybe the better solution is to block the upgrade from beginning if it tries to skip a minor version

hongtaiwang: But I still think upgrade tag should be the last step for upgrade cluster since we should not set the cluster to new version if the upgrade process is not finished completely. 

And of course the impact for addon-manager should be taken into consideration. Do you know if addon-manager could work after the following update?
* set tkr v1 -> v2 during upgrade
* upgrade fail after addon-manager updated, set back v2 -> v1 (maybe addon-manager will be broken now)
* re-run upgrade command, v1 -> v2 again ===> could addon-manager work correctly in this case?

12345lcr: > upgrade tag should be the last step for upgrade cluster 

This was the previous behavior that brought problems. The core addon's upgrade path is `upgrade tkr tag in cluster-> addons manager upgrade core addon's repo -> kapp-controller upgrade core addon`

In this path if the tkr is updated at the final step, then this means the core addons got updated after node upgrade, but if the old addons were not compatible with new version kubernetes nodes, the upgrade will hang on
* Node upgrade stuck because nodes not ready, due to old core addons not compatible
* Addons cannot be upgraded because nodes upgrade was not finished successfully

tkg/client/upgrade_cluster.go
```diff
@@ -181,18 +181,35 @@
 	return c.DoLegacyClusterUpgrade(regionalClusterClient, currentClusterClient, options)
 }
 
-func (c *TkgClient) DoLegacyClusterUpgrade(regionalClusterClient, currentClusterClient clusterclient.Client, options *UpgradeClusterOptions) error {
+func (c *TkgClient) DoLegacyClusterUpgrade(regionalClusterClient, currentClusterClient clusterclient.Client, options *UpgradeClusterOptions) (err error) {
```
imikushin, line 184: **suggestion:** rename `err` to `retErr` to avoid accidental assignment to the return variable in the method body. It's best to _only_ access `retErr` in the _deferred_ func.

In the method body, write return statements as usual (_do not_ assign to `retErr`).

tkg/client/upgrade_cluster.go
```diff
@@ -181,18 +181,35 @@
...
 	log.Info("Verifying kubernetes version...")
-	err := c.verifyK8sVersion(currentClusterClient, options.KubernetesVersion)
+	err = c.verifyK8sVersion(currentClusterClient, options.KubernetesVersion)
 	if err != nil {
 		return errors.Wrap(err, "kubernetes version verification failed")
 	}
 
+	// backup tkr version for current cluster
+	currentClusterInfo := regionalClusterClient.GetClusterStatusInfo(options.ClusterName, options.Namespace, currentClusterClient)
+	currentTkrVersion := currentClusterInfo.ClusterObject.Labels["tanzuKubernetesRelease"]
```
imikushin, line 193: **nit:** Since `currentClusterInfo` only exists to help obtain `currentTkrVersion`, perhaps it would be helpful to introduce a function that would return `currentTkrVersion`. 

Also, I believe we have a constant for `"tanzuKubernetesRelease"`. If it's not convenient to import it here, it's still better to use a constant rather than a string value.

cmd/cli/plugin/cluster/upgrade.go
```diff
@@ -364,6 +361,11 @@
 		return "", errors.Errorf("no available upgrades for cluster '%s', namespace '%s'", cluster.Name, cluster.Namespace)
 	}
 
+	// if tkrname is not specified, return the latest available version
+	if tkrName == "" {
+		return latestTkrVersion(updates.Slice()), nil
+	}
```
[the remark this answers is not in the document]
hongtaiwang: addressed

tkg/client/upgrade_cluster.go
hongtaiwang: done

tkg/client/upgrade_cluster.go
hongtaiwang: done

hongtaiwang:
As discussed, we are not going to restore k8s/tkr version when upgrade failed.

jeffwubj (approved):

cmd/cli/plugin/cluster/upgrade.go
```diff
@@ -115,18 +115,15 @@ func upgradeCluster(server *configapi.Server, clusterName string) error {
 		return err
 	}
 
-	tkrVersion := ""
-	if uc.tkrName != "" {
```
wjun, line 119: Why do we remove this condition here?

cmd/cli/plugin/cluster/upgrade.go
```diff
@@ -352,6 +349,15 @@ func getClusterTKRNameFromClusterLabels(clusterLabels map[string]string) (string
 }
 
 func getValidTKRVersionFromClusterForUpgrade(cluster *capiv1.Cluster, tkrName string) (string, error) {
+	// if tkrname is not specified, return the latest available version
```
wjun, line 352: From the issue description, we should avoid upgrade from 1.22 to 1.24. How is this latest available version satisfy this condition?

hongtaiwang: this label is tagged by tkr-status-controller https://github.com/vmware-tanzu/tanzu-framework/blob/6c7f5cc18a6caeed6724e88ebbe90ccd05e35e7f/tkr/controller/tkr-status/clusterstatus/clusterstatus.go#L183 . It is selected from tkrs within minor+1 version.

cmd/cli/plugin/cluster/upgrade.go
hongtaiwang: because `getValidTkrVersionFromTkrForUpgrade` should return highest available tkr if `tkrName` is not specified. And it should be called if `tkrName` is empty.

wjun (approved):
LGTM

PR #4035, opened 2022-11-29
[DoNotMerge]Add akoo credentials to the meta package input

12345lcr:
Add akoo credentials to the meta package input during upgrade. Since we are reusing the `09_miscellaneous/08_store_data_values.yaml` secret for generating tkg-pkg package data values but this secret explicitly filtered out some credentials like `AVI_PASSWORD` , which is required when installing AKOO composite package.

Fixes #

Wrote ut and passed locally

### Release note

```release-note

```


12345lcr:
Closing this since we are using a different approach to get akoo information from addon secret in https://github.com/vmware-tanzu/tanzu-framework/pull/4048

PR #4039, opened 2022-11-29
Fix featuregates package installation in Clusterclass based AWS clusters

yharish991:

Currently, the hostNetwork is set to true and that is causing the featuregates package installation in cc based AWS clusters to fail as the ingress rule for featuregates is not created in AWS. This PR sets the hostNetwork to false for featuregates pod as there is no need for featuregates pod to use host network. 

This PR also reverts the workaround added by @12345lcr with this pr https://github.com/vmware-tanzu/tanzu-framework/pull/3983

Fixes #

Monitoring ClusterClass CAPA integration tests CI check

### Release note

```release-note

```

### DIFF
--- a/packages/featuregates/bundle/config/values.yaml
+++ b/packages/featuregates/bundle/config/values.yaml
@@ -2,9 +2,8 @@
 ---
 namespace: tkg-system
 deployment:
-  hostNetwork: true
-  nodeSelector:
-    node-role.kubernetes.io/control-plane: ""
+  hostNetwork: false
+  nodeSelector: null
   tolerations: []
   webhookServerPort: 9443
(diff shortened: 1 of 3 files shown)


packages/featuregates/bundle/config/values.yaml
```diff
@@ -2,9 +2,8 @@
 ---
 namespace: tkg-system
 deployment:
-  hostNetwork: true
-  nodeSelector:
-    node-role.kubernetes.io/control-plane: ""
```
rajathagasthya, line 7: Description only mentions `hostNetwork` change. Why are we removing nodeSelector too?

yharish991: that was the initial workaround fix that @12345lcr added with this pr https://github.com/vmware-tanzu/tanzu-framework/pull/3983, so reverting that

rajathagasthya (approved):

12345lcr (approved):
LGTM

PR #3862, opened 2022-11-08
provider: Introduce Pod Security Standard for non-CC and CC based Clusters

chrischdi:
Note: I will be on PTO from 20th November until 4th of December). @ramineni will takeover during that time!

### What this PR does / why we need it

Introduces:

* Namespace labels for CAPZ and CAPD provider to specify privileged pod security admission
* Overlay for ClusterClasses to enable and configure Pod Security Standard for v1.24 clusters
* Overlay for non-ClusterClass clusters to enable and configure Pod Security Standard for v1.24 clusters
* Migration code for non-ClusterClass clusters to enable Pod Security Standard when upgrading to v1.24
* Fixes the audit-log patches for ClusterClass to work via the `kcp-empty-array-overlay.yaml`

A follow-up PR will raises the enablement to v1.24, as soon as v1.24 is available here.

### Which issue(s) this PR fixes

Replaces #3693

Fixes #

### Describe testing done for PR

- Validated in manual CAPA installation (ClusterClass + non ClusterClass)
- vcenter pipelines succeeded
- Tested overlays on local capd (with and without clusterclass)
- Tested overlays using TKG + small changes which enable PSA already for v1.23

### Release note

```release-note
Enable pod security admission per default for Clusters in audit and warn mode. Also introduces variables to the ClusterClasses to make it configurable for ClusterClass based clusters.
```

PR #4050, opened 2022-11-30
WIP:provider: [PSA] Remove the overlay to initialize apiserver/extraVolumes empty array for CAPZ

ramineni:
The PR https://github.com/vmware-tanzu/tanzu-framework/pull/3862 , adds overlay to initialize the apiServer/extraVolumes array for all providers. This is replacing the extraVolumes defined in base.yaml ( https://github.com/vmware-tanzu/tanzu-framework/blob/main/packages/tkg-clusterclass-azure/bundle/config/upstream/base.yaml ) causing management cluster create to fail. This PR is to fix the same, remove initializing extraVolumes array for infrastructure-azure.

Fixes #

### Release note

```release-note

```

### DIFF
--- a/packages/tkg-clusterclass-aws/bundle/config/upstream/overlay.yaml
+++ b/packages/tkg-clusterclass-aws/bundle/config/upstream/overlay.yaml
@@ -35,4 +35,17 @@ spec:
           status: "False"
           timeout: #@ data.values.MHC_FALSE_STATUS_TIMEOUT
         #@ end
-
+  patches:
+    #@overlay/match by=overlay.index(0)
+    #@overlay/insert before=True
+    - definitions:
+      - jsonPatches:
+          - op: add
+            path: /spec/template/spec/kubeadmConfigSpec/clusterConfiguration/apiServer/extraVolumes
+            value: []
+        selector:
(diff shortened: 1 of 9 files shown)


wjun:
@ramineni Please double check if the jsonpatch in the extra secret will make all other jsonpatches in base.yaml, for example,  https://github.com/vmware-tanzu/tanzu-framework/blob/main/packages/tkg-clusterclass-azure/bundle/config/upstream/base.yaml invalid anymore?

ramineni:
> @ramineni Please double check if the jsonpatch in the extra secret will make all other jsonpatches in base.yaml, for example, https://github.com/vmware-tanzu/tanzu-framework/blob/main/packages/tkg-clusterclass-azure/bundle/config/upstream/base.yaml invalid anymore? You could also have a look at the live environment in tkg-15461.

@wjun Thanks for the review. As checked again in the live env , it is appending to the existing array and not changing the existing patches. I have updated to add the empty array only for capd provider as other providers do have initialized in the base template.

ramineni:
@tenczar PTAL

wjun (approved):
LGTM

PR #4054, opened 2022-11-30
Fix the deletion of discovery sources by using DeleteCLIDiscoverySource API in tanzu plugin source delete command

### LINKED ISSUE #4055: Unable to delete a plugin source
**Bug description**

I am no longer able to delete a plugin source.

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

A plugin source should be properly deleted

```
tanzu plugin source list
  NAME     TYPE   SCOPE
  default  local  Standalone
tanzu plugin source add -n stage.25 -t oci -u projects-stg.registry.vmware.com/tkg/packages/standalone-plugins:v0.25.3-tf_vmware.1
✔  successfully added discovery source stage.25
tanzu plugin source list
  NAME      TYPE   SCOPE
  default   local  Standalone
  stage.25  oci    Standalone
tanzu plugin source delete stage.25
✔  deleted discovery source stage.25
tanzu plugin source list
  NAME      TYPE   SCOPE
  default   local  Standalone
  stage.25  oci    Standalone
```
Notice that the plugin source `stage.25` was not actually removed although right above it said it was.

Today's main: e0064860dab49ec9dc406d700078f2cea64fea78

Locally

mpanchajanya:

Fix the deletion of discovery sources.
Uses the Config API method `DeleteCLIDiscoverySource` instead of deprecated `StoreClientConfig`

Fixes #4055 

Unit Testing and Validate the `tanzu plugin source delete` command

```
mpanchajanya@mpanchajany-a01 tanzu-framework % tanzu config get                  
clientOptions:
    features:
        global:
            context-target: "false"
            tkr-version-v1alpha3-beta: "false"
        management-cluster:
            export-from-confirm: "true"
            import: "false"
            custom-nameservers: "false"
        package:
            kctrl-package-command-tree: "true"
        cluster:
            custom-nameservers: "false"
    cli:
        discoverySources:
            - local:
                name: admin-local
                path: admin
            - local:
                name: default
                path: standalone
              contextType: k8s
        edition: tkg
        bomRepo: projects.registry.vmware.com/tkg
```

```
mpanchajanya@mpanchajany-a01 tanzu-framework % tanzu plugin source list
  NAME         TYPE   SCOPE       
  admin-local  local  Standalone  
  default      local  Standalone  
```

```
mpanchajanya@mpanchajany-a01 tanzu-framework % tanzu plugin source delete admin-local
✔  deleted discovery source admin-local
```

```
mpanchajanya@mpanchajany-a01 tanzu-framework % tanzu config get                      
clientOptions:
    features:
        global:
            context-target: "false"
            tkr-version-v1alpha3-beta: "false"
        management-cluster:
            export-from-confirm: "true"
            import: "false"
            custom-nameservers: "false"
        package:
            kctrl-package-command-tree: "true"
        cluster:
            custom-nameservers: "false"
    cli:
        discoverySources:
            - local:
                name: default
                path: standalone
              contextType: k8s
        edition: tkg
        bomRepo: projects.registry.vmware.com/tkg
```

### DIFF
--- a/cli/core/pkg/command/discovery_source.go
+++ b/cli/core/pkg/command/discovery_source.go
@@ -200,25 +200,7 @@ var deleteDiscoverySourceCmd = &cobra.Command{
 	RunE: func(cmd *cobra.Command, args []string) (err error) {
 		discoveryName := args[0]
 
-		// Acquire tanzu config lock
-		configlib.AcquireTanzuConfigLock()
-		defer configlib.ReleaseTanzuConfigLock()
-
-		cfg, err := configlib.GetClientConfigNoLock()
-		if err != nil {
-			return err
-		}
-		if cfg.ClientOptions == nil || cfg.ClientOptions.CLI == nil {
-			return fmt.Errorf("discovery %q unknown", discoveryName)
-		}
-
-		newDiscoverySources, err := deleteDiscoverySource(cfg.ClientOptions.CLI.DiscoverySources, discoveryName)
-		if err != nil {
-			return err
-		}
-
-		cfg.ClientOptions.CLI.DiscoverySources = newDiscoverySources
-		err = configlib.StoreClientConfig(cfg)
+		err = configlib.DeleteCLIDiscoverySource(discoveryName)
 		if err != nil {
 			return err
 		}
--- a/cli/runtime/config/cli_discovery_sources_test.go
+++ b/cli/runtime/config/cli_discovery_sources_test.go
@@ -254,15 +254,14 @@ func TestDeleteCLIDiscoverySource(t *testing.T) {
 	}(f.Name())
 
 	tests := []struct {
-		name    string
-		src     *configapi.ClientConfig
-		input   string
-		deleted bool
-		count   int
-		errStr  string
+		name   string
+		src    *configapi.ClientConfig
+		input  string
+		count  int
+		errStr string
 	}{
 		{
-			name: "should return true on deleting non existing item",
+			name: "should return err on deleting non existing source",
 			src: &configapi.ClientConfig{
 				ClientOptions: &configapi.ClientOptions{
 					CLI: &configapi.CLIOptions{
@@ -278,13 +277,12 @@ func TestDeleteCLIDiscoverySource(t *testing.T) {
 					},
 				},
 			},
-			input:   "test-notfound",
-			deleted: true,
-			count:   1,
-			errStr:  "cli discovery source not found",
+			input:  "test-notfound",
+			count:  1,
+			errStr: "cli discovery source not found",
 		},
 		{
-			name: "should return true on deleting existing item",
+			name: "should delete existing test source",
 			src: &configapi.ClientConfig{
 				ClientOptions: &configapi.ClientOptions{
 					CLI: &configapi.CLIOptions{
@@ -300,12 +298,11 @@ func TestDeleteCLIDiscoverySource(t *testing.T) {
 					},
 				},
 			},
-			input:   "test",
-			count:   0,
-			deleted: true,
+			input: "test",
+			count: 0,
 		},
 		{
-			name: "should return true on deleting existing test2",
+			name: "should delete test2 source",
 			src: &configapi.ClientConfig{
 				ClientOptions: &configapi.ClientOptions{
 					CLI: &configapi.CLIOptions{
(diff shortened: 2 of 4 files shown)


cli/runtime/config/cli_discovery_sources_test.go
```diff
@@ -268,6 +268,7 @@ func TestDeleteCLIDiscoverySource(t *testing.T) {
 		src     *configapi.ClientConfig
 		input   string
 		deleted bool
```
vuil, line 270: nit: I don't see .deleted getting used anywhere

cli/runtime/config/cli_discovery_sources_test.go
```diff
@@ -339,9 +342,46 @@ func TestDeleteCLIDiscoverySource(t *testing.T) {
 					},
 				},
 			},
+			count:   1,
 			input:   "test",
```
vuil, line 346: test2?

cli/runtime/config/cli_discovery_sources_test.go
```diff
@@ -339,9 +342,46 @@ func TestDeleteCLIDiscoverySource(t *testing.T) {
...
+							{
+								Local: &configapi.LocalDiscovery{
+									Name: "default",
+									Path: "standalone",
+								},
+								ContextType: configapi.CtxTypeK8s,
+							},
+							{
+								Local: &configapi.LocalDiscovery{
+									Name: "admin-local",
```
vuil, line 372: Looks like we support sources of same name of different discovery type? If so, would be good to include a test to verify the expected behavior when delete is issued on the name.

cli/core/pkg/command/discovery_source.go
```diff
@@ -193,25 +193,7 @@ var deleteDiscoverySourceCmd = &cobra.Command{
...
-		}
-
-		newDiscoverySources, err := deleteDiscoverySource(cfg.ClientOptions.CLI.DiscoverySources, discoveryName)
-		if err != nil {
-			return err
-		}
-
-		cfg.ClientOptions.CLI.DiscoverySources = newDiscoverySources
-		err = configlib.StoreClientConfig(cfg)
+		err = configlib.DeleteCLIDiscoverySource(discoveryName)
```
vuil, line 196: The intention is to support deleting the first matching source by name? If so it may be a good idea to add a comment to deleteCLIDiscoverySource near where getCLIDiscoverySource is called.

cli/runtime/config/cli_discovery_sources_test.go
anujc25: > Looks like we support sources of same name of different discovery type? 

I don't think we support this AFAIK.

cli/runtime/config/cli_discovery_sources_test.go
mpanchajanya: not needed I have used errStr for the same. removed `deleted`

cli/runtime/config/cli_discovery_sources_test.go
mpanchajanya: test case is valid description was misspelled. Corrected

cli/core/pkg/command/discovery_source.go
mpanchajanya: since discovery source are unique by name itself. There should only be one entry matching with the name.

cli/runtime/config/cli_discovery_sources_test.go
mpanchajanya: discovery source are unique by name itself; Cannot have same name for two different types. There should only be one entry matching with the name.

cli/core/pkg/command/discovery_source.go
vuil: So it is not intentional, but the code does not error out on config with multiple sources of the same name. Which is probably a not a bad thing, since it means the implementation can still be used to correct the discovery sources should the file somehow reach that state. 
I think it warrants at least a comment say at getCLIDiscoverySource that we are making an assumption that there should only be one source with the name, hence we are returning the first match.

mpanchajanya: added the description on GetCLIDiscoverySource()

anujc25 (approved):
Looks good.

PR #4060, opened 2022-12-01
Skip the same cluster name validation if dry-run

zhouke1991:
It is unnecessary to validate if the cluster name is duplicated with the existing cluster when dry run.

### DIFF
--- a/tkg/client/client.go
+++ b/tkg/client/client.go
@@ -83,6 +83,7 @@ type InitRegionOptions struct {
 	CeipOptIn                    bool
 	UseExistingCluster           bool
 	IsInputFileClusterClassBased bool
+	GenerateOnly                 bool
 }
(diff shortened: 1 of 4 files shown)


tkg/client/client.go
```diff
@@ -83,6 +83,7 @@ type InitRegionOptions struct {
 	CeipOptIn                    bool
 	UseExistingCluster           bool
 	IsInputFileClusterClassBased bool
+	GenerateOnly                 bool
```
DanielXiao, line 86: It seems to me DryRun is more straightforward.

zhouke1991: As other options structs have already used this name , I think we'd better to be consistent with other structs.

DanielXiao (approved):

PR #4071, opened 2022-12-01
Discard contexts and currentContext nodes from config.yaml

### LINKED ISSUE #4072: Deleted context can re appear in the config
**Bug description**

Deleted contexts can re appear in the config.

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- (x) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**
Delete context should not re appear in the config

1. Add a context, it appears in config-ng.yaml
2. Run an old plugin: tanzu login version (-> 0.25.3); The context field appears in config.yaml
3. Delete the context, it is removed from config-ng but stays in config.yaml. It is removed from the server list in config.yaml: ok
4. ```tanzu context list``` does not show the deleted context: ok
5. Run an old plugin again: ```tanzu login version ```(-> 0.25.3)
The old plugin syncs the server list with the context list but since the context list has the deleted context, it get re-added as a server
6. Run any tanzu CLI command
- The CLI syncs the server list of the old config.yaml with the context list of the new config-ng
- The deleted context is re-added to the list of context in config-ng
7. ```tanzu context list``` - now shows the deleted context again

Today's main

Local

mpanchajanya:

Discard or Delete the contexts from config.yaml since it could corrupt the config-ng.yaml because of interlink sync between servers and contexts.

### Solution

- Completely discard/ remove contexts from config.yaml

Fixes #4072 

Unit testing and Integration Tests

### DIFF
--- a/cli/runtime/config/config_factory.go
+++ b/cli/runtime/config/config_factory.go
@@ -26,6 +26,11 @@ var LegacyConfigNodeKeys = []string{
 	KeyCurrentServer,
 }
 
+var DiscardedConfigNodeKeys = []string{
+	KeyContexts,
+	KeyCurrentContext,
+}
+
 // getMultiConfig retrieves combined config.yaml and config-ng.yaml
 func getMultiConfig() (*yaml.Node, error) {
 	cfgNode, err := getClientConfig()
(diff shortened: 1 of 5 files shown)


marckhouzam (approved):
Thanks for the quick fix @mpanchajanya!
I tested the usecase before and after this PR, and things get fixed as expected 🎉 

I just want to point out that the PR removes the `contexts` section but not the `currentContext` one.  I don't see a problem with it though but I was curious if it was a conscious decision.

LGTM

cli/runtime/config/config_factory.go
```diff
@@ -26,6 +26,11 @@ var LegacyConfigNodeKeys = []string{
 	KeyCurrentServer,
 }
 
+var DiscardedConfigNodeKeys = []string{
```
vuil, line 29: nit: it's good to have some trace back to reason for the change. 
A comment explaining why might be good. 

Also: please add
```
Fixes #4072 
```
into the commit you plan to retain, that way there is linkage in the git commit history as well.

Fixes: #

vuil (approved):
I assume you plan to squash the commits? 

Just a nit about adding a comment and fixing the commit message, but otherwise lgtm!

PR #4090, opened 2022-12-02
Fix `current` field getting set to tmc context in client config

### LINKED ISSUE #4091: CurrentServer (`current`) field in ClientConfig should not point to TMC context
**Bug description**

current field in the config.yaml is meant to be used to just include managment-cluster/k8s context and it should not include tmc context

For configuring tmc context CLI should use currentContext.mission-control field.

Also, TMC plugins should be using currentContext.mission-control to determine the active tmc context.

This is required for backward compatibility of kubernetes targeted plugins with new context/target functionality

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

CurrentServer (`current`) field in ClientConfig should not point to TMC context.

anujc25:
- Do not set `current` field to point to TMC context in client config

**Details:**
- `current` field in the config.yaml is meant to be used to just include managment-cluster/k8s context and it should not include tmc context
- For configuring tmc context CLI should use `currentContext.mission-control` field.
- Also, TMC plugins should be using `currentContext.mission-control` to determine the active tmc context.

- This is required for backward compatibility of kubernetes targeted plugins with new context/target functionality

Fixes #4091

Tested below commands:
- `tanzu login`
- `tanzu context create` 
- `tanzu context use` 

Above commands should not set the `current` to point to TMC context when creating or setting TMC context and just set it in the `currentContext.mission-control` field.

### Release note

```release-note
Fix `current` field getting set to tmc context in client config (config.yaml)
```

### DIFF
--- a/cli/runtime/config/servers.go
+++ b/cli/runtime/config/servers.go
@@ -6,6 +6,8 @@ package config
 import (
 	"fmt"
 
+	"github.com/pkg/errors"
+
 	configapi "github.com/vmware-tanzu/tanzu-framework/cli/runtime/apis/config/v1alpha1"
 	"github.com/vmware-tanzu/tanzu-framework/cli/runtime/config/nodeutils"
 
@@ -135,7 +137,7 @@ func SetServer(s *configapi.Server, setCurrent bool) error {
 			return err
 		}
 	}
-	if setCurrent {
+	if setCurrent && s.Type == configapi.ManagementClusterServerType {
 		persist, err = setCurrentServer(node, s.Name)
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


cli/runtime/config/servers.go
```diff
@@ -226,6 +228,15 @@ func RemoveServer(name string) error {
 }
 
 func setCurrentServer(node *yaml.Node, name string) (persist bool, err error) {
+	s, err := getServer(node, name)
+	if err != nil {
+		return false, err
+	}
+
+	if s.Type != configapi.ManagementClusterServerType {
+		return false, errors.Errorf("cannot set non management-cluster server as current server")
```
marckhouzam, line 237: Nice!

marckhouzam (approved):
Tested and it works beautifully.  LGTM

One question however.  What if we use `tanzu login <tmc cluster>` with an old version of the plugin?  Is this possible?  Because then it will still set the `current` field to the tmc cluster...

anujc25:
>One question however. What if we use tanzu login <tmc cluster> with an old version of the plugin? Is this possible? Because then it will still set the current field to the tmc cluster...

Yes. Ideally, TMC users should always be using the new CLI to use this feature as we are adding support in upcoming release but If for some reason, user uses old login plugin to login to tmc than it will set the `current` field.

vuil (approved):
lgtm, thanks!

PR #4092, opened 2022-12-03
Add trivy scanning

randomvariable:

Matches the default scanner of Harbor, which is how Tanzu Framework is shipped. Results will apepar in the security tab and can be filtered by PR.

Fixes #

### Release note

```release-note
NONE
```

### DIFF
--- a/.github/workflows/trivy.yaml
+++ b/.github/workflows/trivy.yaml
@@ -1,0 +1,26 @@
+name: Security scan
+on:
+  push:
+    branches:
+    - main
+  pull_request:
+jobs:
+  build:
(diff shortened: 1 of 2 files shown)


randomvariable:
> What does trivy scan in our case? The source files and go dependencies? I've added a bunch of exclusions for things which aren't relevant.

Correct.

We basically do need to fix these anyway as TKG releases are based upon this, and they'll be flagged at release time for fixing. This allows us to be more proactive.

marckhouzam (approved):
Seems like a good idea.
LGTM

randomvariable:
Hmm. need to revisit

PR #4096, opened 2022-12-05
Fix kapp-controller upgrade missing CORE_DNS_IP and tkg-clusterclass-aws upgrade missing AWS_AMI_ID

12345lcr:
1. kapp-controller
During mgmt cluster upgrade, the func `InstallOrUpgradeKappController` is reused to upgrade the kapp-controller. But this func seems designed for cluster creation, which means all the default and user’s configurations have already generated and stored in TKGConfigReaderWriter . But in upgrade case the configurations are not stored locally but should be retrieved from the existing cluster before cluster upgrade in [configureVariablesForProvidersInstallation](https://github.com/vmware-tanzu/tanzu-framework/blob/99a8d0fafe246e6c395994d093774da5926c3fe7/tkg/client/upgrade_region.go#L235-L283) first to store the required fields in local TKGConfigReaderWriter .
Currently some of the kapp-controller input values like CORE_DNS_IP are not retrieved during mgmt cluster upgrade, which leads to the kapp-controller mis-configured. All kapp-controller required fields:
https://github.com/vmware-tanzu/tanzu-framework/blob/8df9dd7953c38988793651434f40530507407944/providers/kapp-controller-values/values.yaml

2. tkg-clusterclass-aws
The `AWS_AMI_ID` configuration for aws-clusterclass-package is missing during cluster upgrade.

Fixes #

1. Add ut and passed locally

2. Created a tkg 1.6.0 env and manually build tanzu cli to upgrade.
- The mgmt and wl cluster upgrade are passed. 
- CORE_DNS_IP and AWS_AMI_ID are exist in the tkg-pkg-tkg-system-values secret. 
- The kapp-controller received the CORE_DNS_IP during upgrade.

### Release note

```release-note

```

### DIFF
--- a/tkg/client/init.go
+++ b/tkg/client/init.go
@@ -215,7 +215,7 @@ func (c *TkgClient) InitRegion(options *InitRegionOptions) error { //nolint:funl
 	// If clusterclass feature flag is enabled then deploy kapp-controller
 	if config.IsFeatureActivated(constants.FeatureFlagPackageBasedCC) {
 		log.Info("Installing kapp-controller on bootstrap cluster...")
-		if err = c.InstallOrUpgradeKappController(bootStrapClusterClient, constants.OperationTypeInstall); err != nil {
+		if err = c.InstallOrUpgradeKappController(bootStrapClusterClient, constants.OperationTypeInstall, false); err != nil {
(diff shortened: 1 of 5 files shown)


wjun (approved):
Thanks for the fixes!

PR #4097, opened 2022-12-05
Bring Feature plugin, client unit tests to >70% coverage, upload to Codecov

codegold79:
Bring Feature plugin and FeatureGate client unit tests to >70% coverage. Upload coverage files to Codecov.

Feature plugin and feature client unit tests have less than 70% coverage. To help ensure quality of code, the change in this PR brings the coverage up to above 70% as shown in the Codecov dashboard.

Fixes #

Checked CodeCov dashboard and ensured coverage was greater than 70% for Feature plugin and FeatureGate client.

Screen Shot 2022-12-08 at 4 54 46 PM

Screen Shot 2022-12-08 at 4 55 14 PM

### Release note

```release-note

```


cmd/cli/plugin/feature/activate.go
```diff
@@ -113,15 +124,19 @@ func userGivesPermissionToVoidWarranty(feature *corev1alpha2.Feature, userAllows
...
 		fmt.Printf("Warning: activating %q, a %s Feature will irrevocably void all support guarantees for this environment. You will need to recreate the environment to return to a supported state.\nWould you like to continue [y/N]?", feature.Name, feature.Spec.Stability)
-		return userAllowsByInteractiveCLI()
+		return userAllowsByInteractiveCLI(nil)
 	}
 	// If the user did pass a value to the flag, use the user's input. No interactive user input is needed.
 	return *userAllowsByFlag, nil
 }
 
-func userAllowsByInteractiveCLI() (bool, error) {
+func userAllowsByInteractiveCLI(in *os.File) (bool, error) {
```
rajathagasthya, line 133: Very nice adjustment for testing!

Suggestion: could we use `io.Reader` here instead of `*os.File`? We can then just use `bytes.Buffer` to test it and don't need to create files.

cmd/cli/plugin/feature/test/k8s_config.kube
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
rajathagasthya, line 1: Please rename this directory `testdata` so the go tool ignores this.

featuregates/client/pkg/featuregateclient/client_test.go
```diff
@@ -71,6 +71,58 @@ func TestGetFeature(t *testing.T) {
...
 	}
 }
+func TestGetFeatureList(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), contextTimeout)
+	defer cancel()
+
+	objs, _, _ := fake.GetTestObjects()
+	s := scheme.Scheme
+	if err := corev1alpha2.AddToScheme(s); err != nil {
+		t.Fatalf("Unable to add config scheme: (%v)", err)
```
rajathagasthya, line 81: nit: start error messages with lower case (there is inconsistency between files here).

rajathagasthya (commented):
Minor comments. Overall looks great!

cmd/cli/plugin/feature/activate.go
codegold79: I chose `*os.File` because the way the Feature plugin works is to get user input via `os.Stdin`. The `os.Stdin` type is `*os.File`. Shouldn't tests most closely act like the real thing? If I use `bytes.Buffer`, it's less like `os.Stdin`. 

To answer your question, "could we use `io.Reader`", well, we are. The `io.Reader` interface can accept many types, including `*os.File`, so I am using it now. What you're probably asking is can we instead `bytes.Reader` which can be created with`reader := bytes.NewReader(data)` to satisfy the `io.Reader` interface. And yes, I could do that, but should I?

My first inclination is to use `*os.File`, but if you think `bytes.Reader` is still more appropriate for the tests, I'll make the change.

Sources: 
- https://pkg.go.dev/os#pkg-variables
- https://pkg.go.dev/os#NewFile

cmd/cli/plugin/feature/deactivate_test.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+func TestExecuteFeatureDectivateCommand(t *testing.T) {
+	test := struct {
+		description string
+		cmd         *cobra.Command
+		featName    string
+		errMsg      string
+	}{
+		description: "FeatureGateClient cannot connect due to bad configuration",
```
yharish991, line 21: can you add a testcase for feature deactivation working successfully?

cmd/cli/plugin/feature/activate_test.go
```diff
@@ -0,0 +1,258 @@
...
+	objs, _, _ := fake.GetTestObjects()
+	s := scheme.Scheme
+	if err := corev1alpha2.AddToScheme(s); err != nil {
+		t.Fatalf("add config scheme: (%v)", err)
+	}
+
+	cl := crclient.NewClientBuilder().WithRuntimeObjects(objs...).Build()
+	fgClient, err := featuregateclient.NewFeatureGateClient(featuregateclient.WithClient(cl))
+	if err != nil {
+		t.Fatalf("get FeatureGate client: %v", err)
```
yharish991, line 110:
```suggestion
		t.Fatalf("could not get FeatureGate client: %v", err)
```

yharish991: can you make this change in other places of this pr as well?

cmd/cli/plugin/feature/deactivate_test.go
codegold79: I added a few more test cases for deactivating a feature.

cmd/cli/plugin/feature/activate_test.go
codegold79: Done.

cmd/cli/plugin/feature/activate.go
rajathagasthya: > Shouldn't code be most restrictive to the type that we know will work? If I allow for bytes.Buffer for tests, I'd have to allow for it in the plugin code, and I don't know if it will work as intended.

Not sure this is a valid concern. You are passing `os.Stdin` to `fmt.Fscanln` which takes in a `io.Reader`. It does not take in an `*os.File`. So we know it should work for all types that implement `io.Reader`. 

> What you're probably asking is can we instead bytes.Reader which can be created withreader := bytes.NewReader(data) to satisfy the io.Reader interface. And yes, I could do that, but should I?

I'm asking if we can use `bytes.Buffer` to simulate reading from a file since that type implements `io.Reader`.

> But if you think bytes.Reader is still more appropriate for the tests, I'll have to make changes to production code (in the plugin's activation logic) and allow for types other than *os.File. Seems risky.

Why would plugin activation logic change if we make this change? The reason I discourage temp files in test is because they tend to not be cleaned up properly

codegold79: > Why would plugin activation logic change if we make this change?

It's because right now, I have in the plugin activation logic a function signature that only accepts the `*os.File` type. Yes, I know I could have opened it up to more possibilities. I can open it up to anything that meets the `io.Reader` interface which is what `fmt.Fscanln` requires. But, I don't really want to think about how opening up the function signature to accept more than `*os.File` affects plugin logic. I wanted to restrict it to `Stdin` (a file) as much a possible so I didn't have to think about other things.

The function I'm talking about is in `activation.go`. You made your review comment in production code, not test code, so you must see how I have to change plugin activation logic to allow for the test to send in `bytes.Buffer` instead of `*os.File`. Unless I'm missing something.

I don't see a problem with file clean up, but if you do, and you think there's no problem with accepting any type in the `userAllowsByInteractiveCLI` function that meets `io.Reader`, then I'll go ahead and make the change to allow for `bytes. Buffer` to be sent in for testing purposes.

```go
func userAllowsByInteractiveCLI(in *os.File) (bool, error) {
        if in == nil {
		in = os.Stdin
	}

	var userAnswer string
	if n, err := fmt.Fscanln(in, &userAnswer); err != nil {
		if n == 0 {
			return false, nil
		}
		return false, fmt.Errorf("could not read user input: %w", err)
	}
       ...
}
```

rajathagasthya: > But, I think tests should be close to what is happening in production. By restricting userAllowsByInteractiveCLI function signature to only use Stdin (a file), tests will be forced to use os.File which is what I want.

You can keep `os.File`, but this is not a valid reason. Using an interface here isn't changing the function's behavior.

rajathagasthya: I'll resolve this as this is a fairly minor issue .

rajathagasthya (approved):

PR #4107, opened 2022-12-06
Bump capz 1.4.5 and TKG-13041 upgrade azure fix

996refuse:


sameerkhan97:
@996refuse Can we get a closure on this since #4132 has already been merged.

PR #4112, opened 2022-12-07
clusterclass: move overlays from tkg-clusterclass to provider specific directories

chrischdi:
This is currently the way to go, there is no central location for providing provider-unspecific overlays.

Follow-up to #3862.
Having the overlays in `tkg-clusterclass` results in:
* not having them included in `tanzu mc create --dry-run`
* makes it hard to debug overlay changes which overlap with making it way more complicated to gather what overlays are added to a clusterclass

Fixes #

### Release note

```release-note

```


packages/tkg-clusterclass-aws/bundle/config/upstream/overlay.yaml
```diff
@@ -35,4 +35,17 @@ spec:
           status: "False"
           timeout: #@ data.values.MHC_FALSE_STATUS_TIMEOUT
         #@ end
-
+  patches:
```
wjun, line 38: This needs to be added into base.yaml at the beginning of the patches, as there are extraVolume patches in base.yaml. The same is for all cloud providers.

wjun (approved):
LGTM

PR #4130, opened 2022-12-09
fix TKG-15754

weiwenli97:
For new created clusterclass based cluster, we cannot read the existing OS information for the os information annotation is missed while it is implemented in legacy cluster.

Fixes #TKG-15754

Create vsphere clusterclass based cluster and os information annotations is set successfully.

### Release note

```release-note

```


wjun (approved):
LGTM

PR #3956, opened 2022-11-21
Decouple the feature gate of control management-cluster and workload-cluster mode

xiujuanx:
### What this PR does / why we need it
Currently, we are using `features.global.package-based-lcm-beta` feature gate to control management-cluster and workload-cluster mode at the same time. However, in the feature, cluster class is the default mode for management-cluster and the workload cluster can be legacy or clusterclass. So we need to decouple mc and wc mode.
- Rename `features.global.package-based-lcm-beta` to `features.management-cluster.package-based-cc`. It only can control MC mode. By default, it true;
- Add a new feature gate `features.cluster.allow-legacy-cluster` for workload cluster mode. By default, it's false;

### Which issue(s) this PR fixes

Fixes #https://github.com/vmware-tanzu/tanzu-framework/issues/3957

### Describe testing done for PR
1. cluster class based management-cluster
- create a cluster class management-cluster(`features.management-cluster.package-based-cc` is true by default)
```
clientOptions:
    features:
        global:
            context-target: "false"
            context-aware-cli-for-plugins: "true"
            tkr-version-v1alpha3-beta: "false"
        cluster:
            dual-stack-ipv6-primary: "false"
            allow-legacy-cluster: "true"
            custom-nameservers: "false"
            dual-stack-ipv4-primary: "false"
        management-cluster:
            package-based-cc: "true"
            export-from-confirm: "true"
            import: "false"
            standalone-cluster-mode: "false"
            aws-instance-types-exclude-arm: "true"
            custom-nameservers: "false"
            dual-stack-ipv4-primary: "false"
            dual-stack-ipv6-primary: "false"
        package:
            kctrl-package-command-tree: "true"
```
cluster class based mc is created successfully:
```
kubo@guvRofrQi1q2i:~$ kubectl get cluster tkg-mgmt-vc -n tkg-system -o json | jq -r '.spec.topology.class'
tkg-vsphere-default
```
- Create cluster class based workload cluster (`features.cluster.allow-legacy-cluster` is false by default)
```
kubo@guvRofrQi1q2i:~$ tanzu cluster list
  NAME    NAMESPACE  STATUS   CONTROLPLANE  WORKERS  KUBERNETES        ROLES   PLAN  TKR
  cc-dev  default    running  1/1           1/1      v1.24.6+vmware.1  <none>        v1.24.6---vmware.1-tkg.1-zshippable
kubo@guvRofrQi1q2i:~$ kubectl get cluster cc-dev -o json | jq -r '.spec.topology.class'
[...]

PR #4134, opened 2022-12-11
Supplement unit tests for pr 3956

xiujuanx:
This pr supplements unit tests for [decouple mgmt and wl cluster feature gates pr](https://github.com/vmware-tanzu/tanzu-framework/pull/3956). Before the code coverage of [PR 3956](https://github.com/vmware-tanzu/tanzu-framework/pull/3956) didn't meet the standards, however, it was urgent to enable classy as the default mode for pipeline testing. So it was merged. Now, this pr is adding the missing unit tests. 

Fixes #

### Release note

```release-note

```

### DIFF
--- a/tkg/client/common.go
+++ b/tkg/client/common.go
@@ -17,7 +17,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/version"
 	apimachineryversion "k8s.io/apimachinery/pkg/version"
 
-	"github.com/vmware-tanzu/tanzu-framework/cli/runtime/config"
 	"github.com/vmware-tanzu/tanzu-framework/tkg/constants"
 	"github.com/vmware-tanzu/tanzu-framework/tkg/log"
 	"github.com/vmware-tanzu/tanzu-framework/tkg/tkgconfigreaderwriter"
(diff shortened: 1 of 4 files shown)


tkg/client/common.go
```diff
@@ -323,9 +322,21 @@ func (c *TkgClient) SetAllowLegacyClusterConfiguration() string {
 		log.V(6).Infof("Setting %v to %q", constants.ConfigVariableAllowLegacyCluster, allowLegacyCluster)
 		c.TKGConfigReaderWriter().Set(constants.ConfigVariableAllowLegacyCluster, allowLegacyCluster)
 	} else {
-		log.V(6).Infof("Info: %v configuration already set to %q", constants.ConfigVariableAllowLegacyCluster, value)
 		// ALLOW_LEGACY_CLUSTER is explicitly set in cluster config file
-		allowLegacyCluster = value
+		log.V(6).Infof("Info: %v configuration already set to %q", constants.ConfigVariableAllowLegacyCluster, value)
+		if value != "true" && value != "false" {
```
DanielXiao, line 327: Can it be defined as a bool variable in the config file?

tkg/client/common.go
```diff
@@ -323,9 +322,21 @@ func (c *TkgClient) SetAllowLegacyClusterConfiguration() string {
...
+			log.Warningf("Seem like you have set %v, but it's not a valid value. It should be true or false", constants.ConfigVariableAllowLegacyCluster)
+			// Set ALLOW_LEGACY_CLUSTER to a valid value
+			if !c.IsFeatureActivated(constants.FeatureFlagAllowLegacyCluster) {
```
DanielXiao, line 330: I am confused why there are feature flag allow-legacy-cluster and variable ALLOW_LEGACY_CLUSTER?

tkg/client/common.go
```diff
@@ -369,7 +381,7 @@ func (c *TkgClient) ShouldDeployClusterClassBasedCluster(isManagementCluster boo
 		log.Warning(constants.YTTBasedClusterWarning)
 	}
 
-	if config.IsFeatureActivated(constants.FeatureFlagForceDeployClusterWithClusterClass) {
+	if c.IsFeatureActivated(constants.FeatureFlagForceDeployClusterWithClusterClass) {
```
DanielXiao, line 384: What 's this feature flag used for? Can you draft a doc about the combination of all related flags, variables?

tkg/client/common.go
xiujuanx: @DanielXiao , according to the [design doc](https://docs.google.com/document/d/1VV_3z1d2OmJC9CoXMszYOvS1cFMziUL1nupMlIJbRM8/edit#), if user wants to create a legacy based cluster, then either use `tanzu config set features.cluster.allow_legacy_cluster true` command, or add variable `ALLOW_LEGACY_CLUSTER: "true"` into config file to active legacy mode.

tkg/client/common.go
xiujuanx: @DanielXiao , yes, it can be defined as a bool variable int he config file. 
The value is a string type is caused by `TKGConfigReaderWriter().Get` func will alway return a string, you can check [here](https://github.com/kubernetes-sigs/cluster-api/blob/main/cmd/clusterctl/client/config/reader_viper.go#L173-L177). 
And `Get` func is calling `GetString` func, [GetString](https://github.com/spf13/viper/blob/master/viper.go#L951-L955) func will alway converts a value to a string type.
```
func GetString(key string) string { return v.GetString(key) }

func (v *Viper) GetString(key string) string {
	return cast.ToString(v.Get(key))
}
```
So I think `ALLOW_LEGACY_CLUSTER`'s  is `true`(bool type) or `"true"`(string type) will cause the same action.

tkg/client/common.go
xiujuanx: `FeatureFlagForceDeployClusterWithClusterClass` is used to deploy classy based cluster even if user has ytt customization to the provider templates. You can find most of feature gates definition [here](https://github.com/vmware-tanzu/tanzu-framework/blob/main/tkg/constants/featureflags.go).

DanielXiao (approved):

wjun (approved):
LGTM

PR #4147, opened 2022-12-12
Add missing `TKG_DEFAULT_IMAGE_REPOSITORY` in CLI Makefile

### LINKED ISSUE #4144: Missing TKG_DEFAULT_IMAGE_REPOSITORY in cli/core/Makefile
**Bug description**

When building the CLI to use the default OCI repository, the build variable `TKG_DEFAULT_IMAGE_REPOSITORY` is missing and the CLI does not properly create the default discoverySource. 

It turns out that the `cli/core/Makefile` sets this the necessary discovery path by using `TKG_DEFAULT_IMAGE_REPOSITORY`, which is not set in that Makefile.

This is not a problem for production builds because that environment variable is set explicitly outside the makefile.
But for a local build that wants to use OCI, that variable should have a default value of "projects-stg.registry.vmware.com/tkg", like it does in the main Makefile.

- ( ) APIs
- ( ) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- (x) Installation
- ( ) Plugin
- ( ) Security
- (x) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

The variable `TKG_DEFAULT_IMAGE_REPOSITORY` should be set in `cli/core/Makefile`

```
$ cd tanzu-framework
$ # Build the CLI with OCI discovery
$ make install-cli-oci
$ tanzu config get | grep -A3 discoverySources

        discoverySources:
            - oci:
                name: default
                image: /packages/standalone-plugins:v0.28.0-dev

$ # Notice that the image is missing the domain name part (projects-stg.registry.vmware.com)
 ```

Today's main: 84cc3e591

Locally

### PR DESCRIPTION
**Pull request by marckhouzam**

Add the missing `TKG_DEFAULT_IMAGE_REPOSITORY` default value in the CLI makefile

Fixes #4144

Before PR:
```
$ cd tanzu-framework
$ # Build the CLI with OCI discovery
$ make install-cli-oci
$ tanzu config get | grep -A3 discoverySources

        discoverySources:
            - oci:
                name: default
                image: /packages/standalone-plugins:v0.28.0-dev

$ # Notice that the image is missing the domain name part (projects-stg.registry.vmware.com)
```
With PR:
```
$ cd tanzu-framework
$ # Build the CLI with OCI discovery
$ make install-cli-oci
$ tanzu config get | grep -A3 discoverySources

tanzu config get | grep -A3 discoverySources
        discoverySources:
            - oci:
                name: default
                image: projects-stg.registry.vmware.com/tkg/packages/standalone-plugins:v0.28.0-dev

$ # Notice that the image is now complete
```

Note that this missing variable is not a problem for production builds because that environment variable is set explicitly outside the makefile.

### DIFF
--- a/cli/core/Makefile
+++ b/cli/core/Makefile
@@ -52,6 +52,9 @@ LD_FLAGS += -X 'github.com/vmware-tanzu/tanzu-framework/cli/core/pkg/buildinfo.D
 LD_FLAGS += -X 'github.com/vmware-tanzu/tanzu-framework/cli/core/pkg/buildinfo.SHA=$(BUILD_SHA)'
 LD_FLAGS += -X 'github.com/vmware-tanzu/tanzu-framework/cli/core/pkg/buildinfo.Version=$(BUILD_VERSION)'
 
+ifndef TKG_DEFAULT_IMAGE_REPOSITORY
+TKG_DEFAULT_IMAGE_REPOSITORY = "projects-stg.registry.vmware.com/tkg"
+endif
 ifndef ENABLE_CONTEXT_AWARE_PLUGIN_DISCOVERY
 ENABLE_CONTEXT_AWARE_PLUGIN_DISCOVERY = "true"
 endif


marckhouzam:
> LGTM. I see `TANZU_CORE_BUCKET` is also missing. another possible bug ? or could be discarded ?

You are right.  But that variable is not even set in the main `Makefile`.  Weird.
@prkalle Do we even support this core bucket anymore?

prkalle:
> You are right. But that variable is not even set in the main Makefile. Weird.
>@prkalle Do we even support this core bucket anymore?
I don't think we supporting the core bucket anymore after we introduced the plugin discovery and context-scope changes.
cc: @anujc25 @vuil

prkalle (approved):
LGTM. Thanks for the changes @marckhouzam!

anujc25:
As @prkalle mentioned above, `TANZU_CORE_BUCKET` can be ignored as we do not use the old repository behavior anymore and just use discovery sources.

anujc25 (approved):

Issue #4148, opened 2022-12-12
Dependabot config is missing most of the modules in the repo


rajathagasthya:
**Bug description**

`dependabot.yml` should include all modules in the repo to ensure automatic updates (including due to CVEs) happen to them. It's currently missing a lot of them. 

**Affected product area (please put an X in all that apply)**

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- (x) Security
- (x) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

Dependabot to create version bump PRs for all modules in the repo.

**Steps to reproduce the bug**

**Version** (include the SHA if the version is not obvious)

**Environment where the bug was observed (cloud, OS, etc)**

**Relevant Debug Output (Logs, manifests, etc)**

vuil:
Thanks for filing the issue to track this. @rajathagasthya

PR #4152, opened 2022-12-12
Update dependabot.yaml to disable version checks correctly

vuil:

Looks like the current way of disabling version updates is no longer valid. Update the configuration to achieve the same effect.

Signed-off-by: Vui Lam <vui@vmware.com>

Related-to: #4148

Monitor CI before merge.
Will monitor dependabot PRs after to verify only security updates are posted.

### Release note
```release-note
None
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,54 +1,69 @@
 # Dependabot config for go mod version upgrades
+#
+# TODO(vuil):
+# All version updates being disabled with
+# 'open-pull-requests-limit: 0'
+# restoring of version updates being tracked in
+# https://github.com/vmware-tanzu/tanzu-framework/issues/4148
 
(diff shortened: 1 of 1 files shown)


codegold79 (approved):
Thanks for the informative comment explaining what `open-pull-requests-limit: 0` does, and referring to issue 4148. 

Also, thanks for quoting [the docs](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates#overriding-the-default-behavior-with-a-configuration-file) in a separate conversation we had. Your quote was,

> If you only require security updates and want to exclude version updates, you can set open-pull-requests-limit to 0 in order to prevent version updates for a given package-ecosystem. For more information, see "open-pull-requests-limit."

prkalle (approved):
LGTM. Thanks!

PR #4157, opened 2022-12-13
transfer storageclass inline value to downstream

m1zzx2:
The in-tree storage class is not needed in the cluster class base, so delete it.


jeffwubj (approved):

imikushin (approved):
LGTM.

PR #4162, opened 2022-12-13
Rename kctrl-command-tree feature flag

praveenrewar:
Rename kctrl-command-tree feature flag.
Use `features.package.kctrl-command-tree` instead of `features.package.kctrl-package-command-tree`. This is done to ensure that the default value is toggled to true even while upgrading from an older version.

Fixes #

### Release note

```release-note
`tanzu config set features.package.kctrl-command-tree false` can be used to turn off kctrl command tree.
```

### DIFF
--- a/cmd/cli/plugin/package/featureflags.go
+++ b/cmd/cli/plugin/package/featureflags.go
@@ -6,7 +6,7 @@ package main
 const (
 	// Package Plugin Kctrl Command Tree determines whether to use the command tree from kctrl. Setting feature flag to
 	// true will allow to use the package command tree from kctrl for package plugin
-	FeatureFlagPackagePluginKctrlCommandTree = "features.package.kctrl-package-command-tree"
+	FeatureFlagPackagePluginKctrlCommandTree = "features.package.kctrl-command-tree"
 )
 
 // DefaultFeatureFlagsForPackagePlugin is used to populate default feature-flags for the package plugin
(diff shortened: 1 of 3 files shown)


marckhouzam (commented):
This looks ok to me.
But we may want to update the string `kctrl-package-command-tree` in the test of `cmd/cli/plugin/package/test/package_plugin_integration_test.go`.

praveenrewar:
> But we may want to update the string kctrl-package-command-tree in the test of cmd/cli/plugin/package/test/package_plugin_integration_test.go.

Good catch! Thanks :)
Not that it matters but I am thinking of updating the usage in `cli/runtime/config/config_factory_test.go` as well.

marckhouzam:
> Good catch! Thanks :) Not that it matters but I am thinking of updating the usage in `cli/runtime/config/config_factory_test.go` as well.

Sounds like a good idea.  Thanks for doing that! 👍

cmd/cli/plugin/package/featureflags.go
```diff
@@ -6,7 +6,7 @@ package main
 const (
 	// Package Plugin Kctrl Command Tree determines whether to use the command tree from kctrl. Setting feature flag to
 	// true will allow to use the package command tree from kctrl for package plugin
-	FeatureFlagPackagePluginKctrlCommandTree = "features.package.kctrl-package-command-tree"
```
shyaamsn, line 9: Is it ok to remove the old feature flag ? @anujc25 Would there be any issue if this flag is present in config.yaml

anujc25: Yes. It is fine to remove the old flag. It shouldn't be an issue if the old flag is present.

maralavi (approved):

PR #4164, opened 2022-12-13
Add check_version make target to the toolchain make step

### LINKED ISSUE #4165: hack/tools/bin out of sync can give odd local results
**Bug description**
Theres alot of compile deps, but the way we call hack/tools/Makefile doesnt clean them out which means that if you have older ytt like 3.0, youll get unobvious errors like

```
ytt: Error: Non-ytt comment at line upstream/metadata.yaml:1: '# maps release series of major.minor to cluster-api contract version':
  Unrecognized comment type (expected '#@' or '#!'). (hint: if this is plain YAML — not a template — consider `--file-mark '<filename>:type=yaml-plain'`)
```

This can cost hours of developer debugging on a local env.  So we need to quickly test YTT (and optionally other versions) in the hack tools make flow, after running it so that we can inform users that they have a wrong version in their toolchain in the hack dir.

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( x ) Test and Release
- ( ) User Experience
- ( x ) Developer Experience

**Expected behavior**

the make tools workflow will ensure that you either have the right tools, or explicitly fail. 

install ytt 0.3.0 in hack/tools/bin and then run `make test`

main 

jayunit100:
Check YTT version before compilation.  Spent a long several hours on this bc my hack/tools dir was out of date.  should prevent similar confusion in the future.

Fixes https://github.com/vmware-tanzu/tanzu-framework/issues/4165

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -1,6 +1,10 @@
 # Copyright 2021 VMware, Inc. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+# Versions for checking installation, synced with common.mk
+# Otherwise developers have to remember to clean hack/tools/bin/ when running!
+YTT_VERSION=v0.40.0
+
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
 GOHOSTOS ?= $(shell go env GOHOSTOS)
@@ -117,7 +121,13 @@ help: ## Display this help (default)
 ##@ Tooling Binaries
 ## --------------------------------------
 
-tools: $(TOOLING_BINARIES) ## Build tooling binaries
+check_tools:
+	# Check YTT Version or fail.
+	# Could do something more clever reusing versions in hack/tools/Makefile someday.
+	$(TOOLS_DIR)/check_version.sh $(TOOLS_BIN_DIR) ytt $(YTT_VERSION)
+
+tools: $(TOOLING_BINARIES) check_tools ## Build tooling binaries
+
 .PHONY: $(TOOLING_BINARIES)
 $(TOOLING_BINARIES):
 	make -C $(TOOLS_DIR) $(@F)
--- a/hack/tools/check_version.sh
+++ b/hack/tools/check_version.sh
@@ -4,7 +4,7 @@
 # exit 0 if the program, when called from path, has the version in its output.
 VERSION=`$1/$2 version` 
 
-# We dont check v bc, we dont care about "v0.40.0" vs "0.40.0" and differnt
+# We dont check v bc, we dont care about "v0.40.0" vs "0.40.0" and some
 # programs print version strings out differently.
 if echo $VERSION | `grep -q $3 | sed 's/v//g'` ; then
     echo "correct version *** $VERSION ***"
--- a/hack/tools/check_version.sh
+++ b/hack/tools/check_version.sh
@@ -1,0 +1,17 @@
+#! /bin/bash
+
+# Takes path , program, version as input
+# exit 0 if the program, when called from path, has the version in its output.
+VERSION=`$1/$2 version` 
+
+# We dont check v bc, we dont care about "v0.40.0" vs "0.40.0" and differnt
+# programs print version strings out differently.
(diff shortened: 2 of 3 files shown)


common.mk
```diff
@@ -117,7 +121,13 @@ help: ## Display this help (default)
 ##@ Tooling Binaries
 ## --------------------------------------
 
-tools: $(TOOLING_BINARIES) ## Build tooling binaries
+check_tools:
+	# Check YTT Version or fail.
+	# Could do something more clever reusing versions in hack/tools/Makefile someday.
```
marckhouzam, line 126: The above two comments are printed to the screen.  If you want to suppress them you need to prefix even the comments with `@`

common.mk
```diff
@@ -117,7 +121,13 @@ help: ## Display this help (default)
 ##@ Tooling Binaries
 ## --------------------------------------
 
-tools: $(TOOLING_BINARIES) ## Build tooling binaries
+check_tools:
```
marckhouzam, line 124: If you want this new target to appear in the output of `make help` you need to suffix it with a `##` command with the text to use in the help.  Then again, you may not want to make it an officially public target.

hack/tools/check_version.sh
```diff
@@ -0,0 +1,17 @@
+#! /bin/bash
+
+# Takes path , program, version as input
+# exit 0 if the program, when called from path, has the version in its output.
+VERSION=`$1/$2 version` 
+
+# We dont check v bc, we dont care about "v0.40.0" vs "0.40.0" and some
+# programs print version strings out differently.
+if echo $VERSION | `grep -q $3 | sed 's/v//g'` ; then
```
marckhouzam, line 9: The backticks look weird here...
And I don't understand what the `sed` is trying to do?  Does it want to remove the `v` from the `$VERSION` variable?  If so maybe it should be line 5?

hack/tools/check_version.sh
```diff
@@ -0,0 +1,17 @@
...
+    echo "correct version *** $VERSION ***"
+    exit 0
+else
+    echo "incorrect version of $0/$1: *** $VERSION *** "
```
marckhouzam, line 13: Should this be `$1/$2` instead?
Maybe setting variable to these parameters would help?

hack/tools/check_version.sh
jayunit100: ytt version returns "0.40.0" but the YTT_VERSION = "v0.40.0"

hack/tools/check_version.sh
jayunit100: ah sorry, old commit pushing update !

common.mk
jayunit100: (no specific urge to suppress) seems like its good debugging info ?  but good to note.

jayunit100:
adding license check to the validation script

common.mk
jayunit100: doesnt really need to be officially public, ok if i resolve ?

hack/tools/check_version.sh
jayunit100: ya , they basically say "dont bother matching the v" 


- `grep -q` will cause a failure if theres no match
- `sed 's/v//g' will replace "v0.40" with "0.40" 
- The final result: It will make sure that 

YTT_VERSION=v0.40 OR YTT_VERSION=0.40

both pass validation :)

common.mk
marckhouzam: Sure. I'll mark as resolved

hack/tools/check_version.sh
marckhouzam: The check never fails for me:
```
$ make check_tools
# Check YTT Version or fail.
# Could do something more clever reusing versions in hack/tools/Makefile someday.
/Users/kmarc/git/tanzu-framework/hack/tools/check_version.sh /Users/kmarc/git/tanzu-framework/hack/tools/bin ytt v0.40.0
correct version *** 0.3.0 ***```

marckhouzam (approved):
LGTM

PR #4170, opened 2022-12-14
add ALLOW_LEGACY_CLUSTER parameter into default config

xiujuanx:
In G release, we still support legacy based workload cluster creation. So we need to add `ALLOW_LEGACY_CLUSTER` parameter into default config file so that user can know how to active legacy mode.

Fixes #

### Release note

```release-note

```

### DIFF
--- a/providers/config_default.yaml
+++ b/providers/config_default.yaml
@@ -33,6 +33,10 @@ CLUSTER_API_SERVER_PORT:
 
 CLUSTER_CLASS:
 
+#! ALLOW_LEGACY_CLUSTER will allow legacy based workload cluster creation.
+#! Supported values are "false", "true". By default, it's "false".
+ALLOW_LEGACY_CLUSTER:
+
 #! Instance size configuration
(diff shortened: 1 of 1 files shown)


DanielXiao (approved):

providers/config_default.yaml
```diff
@@ -33,6 +33,10 @@ CLUSTER_API_SERVER_PORT:
 
 CLUSTER_CLASS:
 
+#! ALLOW_LEGACY_CLUSTER will allow legacy based workload cluster creation.
+#! Support values are 'false', 'true'. By default, it's false.
```
wjun, line 37: Support ->Supported. Please also align it to other parameters: do you mean "false", "true"? If so, please update the comment as:  Supported values are "false", "true". By default, it's "false".

xiujuanx: Changed according to the comment.

wjun (approved):
LGTM

PR #4176, opened 2022-12-15
use ALLOW_LEGACY_CLUSTER to judge create legacy or classy cluster

### LINKED ISSUE #4177: Legacy based workload cluster doesn't create automactically when set `ALLOW_LEGACY_CLUSTER: true` in config file
**Bug description**

Related to [TKG-15894](https://jira.eng.vmware.com/browse/TKG-15894)
Set ALLOW_LEGACY_CLUSTER: "true" in config file,  and `features.cluster.allow-legacy-cluster` is still false. During create workload cluster, it will stop and show us below message:
```
Warning: Use of ytt based cluster templates will be deprecated in favor of ClusterClass templates in a future version of TKG. Please work to move your workloads to a ClusterClass enabled cluster.Legacy configuration file detected. The inputs from said file have been converted into the new Cluster configuration as '/etc/tkg/clusterconfigs/tkg210-azure-wc.yaml'To create a cluster with it, use
    tanzu cluster create --file /etc/tkg/clusterconfigs/tkg210-azure-wc.yaml
```
Although this converted config file is legacy based, this behaviour is unexpected.

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- (x) User Experience
- ( ) Developer Experience

**Expected behavior**
 Expected behaviour is create legacy based cluster directly.

xiujuanx:

Fixes #4177 

### Release note

```release-note

```

### DIFF
--- a/tkg/client/cluster.go
+++ b/tkg/client/cluster.go
@@ -160,7 +160,11 @@ func (c *TkgClient) CreateCluster(options *CreateClusterOptions, waitForCluster 
 			return false, errors.Wrap(err, "unable to get cluster configuration")
 		}
 
-		if !c.IsFeatureActivated(constants.FeatureFlagAllowLegacyCluster) {
+		allowLegacyCluster, err := c.allowLegacyCluster()
+		if err != nil {
+			return false, err
+		}
+		if !allowLegacyCluster {
 			clusterConfigDir, err := c.tkgConfigPathsClient.GetClusterConfigurationDirectory()
 			if err != nil {
 				return false, err
@@ -195,6 +199,16 @@ func (c *TkgClient) CreateCluster(options *CreateClusterOptions, waitForCluster 
 		return false, nil
 	}
 	return true, c.waitForClusterCreation(regionalClusterClient, options)
+}
+
+func (c *TkgClient) allowLegacyCluster() (bool, error) {
+	allowLegacyCluster, err := c.TKGConfigReaderWriter().Get(constants.ConfigVariableAllowLegacyCluster)
+	if err != nil {
+		log.V(6).Infof("failed getting variable ALLOW_LEGACY_CLUSTER, %s", err.Error())
+		return false, err
+	}
+
+	return strconv.ParseBool(allowLegacyCluster)
 }
 
 // getClusterConfigurationBytes returns cluster configuration by taking into consideration of legacy vs clusterclass based cluster creation
--- a/tkg/client/cluster_test.go
+++ b/tkg/client/cluster_test.go
@@ -11,6 +11,7 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/pkg/errors"
 	"gopkg.in/yaml.v3"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/pointer"
@@ -328,6 +329,16 @@ ova: []
 				tkgConfigUpdaterClient.GetPopulatedProvidersChecksumFromFileStub = func() (string, error) {
 					return "FakeFileShaIsSame", nil
 				}
+				tkgConfigReaderWriter.GetCalls(func(key string) (string, error) {
+					configMap := populateConfigMap()
+					if val, ok := configMap[key]; ok {
+						return val, nil
+					}
+					if key == constants.ConfigVariableAllowLegacyCluster {
+						return "true", nil
+					}
+					return "192.168.2.1/16", nil
+				})
 			})
 
 			It("Should go to create a legacy based workload cluster directly", func() {
@@ -350,6 +361,16 @@ ova: []
 				tkgConfigUpdaterClient.GetPopulatedProvidersChecksumFromFileStub = func() (string, error) {
 					return "FakeFileShaIsSame", nil
 				}
+				tkgConfigReaderWriter.GetCalls(func(key string) (string, error) {
+					configMap := populateConfigMap()
+					if val, ok := configMap[key]; ok {
+						return val, nil
+					}
+					if key == constants.ConfigVariableAllowLegacyCluster {
+						return "false", nil
+					}
+					return "192.168.2.1/16", nil
+				})
 			})
 
 			It("Should go to create a classy based workload cluster directly", func() {
@@ -385,6 +406,43 @@ ova: []
 				Expect(result).To(ContainSubstring("To create a cluster with it, use"))
 			})
 
+		})
+
+		When("ALLOW_LEGACY_CLUSTER is failed to set", func() {
+			BeforeEach(func() {
+				tkgConfigReaderWriter.GetCalls(func(key string) (string, error) {
+					configMap := populateConfigMap()
+					if val, ok := configMap[key]; ok {
+						return val, nil
+					}
+					if key == constants.ConfigVariableAllowLegacyCluster {
+						return "false", errors.Errorf("fake error")
+					}
+					return "192.168.2.1/16", nil
+				})
+
+				featureFlagClient.IsConfigFeatureActivatedStub = func(featureFlagName string) (bool, error) {
+					if featureFlagName == constants.FeatureFlagAllowLegacyCluster {
+						return false, nil
+					}
+					return true, nil
+				}
+
+				tkgConfigUpdaterClient.GetProvidersChecksumStub = func() (string, error) {
+					return "FakeFileShaIsSame", nil
+				}
+				tkgConfigUpdaterClient.GetPopulatedProvidersChecksumFromFileStub = func() (string, error) {
+					return "FakeFileShaIsSame", nil
+				}
+			})
+
+			It("Should return an error", func() {
+				options = createClusterOptions(clusterName, "../fakes/config/config.yaml")
+				options.IsInputFileClusterClassBased = false
+				Expect(err).ToNot(HaveOccurred())
+				result := captureOutput(*tkgClient, options, false)
+				Expect(result).To(ContainSubstring("failed getting variable ALLOW_LEGACY_CLUSTER"))
+			})
 		})
 	})
 
--- a/tkg/client/common.go
+++ b/tkg/client/common.go
@@ -310,7 +310,7 @@ func (c *TkgClient) isCustomOverlayPresent() (bool, error) {
 // Sets the appropriate AllowLegacyCluster configuration unless it has been explicitly overridden
 func (c *TkgClient) SetAllowLegacyClusterConfiguration() string {
 	var allowLegacyCluster string
-	value, err := c.TKGConfigReaderWriter().Get(constants.ConfigVariableAllowLegacyCluster)
+	value, err := c.allowLegacyCluster()
 	if err != nil {
 		// ALLOW_LEGACY_CLUSTER doesn't be explicitly set in cluster config file
 		if !c.IsFeatureActivated(constants.FeatureFlagAllowLegacyCluster) {
@@ -325,19 +325,7 @@ func (c *TkgClient) SetAllowLegacyClusterConfiguration() string {
 	} else {
 		// ALLOW_LEGACY_CLUSTER is explicitly set in cluster config file
 		log.V(6).Infof("Info: %v configuration already set to %q", constants.ConfigVariableAllowLegacyCluster, value)
-		if value != "true" && value != "false" {
-			log.Warningf("Seem like you have set %v, but it's not a valid value. It should be true or false", constants.ConfigVariableAllowLegacyCluster)
-			// Set ALLOW_LEGACY_CLUSTER to a valid value
-			if !c.IsFeatureActivated(constants.FeatureFlagAllowLegacyCluster) {
-				allowLegacyCluster = "false"
-			} else {
-				allowLegacyCluster = "true"
-			}
-			log.V(6).Infof("Setting %v to %q", constants.ConfigVariableAllowLegacyCluster, allowLegacyCluster)
-			c.TKGConfigReaderWriter().Set(constants.ConfigVariableAllowLegacyCluster, allowLegacyCluster)
-		} else {
-			allowLegacyCluster = value
-		}
+		allowLegacyCluster = strconv.FormatBool(value)
 	}
 
 	return allowLegacyCluster


tkg/client/cluster.go
```diff
@@ -160,7 +160,13 @@ func (c *TkgClient) CreateCluster(options *CreateClusterOptions, waitForCluster
 			return false, errors.Wrap(err, "unable to get cluster configuration")
 		}
 
-		if !c.IsFeatureActivated(constants.FeatureFlagAllowLegacyCluster) {
+		allowLegacyCluster, err := c.TKGConfigReaderWriter().Get(constants.ConfigVariableAllowLegacyCluster)
```
DanielXiao, line 163: Can you add more context in the commit message? https://cbea.ms/git-commit/

xiujuanx: @DanielXiao , add more explanations into commit message, pls check it again, thanks!

wjun, line 163: Will this get api make sure allow_legacy_cluster set through either feature gate or config file ready to get?

tkg/client/cluster.go
```diff
@@ -160,7 +160,13 @@ func (c *TkgClient) CreateCluster(options *CreateClusterOptions, waitForCluster
...
+		if err != nil {
+			log.Error(err, "ALLOW_LEGACY_CLUSTER variable is set failed")
```
wjun, line 165: What does the variable here mean?

xiujuanx: @wjun 
>I wonder if the error means 'variable is not set' actually?

when get api has some issues will also cause error, so can I modify the error message to `Get ALLOW_LEGACY_CLUSTER variable's value failed`?

tkg/client/cluster.go
xiujuanx: @wjun 
If users don't enable allow_legacy_cluster, then our code will set allow_legacy_cluster according to feature gate's value. By default, it's "false".

tkg/client/cluster.go
DanielXiao: Is err not nil when ALLOW_LEGACY_CLUSTER is absent? Or just it raise an error only when type is not bool?

xiujuanx: @DanielXiao , when err is not nil, that means ALLOW_LEGACY_CLUSTER is absent or get api has some issues. But actually, in this place, if error happens, that only means get api has issues. Because before this check, we have set `ALLOW_LEGACY_CLUSTER` [here](https://github.com/vmware-tanzu/tanzu-framework/blob/main/tkg/client/common.go#L367). [SetAllowLegacyClusterConfiguration](https://github.com/vmware-tanzu/tanzu-framework/blob/main/tkg/client/common.go#L310-L342)  func has checked if ALLOW_LEGACY_CLUSTER is absent or not. If it's absent, then we will set it.

tkg/client/cluster.go
```diff
@@ -160,7 +160,13 @@ func (c *TkgClient) CreateCluster(options *CreateClusterOptions, waitForCluster
...
+			return false, err
+		}
+
+		if allowLegacyCluster != "true" {
```
imikushin, line 169: **suggestion:** It is best not to make an already giant method bigger. Instead, create a function encapsulating the added logic. 

The code that you are replacing is simpler. It would be great to preserve that simplicity.

It will also be easier to test and understand.

tkg/client/cluster.go
imikushin: **suggestion:** Why don't we use something like the following as the message?

```go
fmt.Sprintf("failed getting environment variable %s", constants.ConfigVariableAllowLegacyCluster)
```

tkg/client/cluster_test.go
```diff
@@ -386,6 +387,43 @@ ova: []
...
+					}
+					if key == constants.ConfigVariableAllowLegacyCluster {
+						return "false", errors.Errorf("fake error")
+					}
+					return "192.168.2.1/16", nil
+				})
+
+				featureFlagClient.IsConfigFeatureActivatedStub = func(featureFlagName string) (bool, error) {
+					if featureFlagName == constants.FeatureFlagAllowLegacyCluster {
+						return false, nil
```
imikushin, line 406: **suggestion:** We should test for all combinations of:
- feature gate is true or false
- env var is absent, set to something we can't parse, set to true or false.

This is also why we should use a function to encapsulate our logic and return a clear value.

tkg/client/cluster.go
```diff
@@ -160,7 +160,13 @@ func (c *TkgClient) CreateCluster(options *CreateClusterOptions, waitForCluster
 			return false, errors.Wrap(err, "unable to get cluster configuration")
 		}
 
-		if !c.IsFeatureActivated(constants.FeatureFlagAllowLegacyCluster) {
+		allowLegacyCluster, err := c.TKGConfigReaderWriter().Get(constants.ConfigVariableAllowLegacyCluster)
+		if err != nil {
+			log.Error(err, "ALLOW_LEGACY_CLUSTER variable is set failed")
+			return false, err
```
imikushin, line 166: **question:** Is it catastrophic if we fail to get the ALLOW_LEGACY_CLUSTER env var? Especially if it isn't set. I think, not. Logging the fact is sufficient in my opinion. Then fallback on the feature flag.

tkg/client/cluster_test.go
xiujuanx: @imikushin , about you mentioned tests, they have been done [here](https://github.com/vmware-tanzu/tanzu-framework/blob/main/tkg/client/common_test.go#L97-L148).
So, I think we only need to test the situation where get ALLOW_LEGACY_CLUSTER env var failed.  Does this make sense?

tkg/client/cluster.go
```diff
@@ -197,6 +201,15 @@ func (c *TkgClient) CreateCluster(options *CreateClusterOptions, waitForCluster
 	return true, c.waitForClusterCreation(regionalClusterClient, options)
 }
 
+func (c *TkgClient) getAllowLegacyClusterVariable() (string, error) {
```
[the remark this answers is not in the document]
imikushin: Just noticed: error returned by `c.TKGConfigReaderWriter().Get()` is _only_ returned if the env var is not set. If we didn't rely on setting the env var based on the feature flag, and simply fell back on the feature flag, we wouldn't need to return an error. 

It's not something we can't do anything about (something external, like network error), so we don't _need_ to return an error from this method.

Also, we should call this method everywhere we're now checking the value of ALLOW_LEGACY_CLUSTER, instead. Luckily, it's just in this package.

tkg/client/cluster.go
imikushin: So, `err` is provably _always_ `nil`. Why bother checking it?

xiujuanx: @imikushin , no, `err` is not always `nil`. Because `c.TKGConfigReaderWriter().Get()` method will return a string and an error. like below:
```
// Get returns a configuration value of type string.
// In case the configuration value does not exists, it returns an error.
Get(key string) (string, error)
```
So check err is nil or not is necessary.

imikushin: See my other comment: https://github.com/vmware-tanzu/tanzu-framework/pull/4176#discussion_r1053911842

xiujuanx: @imikushin , error check is needed, because `c.TKGConfigReaderWriter().Get()` return a string and an error. if we want to below func returns a bool type,
```
func (c *TkgClient) allowLegacyCluster() (bool, error)
```
then we need convert the string value is returned by `c.TKGConfigReaderWriter().Get()` to a bool type use [strconv.ParseBool()](https://pkg.go.dev/strconv#ParseBool). Also `strconv.ParseBool` still returns a bool and an error. We can hidden error if happens.

tkg/client/cluster.go
```diff
@@ -160,7 +160,11 @@ func (c *TkgClient) CreateCluster(options *CreateClusterOptions, waitForCluster
 			return false, errors.Wrap(err, "unable to get cluster configuration")
 		}
 
-		if !c.IsFeatureActivated(constants.FeatureFlagAllowLegacyCluster) {
+		allowLegacyCluster, err := c.getAllowLegacyClusterVariable()
+		if err != nil {
+			return false, err
+		}
+		if allowLegacyCluster != "true" {
```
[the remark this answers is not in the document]
xiujuanx: Have changed codes as your suggestions!

tkg/client/cluster.go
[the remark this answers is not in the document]
xiujuanx: >Also, we should call this method everywhere we're now checking the value of ALLOW_LEGACY_CLUSTER, instead. Luckily, it's just in this package.

Good suggestion, modified code according to suggestion.

tkg/client/cluster.go
imikushin: > error check is needed, because `c.TKGConfigReaderWriter().Get()` return a string and an error. 

The error it returns is solely to indicate that the variable is not set. This information can be used (in our decision logic) and error safely ignored (we could log the fact that the env var is not set): the var _not being set by the user_ is not an error condition.

> if we want to below func returns a bool type,
> 
> ```
> func (c *TkgClient) allowLegacyCluster() (bool, error)
> ```
> 
> then we need convert the string value is returned by `c.TKGConfigReaderWriter().Get()` to a bool type use [strconv.ParseBool()](https://pkg.go.dev/strconv#ParseBool). Also `strconv.ParseBool` still returns a bool and an error. We can't hidden error if happens.

If we want to fail cluster creation with an error, if _the user made a mistake_ setting an env var, that is a legitimate option.

But then, we shouldn't modify the env var, because if we do, we're confusing ourselves (whether the value in the env var is input from the user or our interpretation of user preferences set elsewhere), making the code needlessly more complex.

I'm basically saying, we don't need the method `SetAllowLegacyClusterConfiguration()`, and we should use `allowLegacyCluster()`, which should fully encapsulate the decision logic (of whether to allow legacy clusters), instead.

In general, if we can make a decision _without_ mutating state (_especially_ globally accessible state, like an env var), we should do it _without_ mutating state.

imikushin: The same thinking applies everywhere: 
- If the user provides input via an env var, we _should not_ modify this env var, but only read from it.
- If it is not absolutely necessary to mutate state to do something, definitely _do not mutate state_.

xiujuanx: @imikushin , `SetAllowLegacyClusterConfiguration` func is needed, because this func needs to do below actions:
- if user provides the env var, then we will only read it;
- if user doesn't provide the env var, then we will set it according to the feature gate's value (`features.cluster.allow-legacy-cluster`).

`SetAllowLegacyClusterConfiguration` is called during get related cluster configurations, namely `getClusterConfigurationBytes` func. We will generate legacy or classy cluster configurations according to func's return.
After the cluster configuration is generated, we call `allowLegacyCluster()` func again just to get `ALLOW_LEGACY_CLUSTER` env's value to decide if we want to use a converted config file to create cluster or not. We don't modify it's value.
```
bytes, err = c.getClusterConfigurationBytes(&options.ClusterConfigOptions, infraProviderName, isManagementCluster)
if err != nil {
     return false, errors.Wrap(err, "unable to get cluster configuration")
}

allowLegacyCluster, err := c.allowLegacyCluster()
if err != nil {
     return false, err
}
if !allowLegacyCluster {
     ..........
 }
```
In a word, `SetAllowLegacyClusterConfiguration()` will set `ALLOW_LEGACY_CLUSTER`'s value. `allowLegacyCluster()` only read `ALLOW_LEGACY_CLUSTER`'s value.

xiujuanx:
```golang
bytes, err = c.getClusterConfigurationBytes(&options.ClusterConfigOptions, infraProviderName, isManagementCluster)
if err != nil {
     return false, errors.Wrap(err, "unable to get cluster configuration")
}

allowLegacyCluster, err := c.allowLegacyCluster() <======== (Here, I agree with you, we can ignore error.)
if err != nil {
     return false, err
}
if !allowLegacyCluster {
     ..........
 }
```
Because if error happens here, that means some network error or api client error, like you said, we can't do everything.

imikushin: > `SetAllowLegacyClusterConfiguration` func is needed, because this func needs to do below actions:
> 
> * if user provides the env var, then we will only read it;
> * if user doesn't provide the env var, then we will set it according to the feature gate's value (`features.cluster.allow-legacy-cluster`).

Yes, I can confirm it does these things. I'm also saying we don't really need it.

> `SetAllowLegacyClusterConfiguration` is called during get related cluster configurations, namely `getClusterConfigurationBytes` func. We will generate legacy or classy cluster configurations according to func's return. After the cluster configuration is generated, we call `allowLegacyCluster()` func again just to get `ALLOW_LEGACY_CLUSTER` env's value to decide if we want to use a converted config file to create cluster or not. We don't modify it's value.

Okay, I'm looking at the only place where it is used and I see this:

```go
	allowLegacyClusterCreated = c.SetAllowLegacyClusterConfiguration()
	if allowLegacyClusterCreated == "false" {
```

So, basically, its responsibility is to decide if legacy clusters are allowed (beside setting an env var). But don't we already have a good method for that? Yes we do, it is `allowLegacyCluster()`. Now, imagine `SetAllowLegacyClusterConfiguration()` didn't exist. We would then need to put all the logic from `SetAllowLegacyClusterConfiguration()` into `allowLegacyCluster()`, and call `allowLegacyCluster()` instead of `SetAllowLegacyClusterConfiguration()`.

We would end up setting an env var (if it is not set) and then reading from it. We'd naturally want to replace it with using a local variable, wouldn't we?

xiujuanx: @imikushin as we discussed in [Slack](https://vmware.slack.com/archives/C03UACASCFL/p1671428434106079), we may focus on an incorrect path. The `ALLOW_LEGACY_CLUSTER` variable should be a config variable rather than an environment variable, although it can be overridden via set an env var. 
So I still think we need to set it according to the feature gate's value if users don't config it in their cluster config file. Otherwise, we will read it from their config file.
How do you think?

wjun (approved):
LGTM. @imikushin I had a chat with Xiujuan with this pr. Since in most cases, users will use config file parameter or feture gate to set allowlegacycluster, I think it is ok to set and get with different funcs. The error check may look useless in most cases, and may catch some I/O errors like disk is full/broken, etc.

PR #4180, opened 2022-12-15
Skip yttcc and yttcb customization check

### LINKED ISSUE #4187: Ytt overlay customization check should be skipped for yttcc and yttcb
**Bug description**

Related to [TKG-15856]( https://jira.eng.vmware.com/browse/TKG-15856)
The purpose of the checksum check is to detect modifications to legacy overlays that could possibly impact the accuracy of the legacy-to-CC translation of a cluster creation attempt using a legacy configuration file.
However, now, checkSum func is checking all ytt files. This causes some users create a classy based cluster with yttcc customizations will be fail.
We should allow users to create classy cluster with yttcc customizations. So we should skip yttcc and yttcb customizations check. That means we have below scenarios:
- If a user does ytt customizations, but user doesn't set allow-legacy-cluster to be true, then we will report error;
- If a user does ytt customizations, but user has set allow-legacy-cluster to be true, go to create legacy based cluster;
- If a user does yttcc or yttcb customizations, we should go to create classy cluster (by default) or legacy cluster (if allow-legacy-cluster is true).

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

xiujuanx:

Fixes #4187 

### Release note

```release-note

```

### DIFF
--- a/tkg/tkgconfigupdater/helper.go
+++ b/tkg/tkgconfigupdater/helper.go
@@ -212,6 +212,9 @@ func getFilesForChecksum(dirPath string) ([]string, error) {
 			return err
 		}
 
+		if info.IsDir() && (info.Name() == "yttcc" || info.Name() == "yttcb") {
+			return filepath.SkipDir
+		}
 		if !info.IsDir() && includePathForChecksum(path) {
 			files = append(files, path)
 		}
--- a/tkg/tkgconfigupdater/tkgconfigupdater_test.go
+++ b/tkg/tkgconfigupdater/tkgconfigupdater_test.go
@@ -349,6 +349,8 @@ var _ = Describe("GetProvidersChecksum", func() {
 		err      error
 		client   Client
 		checksum string
+		yttCCDir string
+		yttCBDir string
 	)
 
 	BeforeEach(func() {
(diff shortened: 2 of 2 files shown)


tkg/tkgconfigupdater/helper.go
```diff
@@ -212,6 +212,9 @@ func getFilesForChecksum(dirPath string) ([]string, error) {
 			return err
 		}
 
+		if info.IsDir() && (info.Name() == "yttcc" || info.Name() == "yttcb") {
```
DanielXiao, line 215: Can you explain the reason in commit message?

xiujuanx: @DanielXiao Add explanation into the commit message, pls check it, thanks!

wjun:
There are two funcs: GetProvidersChecksum and saveProvidersChecksumToFile that calculate ytt related checksum and compare. Please update both funcs to exclude yttcc/yttcb

xiujuanx:
@wjun 
>There are two funcs: GetProvidersChecksum and saveProvidersChecksumToFile that calculate ytt related checksum and compare. Please update both funcs to exclude yttcc/yttcb. More details can be found in the jira. Or the func of getFilesForChecksum are used by both funcs already?
For `saveProvidersChecksumToFile` func, it also calls `GetProvidersChecksum` to calculate, you can check [here](https://github.com/vmware-tanzu/tanzu-framework/blob/main/tkg/tkgconfigupdater/helper.go#L121-L140). After calculating the sha256, then save them.

DanielXiao (approved):

xiujuanx:
Hi @wjun @DanielXiao , discussed with Yang and Steven, we have another solution for [TKG-15865](https://jira.eng.vmware.com/browse/TKG-15856) issue. Another pr is submitted: https://github.com/vmware-tanzu/tanzu-framework/pull/4201

wjun (approved):
LGTM

PR #4192, opened 2022-12-18
Fix package plugin tests for classy clusters

### LINKED ISSUE #4209: Package plugin integration test fails with Calico CNI
**Bug description**

#4138 has broken package plugin integration test. It fails consistently with the following error:

```
Validating the pre-requisites...
Failure [2.369 seconds]
[BeforeSuite] BeforeSuite 
/datadrive/tanzu/actions-runner/_work/tanzu-framework/tanzu-framework/cmd/cli/plugin/package/test/package_plugin_integration_test.go:149

Unexpected error:
      <*errors.fundamental | 0xc00071a4b0>: {
          msg: "Calico management-cluster creation is forbidden...",
          stack: [0x285b7d5, 0x28624be, 0xaed011, 0xaeca05, 0xaed546, 0xb0f82b, 0xb0f30f, 0xb116f2, 0xb1c629, 0xb1c44a, 0x286d2a5, 0x524162, 0x46da21],
      }
      Calico management-cluster creation is forbidden...
  occurred
``` 

This needs to be fixed to unblock CI.

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- (x) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

shyaamsn:
Fix package plugin tests for classy clusters. Classy clusters use packageInstall CR on management cluster for installing kapp-controller on workload cluster. Legacy clusters use App CR on management cluster for installing kapp-controller on workload cluster. In the test we were pausing App before patching kapp-controller configmap but in Classy clusters this was getting overridden since PackageInstall CR is not paused. This change makes sure to check if packageinstall is found then pause packageinstall else pause app. 

Fixes #4209 

Ran the test on a classy and legacy cluster and it passes

### Release note

```release-note
Fix package plugin tests for classy clusters
```

### DIFF
--- a/cmd/cli/plugin/package/test/package_plugin_integration_test.go
+++ b/cmd/cli/plugin/package/test/package_plugin_integration_test.go
@@ -183,6 +183,20 @@ var _ = Describe("Package plugin integration test", func() {
 
 		if !config.UseExistingCluster {
 			By(fmt.Sprintf("Creating management cluster %q", config.ClusterNameMC))
+
+			// CAPD cluster creation on classy clusters doesnt work. So need to create legacy cluster.
+			// Below hack enables legacy cluster creation.
+			hackCmd := exec.NewCommand(
+				exec.WithCommand("./scripts/legacy_hack.sh"),
+				exec.WithStdout(GinkgoWriter),
+			)
+
+			fmt.Println("Executing the legacy hack script")
+			_, _, err := hackCmd.Run(context.Background())
+			Expect(err).To(BeNil())
+
+			os.Setenv("_ALLOW_CALICO_ON_MANAGEMENT_CLUSTER", "true")
+
 			cli, err := tkgctl.New(tkgctl.Options{
 				ConfigDir: tkgCfgDir,
 				LogOptions: tkgctl.LoggingOptions{
@@ -202,6 +216,8 @@ var _ = Describe("Package plugin integration test", func() {
 			})
 			Expect(err).NotTo(HaveOccurred())
 
+			os.Unsetenv("_ALLOW_CALICO_ON_MANAGEMENT_CLUSTER")
+
 			By(fmt.Sprintf("Creating workload cluster %q", config.ClusterNameWLC))
 			tkgCtlClient, err := tkgctl.New(tkgctl.Options{
 				ConfigDir: tkgCfgDir,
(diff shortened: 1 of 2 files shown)


cmd/cli/plugin/package/test/package_plugin_integration_test.go
```diff
@@ -610,12 +628,33 @@ func pauseKappControllerPackage(clusterName, clusterNamespace, mgmtClusterKubeco
...
+	)
+
 	command := exec.NewCommand(
 		exec.WithCommand("kubectl"),
-		exec.WithArgs("patch", "app", fmt.Sprintf("%s-kapp-controller", clusterName), "-n", clusterNamespace, "--type", "merge", "-p", fmt.Sprintf("{\"spec\":{\"paused\":%t}}", pause), "--context", mgmtClusterKubeCtx, "--kubeconfig", mgmtClusterKubeconfigPath),
+		exec.WithArgs("get", "packageinstall", fmt.Sprintf("%s-kapp-controller", clusterName), "-n", clusterNamespace, "--context", mgmtClusterKubeCtx, "--kubeconfig", mgmtClusterKubeconfigPath),
 		exec.WithStdout(GinkgoWriter),
 	)
 	err := command.RunAndRedirectOutput(context.Background())
+	if err == nil {
```
maralavi, line 642: What if the error be something other than NotFound?

shyaamsn: Unfortunately the error returned by RunAndRedirectOutput is errors.WithStack and it doesnt return the actual error string. So cannot distinguish between NotFound and other error. The likelihood of other error happening is small as well as this is test code. So I think we can live with this for now and try to enhance it later.

praveenrewar: Since we have added the hack to create legacy clusters, maybe we can just skip adding this logic to check for PackageInstall or App for now?

maralavi: > Unfortunately the error returned by RunAndRedirectOutput is errors.WithStack and it doesnt return the actual error string. So cannot distinguish between NotFound and other error. The likelihood of other error happening is small as well as this is test code. So I think we can live with this for now and try to enhance it later.

I see, thanks for the clarification

maralavi (approved):

cmd/cli/plugin/package/test/package_plugin_integration_test.go
shyaamsn: > Since we have added the hack to create legacy clusters, maybe we can just skip adding this logic to check for PackageInstall or App for now?

Downstream tests run on classy clusters. So need to add this logic for packageinstall as well as app.

rajathagasthya (approved):

PR #4193, opened 2022-12-18
Fix clusterbootstrap continuous reconciliation

shyaamsn:
Fix clusterbootstrap continuous reconciliation.

Clusterbootstrap controller was continuously reconciling since the secret containing inline data values was continuously changing due to owner reference being overwritten.

This change ensures the owner reference is not overwritten.

Fixes #<TBC>

Tested cluster creation and ensured the continuous reconciliation is stopped

### Release note
```release-note
Fix clusterbootstrap continuous reconciliation
```


alfredthenarwhal:
> @shyaamsn:  /test install-vc7
Commit: aa158c0

Tests passed! Build no: 3639

maralavi (approved):

PR #4201, opened 2022-12-20
Suppress provider update if developer needs

### LINKED ISSUE #4187: Ytt overlay customization check should be skipped for yttcc and yttcb

xiujuanx:

For our internal development, sometimes, we need to update ytt/yttcc/yttcb template, and test if our changes work.
Also we don't want to build the whole things so that we can save more time for developing. In this situation, we can set SUPPRESS_PROVIDERS_UPDATE env var, and create legacy or classy cluster depends on yourself.

Fixes #4187 

### Release note

```release-note

```

### DIFF
--- a/tkg/client/common.go
+++ b/tkg/client/common.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"os"
 	"os/exec"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


tkg/client/common.go
```diff
@@ -364,11 +365,14 @@ func (c *TkgClient) ShouldDeployClusterClassBasedCluster(isManagementCluster boo
 		return true, nil
 	}
 
+	// Ignore the checksum check if config variable 'SUPPRESS_PROVIDERS_UPDATE' is set.
+	isSuppressProvidersUpdateSet := os.Getenv(constants.SuppressProvidersUpdate) != ""
```
12345lcr, line 369: May I ask why the `SUPPRESS_PROVIDERS_UPDATE` was required to set to number `1` instead of boolean or other values? Does this number contain any meaning?

xiujuanx: @12345lcr , no, `SUPPRESS_PROVIDERS_UPDATE` can be set to any string. As long as it is not an empty string, then `isSuppressProvidersUpdateSet` will be `true`.

12345lcr (approved):

wjun (approved):

imikushin (approved):

PR #4205, opened 2022-12-21
fix aws cluster hardcode ubuntu

m1zzx2:

Fixes #

### Release note

```release-note

```

### DIFF
--- a/providers/infrastructure-aws/v2.0.2/yttcc/overlay.yaml
+++ b/providers/infrastructure-aws/v2.0.2/yttcc/overlay.yaml
@@ -53,8 +53,7 @@ spec:
       replicas: #@ data.values.CONTROL_PLANE_MACHINE_COUNT
       metadata:
         annotations:
-          #! VVV TODO(vui) os-name handling
-          run.tanzu.vmware.com/resolve-os-image: #@ "ami-region={},os-name=ubuntu".format(data.values.AWS_REGION)
+          run.tanzu.vmware.com/resolve-os-image: #@ "ami-region={},os-name={},os-arch={}".format(data.values.AWS_REGION, data.values.OS_NAME, data.values.OS_ARCH)
     workers:
       machineDeployments:
       #@overlay/match by=overlay.index(0)


providers/infrastructure-aws/v2.0.2/yttcc/overlay.yaml
```diff
@@ -54,7 +54,7 @@ spec:
       metadata:
         annotations:
           #! VVV TODO(vui) os-name handling
```
srm09, line 56: consider removing the TODO, now that this is resolved

providers/infrastructure-aws/v2.0.2/yttcc/overlay.yaml
```diff
@@ -54,7 +54,7 @@ spec:
...
-          run.tanzu.vmware.com/resolve-os-image: #@ "ami-region={},os-name=ubuntu".format(data.values.AWS_REGION)
+          run.tanzu.vmware.com/resolve-os-image: #@ "ami-region={},os-name={},os-arch={}".format(data.values.AWS_REGION, data.values.OS_NAME, data.values.OS_ARCH)
```
tenczar, line 57: This works and is exactly what we do in the machine deployments below. I believe it is possible for one or more of these values to be empty which will cause errors on deployment because you'll get nonsense like `un.tanzu.vmware.com/resolve-os-image: ami-region=us-west-2,os-name=,os-arch=amd64` or something like that. We should try to update all of these to only include the key/value pair if the value exists.

tenczar (approved):
This can go in now and fix our current issue, but we should really address the issue from my comment.

providers/infrastructure-aws/v2.0.2/yttcc/overlay.yaml
tenczar: In fact all of these should probably be updated to include os-value. I think I tried adding that proactively back in ~June, but ran into the above issue because OS_VERSION hadn't been set.

providers/infrastructure-aws/v2.0.2/yttcc/overlay.yaml
m1zzx2: ok, removed

providers/infrastructure-aws/v2.0.2/yttcc/overlay.yaml
m1zzx2: just tried it.Default value will be added automatically if not set.

config:

CLUSTER_NAME: tkg-capz-mgmt-azure-11
CLUSTER_PLAN: dev
#OS_ARCH: amd64
INFRASTRUCTURE_PROVIDER: aws
OS_NAME: ''
OS_VERSION: ''
ENABLE_DEFAULT_STORAGE_CLASS: true

run: tanzu mc create -f xxx.yaml --dry-run:

the annotations will be:
run.tanzu.vmware.com/resolve-os-image: ami-region=us-west-2,os-name=ubuntu,os-arch=amd64


So can I merge it directly？

PR #4218, opened 2023-01-03
Add missing avi configurations for clusterclass based cluster

XudongLiuHarold:
Signed-off-by: Xudong Liu <xudongl@vmware.com>

- This PR added missing avi configurations(AVI_LABELS & AVI_NSXT_T1_LR) for clusterclass based clusters

Fixes #

- Unit test cases added
- Manually verified on vSphere env.

### Release note

```release-note

```

### DIFF
--- a/tkg/managementcomponents/helper.go
+++ b/tkg/managementcomponents/helper.go
@@ -105,9 +105,9 @@ func GetTKGPackageConfigValuesFileFromUserConfig(managementPackageVersion, addon
 		},
 	}
 
-	//Auto fill empty fields in AkoOperatorConfig
+	// fill in nsx advanced load balancer(a.k.a avi) config
 	if err := setAkoOperatorConfig(&tkgPackageConfig, userProviderConfigValues, onBootstrapCluster); err != nil {
-		return "", errors.Errorf("Error set ako operator config")
+		return "", err
 	}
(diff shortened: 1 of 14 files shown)


alfredthenarwhal:
Commit: aa5abc1

Tests failed! Build no: 3718

alfredthenarwhal:
Commit: 43ed2d2

Build failed! Build no: 3733

tkg/managementcomponents/helper.go
```diff
@@ -168,16 +168,38 @@ func convertToBool(config interface{}) bool {
...
 }
 
-// autofillAkoOperatorConfig autofills empty fields in AkoOperatorConfig
+// convertAVILabels converts config into string type, and return empty string if config is not set
+func convertAVILabels(config interface{}) (string, error) {
+	if config != nil {
+		switch config.(type) {
+		case string:
+			return config.(string), nil
+		default:
```
lubronzhan, line 177: Hm, would it always be string type and this will never be reached?
maybe we could just do val := config.(string)
And the runtime panic will happen if it's not string type

XudongLiuHarold: Actually not, it could be: 
- string `AVI_LABELS: '{"foo":"bar"}`, 
- map: 
```yaml
AVI_LABELS: |
    foo: bar
```
So it can be touched.

lubronzhan (approved):

PR #4225, opened 2023-01-03
Fix copyright headers in generated files

shyaamsn:
Fix copyright headers to set year to 2023. Without this make verify fails

Fixes #

### Release note

```release-note
Fix copyright headers to set year to 2023
```


miclettej (approved):

danniel1205:
should we update the copyright format to be `xxxx-2023` ?

apis/addonconfigs/cni/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,7 +1,7 @@
 //go:build !ignore_autogenerated
 // +build !ignore_autogenerated
 
-// Copyright 2022 VMware, Inc. All Rights Reserved.
+// Copyright 2023 VMware, Inc. All Rights Reserved.
```
danniel1205, line 4: question: this seems updated the year in generated file only, do we need to update the source file as well ?

shyaamsn: I think doing an audit of copyright headers in all the files and fixing it should be the right thing to do. That seems like a broader exercise.

The goal of this PR is to unblock the CI task `make verify` which unblocks all the PRs.

shyaamsn: I have created this issue: https://github.com/vmware-tanzu/tanzu-framework/issues/4227 to address fixing copyright headers in all source files. 

Resolving this comment now to unblock CI pipelines and developer PRs

PR #3992, opened 2022-11-24
Change the certificate struct for clusterclass, combining all certs into one encoding

lubronzhan:
…er annotation

### What this PR does / why we need it
Change the format of trust.cert to align with different products
```
name: trust
  value:
     additionalTrustedCAs:
         - name: CompanyInternalCA-1
           data: aGVsbG8=
        - name: CompanyInternalCA-2
           data: bHWtcH9=

```

Also when reading the cert and put into annotation, combine all the certs in one encoding
### Which issue(s) this PR fixes

Fixes #

### Release note

```release-note

```

PR #4228, opened 2023-01-04
Fix the certificate processing for clusterclass

lubronzhan:
Previously, we changed the `trust`  structure in https://github.com/vmware-tanzu/tanzu-framework/pull/3992.
But the logic of mapping from cluster variable to legacy variable was not changed, which we should.

Fixes #

### Release note

```release-note

```


alfredthenarwhal:
Commit: 01283f7

Tests passed! Build no: 3731

randomvariable (approved):

PR #2690, opened 2022-06-17
Sync remaining changes from package-based-lcm branch

vuil:
### What this PR does / why we need it
This PR introduces the remaining set of changes from the `package-based-lcm branch`
Includes
- installing core-management-plugins-package as part of TKG composite package
- CCluster.yaml config validation
- CC-side fixes
- Conversion of user provided values into correct types during CC package installationo
- Introduce support for a compatible-tkr config map to preload a bootstrap cluster for MC creation

### Which issue(s) this PR fixes

Fixes #2597, 
Fixes #2589, 
Related to #2592

### Describe testing done for PR

UT ran on individual commits.
Will verify via CI pipelines and triggered downstream tests.

### Release note

```release-note

```

### PR Checklist

- x Squash the commits into one or a small number of logical commits
      
- x Use good commit [messages](https://github.com/vmware-tanzu/tanzu-framework/blob/main/CONTRIBUTING.md)
- x Ensure PR contains terms all contributors can understand and links all contributors can access

PR #4230, opened 2023-01-04
drop ligatures from various manifests

yastij:
Signed-off-by: Yassine TIJANI <ytijani@vmware.com>

This PR drops ligatures from various manifests introduced in https://github.com/vmware-tanzu/tanzu-framework/pull/2690 and , this actually broke cloud-init, specifically Jinja templates rendering in amazon linux 2  (ligatures should only be visual and shouldn't be committed)

Fixes #

### Release note

```release-note

```

### DIFF
--- a/packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml
+++ b/packages/tkg-clusterclass-aws/bundle/config/upstream/base.yaml
@@ -1385,7 +1385,7 @@ spec:
             default file
         preKubeadmCommands:
         #! Ensure a non-empty preKubeadmCommands array is present to facilitate patching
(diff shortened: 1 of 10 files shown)


randomvariable (approved):

randomvariable:
i'm yoloing this one. i just can't see how swapping out ligatures with ... should fail.

randomvariable:
For future reference, we believe the issue is actually in Amazon Linux 2's CentOS 7 base.

Ankitasw:
> For future reference, we believe the issue is actually in Amazon Linux 2's CentOS 7 base.

@randomvariable is it the issue with CAPA AMI or base AMI?

PR #4232, opened 2023-01-04
Skip CLIPlugin CRD installation as part of core-management-plugins package

anujc25:

It was observed that `cliplugins` package reconciliation fails because of the race condition on installing CLIPlugins CRD by `cliplugin` and `core-management-plugins` packages. 

- `core-management-plugins` package should not install the CRD if it is getting deployed through TKG composite package.

Currently, the `values.yaml` file of the TKG composite package specifies `deployCLIPluginCRD: false` however, the same thing is not configured programmatically and hence the `core-management-plugins` package is also trying to deploy that CRD. This PR fixes that logic.

Fixes #

### Release note

```release-note
Skip CLIPlugin CRD installation as part of core-management-plugins package
```


randomvariable:
Thanks

randomvariable (approved):

PR #4243, opened 2023-01-05
refactor retrying code on azure upgrade and create

996refuse:
This PR address tolerating some failures when creating or upgrading an azure cluster, because the cluster status may return 'context deadline exceeded' or 'context canceled' error which is recoverable from capz side.

### DIFF
--- a/tkg/clusterclient/clusterclient.go
+++ b/tkg/clusterclient/clusterclient.go
@@ -581,7 +581,7 @@ func (c *client) WaitForClusterInitialized(clusterName, namespace string) error 
 	maxTimeout := 3 * c.operationTimeout
 	maxTimeoutCounter := 0
 	errorRetry := 0
-	maxErrorRetry := 3
+	maxErrorRetry := 20
 
 	getterFunc := func() (interface{}, error) {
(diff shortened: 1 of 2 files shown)


sonasingh46:
Do we have a check for delete? Please add the same timeout logic in the delete path also.

996refuse:
@sonasingh46 correct me if I'm wrong, deletion function **WaitForClusterDeletion** is different, no need to check the status of the clusters, it returns success if it can't get the cluster resource anymore.

sonasingh46:
@996refuse -- In that case, that is fine.

alfredthenarwhal:
> @yastij:  /test upgrade-azure
Commit: 9086dae

Tests failed! Build no: 3763

alfredthenarwhal:
> @wjun:  /test upgrade-azure
Commit: 9086dae

Tests failed! Build no: 3767

alfredthenarwhal:
> @wjun:  /test upgrade-azure
Commit: 9086dae

Tests failed! Build no: 3771

wjun (commented):
LGTM. Given we have passed bolt upgrade azure tests, we will ignore the capz cc pipeline failure which seems not related to this change.

wjun (approved):

PR #4245, opened 2023-01-05
Trim input values for `tanzu context create`

chandrareddyp:
This PR is to fix a bug in the command `tanzu context create` , trims leading and tailing spaces of input values for the command

Fixes #NA

Testing has done to validate the fix

### Release note

```release-note

```

### DIFF
--- a/cli/core/pkg/command/context.go
+++ b/cli/core/pkg/command/context.go
@@ -204,6 +204,7 @@ func createContextWithKubeconfig() (context *configapi.Context, err error) {
 	} else if kubeConfig == "" {
 		kubeConfig = getDefaultKubeconfigPath()
 	}
+	kubeConfig = strings.TrimSpace(kubeConfig)
 
 	if kubeConfig != "" && kubeContext == "" {
 		err = component.Prompt(
@@ -217,6 +218,7 @@ func createContextWithKubeconfig() (context *configapi.Context, err error) {
 			return
 		}
 	}
+	kubeContext = strings.TrimSpace(kubeContext)
 
 	if ctxName == "" {
 		err = component.Prompt(
@@ -230,6 +232,7 @@ func createContextWithKubeconfig() (context *configapi.Context, err error) {
 			return
 		}
 	}
+	ctxName = strings.TrimSpace(ctxName)
 	exists, err := config.ContextExists(ctxName)
 	if err != nil {
 		return context, err
@@ -266,7 +269,7 @@ func createContextWithEndpoint() (context *configapi.Context, err error) {
 			return
 		}
 	}
-
+	endpoint = strings.TrimSpace(endpoint)
 	if ctxName == "" {
 		err = component.Prompt(
 			&component.PromptConfig{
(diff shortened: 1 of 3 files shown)


cli/core/pkg/command/context.go
```diff
@@ -240,11 +240,11 @@ func createContextWithKubeconfig() (context *configapi.Context, err error) {
 	}
 
 	context = &configapi.Context{
-		Name:   ctxName,
+		Name:   strings.Trim(ctxName," "),
```
marckhouzam, line 243: I think if we do a trim, it should be closer to where the variable was set.  For example `ctxName` is used above to check if the context already exists and it will not work properly if there are spaces around the context name.

Also, there is a `strings.TrimSpace()` which may be a little nicer to use.

cli/core/pkg/command/context.go
```diff
@@ -240,11 +240,11 @@ func createContextWithKubeconfig() (context *configapi.Context, err error) {
...
 		Target: cliv1alpha1.TargetK8s,
 		ClusterOpts: &configapi.ClusterServer{
-			Path:                kubeConfig,
-			Context:             kubeContext,
+			Path:                strings.Trim(kubeConfig," "),
+			Context:             strings.Trim(kubeContext," "),
 			Endpoint:            endpoint,
 			IsManagementCluster: true,
 		},
```
marckhouzam, line 250: Technically, this problem is not limited to this function but will affect (slightly differently) `createContextWithEndpoint()` also.

marckhouzam (changes requested):
Thanks @chandrareddyp.

I think you have uncovered a more general problem about using `component.Prompt`.
The prompt function returns whatever the user typed, including spaces.  I agree that this can be confusing to users so fixing it for these situations seems like a good idea.  I think this PR is a good start and is sufficient in itself.  However, what do you say about opening an issue to document this problem in general so that maybe some day it can be fixed for other scenarios that use `component.Prompt`?

As a side note, IMHO we should not fix this directly in the `component.Prompt` logic because there may be some scenarios where a user will *want* to add a space as part of the input.  We should only fix it when we understand the input and feel there is no reason to have spaces in it, like you are doing in this PR 👍 .

cli/core/pkg/command/context.go
chandrareddyp: agreed, trimmed all input values after reading.

chandrareddyp:
> Thanks @chandrareddyp.
> 
> I think you have uncovered a more general problem about using `component.Prompt`. The prompt function returns whatever the user typed, including spaces. I agree that this can be confusing to users so fixing it for these situations seems like a good idea. I think this PR is a good start and is sufficient in itself. However, what do you say about opening an issue to document this problem in general so that maybe some day it can be fixed for other scenarios that use `component.Prompt`?
> 
> As a side note, IMHO we should not fix this directly in the `component.Prompt` logic because there may be some scenarios where a user will _want_ to add a space as part of the input. We should only fix it when we understand the input and feel there is no reason to have spaces in it, like you are doing in this PR 👍 .

as discussed, I have updated the `component.Prompt` API documentation saying `"Prompt for input, reads input value, without trimming any characters (may include leading/tailing spaces)"`, and its end-user responsibility to take care about input value trimming.

marckhouzam (approved):
Thanks for the update @chandrareddyp.
Assuming CI passes, this LGTM

anujc25 (approved):
LGTM. Thanks for fixing this.

PR #4250, opened 2023-01-06
Azure: Use ManualServicePrincipal instead of ServicePrincipal

randomvariable:

This avoids an indirection where CAPZ requests credentials via IMDS via the deprecated NMI service that injects a DNAT rule into the pod which can interfere with CNIs.

Fixes #

### Release note

```release-note
Deprecated Azure AD Pod Identity service is no longer used for authentication of Cluster API Azure on new clusters.
```

### DIFF
--- a/providers/infrastructure-azure/v1.5.3/ytt/overlay.yaml
+++ b/providers/infrastructure-azure/v1.5.3/ytt/overlay.yaml
@@ -357,5 +357,5 @@ spec:
     name: #@ "{}-identity-secret".format(data.values.CLUSTER_NAME)
     namespace: tkg-system
   tenantID: #@ data.values.AZURE_TENANT_ID
-  type: ServicePrincipal
+  type: ManualServicePrincipal
 #@ end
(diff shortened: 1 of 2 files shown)


yastij (approved):

alfredthenarwhal:
> @randomvariable:  /test upgrade-azure
Commit: c60e3ac

Build failed! Build no: 3781

zjs:
> Deprecated Azure AD Pod Identity service is no longer used for authentication of Cluster API Azure on new clusters.

Is there something that needs to be done to shift existing clusters to use ManualServicePrincipal instead of ServicePrincipal. (Sorry, didn't dig into the details here; just noticed "on new clusters" and wondered about existing clusters.)

randomvariable:
@zjs Oddly, the AzureClusterIdentity CR is mutable (it probably shouldn't be, but we'll take it for now), so we think it's going to work. The upgrade pipeline should tell us.

randomvariable:
Build failed. Will be retrying downstream.

CAPZ integration tests fail because it's using the wrong TKR apparently? Not sure how to change it.

randomvariable:
Initial downstream test shows no real improvement, but not enough logs to tell me what happened. We can probably deprioritise this one, but will retry.

randomvariable:
Test run results:
Both Upgrade and Install jobs, except for Deletion where we know that's due to incorrect tests.

I did temporarily still see StatusCode=0 erros in the Install job in the kind cluster, but once the management cluster was up and running, there were 0 log errors for CAPZ beyond normal reconciliation waits. It does seem like networking was the root cause of these, and this does mark an improvement.

sonasingh46 (approved):

PR #4253, opened 2023-01-06
Install CLIPlugin CRD through core-management-plugins package

### LINKED ISSUE #4258: Race condition between cliplugins and core-management-plugins packages causes management cluster bootstrap to fail in high speed environments
**Bug description**

If the connection between the bootstrap cluster and the source registry is sufficiently fast OR the images are cached on the bootstrap cluster, then
running `tanzu management-cluster create` leads to failures in cliplugins package with

```
Ownership errors:
 - Resource 'customresourcedefinition/cliplugins.cli.tanzu.vmware.com (apiextensions.k8s.io/v1) cluster' is already associated with a different label 'kapp.k14s.io/app=1671445784745865472'}}
```
Both packages include the CRD

To reproduce this, do the following:

Run `tanzu management-cluster create` against a kind cluster, and exit just before the cluster is created.

Then run the following, to clean up the kind cluster, but leave OCI images cached:
```
kubectl delete cluster --all -A
kubectl delete -A clusterclasses.cluster.x-k8s.io --all
kubectl delete -A pkgi --all
kubectl delete packagerepositories.packaging.carvel.dev -n tkg-system tanzu-management
kubectl delete cliplugins -A --all
kubectl delete crd cliplugins.cli.tanzu.vmware.com
```
Then run create again. The race condition then occurs.

In a separate upgrade scenario, we have noticed that once the management-cluster upgrade is successful, the CLIPlugin resources are not available for a brief period of time.

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

anujc25:

- Install CLIPlugin CRD through core-management-plugins package
- Make installation of CLIPlugin CRD through cliplugin package conditional
- When installing packages through TKG composite package, the value of `deployCLIPluginCRD` will be set to `false`
- However, the default value for `deployCLIPluginCRD` within the cliplugin package is `true`

Fixes #4258

- Manual testing was done by running the E2E test for cluster create and upgrade
- Automated test result/logs was verified to see no warning message are displayed now.

### Release note

```release-note
Install CLIPlugin CRD through core-management-plugins package
```


tkg/managementcomponents/types.go
```diff
@@ -79,7 +79,6 @@ type CoreManagementPluginsPackage struct {
 }
 
 type CoreManagementPluginsPackageValue struct {
```
12345lcr, line 81: Do we still need this struct? Since it will be empty after DeployCLIPluginCRD is removed

12345lcr (approved):
LGTM, thanks Anuj!

vuil (approved):
lgtm, thanks for the fix!

prkalle (approved):
LGTM.Thanks!

Issue #4265, opened 2023-01-10
InitRegion is long, lots of side effects


jayunit100:
**Bug description**

The InitRegion function has alot of side effects and is like 500 lines long.  
lets split the code up and try to get rid of global func variables so its easy to debug/add logs for stages and so on, as its one of the most common things that breaks in TKG installation. 

**Affected product area (please put an X in all that apply)**

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

**Steps to reproduce the bug**

**Version** (include the SHA if the version is not obvious)

**Environment where the bug was observed (cloud, OS, etc)**

**Relevant Debug Output (Logs, manifests, etc)**

PR #4266, opened 2023-01-10
Removing global vars and improving logs during complex bootstrapping [WIP]

jayunit100:
Fix to #4265

### DIFF
--- a/tkg/tkgctl/helpers.go
+++ b/tkg/tkgctl/helpers.go
@@ -214,6 +214,9 @@ func getFieldfromUnstructuredObject(obj unstructured.Unstructured, fields ...str
 }
 
 func setVSphereCredentialFromInputfile(legacyVarMap *map[string]string, clusterConfigFile, clusterName, namespace string) error {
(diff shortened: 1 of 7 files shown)


randomvariable:
Looking good so far but we should get feedback from PM/UX about what terminology to use for end users about classy vs legacy.

tkg/tkgctl/helpers.go
```diff
@@ -214,6 +214,9 @@ func getFieldfromUnstructuredObject(obj unstructured.Unstructured, fields ...str
 }
 
 func setVSphereCredentialFromInputfile(legacyVarMap *map[string]string, clusterConfigFile, clusterName, namespace string) error {
+	if clusterName == "" {
```
jayunit100, line 217: fwiw, this happened to me and indira the other day and we got confused :) if the name is empty you otherwise get an ugly lower level failure

jayunit100:
closing bc i cant understand these tests results will reopen

PR #4271, opened 2023-01-11
TKR Resolver: exact TKR version match (if exists) should be the only result

### LINKED ISSUE #4262: TKR resolver does not resolve a TKR by exact version if there are TKRs with that version as their version prefix
**Bug description**
If 2 TKrs exist with the same prefix, and the cluster is set to use TKr with `v1.22.13+vmware.1-tkg.1`, TKr resolver resolves it to v1.22.13+vmware.1-tkg.1-zshippable.

```
v1.22.13---vmware.1-tkg.1              v1.22.13+vmware.1-tkg.1              True    True         6d11h
v1.22.13---vmware.1-tkg.1-zshippable   v1.22.13+vmware.1-tkg.1-zshippable   True    True         6d11h

```

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**
When  Cluster.spec.topology.version is set with a specific TKr version, TKr resolver should use the existing TKr with the
same name/version.
It should have resolved the TKr to v1.22.13---vmware.1-tkg.1  instead

1. Create 2 TKr with the same prefix. 
E.g v1.22.13---vmware.1-tkg.1    and v1.22.13---vmware.1-tkg.1-zshippable
2.  Create cluster with TKr v1.22.13---vmware.1-tkg.1 by setting Cluster.spec.topology.version
3. TKr resolver should resolve this to TKr v1.22.13---vmware.1-tkg.1 

ST environment vSPhere on Supervisor

imikushin:

When the exact TKR version is specified as the version prefix AND such TKR exists, it should be the only result in the returned resolution result set.

This fixes a problem, where there are TKRs whose versions are _longer_ than the provided exactly matching TKR version.

For example, the provided prefix is `v1.22.13+vmware.1-tkg.1`. There are two TKRs with versions satisfying this version prefix:
- `v1.22.13+vmware.1-tkg.1`,
- `v1.22.13+vmware.1-tkg.1-zshippable`.

Of the above two, `v1.22.13+vmware.1-tkg.1` must be used, because it is the exact match, and is least surprising for the user specifying `v1.22.13+vmware.1-tkg.1` as the version prefix.

Fixes #4262 

Updated unit tests with the test case exposing the issue.

### Release note

```release-note
Fixed a problem when resolving TKRs, where there are TKRs whose versions are longer 
than the matching TKR's version.

For example, the provided version prefix is `v1.22.13+vmware.1-tkg.1`. There are two TKRs 
with versions satisfying this prefix:
- `v1.22.13+vmware.1-tkg.1`,
- `v1.22.13+vmware.1-tkg.1-zshippable`.

Of the above, `v1.22.13+vmware.1-tkg.1` will now be returned, because it is the exact match. 
When the version prefix is shorter (e.g. `v1.22.13+vmware.1`), the result will still be 
`v1.22.13+vmware.1-tkg.1-zshippable` because it is considered a "more recent" version from 
the alpha-numerical comparison point of view.
```

### DIFF
--- a/tkr/resolver/internal/resolver_impl.go
+++ b/tkr/resolver/internal/resolver_impl.go
@@ -324,12 +324,21 @@ func (cache *cache) filterOSImageDetails(osImageQuery *data.OSImageQuery) *osIma
 	}
 
 	consideredTKRs := cache.consideredTKRs(*osImageQuery)
+	consideredTKRs = possiblyNarrowToExactMatch(*osImageQuery, consideredTKRs)
 	filteredOSImagesByTKR := cache.filterOSImagesByTKR(*osImageQuery, consideredTKRs)
 
 	return &osImageDetails{
 		tkrs:          filterTKRsWithOSImages(filteredOSImagesByTKR, consideredTKRs),
 		osImagesByTKR: filteredOSImagesByTKR,
 	}
+}
+
+func possiblyNarrowToExactMatch(osImageQuery data.OSImageQuery, consideredTKRs data.TKRs) data.TKRs {
+	name := version.Label(osImageQuery.K8sVersionPrefix)
+	if tkr := consideredTKRs[name]; tkr != nil {
+		return data.TKRs{name: tkr}
+	}
+	return consideredTKRs
 }
 
 // consideredTKRs returns the initial set of TKRs satisfying the query.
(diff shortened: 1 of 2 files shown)


prkalle (approved):
LGTM. Thanks @imikushin!

PR #4272, opened 2023-01-11
fix issue with querying for GVR in capabilities discovery package

yharish991:

When querying for existence of GVR without version, capabilities discovery package returns false even though the resource exists. This PR is for fixing that issue.

Fixes #

Verified that the units tests are passing and wrote a small sample program to test GVR query functionality

### Release note

```release-note
Fixed query GVR functionality when a version is not provided in capabilities discovery package.
```


codegold79 (approved):
Thanks for finding this bug and fixing it so quickly, @yharish991.

danniel1205 (approved):
the changes lgtm, adding my approval. (But we still need to cleanup before merge)

knabben (approved):

rajathagasthya (approved):
LGTM, thanks! Please add some tests in a follow up PR.

PR #3926, opened 2022-11-16
VSphere Cluster CC/Template: Use local_hostname instead of FQDN for node name

randomvariable:
K8s resources are limited to 63 characters, so to ensure nodes can
join, they should not include the complete FQDN and instead be
limited only to the local hostname.

Signed-off-by: Naadir Jeewa <jeewan@vmware.com>

### Which issue(s) this PR fixes

Fixes #

### Release note

```release-note
In all vSphere clusterclasses, we now use local_hostname instead of FQDN for node registration to avoid hitting 63 character limit.
```

PR #4276, opened 2023-01-12
Revert "VSphere Cluster CC/Template: Use local_hostname instead of FQDN for node name (#3926)

XudongLiuHarold:
This reverts commit 22d2710b2400fe5a3554489493e0b1477e3fc5b3.

This PR reverts #3926 because we found that sometimes the DNS name is not equivalent to the node name in vSphere env. 
CPI use node name as DNS name to find VM at the beginning. Because it think the node name is the identifier.
It doesn't use node name as VM name to fetch VM because it needs to support multi-VC case, while VM name could collide in different VC.

Fixes #

### Release note

```release-note

```


alfredthenarwhal:
Commit: 9601cde

Tests failed! Build no: 3843

randomvariable (approved):

PR #4287, opened 2023-01-17
make 'go mod vendor' idempotent, add required providers.zip

navidshaikh:
- Adds `tkg/tkgctl/client/manifest/providers.zip` file required for `go mod vendor` to pass
- How? (all *.zip are ignored in .gitignore file)
      - update .gitignore to allow adding `tkg/tkgctl/client/manifest/providers.zip` (explicitly) while ignoring all other `providers.zip`
      - add `tkg/tkgctl/client/manifest/providers.zip` file, commit it
      - revert `.gitignore` changes in step 1 and ignore all subsequent changes to providers.zip files
 - Verify `go mod vendor` check in build step of CI passes 
    
Fixes #

### Release note

```release-note

```


DanielXiao:
Why do you add a providers.zip to source code? If this is just for CI, you can run a command to make the zip bundle.

alfredthenarwhal:
> @navidshaikh:  /test install-vc7
Commit: 186ff21

Tests passed! Build no: 3874

vuil:
nit: If the goal is just to make 'go mod vendor' complete cleaning. An empty file for providers.zip would do as well.

navidshaikh:
thanks for review!, handling the fix elsewhere in the pipeline, this change is not needed anymore.

PR #4296, opened 2023-01-26
Update the image path of kube-vip

lubronzhan:
Currently, kube-vip imageRepository was fetched from `TKR_DATA.value.(k8s-version).kubernetesSpec.kube-vip.imageRepository`
```
- name: TKR_DATA
        value:
          v1.24.9+vmware.1:
            kubernetesSpec:
              coredns:
                imageRepository: projects-stg.registry.vmware.com/tkg
                imageTag: v1.8.6_vmware.15
              etcd:
                imageRepository: projects-stg.registry.vmware.com/tkg
                imageTag: v3.5.6_vmware.3
              imageRepository: projects-stg.registry.vmware.com/tkg   
              kube-vip:
                imageRepository: projects-stg.registry.vmware.com/tkg  <- ------ here   
                imageTag: v0.5.7_vmware.1
              pause:
                imageRepository: projects-stg.registry.vmware.com/tkg
                imageTag: "3.7"
              version: v1.24.9+vmware.1
```
But this value won't be there if `TKG_CUSTOM_IMAGE_REPOSITORY ` is not set
```
              coredns:
                imageTag: v1.8.6_vmware.15
              etcd:
                imageTag: v3.5.6_vmware.3
              imageRepository: projects.registry.vmware.com/tkg
              kube-vip:
                imageTag: v0.5.7_vmware.1
              pause:
                imageTag: "3.7"
              version: v1.24.9+vmware.1
            labels:
```
Since these two values will always be the same, kube-vip should also extract imageRepository from top level

Fixes #

### Release note

```release-note

```

### DIFF
--- a/providers/infrastructure-vsphere/v1.5.1/cconly/base.yaml
+++ b/providers/infrastructure-vsphere/v1.5.1/cconly/base.yaml
@@ -1038,7 +1038,7 @@ spec:
                     value: "20"
                   - name: vip_retryperiod
                     value: "4"
-                  image: {{(index (index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec "kube-vip").imageRepository}}/kube-vip:{{(index (index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec "kube-vip").imageTag}}
+                  image: {{(index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec.imageRepository}}/kube-vip:{{(index (index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec "kube-vip").imageTag}}
                   imagePullPolicy: IfNotPresent
                   name: kube-vip
                   resources: {}
--- a/packages/tkg-clusterclass-vsphere/bundle/config/upstream/base.yaml
+++ b/packages/tkg-clusterclass-vsphere/bundle/config/upstream/base.yaml
@@ -1038,7 +1038,7 @@ spec:
                     value: "20"
                   - name: vip_retryperiod
                     value: "4"
-                  image: {{ if (index (index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec "kube-vip").imageRepository }}{{(index (index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec "kube-vip").imageRepository}}{{ else }}{{(index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec.imageRepository}}{{ end }}/kube-vip:{{(index (index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec "kube-vip").imageTag}}
+                  image: {{(index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec.imageRepository}}/kube-vip:{{(index (index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec "kube-vip").imageTag}}
                   imagePullPolicy: IfNotPresent
                   name: kube-vip
                   resources: {}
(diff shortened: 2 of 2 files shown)


providers/infrastructure-vsphere/v1.5.1/cconly/base.yaml
```diff
@@ -1038,7 +1038,7 @@ spec:
                     value: "20"
                   - name: vip_retryperiod
                     value: "4"
-                  image: {{(index (index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec "kube-vip").imageRepository}}/kube-vip:{{(index (index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec "kube-vip").imageTag}}
+                  image: {{(index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec.imageRepository}}/kube-vip:{{(index (index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec "kube-vip").imageTag}}
```
zjs, line 1041: I think the ideal behavior would be:
1. Check for `TKR_DATA.value.(k8s-version).kubernetesSpec.kube-vip.imageRepository`. If set, use that.
2. Otherwise, use `TKR_DATA.value.(k8s-version).kubernetesSpec.imageRepository`.

We missed a scenario by always doing (1). I worry we may miss some other scenario by always doing (2).

(I don't _know of_ any such scenario though; I just worry.)

Is the extra complexity of the conditional logic worth considering?

alfredthenarwhal:
Commit: 04c6544

Tests failed! Build no: 3939

providers/infrastructure-vsphere/v1.5.1/cconly/base.yaml
srm09: I will try to look at the behavior of the other images and see how they handle it, but totally 💯 to checking if `kube-vip` has that value set, then use that otherwise use the top level image repository info.

lubronzhan: This would be the new sentence
```
image: {{ if (index (index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec "kube-vip").imageRepository }}{{(index (index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec "kube-vip").imageRepository}}{{ else }}{{(index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec.imageRepository}}{{ end }}/kube-vip:{{(index (index .TKR_DATA .builtin.controlPlane.version).kubernetesSpec "kube-vip").imageTag}}
```
Readability is bad. I think we could get confirmation from @ivan whether these <component>.imageRepository are really needed

srm09: https://github.com/vmware-tanzu/tanzu-framework/blob/949c6d04ba7496537b4192fc129f65fe0b1b2a3d/providers/infrastructure-vsphere/v1.5.1/cconly/base.yaml#L951-L970
None of the above do the conditional check

zjs: Good point, @srm09. If we're not doing the conditional check in other places, perhaps for now we should consistently not do the conditional check.

If there's potentially a use case for conditional logic we could explore that separately/holistically.

zjs: Idea from @imikushin: Instead of putting the conditional logic here on the "consumer" side, teach the the tkr webhook to cascade the top-level {{imageRepository}} value down to each component when its unset at the component level.

That seems like a good, consistent medium-term fix.

zjs (approved):

providers/infrastructure-vsphere/v1.5.1/cconly/base.yaml
imikushin: See #4298 for the above.

imikushin (approved):
LGTM.

PR #4306, opened 2023-01-30
raise error for tkr controller if image path is empty

hongtaiwang:
to avoid stack track error when calling `ListImageTags`

Fixes #

### Release note

```release-note

```

### DIFF
--- a/tkr/controller/tkr-source/main.go
+++ b/tkr/controller/tkr-source/main.go
@@ -71,6 +71,20 @@ func init() {
 	flag.IntVar(&continuousTKRDiscoverFreq, "continuous-discover-frequency", 600, "Continuous TKR discovery frequency in seconds")
 	flag.Parse()
 
+	// validate image path vars
+	if bomImagePath == "" {
+		setupLog.Error(fmt.Errorf("unable to set up controller manager"), "bom image path should not be empty")
+		os.Exit(1)
+	}
+	if bomMetadataImagePath == "" {
+		setupLog.Error(fmt.Errorf("unable to set up controller manager"), "bom metadata image path should not be empty")
+		os.Exit(1)
+	}
+	if tkrRepoImagePath == "" {
+		setupLog.Error(fmt.Errorf("unable to set up controller manager"), "tkr repo image path should not be empty")
+		os.Exit(1)
+	}
+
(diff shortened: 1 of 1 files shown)


tkr/controller/tkr-source/fetcher/fetcher.go
```diff
@@ -176,6 +176,9 @@ func (f *Fetcher) saveTKRCompatibilityCM(ctx context.Context, ns string, metadat
 }
 
 func (f *Fetcher) fetchCompatibilityMetadata() (*tkrv1.CompatibilityMetadata, error) {
+	if f.Config.BOMMetadataImagePath == "" {
+		return nil, errors.New("compatibility metadata image path should not be empty")
+	}
```
imikushin, line 181: Can we insert this check _before_ starting the controller? It feels like, at this point in the code, it's already too late.

So, maybe in main.go where we just got these config values and are creating Config objects.

tkr/controller/tkr-source/fetcher/fetcher.go
```diff
@@ -231,6 +234,10 @@ func (f *Fetcher) fetchTKRBOMConfigMaps(ctx context.Context) error {
 		return err
 	}
 
+	if f.Config.BOMImagePath == "" {
+		return errors.New("bom image should not be empty")
+	}
+
```
imikushin, line 240: Same.

tkr/controller/tkr-source/fetcher/fetcher.go
```diff
@@ -334,6 +341,10 @@ func (f *Fetcher) fetchTKRPackages(ctx context.Context) error {
 		return err
 	}
 
+	if f.Config.TKRRepoImagePath == "" {
+		return errors.Wrap(err, "failed to list current available TKR Package Repository image tags, should not be empty")
+	}
+
```
imikushin, line 347: Same.

tkr/controller/tkr-source/fetcher/fetcher.go
hongtaiwang: sure. updated

imikushin (approved):
LGTM. Thanks @hongtaiwang !

PR #4310, opened 2023-01-31
Set proxy env variable where tanzu cli is running, because now instal…

lubronzhan:
…ling KAPP requires talking to repository now

Right now with CC, need to install KAPP controller by pulling the source code from repository directly. When customer has env that needs proxy to access repository, it will fail
```
Kapp-controller values-file: /tmp/650094154.yaml
Deleting kind cluster: tkg-kind-cfc50dmlcgd73h3u8ktg
Error: unable to set up management cluster: unable to install kapp-controller to bootstrap cluster: failed to get resource files from discovery: error downloading bundle: Checking if image is bundle: Fetching image: Get "https://projects.registry.vmware.com/v2/": dial tcp 10.188.25.227:443: connect: connection refused
```

So proxy environment variable is needed. In CLI

Added code to set proxy variable if presented in customer's config. When creating or upgrading management cluster, `configureVariablesForProvidersInstallation` will be called, these environmement variables will be set

Fixes #

### Release note

```release-note

```

### DIFF
--- a/tkg/client/upgrade_region.go
+++ b/tkg/client/upgrade_region.go
@@ -6,6 +6,7 @@ package client
 import (
 	"context"
 	"fmt"
+	"os"
 	"strconv"
 	"strings"
 	"time"
@@ -317,6 +318,9 @@ func (c *TkgClient) configureVariablesForProvidersInstallation(regionalClusterCl
 		// no variable configuration is needed to deploy Docker provider as
 		// infrastructure-components.yaml for docker does not require any variable
 	}
+
+	c.SetProxyEnvironmentVariable()
+
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


12345lcr:
Hey @lubronzhan , thanks for the fix. I have 3 questions that I am not very clear.
1. Is this func similar to `setupProxyConfigurationForKindCluster`? Can we reuse this?
    https://github.com/vmware-tanzu/tanzu-framework/blob/e897c6b988dae116180d39e3f16891a3f897263e/tkg/kind/client.go#L445-L478
2. I don't understand why the management cluster creation was failing before this fix, because this `setupProxyConfigurationForKindCluster` should handle the proxy env settings during kind cluster creation, right?
3. `configureVariablesForProvidersInstallation` returned right after `configureImageTagsForProviderInstallation` during cluster creation(empty mgmt client). So this `SetProxyEnvironmentVariable` will not be executed for cluster creation, only upgrade is covered. Is this expected?

lubronzhan:
Thanks @12345lcr . let me check what does `setupProxyConfigurationForKindCluster ` do
Fo question 3, 
`configureVariablesForProvidersInstallation` is executed right before installing kapp. It's only needed when installing kapp for management cluster and bootstrapcluster. so no need to execute before workload cluster creation
https://github.com/vmware-tanzu/tanzu-framework/blob/main/tkg/client/init.go#L211

tkg/client/upgrade_region.go
```diff
@@ -317,6 +318,9 @@ func (c *TkgClient) configureVariablesForProvidersInstallation(regionalClusterCl
 		// no variable configuration is needed to deploy Docker provider as
 		// infrastructure-components.yaml for docker does not require any variable
 	}
+
+	c.SetProxyEnvironmentVariable()
```
12345lcr, line 322: Thanks @lubronzhan , for the question 3 above what I meant is I think this line will not be called during management cluster creation, because `configureVariablesForProvidersInstallation` will be returned above in here
https://github.com/vmware-tanzu/tanzu-framework/blob/b6c690beafcb3269ce348f7336b9f8a2ded80bde/tkg/client/upgrade_region.go#L276-L278

Since `configureVariablesForProvidersInstallation` is shared by management cluster creation and upgrade, during creation the client is `nil`

https://github.com/vmware-tanzu/tanzu-framework/blob/b6c690beafcb3269ce348f7336b9f8a2ded80bde/tkg/client/init.go#L211

lubronzhan:
Ok got it thanks. I thought that was for management config validation. looks like it should be [here](https://github.com/vmware-tanzu/tanzu-framework/blob/main/tkg/tkgctl/init.go#L176) 
Let me first figure out why the code you showed doesn't work.
BTW, do you know any faster way to compile CLI locally?
I saw this note https://github.com/vmware-tanzu/tanzu-framework/blob/9dd14f9a58e4061413f61fe005cca305c6521bb3/docs/oracle-mvp.md#build-tanzu-cli
But it requires setting all the management repository variables, not sure are those needed.

12345lcr:
Hi @lubronzhan , `make build-install-cli-local` directly should be enough for cli development. Other things are not required. The built cli is in the path like `artifacts/linux/amd64/cli/core/<version>/tanzu-core-linux_amd64`

FYI Unless you need to touch the meta packages component, you can extract the result of pr pipeline `TKG Integration Tests ClusterClass / Build and Publish Carvel Packages (pull_request) ` to use the packages built based on your pr changes
```
export _MANAGEMENT_PACKAGE_REPO_IMAGE=gcr.io/eminent-nation-87317/tanzu_framework/github-actions/3865/management:v0.21.0
export _ADDONS_MANAGER_PACKAGE_VERSION=v0.21.0
```

And if you need to use another version of compatibility file(tkg-bom, tkr-bom) other than the default in `projects-stg.registry.vmware.com/tkg/framework-zshippable/tkg-compatibility`, you can export your specific version of compatibility image 
```
export TKG_CUSTOM_IMAGE_REPOSITORY='projects-stg.registry.vmware.com/tkg'
export TKG_CUSTOM_COMPATIBILITY_IMAGE_PATH='framework-zshippable/tkg-compatibility'
```

lubronzhan:
https://forum.golangbridge.org/t/export-environment-variables-using-go-program-not-working-in-bash/18032
Ok learnt from here that you can't alter the parent process environment variable through os.SetEnv.
Unless you do 
```
  fmt.Println("export KEY_ID=Hello")
  fmt.Println("export SECRET_ID=world")
```
But this will impact customer's environment, also could impact other test. We should not do this. Instead we should ask customer to export these variables

PR #4313, opened 2023-01-31
enable delete of cluster when tkr is not present on supervisor

### LINKED ISSUE #4304: Enable delete of workload cluster on supervisor if TKR is not available
**Bug description**

Today in TKGs (vSphere with supervisor) user cannot delete TKC/Classy cluster when the TKR associated to the cluster is no longer available on the supervisor.

If the VMImage associated to a TKR is no longer available in the content library associated to the TKG service on supervisor, the TKR is no longer available on the supervisor cluster. 

In such cases, the addons controller does not have access to the TKR as a result blocking the reconcile loops which in turn blocks the delete cluster workflow.

This issue was fixed on the supervisor in guest-cluster-controller as part of https://gitlab.eng.vmware.com/core-build/guest-cluster-controller/-/merge_requests/958

A similar change needs to be done in addons controller to ensure the reconcile loops in addons controller can handle delete cluster workflow when TKR is not available on the supervisor

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**
Should be able to delete workload cluster in cases where TKR resource associated to the cluster is not available on the supervisor
On supervisor
1. create workload cluster (TKC/Classy) with say version 1.23.8
2. after successful bring up of the cluster, delete the associated vmimage from the content library on vSphere
3. Once the vmimage is removed from CL, the TKR definition is no longer available in the supervisor
4. Delete workload cluster (TKC/Classy cluster)

Delete of cluster is stuck as predelete hook of addons controller is stuck as the TKR resource is no longer available

jvrahav:

Today in TKGs (vSphere with supervisor) user cannot delete TKC/Classy cluster when the TKR associated to the cluster is no longer available on the supervisor.

If the VMImage associated to a TKR is no longer available in the content library associated to the TKG service on supervisor, the TKR is no longer available on the supervisor cluster.

In such cases, the addons controller does not have access to the TKR as a result blocking the reconcile loops which in turn blocks the delete cluster workflow

Fixes #4304 

Patched addons controller with this change on a testbed.
1. created TKG 2.0 cluster
2. on successfull bringup of the cluster, deleted the TKR resource on supervisor
3. was able to successfully delete the TKG 2.0 cluster

### Release note

```release-note

```

### DIFF
--- a/addons/controllers/cluster_metadata_controller.go
+++ b/addons/controllers/cluster_metadata_controller.go
@@ -92,7 +92,7 @@ func (r *ClusterMetadataReconciler) Reconcile(ctx context.Context, req reconcile
 		}
 	}
 
-	tkr, err := util.GetTKRByNameV1Alpha3(r.context, r.Client, tkrName)
+	tkr, err := util.GetTKRByNameV1Alpha3(r.context, r.Client, cluster, tkrName)
 	if err != nil {
 		log.Error(err, "unable to fetch TKR object", "name", tkrName)
 		return ctrl.Result{}, err
--- a/addons/controllers/clusterbootstrap_controller.go
+++ b/addons/controllers/clusterbootstrap_controller.go
@@ -150,7 +150,7 @@ func (r *ClusterBootstrapReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return ctrl.Result{}, nil
 	}
 
-	tkr, err := util.GetTKRByNameV1Alpha3(r.context, r.Client, tkrName)
+	tkr, err := util.GetTKRByNameV1Alpha3(r.context, r.Client, cluster, tkrName)
 	if err != nil {
 		log.Error(err, "unable to fetch TKR object", "name", tkrName)
 		return ctrl.Result{}, err
--- a/addons/controllers/packageinstallstatus_controller.go
+++ b/addons/controllers/packageinstallstatus_controller.go
@@ -138,7 +138,7 @@ func (r *PackageInstallStatusReconciler) Reconcile(_ context.Context, req reconc
 		return ctrl.Result{}, nil
 	}
 
-	tkr, err := util.GetTKRByNameV1Alpha3(r.ctx, r.Client, tkrName)
+	tkr, err := util.GetTKRByNameV1Alpha3(r.ctx, r.Client, cluster, tkrName)
 	if err != nil {
 		return ctrl.Result{}, errors.Wrapf(err, "unable to fetch TKR object '%s'", tkrName)
 	}
--- a/addons/pkg/util/tkr_util.go
+++ b/addons/pkg/util/tkr_util.go
@@ -4,9 +4,15 @@
 package util
 
 import (
+	"bytes"
+	"compress/gzip"
 	"context"
+	"encoding/base64"
 	"fmt"
+	"io"
 	"strings"
+
+	"gopkg.in/yaml.v3"
 
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -38,7 +44,7 @@ func GetTKRByNameV1Alpha1(ctx context.Context, c client.Client, tkrName string) 
 }
 
 // GetTKRByNameV1Alpha3 gets v1Alpha3 TKR object given a TKR name
-func GetTKRByNameV1Alpha3(ctx context.Context, c client.Client, tkrName string) (*runtanzuv1alpha3.TanzuKubernetesRelease, error) {
+func GetTKRByNameV1Alpha3(ctx context.Context, c client.Client, cluster *clusterapiv1beta1.Cluster, tkrName string) (*runtanzuv1alpha3.TanzuKubernetesRelease, error) {
 	tkrV1Alpha3 := &runtanzuv1alpha3.TanzuKubernetesRelease{}
 
 	if tkrName == "" {
@@ -49,6 +55,11 @@ func GetTKRByNameV1Alpha3(ctx context.Context, c client.Client, tkrName string) 
 
 	if err := c.Get(ctx, tkrNamespaceName, tkrV1Alpha3); err != nil {
 		if apierrors.IsNotFound(err) {
+			tkrV1Alpha3, err = getTKRFromAnnotation(cluster.Annotations)
+			if tkrV1Alpha3 != nil {
+				return tkrV1Alpha3, nil
+			}
+
 			return nil, nil
 		}
 		return nil, err
@@ -72,7 +83,7 @@ func GetBootstrapPackageNameFromTKR(ctx context.Context, clt client.Client, pkgR
 	}
 
 	// get TKR object associated with the cluster
-	tkr, err := GetTKRByNameV1Alpha3(ctx, clt, tkrName)
+	tkr, err := GetTKRByNameV1Alpha3(ctx, clt, cluster, tkrName)
 	if err != nil || tkr == nil {
 		return "", pkgNamePrefix, fmt.Errorf("unable to fetch TKR object '%s'", tkrName)
 	}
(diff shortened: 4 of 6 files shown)


addons/pkg/util/tkr_util.go
```diff
@@ -49,6 +54,15 @@ func GetTKRByNameV1Alpha3(ctx context.Context, c client.Client, tkrName string)
 
 	if err := c.Get(ctx, tkrNamespaceName, tkrV1Alpha3); err != nil {
 		if apierrors.IsNotFound(err) {
+			if cluster == nil {
```
adduarte, line 57: What about the case where err == nil *and* cluster == nil ? 
do we have to handle that? 
not from the current application, but from the point of view that GetTKRByNameAlpha3 is exported and can be called by anything, which means it could get a call with cluster=nil?  and tkrNmae points to an existing tkr.

addons/pkg/util/tkr_util.go
```diff
@@ -90,3 +104,36 @@ func GetBootstrapPackageNameFromTKR(ctx context.Context, clt client.Client, pkgR
 
 	return "", pkgNamePrefix, fmt.Errorf("no bootstrap package prefixed with '%s' is found in the TKR object", pkgNamePrefix)
 }
+
+func FetchTKRSpecFromAnnotations(tkc map[string]string) (*runtanzuv1alpha3.TanzuKubernetesRelease, error) {
+	controlPlaneTKR := &runtanzuv1alpha3.TanzuKubernetesRelease{}
+
+	tkrSpec, ok := tkc["run.tanzu.vmware.com/tkr-spec"]
```
adduarte, line 111: "run.tanzu.vmware.com/tkr-spec" should be defined as a constant in addons/pkg/constants/constants.go

adduarte, line 111: "run.tanzu.vmware.com/tkr-spec" should be defined as a constant in addons/pkg/constants/constants.go

addons/pkg/util/tkr_util.go
```diff
@@ -90,3 +104,36 @@ func GetBootstrapPackageNameFromTKR(ctx context.Context, clt client.Client, pkgR
 
 	return "", pkgNamePrefix, fmt.Errorf("no bootstrap package prefixed with '%s' is found in the TKR object", pkgNamePrefix)
 }
+
+func FetchTKRSpecFromAnnotations(tkc map[string]string) (*runtanzuv1alpha3.TanzuKubernetesRelease, error) {
```
adduarte, line 108: you are not returning "TKRSpec", you are returning a TanzuKubernetesRlease. 
Change name of function to a more generic form. The returned object is clear from the function signature you don't need to put it in the name. 

The word "Fetch" is obvious from the signature and call to the function as well. 
something like l

func TKRFromAnnotation(tkc map[string]string) (*runtanzuv1alpha3.TanzuKubernetesRelease, error)

not sure... but definetily something less specific than FetchTKRSpecFromAnnotations


I guess following precedence of other functions in the file
func GetTKRByAnnotation 
would work too. 


and it does not seem you need to export this function (begins with upper case letter). The only place it is called from is the utils package.   It would be best to make it private, (lower case) unless there is a place where it is called from outside the package, and I don't see that in this patch.

addons/pkg/util/tkr_util.go
adduarte: Also is this the right place to check for cluster==nil?  could you check it before calling GetTKRByNameV1Alpha3? then you don't have to pass in cluster.  seems forced to pass in a new parameter to the function just to check if it is nill.. 

In other words, what is the expected behavior (in general) if cluster == nil? 

seems to me there is a few unit tests missing covering those scnearios. 

also, from places where GetTKRByNameV1Alpha3 is called in this patch, cluster will never == nil. It is always set to something (cluster := &clusterapiv1beta1.Cluster{}). 

Again may be remove this check?

addons/controllers/clusterbootstrap_controller.go
```diff
@@ -150,7 +150,7 @@ func (r *ClusterBootstrapReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return ctrl.Result{}, nil
 	}
 
-	tkr, err := util.GetTKRByNameV1Alpha3(r.context, r.Client, tkrName)
+	tkr, err := util.GetTKRByNameV1Alpha3(r.context, r.Client, cluster, tkrName)
```
adduarte, line 153: cluster will never be nil. so why pass it in just to check if it is nil?

addons/controllers/cluster_metadata_controller.go
```diff
@@ -92,7 +92,7 @@ func (r *ClusterMetadataReconciler) Reconcile(ctx context.Context, req reconcile
 		}
 	}
 
-	tkr, err := util.GetTKRByNameV1Alpha3(r.context, r.Client, tkrName)
+	tkr, err := util.GetTKRByNameV1Alpha3(r.context, r.Client, cluster, tkrName)
```
adduarte, line 95: cluster will never be nil. so why pass it in just to check if it is nil?

addons/controllers/packageinstallstatus_controller.go
```diff
@@ -138,7 +138,7 @@ func (r *PackageInstallStatusReconciler) Reconcile(_ context.Context, req reconc
 		return ctrl.Result{}, nil
 	}
 
-	tkr, err := util.GetTKRByNameV1Alpha3(r.ctx, r.Client, tkrName)
+	tkr, err := util.GetTKRByNameV1Alpha3(r.ctx, r.Client, cluster, tkrName)
```
adduarte, line 141: cluster will never be nil. so why pass it in just to check if it is nil?

addons/pkg/util/tkr_util_test.go
```diff
@@ -138,14 +142,74 @@ var _ = Describe("TKR utils", func() {
...
+				tkrV1Alpha3, err = GetTKRByNameV1Alpha3(ctx, crtCtl, nil, tkrName)
 			})
 
 			It("should return nil TKR", func() {
 				Expect(err).ShouldNot(HaveOccurred())
 				Expect(tkrV1Alpha3).ShouldNot(BeNil())
 			})
 		})
+
+		When("tkr spec is available as annotations on cluster resource", func() {
```
adduarte, line 154: you call it "tkr spec" here but  just "tkr" in the next :) . The correct term should be tkr...
tkr spec sounds to me like the .spec of the TKR CR which is not what you mean

addons/pkg/util/tkr_util_test.go
```diff
@@ -138,14 +142,74 @@ var _ = Describe("TKR utils", func() {
...
+				tkrName = testTKR
+				crtCtl.GetReturns(apierrors.NewNotFound(schema.GroupResource{Resource: "TanzuKubernetesRelease"}, testTKR))
+				tkrV1Alpha3, err = GetTKRByNameV1Alpha3(ctx, crtCtl, cluster, tkrName)
+			})
+
+			It("should return non nil TKR", func() {
+				Expect(err).ShouldNot(HaveOccurred())
+				Expect(tkrV1Alpha3).ShouldNot(BeNil())
+			})
+		})
```
adduarte, line 212: you need to add the tests here covering the same but with cluster ==nil

adduarte (changes requested):
need to change some function signatures
need to add missing unit tests. 
and also need to check if you really need to check cluster==nil, seems unecessary and adds twice as much code. If you can get away with not checking if cluster==nil then the patch gets cut in half. less code = better codebase.

addons/controllers/cluster_metadata_controller.go
jvrahav: we are need the annotations in the cluster hence we are passing the cluster here

addons/controllers/packageinstallstatus_controller.go
jvrahav: answered in the previous comment

addons/pkg/util/tkr_util_test.go
jvrahav: given that cluster resource will never be nil and the check is removed, the test for cluster==nil seems irrelevant now

addons/pkg/util/tkr_util.go
```diff
@@ -38,7 +44,7 @@ func GetTKRByNameV1Alpha1(ctx context.Context, c client.Client, tkrName string)
 }
 
 // GetTKRByNameV1Alpha3 gets v1Alpha3 TKR object given a TKR name
-func GetTKRByNameV1Alpha3(ctx context.Context, c client.Client, tkrName string) (*runtanzuv1alpha3.TanzuKubernetesRelease, error) {
+func GetTKRByNameV1Alpha3(ctx context.Context, c client.Client, cluster *clusterapiv1beta1.Cluster, tkrName string) (*runtanzuv1alpha3.TanzuKubernetesRelease, error) {
 	tkrV1Alpha3 := &runtanzuv1alpha3.TanzuKubernetesRelease{}
```
adduarte, line 48: I am still not comfortable with changing a published function. But I guess no one else is using GetTKRByNameV1Alpha3 but us, so we could get away with it.

adduarte (approved):
Looks good if tests pass

alfredthenarwhal:
> @jvrahav:  /test
Commit: 70c081b

Tests failed! Build no: 3980

alfredthenarwhal:
> @jvrahav:  /test
Commit: fc2ec6f

Tests passed! Build no: 3991

PR #4316, opened 2023-02-01
[tanzu cli] use tkrVersion in topology.version for cc clusters

hongtaiwang:
if `--tkr` flag is configured as tkr name, should set topology.version to tkr version to make it effective.

tkr resolver will resolve k8s version for cluster cr.

tkr resolver resolves tkr version based on `topology.version` field in cluster cr. it is expected to consume tkr prefix or tkr version instead of k8s version.

If multiple tkrs with same k8s version exists in mc, k8s version(as a prefix for tkr version) is not enough for tkr resolver to resolve the desired tkr version. 

Fixes #

cluster created successfully with tkr prefix /exact tkr name/ default tkr as input in vsphere testbed
could upgrade to specific tkr version as well

### Release note

```release-note

```


imikushin (approved):
LGTM. Thanks @hongtaiwang !

PR #4333, opened 2023-02-06
change the capabilities deployment update strategy to fallback-on-replace

### LINKED ISSUE #4341: Change the deployment upgrade strategy in capabilities package
**Bug description**
We were previously installing the capabilities component using CRS and we later packaged the capabilities component as a package. To handle the migration of resources via CRS way to kapp app when the package is being installed, we used `kapp.k14s.io/disable-default-label-scoping-rules: ""` annotation on the capabilities deployment resource. But other products have already consumed the capabilities package and adding this annotation is resulting in the below error

```
Deployment.apps "tanzu-capabilities-controller-manager" is invalid: spec.selector:
Invalid value: v1.LabelSelector{MatchLabels:map[string]string
{"app":"tanzu-capabilities-manager"}
, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable (reason: Invalid) 
```

We need to use `kapp.k14s.io/update-strategy: "fallback-on-replace"` annotation to handle the above error

- ( ) APIs
- (x) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**
Successful capabilities package upgrade 

yharish991:

Previously we used `kapp.k14s.io/disable-default-label-scoping-rules: ""` annotation to handle the upgrade case for capabilities deployment, this was causing upgrade problems in other products consuming capabilities package. The other option to handle the upgrade case is to add `kapp.k14s.io/update-strategy: "fallback-on-replace"` annotation to capabilities deployment.

https://carvel.dev/blog/migrate-existing-resources-to-a-new-kapp-app/#option-2

Fixes https://github.com/vmware-tanzu/tanzu-framework/issues/4341

Installed the older version of Capabilities package on the cluster and then installed the new capabilities package to verify that the package upgrade is not causing issues

### Release note

```release-note
Changed the capabilities deployment update strategy to fallback-on-replace
```

### DIFF
--- a/packages/capabilities/bundle/config/upstream/tanzu-capabilities-controller-manager.yaml
+++ b/packages/capabilities/bundle/config/upstream/tanzu-capabilities-controller-manager.yaml
@@ -10,7 +10,7 @@ metadata:
   name: tanzu-capabilities-controller-manager
   namespace: #@ data.values.namespace
   annotations:
-    kapp.k14s.io/disable-default-label-scoping-rules: ""
+    kapp.k14s.io/update-strategy: "fallback-on-replace"
 spec:
   replicas: 1
   selector:


codegold79 (approved):
I like these small changes for reviews.

rajathagasthya (approved):

PR #4336, opened 2023-02-07
Add version check for antrea config fields marshal

hangyan:
Signed-off-by: Hang Yan <yhang@vmware.com>

This pr makes t-f will works with antrea 1.7.2/1.5.3

Fixes #

Tested on 1.24 TKR with antrea 1.5.3/antrea1.7.2 version

### Release note

```release-note

```

### DIFF
--- a/addons/controllers/antrea/antreaconfig_controller.go
+++ b/addons/controllers/antrea/antreaconfig_controller.go
@@ -8,6 +8,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/pkg/errors"
+
 	"github.com/go-logr/logr"
 	yaml "gopkg.in/yaml.v3"
 	corev1 "k8s.io/api/core/v1"
@@ -23,6 +25,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	addontypes "github.com/vmware-tanzu/tanzu-framework/addons/pkg/types"
 
 	cutil "github.com/vmware-tanzu/tanzu-framework/addons/controllers/utils"
 	addonconfig "github.com/vmware-tanzu/tanzu-framework/addons/pkg/config"
@@ -69,6 +73,12 @@ func (r *AntreaConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	if _, ok := annotations[constants.TKGAnnotationTemplateConfig]; ok {
 		log.Info(fmt.Sprintf("resource '%v' is a config template. Skipping reconciling", req.NamespacedName))
 		return ctrl.Result{}, nil
+	}
+
+	labels := antreaConfig.GetLabels()
+	if _, ok := labels[addontypes.PackageNameLabel]; !ok {
+		r.Log.Info(fmt.Sprintf("AntreaConfig resource '%v' does not contains package name label", req.NamespacedName))
+		return ctrl.Result{}, errors.New("AntreaConfig does not contains package name label")
 	}
 
 	// deep copy AntreaConfig to avoid issues if in the future other controllers where interacting with the same copy
--- a/addons/controllers/antrea/antreaconfig_util.go
+++ b/addons/controllers/antrea/antreaconfig_util.go
@@ -6,8 +6,11 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
+
+	"golang.org/x/mod/semver"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clusterv1beta1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	clusterapiutil "sigs.k8s.io/cluster-api/util"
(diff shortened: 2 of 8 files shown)


hangyan:
@edwardbadboy

addons/controllers/antrea/antreaconfig_controller.go
```diff
@@ -71,6 +73,12 @@ func (r *AntreaConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil
 	}
 
+	labels := antreaConfig.GetLabels()
+	if _, ok := labels[constants.PackageNameLabel]; !ok {
+		r.Log.Info(fmt.Sprintf("AntreaConfig resource '%v' does not contains package name label", req.NamespacedName))
+		return ctrl.Result{}, errors.New("AntreaConfig does not contains package name label")
```
edwardbadboy, line 79: I suppose that the framework will retry automatically, right?

addons/controllers/antrea/antreaconfig_util.go
```diff
@@ -162,6 +165,10 @@ func (r *AntreaConfigReconciler) ClusterToAntreaConfig(o client.Object) []ctrl.R
 }
 
 func mapAntreaConfigSpec(cluster *clusterv1beta1.Cluster, config *cniv1alpha1.AntreaConfig) (*antreaConfigSpec, error) {
+
+	packageName := config.GetLabels()[constants.PackageNameLabel]
+	version := strings.TrimPrefix(strings.Split(packageName, "---")[0], "antrea.tanzu.vmware.com.")
```
edwardbadboy, line 170: I think it's better to fetch version from Package spec.

addons/pkg/constants/constants.go
```diff
@@ -235,6 +235,9 @@ const (
 	// ClusterBootstrapManagedSecret is the name for the secrets that are managed by ClusterBootstrapController
 	ClusterBootstrapManagedSecret = "clusterbootstrap-secret"
 
+	// PackageNameLabel is the label key for package name
+	PackageNameLabel = "tkg.tanzu.vmware.com/package-name"
```
edwardbadboy, line 239: Maybe we can reuse addons/pkg/types/addons.go PackageNameLabel constant.

addons/controllers/antrea/antreaconfig_controller.go
hangyan: yes

addons/controllers/antrea/antreaconfig_util.go
hangyan: this lable is also used by other component, so i think make sense to reuse it for now.

addons/pkg/constants/constants.go
hangyan: done

addons/controllers/antrea/antreaconfig_util.go
```diff
@@ -208,9 +215,15 @@ func mapAntreaConfigSpec(cluster *clusterv1beta1.Cluster, config *cniv1alpha1.An
 	configSpec.Antrea.AntreaConfigDataValue.TLSCipherSuites = config.Spec.Antrea.AntreaConfigDataValue.TLSCipherSuites
 	configSpec.Antrea.AntreaConfigDataValue.DisableUDPTunnelOffload = config.Spec.Antrea.AntreaConfigDataValue.DisableUDPTunnelOffload
 	configSpec.Antrea.AntreaConfigDataValue.DefaultMTU = config.Spec.Antrea.AntreaConfigDataValue.DefaultMTU
-	configSpec.Antrea.AntreaConfigDataValue.DNSServerOverride = config.Spec.Antrea.AntreaConfigDataValue.DNSServerOverride
-	configSpec.Antrea.AntreaConfigDataValue.DisableTXChecksumOffload = config.Spec.Antrea.AntreaConfigDataValue.DisableTXChecksumOffload
-	configSpec.Antrea.AntreaConfigDataValue.EnableBridgingMode = config.Spec.Antrea.AntreaConfigDataValue.EnableBridgingMode
+
+	if semver.Compare(version, "v1.7.1") >= 0 {
+		configSpec.Antrea.AntreaConfigDataValue.MultiCluster.Enable = config.Spec.Antrea.AntreaConfigDataValue.MultiCluster.Enable
```
knabben, line 220: do we need to omit the configuration values as well / convert to pointers?

addons/controllers/antrea/antreaconfig_util.go
edwardbadboy: Do you mean other component also extract the Package version according to the label value? My suggestion is that it get the Package name from the label value, and do a get call to get the Package spec, and get the version from Package spec.

hangyan: I get your point, i just mean this label is used by other part of code too, so it's kind of reliable for a while. I would like to open a catchup pr later to refactor this part, if we don't have enough time for this fix.

addons/controllers/antrea/antreaconfig_util.go
hangyan: they already did, featuregate is the only ones that are missing..

knabben (approved):

edwardbadboy (approved):
The fix looks OK.

wjun (approved):
LGTM. Once the capz clusterclass pipeline passes management cluster creation, you can continue to add 'ok-to-merge' label and merge.

PR #4344, opened 2023-02-09
Add option to trust custom tdnf repository's cert

lubronzhan:
We need way to trust private tdnf repository cert,  also way to resolve it's fqdn if dns server is not configured

Fixes #

### Release note

```release-note

```


tkg/constants/clusterclass_mapping.go
```diff
@@ -245,6 +245,10 @@ var ClusterAttributesToLegacyVariablesMapVsphere = map[string]string{
 	TopologyWorkersMachineDeploymentsName2:          "",
 	TopologyWorkersMachineDeploymentsReplicas2:      ConfigVariableWorkerMachineCount2,
 	TopologyWorkersMachineDeploymentsFailureDomain2: ConfigVariableVsphereAz2,
+
+	"spec.topology.variables.customTDNFRepository.ip":          ConfigVariableCustomTDNFRepositoryIP,
+	"spec.topology.variables.customTDNFRepository.fqdn":        ConfigVariableCustomTDNFRepositoryFQDN,
```
DanielXiao, line 250: The mapping consts are for reusing legacy variable validation. Should we perform format check on them?

lubronzhan: This is just a hack that is temporarily provided to TCA. Don't wanna add more time to that

lubronzhan: ALos if I don't add it here, when user use two steps to create workload cluster, it will fail.

DanielXiao (approved):

PR #4362, opened 2023-02-13
Handle equal sign in Extra Args

DanielXiao:
Various EXTRA_ARGS are required of format "key1=value1;key2=value2", and some of arg value also contains equal sign, for example: feature-gates=RotateKubeletServerCertificate=true,QOSReserved=true We should take everything after the the first equal sign as its value in this case.

Move converting legacy variables to Cluster variables to the common method get_cluster_variables() because there is no difference among providers.

Validate legacy variables on various EXTRA_ARGS.

Fixes #

### Release note

```release-note

```

### DIFF
--- a/tkg/client/cluster.go
+++ b/tkg/client/cluster.go
@@ -738,11 +738,44 @@ func (c *TkgClient) ConfigureAndValidateWorkloadClusterConfiguration(options *Cr
 		return err
 	}
 
+	if err = c.ValidateExtraArgs(); err != nil {
+		return err
+	}
+
 	if err = c.ValidateKubeVipLBConfiguration(TkgLabelClusterRoleWorkload); err != nil {
 		return NewValidationError(ValidationErrorCode, err.Error())
 	}
 
 	return c.ValidateSupportOfK8sVersionForManagmentCluster(clusterClient, options.KubernetesVersion, skipValidation)
+}
+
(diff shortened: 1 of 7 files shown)


tkg/client/cluster.go
```diff
@@ -738,13 +738,46 @@ func (c *TkgClient) ConfigureAndValidateWorkloadClusterConfiguration(options *Cr
...
+	}
+
 	if err = c.ValidateKubeVipLBConfiguration(TkgLabelClusterRoleWorkload); err != nil {
 		return NewValidationError(ValidationErrorCode, err.Error())
 	}
 
 	return c.ValidateSupportOfK8sVersionForManagmentCluster(clusterClient, options.KubernetesVersion, skipValidation)
 }
 
+func (c *TkgClient) ConfigAndValidateExtraArgs() error {
```
hongtaiwang, line 752: seems we only do validation for ExtraArgs in this func, what about renaming the func to avoid confusion?

DanielXiao: We also set empty string if not found

hongtaiwang (approved):
lgtm

zhouke1991:
LGTM!

wjun (approved):

alfredthenarwhal:
Created cherry-pick PR

PR #4343, opened 2023-02-08
workflow: Add cherry-pick job for release-0.28

navidshaikh:
### What this PR does / why we need it
- Add github workflow for cherry-picking to release-0.28 branch.
- Workflow triggers for your merged PR if it has 'cherry-pick/release-0.28' label
- It creates a cherry-pick PR against release-0.28 branch and adds 'backports/release-0.28' label
- Titles the cherry-pick PR referring to original PR number and its title with [release-0.28] branch prefix
- Refers to original PR number in the PR body
- It uses bot 'Alfred the Narwhal' as the cherry-pick PR author, so the CLA could pass, github-actions bot is not added to our CLA
- Workflow uses `ALFRED_PAT` token, the respective token is created and configured in the repo
- `cherry-pick/release-0.28` and `backports/release-0.28` labels are created in t-f repo
- add docs

Signed-off-by: Navid Shaikh <navids@vmware.com>

### Which issue(s) this PR fixes

Fixes #

### Describe testing done for PR
See test PR created by https://github.com/navidshaikh/tanzu-framework/pull/31

### Release note

```release-note

```

PR #4375, opened 2023-02-13
Add comment notifications for cherry-pick workflow result

navidshaikh:
- Add a comment if cherry-pick workflow operation was succesful, failed or cancelled
- Links the created PR if successful
- Update docs
- follow up to https://github.com/vmware-tanzu/tanzu-framework/pull/4343

Signed-off-by: Navid Shaikh <navids@vmware.com>

Fixes #

Success case: https://github.com/navidshaikh/tanzu-framework/pull/35#issuecomment-1428735654
Failure case: https://github.com/navidshaikh/tanzu-framework/pull/37#issuecomment-1428739255

### Release note

```release-note

```

### DIFF
--- a/.github/workflows/cherry-pick.yaml
+++ b/.github/workflows/cherry-pick.yaml
@@ -1,3 +1,5 @@
+name: Cherry-pick
+
 on:
   pull_request:
     branches:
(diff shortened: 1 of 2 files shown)


.github/workflows/cherry-pick.yaml
```diff
@@ -7,15 +9,18 @@ on:
 jobs:
   cherry_pick_release_v0_28:
     runs-on: ubuntu-latest
-    name: Cherry pick into release-0.28 brach
+    name: Cherry-pick into release-0.28 brach
     if: ${{ contains(github.event.pull_request.labels.*.name, 'cherry-pick/release-0.28') && github.event.pull_request.merged == true }}
```
sriramandev, line 13: Cant we rationalize it to use any branch (cherry-pick/*). This will make sure we can support cherry pick to any branch and as when starts for GH, we do not need to modify and also this can be used in future as well without changes.

rajaskakodkar: +1 to this scenario but we would have to create the label first I guess to support (cherry-pick/*)

rajaskakodkar: What about having a comment like `/cherry-pick <branch-name>` instead of label because we have to create the label every time.

sriramandev: @rajaskakodkar problem with argument passing is there can be possibility of the branch passed not existing. Using labels we can to some extend/reasonably assume the target branch is available

sriramandev (approved):
lgtm... the comments could be follow ups...

.github/workflows/cherry-pick.yaml
navidshaikh: re: rationalise it use any branch
will revisit generalising this to work with different release branches

re: we would have to create the label first
yes (created only labels for which cherry-picks are open for now) 

re: /cherry-pick <branch-name>
think it can be done too, could've been better if the issue_comment activity is parsed elsewhere, for now kept it label based so only merge of the PR triggers an event

thanks for reviews! will revisit the suggestions in follow up.

PR #4381, opened 2023-02-14
azure(clusterclass) add debug info

m1zzx2:
### Which issue(s) this PR fixe
print capz/capa logs  for debug


hongtaiwang:
not sure if there will be too much contents to print. 
Is it possible to upload the log bundle to somewhere on DBC and print the DBC link if pipeline fails?

tkg/test/tkgctl/shared/e2e_common_cc.go
```diff
@@ -148,6 +148,11 @@ func E2ECommonCCSpec(ctx context.Context, inputGetter func() E2ECommonCCSpecInpu
 	})
 
 	AfterEach(func() {
+		framework.PrintPodsLogs(mngProxy, "capa-system")
```
wjun, line 151: Is there a way to check only when the test case fails, we print the pod logs?

m1zzx2:
> 



> not sure if there will be too much contents to print. Is it possible to upload the log bundle to somewhere on DBC and print the DBC link if pipeline fails?

I think we better not expose private cloud links in public cloud

tkg/test/tkgctl/shared/e2e_common_cc.go
m1zzx2: Ok, it will be corrected, the output log will be a lot each time

PR #4385, opened 2023-02-15
Ignore the error if unable to retrieve the complete list of server APIs

DanielXiao:
When configurations includes STIG compliance settings, it requires certificate approval post cluster creation. Metrics server goes down because kubelet certs are not available, which lead to its API service down. We will see the following error when we list API resources through method ServerPreferredResources(). We can ignore it because we can still retrieve other available API resources.

"unable to retrieve the complete list of server APIs: metrics.k8s.io/v1beta1: the server is currently unable to handle the request"

Fixes #

### Release note

```release-note

```


addons/webhooks/clusterbootstrap_webhook.go
```diff
@@ -292,7 +292,11 @@ func (wh *ClusterBootstrap) getGVR(gk schema.GroupKind) (*schema.GroupVersionRes
 	}
 	apiResourceList, err := wh.cachedDiscoveryClient.ServerPreferredResources()
 	if err != nil {
-		return nil, err
+		// Sometimes it is unable to retrieve the complete list of server APIs i.g. an aggregated API server is down,
+		// ErrGroupDiscoveryFailed is returned in this case. We can still get other API resources and ignore this error.
+		if _, ok := err.(*discovery.ErrGroupDiscoveryFailed); !ok {
+			return nil, err
+		}
 	}
```
adduarte, line 300: Please add unit test to cover this line.

addons/webhooks/clusterbootstrap_webhook.go
```diff
@@ -292,7 +292,11 @@ func (wh *ClusterBootstrap) getGVR(gk schema.GroupKind) (*schema.GroupVersionRes
...
 	apiResourceList, err := wh.cachedDiscoveryClient.ServerPreferredResources()
 	if err != nil {
-		return nil, err
+		// Sometimes it is unable to retrieve the complete list of server APIs i.g. an aggregated API server is down,
+		// ErrGroupDiscoveryFailed is returned in this case. We can still get other API resources and ignore this error.
+		if _, ok := err.(*discovery.ErrGroupDiscoveryFailed); !ok {
+			return nil, err
```
adduarte, line 298: Is this logged anywhere?  We need to log this if we re going to ignore the error.

addons/webhooks/clusterbootstrap_webhook.go
```diff
@@ -292,7 +292,11 @@ func (wh *ClusterBootstrap) getGVR(gk schema.GroupKind) (*schema.GroupVersionRes
 	}
 	apiResourceList, err := wh.cachedDiscoveryClient.ServerPreferredResources()
 	if err != nil {
-		return nil, err
+		// Sometimes it is unable to retrieve the complete list of server APIs i.g. an aggregated API server is down,
+		// ErrGroupDiscoveryFailed is returned in this case. We can still get other API resources and ignore this error.
```
adduarte, line 296: Remove unecessary comments. They might make sense to you right now but someone coming in 6 months later they are out of context, and is something else to mantain. 
Add any explanation of why we are ignoring the error to  logging. (info, debut, or, etc...)

addons/webhooks/clusterbootstrap_webhook.go
```diff
@@ -292,7 +292,11 @@ func (wh *ClusterBootstrap) getGVR(gk schema.GroupKind) (*schema.GroupVersionRes
...
+		if _, ok := err.(*discovery.ErrGroupDiscoveryFailed); !ok {
```
adduarte, line 297: This check is in the wrong function. 
This function (getGVR) one and only purpose is to return a groupversionresource, or to return error. 

"func (wh *ClusterBootstrap) getGVR(gk schema.GroupKind) (*schema.GroupVersionResource, error)"

Make the check after the call to this. here: https://github.com/vmware-tanzu/tanzu-framework/blob/ba6b6f70b1873e1fbc0fb00f9b8eb47eb42367ee/addons/webhooks/clusterbootstrap_webhook.go#L260

for example:
```
		if err != nil &&  !err.(*discovery.ErrGroupDiscoveryFailed) {
			return field.Invalid(fldPath.Child("ProviderRef"), valuesFrom.ProviderRef, err.Error())
		}
```
I think that would make it more ovious and perhaps not even require  log message, but it would be nice just a couple more lines.

adduarte (changes requested):
- Please add unit tests  to verify new code and to verify no nil pointer access of apiResourceList.  The implication of ignoring the error need to be well covered with tests.  For example, what if the discovery.ErrGroupDiscoveryFailed is due to something else? Can it be caused by something else? and if so do we handle that correctly. Silently ignoring errors is dangerous and we need to make sure the test coverage is sufficient. 

- create issue in tz and fill in the "fix#" part in the message.

addons/webhooks/clusterbootstrap_webhook.go
adduarte: If not, then you need too make sure apiResourceList is not nil

adduarte: Also is this check precise enough to only catch the Metrics server goes down because kubelet certs. 
Would other failures possibly sneak by due to this?  need a unit test for that as well.

DanielXiao: apiResourceList will be at least an empty list according to implementation in method ServerPreferredResources(). I agree to check nil in case there are changes in the future.

addons/webhooks/clusterbootstrap_webhook.go
DanielXiao: point taken.

addons/webhooks/clusterbootstrap_webhook.go
DanielXiao: Point taken

addons/webhooks/clusterbootstrap_webhook.go
DanielXiao: Point taken.

DanielXiao:
> Please add unit tests to verify new code and to verify no nil pointer access of apiResourceList. The implication of ignoring the error need to be well covered with tests. For example, what if the discovery.ErrGroupDiscoveryFailed is due to something else? Can it be caused by something else? and if so do we handle that correctly. Silently ignoring errors is dangerous and we need to make sure the test coverage is sufficient.
Based on current controller-runtime framework, it is impossible to verify nil pointer of apiResourceList. To make ServerPreferredResources() return (nil, err), we have to bring down k8s apiserver right before we call ServerPreferredResources().  ErrGroupDiscoveryFailed returns only when there is a failed API group and apiResourceList will never be nil. In addition `for _, apiResource := range apiResourceList ` produces no iterations even if apiResourceList is nil. I added a test to cover the real ErrGroupDiscoveryFailed case with other APIResources successfully retrieved. If we want to cover the nil corner case we need to make getGVR as a public method and write unit tests with fake implementations of cachedDiscoveryClient. I don't think it worth the effort since we never hit this case in the real world.
 
> create issue in tz and fill in the "fix#" part in the message.
There is a jira ticket tracking this issue, do we need to duplicated one in github?

PR #4415, opened 2023-02-21
error out when getting virtual machine images fails for context related error

hongtaiwang:
if errors like context deadline exceeded happens, we need correct error message instead of partially retrieving results

if context deadline exceeded in the mid way of retrieving vm images, tkr vsphere resolver could only get part of vm images, and if the target vm template is not included in the response, it errors out like
```
1.6768841257280843e+09    DEBUG    controller-runtime.webhook.webhooks    wrote response    {"webhook": "/resolve-template", "code": 403, "reason": "unable to find VM Template associated with OVA Version v1.24.9+vmware.1-tkg.1-f5e94dab9dfbb9988aeb94f1ffdc6e5e. Please upload at least one VM Template to continue", "UID": "15a0500e-2ae9-4089-832f-76ef27f92e91", "allowed": false}
```
which is incorrect.

Fixes #

could get the correct message from tkr vsphere resolver
```
1.6769495279332488e+09	DEBUG	controller-runtime.webhook.webhooks	wrote response	{"webhook": "/resolve-template", "code": 403, "reason": "failed to get K8s VM templates: Post \"https://10.202.228.120/sdk\": context canceled", "UID": "19e0b414-64f9-4895-b675-f7f32c490751", "allowed": false}
```

### Release note

```release-note

```

### DIFF
--- a/tkg/vc/client.go
+++ b/tkg/vc/client.go
@@ -697,6 +697,9 @@ func (c *DefaultClient) GetVirtualMachineImages(ctx context.Context, datacenterM
 		if ovaVersion, distroName, distroVersion, distroArch := c.getVMMetadata(&vms[i]); ovaVersion != "" {
 			path, _, err := c.GetPath(ctx, vms[i].Self.Value)
 			if err != nil {
+				if ctx.Err() != nil {
+					return results, err
+				}
 				continue
 			}
 			obj := &tkgtypes.VSphereVirtualMachine{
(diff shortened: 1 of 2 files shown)


tkg/vc/client.go
```diff
@@ -697,6 +697,9 @@ func (c *DefaultClient) GetVirtualMachineImages(ctx context.Context, datacenterM
 		if ovaVersion, distroName, distroVersion, distroArch := c.getVMMetadata(&vms[i]); ovaVersion != "" {
 			path, _, err := c.GetPath(ctx, vms[i].Self.Value)
 			if err != nil {
+				if strings.Contains(err.Error(), "context canceled") {
```
DanielXiao, line 700: It 's better to check error instance `err == context.Canceled || err == context.DeadlineExceeded` or check context `ctx.Err() != nil`

hongtaiwang: updated. thanks!

imikushin (approved):

PR #4416, opened 2023-02-21
check cluster name length when ako needs to be deployed in the cluster

XudongLiuHarold:

- Because carvel package use `label` instead of `annotation`, so the package name has maximum 64 character limitation, if it exceeds 64, the package deployment will fail.
- This PR checks if the cluster name will cause AKO package deployment fail.
- This is a temp solution, we still need a long-term and generic solution for not only ako package, but all packages.

Fixes #

### Release note

```release-note
Tanzu Cli will throw out errors when cluster name length is longer than the max allowed in avi enabled workload cluster.
```

### DIFF
--- a/tkg/client/cluster.go
+++ b/tkg/client/cluster.go
@@ -746,6 +746,10 @@ func (c *TkgClient) ConfigureAndValidateWorkloadClusterConfiguration(options *Cr
 		return NewValidationError(ValidationErrorCode, err.Error())
 	}
 
+	if err = c.ValidateAviMaxAllowedClusterNameLength(TkgLabelClusterRoleWorkload, options, clusterClient); err != nil {
+		return NewValidationError(ValidationErrorCode, err.Error())
+	}
+
 	return c.ValidateSupportOfK8sVersionForManagmentCluster(clusterClient, options.KubernetesVersion, skipValidation)
 }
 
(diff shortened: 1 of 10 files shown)


alfredthenarwhal:
Commit: 00ae985

Tests failed! Build no: 4174

alfredthenarwhal:
Commit: 39fa5f1

Tests failed! Build no: 4235

alfredthenarwhal:
Commit: bf46119

Tests failed! Build no: 4250

tkg/client/cluster.go
```diff
@@ -952,6 +956,44 @@ func (c *TkgClient) ValidateKubeVipLBConfiguration(clusterRole string) error {
...
+		}
+		// mgmt avi labels is specified & workload cluster avi labels not specify, workload cluster doesn't enable Avi, skip
+		if !utils.IsAviInputEmpty(configValues[constants.ConfigVariableAviLabels]) && aviLabels == "" {
+			return nil
+		}
+		if len(options.ClusterName) > constants.AkoMaxAllowedClusterNameLen {
+			return errors.Errorf("%s contains more than %d character, which will cause AKO package deployment failure", options.ClusterName, constants.AkoMaxAllowedClusterNameLen)
+		}
+	}
+	return nil
```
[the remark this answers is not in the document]
XudongLiuHarold: > Do we handle the case when workload cluster is not clusterclass

If cluster is not clusterclass based, akoo secret will not be found, then there will be no validation on the legacy cluster, which should be good.

> if akooAddonSecretValues not found, we should return nil or error?

We should return nil because when the mgmt cluster doesn't have AVI enabled, it should be ok to not have akoo pod in the management cluster.

chenlin07 (approved):

alfredthenarwhal:
Commit: 24f1fa9

Tests failed! Build no: 4337

alfredthenarwhal:
Commit: 0ebaccd

Tests failed! Build no: 4350

tkg/client/validate.go
```diff
@@ -1874,12 +1874,16 @@ func getDockerBridgeNetworkCidr() (string, error) {
...
 // ConfigureAndValidateAviConfiguration validates the configuration inputs of Avi aka. NSX Advanced Load Balancer
-func (c *TkgClient) ConfigureAndValidateAviConfiguration() error {
+func (c *TkgClient) ConfigureAndValidateAviConfiguration(clusterName string) error {
 	aviEnable, _ := c.TKGConfigReaderWriter().Get(constants.ConfigVariableAviEnable)
 	// ignoring error because AVI_ENABLE is an optional configuration
 	if aviEnable == "" || aviEnable == "false" {
 		return nil
 	}
+
+	if len(clusterName) > constants.AkoMaxAllowedClusterNameLen {
```
lubronzhan, line 1884: Do we check if it's clusterclass type cluster here?

XudongLiuHarold: We don't need to do that, if cluster is not clustercalss based, ako operator sercert can't find, since it is only trying to retrieve akoo v2 addon secret content

tkg/constants/cluster_internal.go
```diff
@@ -62,12 +62,13 @@ const (
...
-	AkoCleanUpAnnotationKey  = "AviObjectDeletionStatus"
-	AkoCleanUpFinishedStatus = "Done"
-	AkoOperatorName          = "ako-operator"
+	AkoStatefulSetName          = "ako"
+	AkoAddonName                = "load-balancer-and-ingress-service"
+	AkoNamespace                = "avi-system"
+	AkoCleanUpAnnotationKey     = "AviObjectDeletionStatus"
+	AkoCleanUpFinishedStatus    = "Done"
+	AkoOperatorName             = "ako-operator"
+	AkoMaxAllowedClusterNameLen = 29
```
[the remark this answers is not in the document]
XudongLiuHarold: Yes, you are right, it is `kapp.k14s.io/app-change-app: tkg-mgmt-vc-load-balancer-and-ingress-service.app`, will update the code

tkg/client/validate.go
XudongLiuHarold: They have different names

lubronzhan (approved):

PR #4421, opened 2023-02-22
WIP: enable user to configure netpermission while create w cluster

nikhilbarge:
This pr will enable tkgm user to configure netpermission while create wl clusters

Fixes #

### Release note

```release-note
This pr will enable tkgm user to confgure csi net-permissions 
```

### DIFF
--- a/providers/config_default.yaml
+++ b/providers/config_default.yaml
@@ -155,6 +155,15 @@ VSPHERE_AZ_0:
 VSPHERE_AZ_1:
 VSPHERE_AZ_2:
 
+#! Settings to provide netpermissions for vpshere-csi file volumes
+#! example:
+#! NETPERMISSIONS:
+#!   A:
+#!     ips: "*"
+#!     permissions: READ_WRITE
+#!     rootsquash: false
+NETPERMISSIONS:
+
 #! FQDN or IP address to vCenter
 VSPHERE_SERVER:
 
--- a/providers/config_default.yaml
+++ b/providers/config_default.yaml
@@ -164,6 +164,9 @@ VSPHERE_AZ_2:
 #!     rootsquash: false
 NETPERMISSIONS:
 
+#! The namespace csi components are to be deployed in
+CSINAMESPACE: ""
+
 #! FQDN or IP address to vCenter
 VSPHERE_SERVER:
 
--- a/providers/config_default.yaml
+++ b/providers/config_default.yaml
@@ -155,17 +155,8 @@ VSPHERE_AZ_0:
 VSPHERE_AZ_1:
 VSPHERE_AZ_2:
 
-#! Settings to provide netpermissions for vpshere-csi file volumes
-#! example:
-#! NETPERMISSIONS:
-#!   A:
-#!     ips: "*"
-#!     permissions: READ_WRITE
-#!     rootsquash: false
+#! Settings to provide netpermissions for vpshere-csi file volumes as a base64-encoded string
 NETPERMISSIONS:
-
-#! The namespace csi components are to be deployed in
-CSI_NAMESPACE: ""
 
 #! FQDN or IP address to vCenter
 VSPHERE_SERVER:
--- a/providers/config_default.yaml
+++ b/providers/config_default.yaml
@@ -165,7 +165,7 @@ VSPHERE_AZ_2:
 NETPERMISSIONS:
 
 #! The namespace csi components are to be deployed in
-CSINAMESPACE: ""
+CSI_NAMESPACE: ""
 
 #! FQDN or IP address to vCenter
 VSPHERE_SERVER:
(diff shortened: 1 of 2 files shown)


alfredthenarwhal:
> @nikhilbarge:  /test install-vc7 --graph tkg-v2.2.0-zshippable.yaml
Commit: ebce04f

Tests passed! Build no: 4209

providers/config_default.yaml
```diff
@@ -155,6 +155,18 @@ VSPHERE_AZ_0:
...
+#! example:
+#! NETPERMISSIONS:
+#!   A:
+#!     ips: "*"
+#!     permissions: READ_WRITE
+#!     rootsquash: false
+NETPERMISSIONS:
+
+#! The namespace csi components are to be deployed in
+CSINAMESPACE: ""
```
jeffwubj, line 168: nit: CSI_NAMESPACE

nikhilbarge: changed to CSI_NAMESPACE

jeffwubj (approved):

wjun:
What integration tests have been done for this pr?

nikhilbarge:
to test this i have run pipeline  install-vc7  also manually tested

providers/config_default.yaml
```diff
@@ -155,6 +155,18 @@ VSPHERE_AZ_0:
...
 VSPHERE_AZ_2:
 
+#! Settings to provide netpermissions for vpshere-csi file volumes
+#! example:
+#! NETPERMISSIONS:
+#!   A:
+#!     ips: "*"
+#!     permissions: READ_WRITE
+#!     rootsquash: false
+NETPERMISSIONS:
```
DanielXiao, line 165: A map object does not work here and you can expose 3 variables.

providers/config_default.yaml
```diff
@@ -155,6 +155,18 @@ VSPHERE_AZ_0:
...
+
+#! The namespace csi components are to be deployed in
+CSI_NAMESPACE: ""
```
DanielXiao, line 168: CSI namespace is controlled by addon manager and should not expose to users

DanielXiao (changes requested):
Please split the map variable into multiple variables.

providers/config_default.yaml
nikhilbarge: will update as per suggestion

providers/config_default.yaml
nikhilbarge: will remove

providers/config_default.yaml
wjun: @DanielXiao @nikhilbarge this is a typical case that the addon config will include a map struct. I do not think it should be exposed to config file with 3 variables. Instread, workload cluster users should create a new clusterbootstrap to include this optional netpermissions when creating a workload cluster, or edit the vpsherecsiconfig to include this optional netpermissions after creating a management cluster or a workload cluster. + @shyaamsn @adduarte @12345lcr

DanielXiao: Got it, it is a map[string]*NetPermissionConfigs type. We can also use a legacy style variable NETPERMISSIONS_BASE64 to encode the yaml object then decode it to the field netPermissions. It is also acceptable if it is not supported through legacy configuration file.

alfredthenarwhal:
> @nikhilbarge:  /test install-vc7 --graph tkg-v2.2.0-zshippable.yaml
Commit: 08674dc

Tests failed! Build no: 4228

DanielXiao (approved):

DanielXiao (changes requested):
Could you add format validation code in tanzu cli on NETPERMISSIONS_BASE64 to ensure it can be converted to map[string]*NetPermissionConfig? Please also build tanzu cli and manually test the following scenarios:
1. Input configuration file does not contain NETPERMISSIONS_BASE64.
2. NETPERMISSIONS_BASE64 is a valid format.
3. NETPERMISSIONS_BASE64 is not a valid format.

providers/config_default.yaml
```diff
@@ -155,6 +155,9 @@ VSPHERE_AZ_0:
 VSPHERE_AZ_1:
 VSPHERE_AZ_2:
 
+#! Settings to provide netpermissions for vpshere-csi file volumes as a base64-encoded string
+NETPERMISSIONS_BASE64:
```
DanielXiao, line 159: The default value is none rather than empty string.

wjun:
@nikhilbarge If you have tested the clusterbootstrap approach, please drop this pr as cli config file is not a place for optional addon config parameters.

wjun (changes requested):
@nikhilbarge If you have tested the clusterbootstrap approach, please drop this pr as cli config file is not a place for optional addon config parameters.

nikhilbarge:
> @nikhilbarge If you have tested the clusterbootstrap approach, please drop this pr as cli config file is not a place for optional addon config parameters.
ok will close this pr

alfredthenarwhal:
Created cherry-pick PR

PR #4450, opened 2023-03-03
Use build-tooling to build FeatureGates component

### LINKED ISSUE #4449: Use build tooling to build FeatureGates component
This issue is to track the effort to consume [build tooling](https://github.com/vmware-tanzu/build-tooling-for-integrations) to build featuregates component.

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- (X) Test and Release
- ( ) User Experience
- (X) Developer Experience

**Additional context**

avi-08:
This PR uses [build tooling](https://github.com/vmware-tanzu/build-tooling-for-integrations) to build featuregates component. Changes:
- Template the base Dockerfile for every component, so that dependencies within the project can be resolved locally
- Add another layer for copying component's module code, instead of mounting the entire context.
- Revert changes to capabilities go.mod files so that it refers to local dependencies.

Fixes #4449

Ran `make docker-build-all -f build-tooling.mk` from project root and verified successful creation of the FeatureGates component images.

### Release note

```release-note

```


featuregates/controller/go.mod
```diff
@@ -2,23 +2,16 @@ module github.com/vmware-tanzu/tanzu-framework/featuregates/controller
...
 
 require (
 	github.com/go-logr/logr v1.2.3
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.20.2
-	github.com/vmware-tanzu/tanzu-framework/apis/config v0.0.0-20220824221239-af5a644ffef7
-	github.com/vmware-tanzu/tanzu-framework/apis/core v0.0.0-00010101000000-000000000000
-	github.com/vmware-tanzu/tanzu-framework/cli/runtime v0.0.0-00010101000000-000000000000
-	github.com/vmware-tanzu/tanzu-framework/featuregates/client v0.0.0-20221024130358-59eae49d96aa
+	github.com/vmware-tanzu/tanzu-framework/apis/config v0.0.0-20230302191204-eb0dfd6919e5
```
sathyanarays, line 11: This makes the controller depend on older version of APIs. If we need to make an API change, the API change has to be check-in first; followed by the controller change check-in. Do we aware of this limitation? Is it possible to make the build tooling to use the config from main itself?

featuregates/controller/go.mod
```diff
@@ -2,23 +2,16 @@ module github.com/vmware-tanzu/tanzu-framework/featuregates/controller
 
 go 1.18
 
-replace (
-	github.com/vmware-tanzu/tanzu-framework/apis/cli => ../../apis/cli
-	github.com/vmware-tanzu/tanzu-framework/apis/config => ../../apis/config
-	github.com/vmware-tanzu/tanzu-framework/apis/core => ../../apis/core
-	github.com/vmware-tanzu/tanzu-framework/cli/runtime => ../../cli/runtime
-	github.com/vmware-tanzu/tanzu-framework/featuregates/client => ../client
-	github.com/vmware-tanzu/tanzu-framework/util => ../../util
```
rajathagasthya, line 11: Why are we removing relative references and pinning it to a version below?

featuregates/client/go.mod
```diff
@@ -2,19 +2,13 @@ module github.com/vmware-tanzu/tanzu-framework/featuregates/client
 
 go 1.18
 
-replace (
-	github.com/vmware-tanzu/tanzu-framework/apis/cli => ../../apis/cli
-	github.com/vmware-tanzu/tanzu-framework/apis/config => ../../apis/config
-	github.com/vmware-tanzu/tanzu-framework/apis/core => ../../apis/core
-	github.com/vmware-tanzu/tanzu-framework/cli/runtime => ../../cli/runtime
-	github.com/vmware-tanzu/tanzu-framework/util => ../../util
-)
```
rajathagasthya, line 11: Why are we removing relative references and pinning it to a version here?

build-tooling.mk
```diff
@@ -13,7 +13,7 @@ IMG_VERSION_OVERRIDE ?= $(IMG_DEFAULT_TAG)
 GOPROXY ?= "https://proxy.golang.org,direct"
 PLATFORM=local
 # check out step 2 in this documentation on how to set the COMPONENTS variable https://github.com/vmware-tanzu/build-tooling-for-integrations/blob/main/docs/build-tooling-getting-started.md
-COMPONENTS ?= capabilities/client capabilities/controller.capabilities-controller-manager.capabilities
+COMPONENTS ?= capabilities/client capabilities/controller.capabilities-controller-manager.capabilities featuregates/client featuregates/controller.featuregates-controller-manager.featuregates
```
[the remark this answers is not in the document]
avi-08: Makes sense, will modify this.

featuregates/controller/go.mod
avi-08: Since current version of build tooling scripts only use the specified "COMPONENT" directory for builds, resolving the relative paths specified with `replace` directive fails during `go mod download` operation. Thus, replaced relative references with latest available package versions.

build-tooling.mk
```diff
@@ -144,37 +152,37 @@ publish:
...
 # Build the binary
 binary-build:
-	$(DOCKER) build . -f Dockerfile --build-arg LD_FLAGS="$(LD_FLAGS)" --target bin --output build/$(COMPONENT)/bin --platform ${PLATFORM} --build-arg COMPONENT=$(COMPONENT) --build-arg GOPROXY_ARG=$(GOPROXY)
+	$(DOCKER) build . -f $(COMPONENT_DOCKERFILE) --build-arg LD_FLAGS="$(LD_FLAGS)" --target bin --output build/$(COMPONENT)/bin --platform ${PLATFORM} --build-arg COMPONENT=$(COMPONENT) --build-arg GOPROXY_ARG=$(GOPROXY)
 
 .PHONY: docker-build
 # Build docker image
 docker-build:
-	$(DOCKER) build . -t $(IMAGE) -f Dockerfile --target image --platform linux/amd64 --build-arg LD_FLAGS="$(LD_FLAGS)" --build-arg COMPONENT=$(COMPONENT) --build-arg GOPROXY_ARG=$(GOPROXY)
+	$(DOCKER) build . -t $(IMAGE) -f $(COMPONENT_DOCKERFILE) --target image --platform linux/amd64 --build-arg LD_FLAGS="$(LD_FLAGS)" --build-arg COMPONENT=$(COMPONENT) --build-arg GOPROXY_ARG=$(GOPROXY)
```
sathyanarays, line 186: Do we have a clean-up step where we remove the generated component dockerfiles?

template-dockerfile.sh
```diff
@@ -0,0 +1,51 @@
...
+# Copy base Dockerfile to create component-specific Dockerfile
+# eg: featuregates/controller -> featuregates.controller.Dockerfile
+COMPONENT_DOCKERFILE=${COMPONENT////.}.Dockerfile
+cp ${BASE_DOCKERFILE} ${COMPONENT_DOCKERFILE}
+
+# Get relative references from replace directive statements
+# Only those replace statements that start with '.'
+replace_refs=($(awk -F ' => ' '{ if($2~/^\./) print $2 }' ${go_mod_file}))
+
+if [ ${#replace_refs[@]} -eq 0 ]; then
```
sathyanarays, line 16: Do we need this special case?

build-tooling.mk
avi-08: These files are added in dockerignore and gitignore. It can also be useful for debugging if something fails.

template-dockerfile.sh
```diff
@@ -0,0 +1,51 @@
...
+done
+
+# Add COPY commands into component-specific Dockerfile
+printf -v copy_str "%s," "${copy_statements[@]}"
+copy_str=${copy_str%?}
+
+awk -v text="${copy_str}" '!found && /COPY/ { 
+	split(text,lines,",");
+	for (i in lines) print lines[i]; 
+	found = 1 } 1' ${COMPONENT_DOCKERFILE} > tmp_file && mv tmp_file ${COMPONENT_DOCKERFILE}
```
sathyanarays, line 49: Add comments explaining this operation. Does this look for the first copy statement and append the extra copy statements below it?

sathyanarays: Can we make this simpler?

avi-08: Yes, will add comment explaining this.

template-dockerfile.sh
avi-08: Its just an early exit condition to avoid additional computations if no valid replace statements are found. Since the awk command in line 46 will read the entire file to find the COPY pattern and replace nothing.

build-tooling.mk
yharish991: i think we can have a flag or something to keep the intermediate dockerfiles for debugging purposes, but let's delete the intermediate dockerfiles after the build process to keep the project root directory clean

template-dockerfile.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
```
yharish991, line 1: I like the approach, but my only worry is, will this tooling work on a windows machine? we are using shell scripts with Linux commands to generate dockerfiles

Dockerfile
```diff
@@ -11,36 +11,32 @@ ARG COMPONENT
 ARG GOPROXY_ARG
 ENV GOPROXY=${GOPROXY_ARG}
 WORKDIR /workspace
-COPY "$COMPONENT"/go.* ./
+COPY "$COMPONENT" "$COMPONENT"
```
yharish991, line 14: can we do `COPY "$COMPONENT"/go.* ./"$COMPONENT"/`, that way this base stage gets invalidated only when the module dependencies or the client module code changes

template-dockerfile.sh
avi-08: The current build tooling [Makefile](https://github.com/vmware-tanzu/build-tooling-for-integrations/blob/main/templates/Makefile) fails to run in a windows environment. Is the expectation to enable build-tooling cross-platform?

build-tooling.mk
avi-08: Makes sense. I'll add a flag to toggle file deletions and delete by default.

template-dockerfile.sh
avi-08: Added comments, please verify.

build-tooling.mk
avi-08: Added the `KEEP_COMPONENT_DOCKERFILE` arg (defaults to false). If set to true, retains the component Dockerfiles.

Dockerfile
avi-08: If we consider the "base" layer to provide set of prerequisites to subsequent layers, then I think it may be good to copy the entire component directory. If we think of the "base" layer to be strictly used for module dependency resolutions, then every other layer will need to COPY the directory. Thoughts?

template-dockerfile.sh
sathyanarays: If the build tooling in its current state is not compatible with the windows environment, I think, making it compatible should be tracked as a different GH issue.

avi-08: +1

yharish991: we have an issue created for that https://github.com/vmware-tanzu/build-tooling-for-integrations/issues/51

Dockerfile
avi-08: Added changes to copy component code within a separate layer to avoid mounting entire context and keeping the base layer strictly around dependency resolution.

template-dockerfile.sh
sathyanarays: Thanks @yharish991! So, should we track the task of making the newly added script windows compatible in https://github.com/vmware-tanzu/build-tooling-for-integrations/issues/51

build-tooling.mk
sathyanarays: Let's say we are running this in pipelines and want to debug issues with the generated dockerfiles, simply keeping them might not be enough. Even if we keep, how do we access it? Instead, can we print the generated dockerfiles for each component on the console?

template-dockerfile.sh
sathyanarays: Looks good!

Dockerfile
```diff
@@ -11,47 +11,46 @@ ARG COMPONENT
 ARG GOPROXY_ARG
```
sathyanarays, line 11: Please add a GH issue in the build toolings project to backport these changes

build-tooling.mk
avi-08: Makes sense, I think a better way would be to get these files as pipeline artifacts. Since, the flag can be toggled only if we configure the argument to true, which happens in pipeline configs. Thoughts ?

sathyanarays: Thanks! If there are ways to get the generated files as artifacts, then we are good. No need of printing in console.

Dockerfile
sathyanarays: Thank you!

sathyanarays (approved):

build-tooling.mk
```diff
@@ -108,10 +118,10 @@ $(COMPONENTS):
 	@if [ "$(PUBLISH_IMAGES)" = "true" ]; then \
 		if [ "$(call check_main_go,$(COMPONENT))" = "Found" ]; then \
 			$(MAKE) validate-component IMAGE_NAME=$(IMAGE_NAME) PACKAGE_PATH=$(PACKAGE_PATH) || exit 1; \
-			$(MAKE) publish IMAGE=$(IMAGE) DEFAULT_IMAGE=$(DEFAULT_IMAGE) PACKAGE_PATH=$(PACKAGE_PATH) BUILD_BIN=$(BUILD_BIN); \
+			$(MAKE) publish cleanup-component-dockerfile IMAGE=$(IMAGE) DEFAULT_IMAGE=$(DEFAULT_IMAGE) PACKAGE_PATH=$(PACKAGE_PATH) BUILD_BIN=$(BUILD_BIN); \
```
yharish991, line 121: I think we won't have a need to cleanup the component dockerfile here

build-tooling.mk
```diff
@@ -144,37 +154,43 @@ publish:
...
 
 .PHONY: vet
 # Perform static analysis of code
 vet:
-	$(DOCKER) build . -f Dockerfile --target vet --build-arg COMPONENT=$(COMPONENT) --build-arg GOPROXY_ARG=$(GOPROXY)
+	$(DOCKER) build . -f $(COMPONENT_DOCKERFILE) --target vet --build-arg COMPONENT=$(COMPONENT) --build-arg GOPROXY_ARG=$(GOPROXY)
 
 .PHONY: test
 # Run tests
 test: fmt vet
```
yharish991, line 175: previously if I wanted to run individual make targets like for example `test`, I could easily do that by running `make test COMPONENT=/path/to/foo`, but now I also need to pass the path to component dockerfile by setting `COMPONENT_DOCKERFILE` var which is computed somewhere in the top level make target, can we provide a similar experience as before to run individual make targets?

avi-08: Added changes to support running targets individually. IMO, this is a bit complicated scenario. 
Due to multiple permutations and nested order in which last mile targets like "fmt" and "vet" could be run:

- `make fmt`
- `make test` -> Triggers fmt
- `make docker-build-all` -> Triggers build -> test -> fmt

It's difficult to maintain and pass a state flag like `KEEP_COMPONENT_DOCKERFILE`, which should be conditionally set based on which nested level the target was triggered at. Eg, on triggering `make test` directly, fmt target should not perform dockerfile cleanup, even if `KEEP_COMPONENT_DOCKERFILE` is set to false globally. To determine whether current target is the final state / its yet another level in the nested invocation (docker-build-all) is not feasible using a single state variable. Thus, introduced another flag `SKIP_INTERMEDIATE_CLEANUP` to handle this.

avi-08: Another option as suggested by @sathyanarays is to make the `cleanup-component-dockerfile` target independent of other target invocations, and we expect the users to explicitly trigger it. I think this is a valid solution to keep things simple.

build-tooling.mk
avi-08: Resolving as these changes are not required now

template-dockerfile.sh
avi-08: Resolving as these changes are not required now

yharish991 (approved):
LGTM!! thanks for the changes

build-tooling.mk
avi-08: Added these changes into build-tooling repo branch for future reference: https://github.com/vmware-tanzu/build-tooling-for-integrations/commit/827a5ae9e162201e422ca603cc6565f2e41a25d7

PR #4453, opened 2023-03-06
Adding env test to featuregate controller

sathyanarays:
Adds env test for featuregate controller

Fixes #

Tested locally and will be tested in pipelines

### Release note

```release-note

```


yharish991 (commented):
looks like you added env tests for controller reconciling on Feature and FeatureGate resources from config API group, but we deprecated those APIs. Can you add env tests for controller reconciling on Feature and FeatureGate resources from core API group?

codegold79 (dismissed):
Test looks good to me. Thanks for putting this in.

featuregates/Makefile
```diff
@@ -19,10 +19,22 @@ CRD_OPTIONS ?= "crd"
 
 all: manager
 
+ENVTEST_K8S_VERSION = 1.26.1
+ENVTEST ?= $(LOCALBIN)/setup-envtest
+
+LOCALBIN ?= $(shell pwd)/bin
```
yharish991, line 25: binaries should be downloaded to `hack/tools/bin` in project root directory, that's where we have been downloading the binaries needed for tanzu-framework

featuregates/controller/pkg/feature/feature_suite_test.go
```diff
@@ -0,0 +1,396 @@
...
+	ctx, cancel = context.WithCancel(context.TODO())
+
+	By("bootstrapping test environment")
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths:     []string{filepath.Join("..", "..", "..", "..", "apis", "core", "config", "crd", "bases")},
+		ErrorIfCRDPathMissing: true,
+	}
+
+	var err error
+	cfg, err = testEnv.Start()
```
yharish991, line 55: can you also configure the API server in the testEnv with featuregate webhooks?

featuregates/controller/pkg/feature/feature_suite_test.go
```diff
@@ -0,0 +1,396 @@
...
+			Name: fmt.Sprintf("featuregate-%v", rand.Intn(10000)),
+		},
+		Spec: corev1alpha2.FeatureGateSpec{
+			Features: []corev1alpha2.FeatureReference{},
+		},
+		Status: corev1alpha2.FeatureGateStatus{},
+	}
+}
+
+func getTestFeature(namespace string) *corev1alpha2.Feature {
```
yharish991, line 105: Feature is a cluster scoped resource, so namespace is not needed

featuregates/controller/pkg/feature/feature_suite_test.go
```diff
@@ -0,0 +1,396 @@
...
+	}
+}
+
+var _ = Describe("Featuregate controller", func() {
+	It("Should not activate experimental features by default", func() {
+		feature := getTestFeature("default")
+		Expect(k8sClient.Create(ctx, feature)).Should(Succeed())
+
+		Eventually(func() bool {
+			err := k8sClient.Get(ctx, types.NamespacedName{Name: feature.Name, Namespace: "default"}, feature)
```
yharish991, line 123: same, namespace is not needed, applies to other places where namespace is used

featuregates/controller/pkg/feature/feature_suite_test.go
```diff
@@ -0,0 +1,396 @@
...
+			return err == nil && feature.Status.Activated == false
+		}, timeout, interval).Should(BeTrue())
+
+		Expect(k8sClient.Delete(ctx, feature)).Should(BeNil())
+		Expect(k8sClient.Delete(ctx, featureGate)).Should(BeNil())
+	})
+
+	It("Should activate and de-activate experimental features when permenantly voided all support guarantees", func() {
```
yharish991, line 165: nit:
```suggestion
	It("Should activate and de-activate experimental features when permanently voided all support guarantees", func() {
```

yharish991 (commented):
can you also configure the validation webhooks in the env test to depict the actual behavior of Feature and FeatureGate API?

featuregates/Makefile
sathyanarays: Done

featuregates/controller/pkg/feature/feature_suite_test.go
sathyanarays: Done. Please verify once if the changes look ok!

featuregates/controller/pkg/feature/feature_suite_test.go
sathyanarays: Done

featuregates/controller/pkg/feature/feature_suite_test.go
sathyanarays: removed

featuregates/controller/pkg/feature/feature_suite_test.go
```diff
@@ -0,0 +1,553 @@
...
+	testEnv   *envtest.Environment
+	ctx       context.Context
+	cancel    context.CancelFunc
+	setupLog  = ctrl.Log.WithName("controllers").WithName("FeatureGates")
+
+	timeout  = 5 * time.Second
+	interval = 100 * time.Millisecond
+)
+
+func generateCertificateAndManifests() error {
```
rajathagasthya, line 65: Are there any utilities to create certs? Addons module does something similar, so you might want to look into it.

apis/core/v1alpha2/featuregate_types.go
```diff
@@ -75,6 +76,7 @@ type FeatureGate struct {
 	Spec FeatureGateSpec `json:"spec,omitempty"`
 	// Status reports activation state and availability of features in the system.
 	Status FeatureGateStatus `json:"status,omitempty"`
+	config *rest.Config      `json:"-"`
```
rajathagasthya, line 79: We should not have any API changes to support tests.

featuregates/controller/pkg/feature/feature_suite_test.go
```diff
@@ -0,0 +1,553 @@
...
+	for i, line := range lines {
+		if strings.Contains(line, "Cg==") {
+			lines[i] = strings.Replace(lines[i], "Cg==", base64.StdEncoding.EncodeToString(caBundleBytes), 1)
+		}
+	}
+	output := strings.Join(lines, "\n")
+	err = ioutil.WriteFile("/tmp/fearuregate-webhook.yaml", []byte(output), 0644)
+	if err != nil {
+		log.Fatalln(err)
+	}
```
rajathagasthya, line 134: I think a cleaner approach would be to craft appropriate Go objects in memory and manipulate them. We are reading and writing objects to a file only to decode them later as unstructured objects for creation.

rajathagasthya (changes requested):
First pass review. I haven't verified the webhook stuff fully, but it is in the right direction.

featuregates/controller/pkg/feature/feature_suite_test.go
sathyanarays: That's correct! I will explore options.

apis/core/v1alpha2/featuregate_types.go
sathyanarays: It seems not possible to have envtest without this change. 

When we create a feature gate resource from envtest, [ValidateCreate](https://github.com/vmware-tanzu/tanzu-framework/blob/99dd6215d05ae98aa5b21f0553f4223a2c25e608/apis/core/v1alpha2/featuregate_webhook.go#LL69C23-L69C37) gets called. From `ValidateCreate`, [getClient](https://github.com/vmware-tanzu/tanzu-framework/blob/99dd6215d05ae98aa5b21f0553f4223a2c25e608/apis/core/v1alpha2/featuregate_webhook.go#L40) gets called. In `getClient`, the [in-cluster config](https://github.com/vmware-tanzu/tanzu-framework/blob/99dd6215d05ae98aa5b21f0553f4223a2c25e608/apis/core/v1alpha2/featuregate_webhook.go#LL50C2-L50C2) is obtained for the client. 

So, when we run envtest without this change, the webhook gets the client using in-cluster config and tries to query FeatureGate resource. That errors out. So, this change seems to be necessary. Please let me know if there are alternate approaches. 

Also, this change does not export any new fields. So, this should be fine.

featuregates/controller/pkg/feature/feature_suite_test.go
sathyanarays: Done. Please verify if it makes sense.

apis/core/v1alpha2/featuregate_types.go
sathyanarays: I think it is a valid point. I looked at other webhooks. They maintain two different structs for webhook and the API types so that any data required for webhook does not bring in API change. Let me have another look at this problem.

sathyanarays: Reverted API changes.

featuregates/controller/pkg/feature/feature_suite_test.go
sathyanarays: The util used by addons is in addons module. If we import, we will be adding a go mod dependency on addon module and that is not necessary. I think what we are doing here to create the certificates are straightforward and no need of using utils from other modules. Let me know if there are other approaches.

yharish991: I think it's better we don't import stuff from addons module as it would be evicted from t-f soon, maybe we can move this code to [util module](https://github.com/vmware-tanzu/tanzu-framework/tree/main/util) for reusability purposes in other controllers

featuregates/controller/pkg/feature/feature_suite_test.go
```diff
@@ -0,0 +1,565 @@
...
+		Expect(feature.Status.Activated).Should(Equal(false))
+
+		featureGate.Spec.Features[0].Activate = true
+		Expect(k8sClient.Update(ctx, featureGate)).ShouldNot(Succeed())
+
+		Expect(k8sClient.Delete(ctx, feature)).Should(BeNil())
+		Expect(k8sClient.Delete(ctx, featureGate)).Should(BeNil())
+	})
+
+	It("Should activate and de-activate experimental features when permanently voided all support guarantees", func() {
```
yharish991, line 297: can you also add a test where you try to set `PermanentlyVoidAllSupportGuarantees` to false after setting it to true initially?

featuregates/controller/pkg/feature/feature_suite_test.go
```diff
@@ -0,0 +1,565 @@
...
+		featureGate.Spec.Features = append(featureGate.Spec.Features, corev1alpha2.FeatureReference{
+			Name:     feature.Name,
+			Activate: false,
+		})
+		Expect(k8sClient.Create(ctx, featureGate)).ShouldNot(Succeed())
+
+		Expect(k8sClient.Delete(ctx, feature)).Should(BeNil())
+	})
+
+	It("Should activate deprecated features by default and stable features should be mutable", func() {
```
yharish991, line 424:
```suggestion
	It("Should activate deprecated features by default and deprecated features should be mutable", func() {
```

featuregates/controller/pkg/feature/feature_suite_test.go
```diff
@@ -0,0 +1,565 @@
...
+		Eventually(func() bool {
+			err := k8sClient.Get(ctx, types.NamespacedName{Name: feature.Name}, feature)
+			return err == nil && feature.Status.Activated == false
+		}, timeout, interval).Should(BeTrue())
+
+		Expect(k8sClient.Delete(ctx, feature)).Should(BeNil())
+		Expect(k8sClient.Delete(ctx, featureGate)).Should(BeNil())
+	})
+
+	It("Should remove feature from featuregate status when reference is removed", func() {
```
yharish991, line 467: this test should also check if the status of the feature is the default activation state

featuregates/controller/pkg/feature/feature_suite_test.go
```diff
@@ -0,0 +1,565 @@
...
+			return err == nil && len(featureGate.Status.FeatureReferenceResults) == 0
+		}, timeout, interval).Should(BeTrue())
+
+		Expect(k8sClient.Delete(ctx, feature)).Should(BeNil())
+		Expect(k8sClient.Delete(ctx, featureGate)).Should(BeNil())
+	})
+})
+
+func createResources(cfg *rest.Config, dynamicClient dynamic.Interface) error {
+	data := generatedWebhookManifestBytes
```
yharish991, line 508: question: is there any reason to assign generatedWebhookManifestBytes to a new variable here?

yharish991 (changes requested):

featuregates/controller/pkg/feature/feature_suite_test.go
sathyanarays: Done

featuregates/controller/pkg/feature/feature_suite_test.go
```diff
@@ -0,0 +1,565 @@
...
+	tmpDir, err = os.MkdirTemp("/tmp", "featuregate-test")
+	if err != nil {
+		return err
+	}
+	err = os.WriteFile(path.Join(tmpDir, "feature-webhook.crt"), caBundleBytes, 0644)
+	if err != nil {
+		return err
+	}
+
+	err = os.WriteFile(path.Join(tmpDir, "feature-webhook.key"), keyPem.Bytes(), 0644)
```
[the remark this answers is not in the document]
sathyanarays: Done

featuregates/controller/pkg/feature/feature_suite_test.go
sathyanarays: Not applicable since moved the code to util

featuregates/controller/pkg/feature/feature_suite_test.go
```diff
@@ -0,0 +1,565 @@
...
+			Name: fmt.Sprintf("featuregate-%v", randomNumber),
+		},
+		Spec: corev1alpha2.FeatureGateSpec{
+			Features: []corev1alpha2.FeatureReference{},
+		},
+		Status: corev1alpha2.FeatureGateStatus{},
+	}
+}
+
+func getTestFeature() *corev1alpha2.Feature {
```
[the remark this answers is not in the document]
sathyanarays: Why is this required?

featuregates/controller/pkg/feature/feature_suite_test.go
sathyanarays: Done

featuregates/controller/pkg/feature/feature_suite_test.go
sathyanarays: Done

featuregates/controller/pkg/feature/feature_suite_test.go
sathyanarays: Done

featuregates/controller/pkg/feature/feature_suite_test.go
```diff
@@ -0,0 +1,565 @@
...
+		CRDDirectoryPaths:     []string{filepath.Join("..", "..", "..", "..", "apis", "core", "config", "crd", "bases")},
+		ErrorIfCRDPathMissing: true,
+	}
+
+	var err error
+	cfg, err = testEnv.Start()
+	Expect(err).NotTo(HaveOccurred())
+	Expect(cfg).NotTo(BeNil())
+
+	testEnv.ControlPlane.APIServer.Configure().Append("admission-control", "MutatingAdmissionWebhook")
```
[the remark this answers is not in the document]
sathyanarays: Done

featuregates/controller/pkg/feature/feature_suite_test.go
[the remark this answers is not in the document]
yharish991: so that you don't have to do [explicitly set the stability level](https://github.com/vmware-tanzu/tanzu-framework/pull/4453/files#diff-d067ce787aa01aa3bc0f7055d55a60592af39dd232a5e3697e1b8bfb1222a8f7R303) in each test case, you can just pass the stability level to the function and the function will return feature with that stability level, it also makes sense to rename this function to `getTestFeatureWithStabilityLevel` or something similar

featuregates/controller/pkg/feature/feature_suite_test.go
```diff
@@ -0,0 +1,458 @@
...
+	RunSpecs(t, "Featuregate Suite")
+}
+
+var (
+	cfg       *rest.Config
+	k8sClient client.Client
+	testEnv   *envtest.Environment
+	ctx       context.Context
+	cancel    context.CancelFunc
+	setupLog  = ctrl.Log.WithName("controllers").WithName("FeatureGates")
```
yharish991, line 51:
```suggestion
	setupLog  = ctrl.Log.WithName("controllers").WithName("Features")
```
since this is for Feature controller

featuregates/controller/pkg/feature/feature_suite_test.go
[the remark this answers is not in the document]
sathyanarays: Done

featuregates/controller/pkg/feature/feature_suite_test.go
sathyanarays: Done

util/cert/cert_util.go
```diff
@@ -0,0 +1,66 @@
...
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"math/big"
+	"time"
+)
+
+func GenerateSelfSignedCertsForTest(dnsName string) ([]byte, []byte, error) {
```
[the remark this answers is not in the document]
sathyanarays: Done

avi-08 (approved):
LGTM

yharish991 (approved):
LGTM 👍

rajathagasthya (approved):
Thanks for the changes!

rajathagasthya:
`make verify` is failing. Looks like it needs a `go mod tidy`.

PR #4458, opened 2023-03-07
Fix tanzu-auth-controller user to refer by id instead of name and fix psa profile

chrischdi:

Kubernetes is only able to detect runAsNonroot if either the given user in the image is referred by id or by specifying the user id directly in the pod spec.

Also, if `hostNetwork: true` is set, the pod security profile used will be privilege. Because of that I added the required label to the namespace so the workload does not get blocked by cluster-wide configuration.

Follow-up for #4451

Fixes #

### Release note

```release-note

```

### DIFF
--- a/packages/tanzu-auth/bundle/config/namespace.yaml
+++ b/packages/tanzu-auth/bundle/config/namespace.yaml
@@ -6,4 +6,7 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: #@ data.values.namespace
+  labels:
+    #@ if/end data.values.deployment.hostNetwork:
+    pod-security.kubernetes.io/enforce: privileged
 #@ end
--- a/pinniped-components/tanzu-auth-controller-manager/Dockerfile
+++ b/pinniped-components/tanzu-auth-controller-manager/Dockerfile
@@ -38,5 +38,5 @@ RUN --mount=type=cache,target=/root/.cache/go-build --mount=type=cache,target=/r
 FROM $DISTROLESS_BASE_IMAGE
 WORKDIR /
 COPY --from=builder /workspace/tanzu-auth-controller-manager .
-USER nonroot:nonroot
+USER 65532:65532
 ENTRYPOINT ["/tanzu-auth-controller-manager"]


alfredthenarwhal:
> @chrischdi:  /test install-vc7
Commit: 45ddaed

Tests failed! Build no: 4368

chrischdi:
> > @chrischdi:  /test install-vc7
> > Commit: [45ddaed](https://github.com/vmware-tanzu/tanzu-framework/commit/45ddaed4a77f2ee65790058d954dc3f2b5f1775d)
> 
> Tests failed! Build no: 4368

retry-build + test succeeded 🎉

joshuatcasey (approved):
This looks good to me. See https://github.com/vmware-tanzu/tanzu-framework/pull/4451/files for some additional PSA changes for the `tanzu-auth-controller`.

chrischdi:
> This looks good to me. See https://github.com/vmware-tanzu/tanzu-framework/pull/4451/files for some additional PSA changes for the `tanzu-auth-controller`.

Thanks for taking a look @joshuatcasey .

Regarding `tanzu-auth`: I think it should also have the namespace label part depending on the hostNetwork setting at https://github.com/vmware-tanzu/tanzu-framework/blob/664cdef2e73d42fbf74add77c8ab788181a26653/packages/tanzu-auth/bundle/config/namespace.yaml:

```yaml
  labels:
    #@ if/end data.values.deployment.hostNetwork:
    pod-security.kubernetes.io/enforce: privileged
```

I'd propose to do this in a separate PR to unblock tanzu-framework with this PR.

joshuatcasey (changes requested):
I think this can be `restricted` instead of `privileged` since nothing in this namespace requires privileges.

joshuatcasey:
@chrischdi I did request a change to the label since nothing in this namespace requires privileges.

chrischdi:
> @chrischdi I did request a change to the label since nothing in this namespace requires privileges.

The idea is:
* when the pod gets set `hostNetwork: true`
* the namespace should get labeled to the privileged profile.

Becuase `hostNetwork: true` is not allowed in the `baseline` or `restricted` profile and thus only allowed in the `privileged` PSA profile.

Please note: privileged PSA Profile != `privileged: true`

chrischdi:
@joshuatcasey TL/DR is: when you set `hostNetwork: true` it **must** be the `privileged` profile on the namespace. Otherwise the pod gets blocked from running, if a customer enforces `restricted` cluster-wide.

wjun (approved):
LGTM

joshuatcasey (approved):

benjaminapetersen (approved):
Spiffy. 🥳

PR #4459, opened 2023-03-07
getGVR should only fail if no GroupVersionResource is found

### LINKED ISSUE #4460: ClusterBootstrap Webhook deny create request if any API service is not available
**Bug description**
If not all gvr are available at the clusterbootrap create validation, the webhook erronously denies the request. 

- ( ) APIs
- ( x) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

adduarte:
The call to ServerPreferredResources could return an err != nil even though the resource list is not empty.

getGVR should not report a failure if the resource list does contain a matching GroupVersionResource, even if ServerPreferredResources returned err != nil.

This patch modifies  getGVR so it  only return a error (failure) if it cannot find a matching GroupVersionResource. 

Fixes #4460

Unit test created and executed.
ran deployed to vsphere 7 platform. 

### Release note

```release-note
Fixes bug in clusterbootstrap webhook which incorrectly failed create validation if not all api service endpoints are running at the time of creation.
```


addons/webhooks/clusterbootstrap_webhook.go
```diff
@@ -292,7 +292,12 @@ func (wh *ClusterBootstrap) getGVR(gk schema.GroupKind) (*schema.GroupVersionRes
 	}
 	apiResourceList, err := wh.cachedDiscoveryClient.ServerPreferredResources()
 	if err != nil {
-		return nil, err
+		if apiResourceList == nil {
+			return nil, err
+		}
```
adduarte, line 297: The comment by codecov is incorrect. The lines are covered by test added in this patch. Verified by stepping through with debugger

addons/webhooks/clusterbootstrap_webhook.go
```diff
@@ -292,7 +292,12 @@ func (wh *ClusterBootstrap) getGVR(gk schema.GroupKind) (*schema.GroupVersionRes
...
+		if _, ok := err.(*discovery.ErrGroupDiscoveryFailed); !ok {
+			return nil, err
+		}
```
adduarte, line 300: The comment by codecov is incorrect. The lines are covered by test added in this patch. Verified by stepping through with debugger

addons/webhooks/clusterbootstrap_webhook.go
```diff
@@ -292,7 +292,12 @@ func (wh *ClusterBootstrap) getGVR(gk schema.GroupKind) (*schema.GroupVersionRes
 	}
 	apiResourceList, err := wh.cachedDiscoveryClient.ServerPreferredResources()
 	if err != nil {
-		return nil, err
+		if apiResourceList == nil {
```
adduarte, line 295: I think we can skip this check since we do iterate over apiResourceList further down the code. and if apiresourceList == nil, we return an error there.

addons/webhooks/clusterbootstrap_webhook.go
```diff
@@ -291,7 +291,7 @@ func (wh *ClusterBootstrap) getGVR(gk schema.GroupKind) (*schema.GroupVersionRes
 		return gvr, nil
 	}
 	apiResourceList, err := wh.cachedDiscoveryClient.ServerPreferredResources()
-	if err != nil {
+	if _, ok := err.(*discovery.ErrGroupDiscoveryFailed); err != nil && !ok {
```
adduarte, line 294: if the error is ErrGroupDiscoveryFailed, we can safely continue, since an apiresult == nil will be catch later down the code and we will return the correct result if we don't find what we are looking for in the contents of apiresult

shyaamsn (commented):
LGTM

addons/webhooks/clusterbootstrap_webhook.go
maralavi: Since you didn't check the error on line 293, wouldn't the following cast fail in case of err==nil?
```err.(*discovery.ErrGroupDiscoveryFailed)````

adduarte: no. I checked by running this code
```func main() {
	var err error
	if _, ok := err.(*discovery.ErrGroupDiscoveryFailed); !ok && err != nil {
		println("Error is not be ")
	}
	if err == nil {
		println("Error is nil")
	}
}

maralavi: Agreed, you are right, ```ok``` captures that case here

alfredthenarwhal:
> @adduarte:  /test install-vc7
Commit: 9a52c4c

Tests failed! Build no: 4398

alfredthenarwhal:
> @adduarte:  /test install-vc7
Commit: 9a52c4c

Tests failed! Build no: 4404

alfredthenarwhal:
> @adduarte:  /test install-vc7
Commit: 85afe58

Tests failed! Build no: 4440

alfredthenarwhal:
> @adduarte:  /test install-vc7
Commit: 85afe58

Tests failed! Build no: 4443

alfredthenarwhal:
> @adduarte:  /test install-vc7
Commit: fc33b8f

Tests passed! Build no: 4446

adduarte:
install-vc7 test succeeded: https://gitlab.eng.vmware.com/TKG/bolt/bolt-release-yamls/-/merge_requests/4446

maralavi (approved):

PR #4474, opened 2023-03-14
Fix deadlock for mgmt cluster upgrade and tkr fetch

12345lcr:
-  Fix the management cluster upgrade deadlock that
     A. tkr-source-controller is fetching tkrs which are compatible with old tkg version, since the TKGVERSION annotation is not updated.
     B. The TKGVERSION annotation should be updated after cluster upgrade, but the cluster cannot be upgraded due to the missing of the target new tkrs
- Extend the k8s version patch timeout to 15 mins which is longer than the fetcher's frequency of pulling tkrs.
- Put the configmap update before the deployment in tkr-source-controller, to prevent old configurations in configmaps got loaded into the updated  tkr-source-controller pods.

Fixes #

### Release note

```release-note

```

### DIFF
--- a/packages/tkr-source-controller/bundle/config/upstream/tkr-source-controller-deployment.yaml
+++ b/packages/tkr-source-controller/bundle/config/upstream/tkr-source-controller-deployment.yaml
@@ -17,6 +17,7 @@ metadata:
   annotations:
     kapp.k14s.io/change-rule.0: "upsert after upserting tkr-source-controller.tanzu.vmware.com/ClusterRoleBinding"
     kapp.k14s.io/change-rule.1: "delete before deleting tkr-source-controller.tanzu.vmware.com/ClusterRoleBinding"
+    kapp.k14s.io/change-rule.3: "upsert after upserting tkr-source-controller.tanzu.vmware.com/ConfigMap"
 spec:
   replicas: 1
   selector:
--- a/tkg/client/upgrade_cluster_clusterclass.go
+++ b/tkg/client/upgrade_cluster_clusterclass.go
@@ -5,6 +5,7 @@ package client
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/pkg/errors"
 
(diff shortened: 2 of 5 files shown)


alfredthenarwhal:
> @12345lcr:  /test upgrade-vc7
Commit: 1268293

Tests failed! Build no: 4416

alfredthenarwhal:
> @12345lcr:  /test upgrade-vc7
Commit: de902d1

alfredthenarwhal:
> @12345lcr:  /test upgrade-vc7
Commit: de902d1

Tests failed! Build no: 4439

alfredthenarwhal:
> @12345lcr:  /test upgrade-vc7
Commit: c4aa8f3

Tests failed! Build no: 4460

tkg/client/upgrade_cluster_clusterclass.go
```diff
@@ -21,7 +22,10 @@ func (c *TkgClient) DoClassyClusterUpgrade(regionalClusterClient clusterclient.C
 	log.Infof("Upgrading kubernetes cluster to `%v` version, tkr version: `%s`", kubernetesVersion, tkrVersion)
 	patchJSONString := fmt.Sprintf(`{"spec": {"topology": {"version": "%v"}}}`, tkrVersion)
 
-	err := regionalClusterClient.PatchClusterObject(options.ClusterName, options.Namespace, patchJSONString)
+	// Timeout set to 15 minutes because the continuousTKRDiscoverFreq for tkr-source-controller's fetcher is 10 minutes.
```
wjun, line 25: This PR looks good to me. I just cannot understand how timeout is related to the tkr fetcher's frequency.

12345lcr: https://github.com/vmware-tanzu/tanzu-framework/blob/decac8230fd187b1e9ac19d786f32358b37298e2/tkr/controller/tkr-source/main.go#L71

In tkr-source-controller the fetcher tried to sync the tkrs from registry every 10 mins. This part was meant to cover at least 1 frequency period to let fetcher do a fetch action(and download) if the desired tkrs are not found.

wjun: @hongtaiwang @12345lcr  why the timeout is related to the fetcher frequency here?

packages/tkr-source-controller/bundle/config/upstream/tkr-source-controller-deployment.yaml
```diff
@@ -17,6 +17,7 @@ metadata:
   annotations:
     kapp.k14s.io/change-rule.0: "upsert after upserting tkr-source-controller.tanzu.vmware.com/ClusterRoleBinding"
     kapp.k14s.io/change-rule.1: "delete before deleting tkr-source-controller.tanzu.vmware.com/ClusterRoleBinding"
+    kapp.k14s.io/change-rule.3: "upsert after upserting tkr-source-controller.tanzu.vmware.com/ConfigMap"
```
hongtaiwang, line 20: should we add `kapp.k14s.io/change-rule.3: "delete before deleting tkr-source-controller.tanzu.vmware.com/ConfigMap"` as well?

tkg/client/upgrade_cluster_clusterclass.go
hongtaiwang: By patching `TKGVERSION` annotation before tkr-source-controller creation, we do not need to extend this time out, right?

wjun (approved):
LGTM.

packages/tkr-source-controller/bundle/config/upstream/tkr-source-controller-deployment.yaml
wjun: Per discussion, there is no deletion triggered.

tkg/client/upgrade_cluster_clusterclass.go
wjun: Per discussion, in worst case, this timeout change can still make the tkr image download safer.

PR #4480, opened 2023-03-15
Change capz pipeline to create cc mc and legacy wc

wjun:
Since we will not support legacy management cluster anymore, change capz pipeline
to create a cc based mc with legacy workload clusters.
Update cronjobs version from v1beta1 to v1 as v1beta is deprecated in k8s 1.25.


imikushin (approved):
Looks good

PR #4502, opened 2023-03-17
Configure Dex to issue refresh tokens with an expiry of 9h

joshuatcasey:
Configure Dex to issue refresh tokens with an expiry of 9h.

This is to support using the more recent versions of Pinniped, which need a relatively long-lived refresh token from Dex so that it can perform refreshes to verify authentication information.


benjaminapetersen (approved):
Fantastic.

benjaminapetersen (approved):

PR #4503, opened 2023-03-19
[WIP]Implement packages installation APIs for cluster essential

Vandy-P:
Implementing APIs for cluster essential installation 
 
Fixes #

### Release note

```release-note

```


util/kube/kube_util.go
Vandy-P, line 1: Will remove this file once @sathyanarays  [PR](https://github.com/vmware-tanzu/tanzu-framework/pull/4453) is merged

cluster-essentials/cluster_init.go
```diff
@@ -0,0 +1,76 @@
...
+		fileName, err := generateTemplateFiles(component)
+		if err != nil {
+			return errors.Wrapf(err, "unable to generate template")
+		}
+		input, err := os.ReadFile(fileName)
+		if err != nil {
+			return errors.Wrapf(err, "unable to read file %s", fileName)
+		}
+		generatedManifestBytes := []byte(string(input))
+		err = kube.CreateResourcesFromManifest(generatedManifestBytes, restConfig, dynamicClient)
```
avi-08, line 70: This function will invoke an equivalent of the `kubectl create` command which is not an idempotent operation. Lets say the `secretgen-controller` install fails due to some reason, and user triggers the cluster init flow again after fixing it, the kapp-controller `create` operation will fail, since resource already exists. Why not use `kubectl apply` equivalent instead?

cluster-essentials/cluster_init.go
avi-08, line 1: As per go naming conventions, underscores in filenames usually show a test file / platform specific file. This can be renamed to `clusterinit.go`

cluster-essentials/cluster_init.go
```diff
@@ -0,0 +1,76 @@
+// Copyright 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package clusteressential has APIs for bootstrapping kubernetes cluster
+package clusteressential
```
avi-08, line 5: Should be `clusteressentials`

cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,107 @@
...
+	dynamicClient, err = dynamic.NewForConfig(restConfig)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "unable to set up rest config")
+	}
+	return restConfig, dynamicClient, nil
+}
+
+// nolint: stylecheck
+func InitClusterEssentials(clusterKubeconfigPath, clusterEssentialBundlePath string) error {
+	restConfig, dynamicClient, err := getClientForConfig(clusterKubeconfigPath)
```
sathyanarays, line 58: Don't initialize the restConfig here. Take it as an argument.

cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,107 @@
...
+			if err != nil {
+				_ = UninstallClusterEssential(clusterKubeconfigPath, "tmp")
+				return errors.Wrapf(err, "unable to install %s", item.Name())
+			}
+		}
+	}
+	return nil
+}
+
+func UninstallClusterEssential(clusterKubeconfigPath, clusterEssentialBundlePath string) error {
```
sathyanarays, line 86: As discussed, this is not required.

cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/vmware-tanzu/tanzu-framework/util/carvelhelpers"
+	"github.com/vmware-tanzu/tanzu-framework/util/kube"
+)
+
+const dotCharacter = 46
+
+func downloadClusterEssentialBundle(clusterEssentialBundlePath string) error {
+	var outputBuf, errorBuf bytes.Buffer
+	writerUI := ui.NewWriterUI(&outputBuf, &errorBuf, nil)
+	outputDir := "tmp"
```
sathyanarays, line 28: This is being used in multiple places. Avoid that and use a constant so that any change to this can be done at a single place.

cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// nolint: stylecheck
+func InitClusterEssentials(clusterKubeconfigPath, clusterEssentialBundlePath string) error {
+	restConfig, dynamicClient, err := getClientForConfig(clusterKubeconfigPath)
+	if err != nil {
+		return err
+	}
+
+	err = downloadClusterEssentialBundle(clusterEssentialBundlePath)
```
sathyanarays, line 63: How is it going to be handled in unit tests?

util/carvelhelpers/package.go
```diff
@@ -0,0 +1,130 @@
...
+	// `config` directory contains ytt files
+	// `.imgpkg` directory contains ImageLock configuration for ImageResolution
+	configDir := filepath.Join(pkgDir, "config")
+	file := []string{configDir}
+	bytes, err := ProcessYTTPackage(file...)
+	if err != nil {
+		return nil, errors.Wrap(err, "error while running ytt")
+	}
+
+	f, err := os.CreateTemp("", "ytt-processed")
```
codegold79, line 57: Is the variable `f` short for `file`? If so, is it the same type as `file` in line 51? If it is, I'd say, just use `file`. Single letter variables should only be for very obvious things, and I'm not so familiar with `os.CreateTemp` to know what `f` is supposed to be.

Oh, I see that `file` in line 51 is a slice of string. In that case, that line 51 variable should be plural, `files := []string{configDir}`. And that frees up `file` to then be used here in line 57.

util/carvelhelpers/package.go
```diff
@@ -0,0 +1,130 @@
...
+		inputFilesForImageResolution = append(inputFilesForImageResolution, imgpkgDir)
+	}
+	return ResolveImagesInPackage(inputFilesForImageResolution)
+}
+
+// SaveFile saves the file to the provided path
+// Also creates missing directories if any
+func SaveFile(filePath string, data []byte) error {
+	dirName := filepath.Dir(filePath)
+	if _, serr := os.Stat(dirName); serr != nil {
```
codegold79, line 81: I can't tell, but looks like the `serr` information doesn't get passed along. Do you think that's all right? Maybe people don't need to know what `serr` is because you pass along `merr` later? Kind of strange, but I guess it can  work.

codegold79 (approved):
Seems pretty straight forward to me. I had a couple questions and minor, optional feedback. None block my approval, so I approve.

util/carvelhelpers/package.go
Vandy-P: @codegold79  using files as variable will cause lint error 
**_importShadow: shadow of imported from 'github.com/vmware-tanzu/carvel-ytt/pkg/files' package 'files'_**

cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/vmware-tanzu/tanzu-framework/util/kube"
+)
+
+var (
+	downloadPackage = downloadClusterEssentialBundle
+)
+
+const tmpDir = "tmp"
+
+func downloadClusterEssentialBundle(clusterEssentialBundlePath string) error {
```
avi-08, line 27: If downloading imgpkg bundles is also needed in operations like upgrade, it'd be good to create this as a utility function.

cluster-essentials/test/package1/config/overlay.yaml
sathyanarays, line 1: The folder name can be `testdata`

cluster-essentials/clusterinit_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+func downloadClusterEssentialBundleFailMock(clusterEssentialBundlePath string) error {
+	return fmt.Errorf("image not found")
+}
+
+func downloadClusterEssentialBundleSuccessMock(clusterEssentialBundlePath string) error {
+	return nil
+}
+
+func Test_InitClusterEssentials(t *testing.T) {
```
sathyanarays, line 26: Tests does not cover the success path.

cluster-essentials/clusterinit.go
sathyanarays: +1

danniel1205:
Question: Why do we have go mod change in other modules ? (Those modules seem unrelated to your change)

Vandy-P:
@danniel1205  We created  kube utility function [util/kube/kube_util.go ] which needs to use client-go v0.25.0 hence all the dependent packages which are using util package got updated with client-go version v0.25.0.

cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"k8s.io/client-go/rest"
+
+	clusterutil "github.com/vmware-tanzu/tanzu-framework/cluster-essentials/utilinterface"
+)
+
+type ClusterEssentialsConfiguration struct {
+	ClusterEssentialBundlePath string
+	RestConfig                 *rest.Config
+	PkgClient                  clusterutil.Client
```
sathyanarays, line 20: Do we need to export these variables? Can we have a function `NewClusterEssentialsClient` that takes in these arguments and returns `ClusterEssentialsConfiguration`?

cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,47 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"k8s.io/client-go/rest"
+
+	clusterutil "github.com/vmware-tanzu/tanzu-framework/cluster-essentials/utilinterface"
+)
+
+type ClusterEssentialsConfiguration struct {
```
sathyanarays, line 17: Can we rename `ClusterEssentialsConfiguration` to `ClusterEssentialsClient`

cluster-essentials/go.mod
```diff
@@ -0,0 +1,121 @@
+module github.com/vmware-tanzu/tanzu-framework/cluster-essentials
+
+go 1.18
+
+replace (
+	github.com/vmware-tanzu/tanzu-framework/cluster-essentials => ./
```
sathyanarays, line 6: Why is this required?

cluster-essentials/utilinterface/client.go
```diff
@@ -0,0 +1,28 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package utilinterface utilinterface defines util functions for cluster essential
+package utilinterface
```
sathyanarays, line 5: The package name doesn't look convincing. Can it be `util` or `helper`?

cluster-essentials/utilinterface/interface.go
```diff
@@ -0,0 +1,15 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+// Package utilinterface Client defines utility functions for cluster essential
+package utilinterface
+
+import (
+	"k8s.io/client-go/rest"
+)
+
+type Client interface {
```
sathyanarays, line 11: We can move both interface and implementation to the same file.

cluster-essentials/test/clusterinit_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/stretchr/testify/require"
+
+	clusteressentials "github.com/vmware-tanzu/tanzu-framework/cluster-essentials"
+	"github.com/vmware-tanzu/tanzu-framework/cluster-essentials/fakes"
+)
+
+var clusterConfig = &clusteressentials.ClusterEssentialsConfiguration{}
+var fake = &fakes.FakeClient{}
+
+func Test_InitClusterEssentials(t *testing.T) {
```
sathyanarays, line 22: Please add a test when reading from temp directory fails

cluster-essentials/test/clusterinit_test.go
```diff
@@ -0,0 +1,70 @@
...
+		fake.DownloadClusterEssentialBundleReturns(errors.New("could not parse reference"))
+		clusterConfig.PkgClient = fake
+		err := clusterConfig.InitClusterEssentials()
+		require.Error(t, err)
+		require.ErrorContains(t, err, "unable to download cluster essential manifest")
+	})
+	t.Run("fails when unable to process downloaded cluster essential bundle ", func(t *testing.T) {
+		fake.DownloadClusterEssentialBundleReturns(nil)
+		var result1 []byte
+		fake.CarvelPackageProcessorReturns(result1, errors.New("error while running ytt"))
```
sathyanarays, line 36: can we use `nil` here?

cluster-essentials/test/clusterinit_test.go
```diff
@@ -0,0 +1,70 @@
...
+		clusterConfig.PkgClient = fake
+		err := clusterConfig.InitClusterEssentials()
+		require.Error(t, err)
+		require.ErrorContains(t, err, "unable to process carvel package")
+	})
+	t.Run("fails when unable to process downloaded cluster essential bundle ", func(t *testing.T) {
```
sathyanarays, line 42: What's the difference between this test case and the previous one?

cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,47 @@
...
+	if err != nil {
+		return err
+	}
+	for _, item := range packages {
+		packageDir := filepath.Join(tmpDir, item.Name())
+		generatedManifestBytes, err := c.PkgClient.CarvelPackageProcessor(packageDir)
+		if err != nil {
+			return fmt.Errorf("unable to process carvel package: %w", err)
+		}
+		err = c.PkgClient.ApplyResourcesFromManifest(generatedManifestBytes, c.RestConfig)
```
sathyanarays, line 41: Let's say there are two packages in the bundle & the second one is corrupted. This results in the cluster having resources belonging to first package installed resulting in a partially installed state. Instead, we can start creating resources only if all the packages are processed successfully.

cluster-essentials/test/clusterinit_test.go
```diff
@@ -0,0 +1,70 @@
...
+		fake.DownloadClusterEssentialBundleReturns(nil)
+		fake.CarvelPackageProcessorReturns(nil, nil)
+		fake.ApplyResourcesFromManifestReturns(errors.New("failed to load kubeconfig"))
+		clusterConfig.PkgClient = fake
+		err := clusterConfig.InitClusterEssentials()
+		require.Error(t, err)
+		require.ErrorContains(t, err, "failed to load kubeconfig")
+	})
+
+	t.Run("Suessful ", func(t *testing.T) {
```
avi-08, line 61: typo

cluster-essentials/go.mod
Vandy-P: Forgot cleanup after doing some experiment.

cluster-essentials/test/clusterinit_test.go
Vandy-P: No difference , copy paste error.

util/carvelhelpers/package.go
codegold79: Oh, I see. Carvel stole the name `files` by naming one of their packages that. How unfortunate. Well, looks like what you got is the best I can think of to do. Resolving.

cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,58 @@
...
+
+type ClusterEssentialsClient struct {
+	clusterEssentialBundlePath string
+	restConfig                 *rest.Config
+	pkgClient                  clusterutil.Client
+}
+
+func NewClusterEssentialsClient(clusterEssentialBundlePath string, restConfig *rest.Config, pkgClient clusterutil.Client) *ClusterEssentialsClient {
+	return &ClusterEssentialsClient{
+		clusterEssentialBundlePath: clusterEssentialBundlePath,
```
sathyanarays, line 25: This should be an argument to `InitClusterEssentials`

cluster-essentials/test/clusterinit_test.go
```diff
@@ -0,0 +1,68 @@
...
+		require.Error(t, err)
+		require.ErrorContains(t, err, "unable to download cluster essential manifest")
+	})
+	t.Run("fails when unable to process downloaded cluster essential bundle ", func(t *testing.T) {
+		fake.DownloadClusterEssentialBundleReturns(nil)
+		fake.CarvelPackageProcessorReturns(nil, errors.New("error while running ytt"))
+		clusterConfig := clusteressentials.NewClusterEssentialsClient(bundleName, restConfig, fake)
+		err := clusterConfig.InitClusterEssentials()
+		require.Error(t, err)
+		require.ErrorContains(t, err, "unable to process carvel package")
```
sathyanarays, line 41: ensure that `ApplyResourcesFromManifestReturns` is not called

sathyanarays (changes requested):

cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,58 @@
...
+		generatedManifestBytes = append(generatedManifestBytes, generatedManifestByte...)
+		if err != nil {
+			return fmt.Errorf("unable to process carvel package %s: %w", item.Name(), err)
+		}
+	}
+
+	err = c.pkgClient.ApplyResourcesFromManifest(generatedManifestBytes, c.restConfig)
+	if err != nil {
+		return fmt.Errorf("unable to install : %w", err)
+	}
```
rajathagasthya, line 56: This does not look like a blocking API. This method needs to block i.e. not return until everything is installed within a timeout or error out. Otherwise, consuming code wouldn't know things are successful or not.

Ideally, we need to have 2 APIs like:

```go
// Install with default timeout
func Install() error {
    // 1. Apply manifests
    // 2. Pick a default timeout
    // 3. Wait for results
}

// Install with option to provide a timeout
func InstallWithTimeout(timeout time.Duration) error {
    // 1. Apply manifests
    // 2. Wait for results
}
```

cluster-essentials/test/clusterinit_test.go
```diff
@@ -0,0 +1,68 @@
+// Copyright 2023 VMware, Inc. All Rights Reserved.
```
rajathagasthya, line 1: Test file should be in the same directory as the file being tested. No need for a new package.

cluster-essentials/util/clientInterface.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"k8s.io/client-go/rest"
+
+	"github.com/vmware-tanzu/tanzu-framework/util/carvelhelpers"
+	"github.com/vmware-tanzu/tanzu-framework/util/imgpkg"
+	"github.com/vmware-tanzu/tanzu-framework/util/kube"
+)
+
+type Client interface {
```
rajathagasthya, line 15: Why do we have this interface? I'm not sure I see the necessity.

util/carvelhelpers/kbld.go
```diff
@@ -0,0 +1,34 @@
+// Copyright 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package carvelhelpers implements wrapper functions to use carvel tooling
+package carvelhelpers
```
rajathagasthya, line 5: These don't belong in the `util` module (yet). If you think it does, please provide justification as to how it might be used by multiple modules. Otherwise, just move this to `cluster-essentials/pkg`. 

Functions/methods in the `util` module need to have much more friendlier APIs (these are somewhat hard to understand) and have test cases.

util/imgpkg/imgpkg.go
```diff
@@ -0,0 +1,35 @@
+// Copyright 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package imgpkg has utility functions for imgpkg
+package imgpkg
```
rajathagasthya, line 5: Same comment as above. Move to `cluster-essentials`.

util/kube/kube_util.go
```diff
@@ -0,0 +1,95 @@
+// Copyright 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package kube has utility functions for interacting with kubernetes cluster using client-go library
+package kube
```
rajathagasthya, line 5: Same comment as above. Move to cluster-essentials.

This one has _some_ merit to be here. But it's a no-go without tests.

rajathagasthya (changes requested):
First pass: looked at the API and code organization. I'm not convinced the current API is the right one.

pkg/client/cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,81 @@
...
+	ClusterEssentialBundlePath string
+	BundleDestDir              string
+	RestConfig                 *rest.Config
+	BlockingCallEnable         bool
+	Calltimeout                time.Duration
+}
+
+var downloadBundle = utils.DownloadBundle
+
+func PreInstallValidation(restConfig *rest.Config) error {
```
rajathagasthya, line 28: Function names must be verbs, so please rename this to `validatePreInstall`. Also need to be non-exported function.

pkg/client/cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,81 @@
...
+	// TODO implementation
+	return nil
+}
+
+func PostInstallValidation(restConfig *rest.Config, blockingEnable bool, timeout time.Duration) error {
```
rajathagasthya, line 33: Same as above. Rename to `validatePostInstall`.

pkg/client/cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,81 @@
...
+	"k8s.io/client-go/rest"
+
+	"github.com/vmware-tanzu/tanzu-framework/pkg/client/cluster-essentials/pkg/utils"
+)
+
+type ClusterEssentialsClient struct {
+	ClusterEssentialBundlePath string
+	BundleDestDir              string
+	RestConfig                 *rest.Config
+	BlockingCallEnable         bool
```
rajathagasthya, line 22: Why do we need this field? When would a consumer not make a blocking call?

pkg/client/cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,81 @@
...
+	"time"
+
+	"k8s.io/client-go/rest"
+
+	"github.com/vmware-tanzu/tanzu-framework/pkg/client/cluster-essentials/pkg/utils"
+)
+
+type ClusterEssentialsClient struct {
+	ClusterEssentialBundlePath string
+	BundleDestDir              string
```
rajathagasthya, line 20: What is the advantage in exposing this field? It's just unpacking the bundle in a (temp) directory.

pkg/client/cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,81 @@
+// Copyright 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package clusteressentials has APIs for bootstrapping kubernetes cluster
+package clusteressentials
```
rajathagasthya, line 5: I think you probably misunderstood my previous comment. We want a directory structure like this:

```
cluster-essentials
├── go.mod
└── pkg
    ├── client
    │   ├── fake.go
    │   ├── helpers.go
    │   ├── install.go
    │   └── install_test.go
    └── testdata
```

pkg/client/cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,81 @@
...
+	// TODO implementation
+	return nil
+}
+
+func PostInstallValidation(restConfig *rest.Config, blockingEnable bool, timeout time.Duration) error {
+	// TODO implementation
+	return nil
+}
+
+func Install(c ClusterEssentialsClient) error {
```
rajathagasthya, line 38: There are only 3 things you really need. You might as well just pass them as function arguments instead of using a `ClusterEssentialsClient` struct to hold them.

```go
func Install(config *rest.Config, imgpkgBundlePath string, timeout time.Duration) {
}
```

pkg/client/cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	if c.BundleDestDir == "" {
+		c.BundleDestDir = "tmp"
+	}
+
+	bundleDir, err := os.MkdirTemp(c.BundleDestDir, "cluster-essential")
+	if err != nil {
+		return err
+	}
```
rajathagasthya, line 51: Please move the `os.RemoveAll` from L79 here and inside a `defer` call.

pkg/client/cluster-essentials/clusterinit_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+var bundleName = "test.image"
+var bundleDir = "tmp"
+
+func Test_InitClusterEssentials(t *testing.T) {
+	c := ClusterEssentialsClient{
+		ClusterEssentialBundlePath: bundleName,
+		BundleDestDir:              bundleDir,
+	}
+	_ = os.MkdirAll(bundleDir, os.ModePerm)
```
rajathagasthya, line 24: `os.RemoveAll` needs to be inside a `defer` here.

pkg/client/cluster-essentials/go.mod
```diff
@@ -0,0 +1,100 @@
+module github.com/vmware-tanzu/tanzu-framework/pkg/client/cluster-essentials
+
+go 1.18
```
rajathagasthya, line 3: Let's use a newer Golang version.

pkg/client/cluster-essentials/pkg/fake/fakes.go
```diff
@@ -0,0 +1,35 @@
+// Copyright 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package fake contain cluster essential util APIs
+package fake
```
rajathagasthya, line 5: I don't think there is a need for a separate `fake` package. Just move this file to the `client` package alongside the install files.

pkg/client/cluster-essentials/pkg/utils/utils.go
```diff
@@ -0,0 +1,265 @@
+// Copyright 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package utils contain cluster essential util APIs
+package utils
```
rajathagasthya, line 5: Looking at the usage of functions in this package, all of it is in `clusterinit.go`. Why not just move this file to that package and get rid of `utils` package?

rajathagasthya (changes requested):
Round 2: API looks much better, but some changes are still necessary.

pkg/client/cluster-essentials/go.mod
```diff
@@ -0,0 +1,100 @@
+module github.com/vmware-tanzu/tanzu-framework/pkg/client/cluster-essentials
```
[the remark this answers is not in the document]
rajathagasthya: Please address all CVE alerts.

pkg/client/cluster-essentials/pkg/fake/fakes.go
rajathagasthya: Please https://github.com/vmware-tanzu/tanzu-framework/pull/4503#discussion_r1157914392 for desired package structure.

pkg/client/cluster-essentials/pkg/utils/utils.go
rajathagasthya: Please see https://github.com/vmware-tanzu/tanzu-framework/pull/4503#discussion_r1157914392 for desired package structure.

pkg/client/cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,81 @@
+// Copyright 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package clusteressentials has APIs for bootstrapping kubernetes cluster
```
avi-08, line 4: I think this needs to say `bootstrapping tanzu cluster-essentials on a kubernetes cluster`

pkg/client/cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/vmware-tanzu/tanzu-framework/pkg/client/cluster-essentials/pkg/utils"
+)
+
+type ClusterEssentialsClient struct {
+	ClusterEssentialBundlePath string
+	BundleDestDir              string
+	RestConfig                 *rest.Config
+	BlockingCallEnable         bool
+	Calltimeout                time.Duration
```
avi-08, line 23: Typo `CallTimeout`

pkg/client/cluster-essentials/clusterinit.go
sathyanarays, line 38: The API consumers will usually use retries to handle transient failures. Let's assume the case of the statement `utils.AppplyResourcesFromManifest` succeeds without error. And assume that while calling `PostInstallValidation` there is some network glitch in reaching the API server and the call fails with an error. If there is a retry loop implemented at consumer side, will the next retry resume `PostInstallValidation`? If not, how do we enforce that?

pkg/client/cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func Install(c ClusterEssentialsClient) error {
+	err := PreInstallValidation(c.RestConfig)
+	if err != nil {
+		return fmt.Errorf("pre install validation is failed with error: %w", err)
+	}
+
+	if c.BundleDestDir == "" {
+		c.BundleDestDir = "tmp"
```
avi-08, line 45: Can we create & use "tmp" as a named constant `defaultBundleDir`

pkg/client/cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,81 @@
...
+	err = downloadBundle(c.ClusterEssentialBundlePath, bundleDir)
+	if err != nil {
+		return fmt.Errorf("unable to download cluster essential manifest: %w", err)
+	}
+	packages, err := os.ReadDir(bundleDir)
+	if err != nil || len(packages) == 0 {
+		return fmt.Errorf("error while reading the bundleDir %s", bundleDir)
+	}
+	var generatedManifestBytes []byte
+	for _, item := range packages {
```
avi-08, line 61: We should only process directories within the bundleDir, just to be on the safe side.

pkg/client/cluster-essentials/clusterinit.go
avi-08: We won't need to have a restart from last failed state, because all operations within `Install` are supposed to be idempotent. More importantly, in this scenario, we should exit at the pre-validation since required resources already exist.

pkg/client/cluster-essentials/clusterinit.go
```diff
@@ -0,0 +1,81 @@
...
+	return nil
+}
+
+func PostInstallValidation(restConfig *rest.Config, blockingEnable bool, timeout time.Duration) error {
+	// TODO implementation
+	return nil
+}
+
+func Install(c ClusterEssentialsClient) error {
+	err := PreInstallValidation(c.RestConfig)
```
avi-08, line 39: How do we run pre validations before downloading the bundle ?

Vandy-P: We are validating the cluster to check if "we can install cluster essential or not "

pkg/client/cluster-essentials/clusterinit.go
Vandy-P: We are processing directories which are inside the bundledir , Please check L56

pkg/client/cluster-essentials/clusterinit.go
Vandy-P: Discussed.

pkg/client/cluster-essentials/go.mod
Vandy-P: if we use greater than 1.18 version make verify (make module) pipeline will fail with 
**"go: go.mod file indicates go 1.19, but maximum version supported by tidy is 1.18"**

avi-08 (approved):
LGTM

cluster-essentials/pkg/client/fakes.go
```diff
@@ -0,0 +1,34 @@
...
+func DownloadCorruptedBundleSuccess(bundlePath, outputDir string) error {
+	outputDir = filepath.Join(outputDir, "package")
+	_ = os.MkdirAll(outputDir, os.ModePerm)
+	_ = cp.Copy("../testdata/package1", outputDir)
+	return nil
+}
+
+func DownloadBundleFail(bundlePath, outputDir string) error {
+	return fmt.Errorf("image not found")
+}
```
rajathagasthya, line 34: Please un-export these functions. Ideally, they can live in `install_test.go` since that's the only place it's used.

cluster-essentials/pkg/client/helpers.go
```diff
@@ -0,0 +1,264 @@
...
+			} else {
+				return err
+			}
+		}
+		accessor, err := meta.Accessor(unstructuredObj)
+		name := accessor.GetName()
+		if err != nil {
+			return err
+		}
+		_, err = resource.Apply(context.Background(), name, unstructuredObj, metav1.ApplyOptions{FieldManager: name})
```
rajathagasthya, line 91: This `context` need to be passed down all the way from `Install`.

cluster-essentials/pkg/client/helpers.go
```diff
@@ -0,0 +1,264 @@
...
+		return nil, err
+	}
+	result, err := loader.Eval(valuesDoc, libraryValueDoc, []*datavalues.SchemaEnvelope{})
+	if err != nil {
+		return nil, err
+	}
+	return result.DocSet.AsBytes()
+}
+
+// CarvelPackageProcessor processes a carvel package and returns a configuration YAML file
```
rajathagasthya, line 159: Can you expand what "processing" and "configuration YAML" means here?

cluster-essentials/pkg/client/helpers.go
```diff
@@ -0,0 +1,264 @@
...
+	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
+	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+)
+
+const ConfigFilePermissions = 0o600
+
+func DownloadBundle(bundlePath, outputDir string) error {
```
rajathagasthya, line 42: Is there a reason all the functions in this file are exported? They all look like something you just use in `Install`. Please un-export if there is no usage outside this package.

cluster-essentials/pkg/client/install.go
```diff
@@ -0,0 +1,84 @@
...
+	// TODO implementation
+	return nil
+}
+
+// Install install the cluster essentials packages
+// It take kubernetes rest config of cluster (config), cluster essential bundler path (imgpkgBundlePath)
+// timeout for blocking call (if zero is given, it will use default timeout of 15 min),
+// bundleDestDir to use as temporary dir to download bundles(if empty, it will use /tmp as default path)
+// It return any error encountered
+func Install(config *rest.Config, imgpkgBundlePath string, timeout time.Duration, bundleDestDir string) error {
```
rajathagasthya, line 36: Please add `ctx context.Context` as the first parameter and pass it down to functions that require a context.

cluster-essentials/pkg/client/install.go
```diff
@@ -0,0 +1,84 @@
...
+func validatePostInstall(restConfig *rest.Config, timeout time.Duration) error {
+	// TODO implementation
+	return nil
+}
+
+// Install install the cluster essentials packages
+// It take kubernetes rest config of cluster (config), cluster essential bundler path (imgpkgBundlePath)
+// timeout for blocking call (if zero is given, it will use default timeout of 15 min),
+// bundleDestDir to use as temporary dir to download bundles(if empty, it will use /tmp as default path)
+// It return any error encountered
```
rajathagasthya, line 35:
```suggestion
// Install installs Cluster Essentials packages.
// This function attempts to install packages from the given imgpkg bundle path
// and waits until install is complete or the given timeout is reached, whichever
// occurs first.
```

rajathagasthya (changes requested):
Suggested some minor changes.

sathyanarays:
Should this be merged with main or feature/runtime-core?

cluster-essentials/pkg/client/install_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/stretchr/testify/require"
+	"k8s.io/client-go/rest"
+)
+
+var bundleName = "test.image"
+var bundleDir = "tmp"
+var config *rest.Config
+
+func Test_InitClusterEssentials(t *testing.T) {
```
sathyanarays, line 19: Please add a success case here.

cluster-essentials/pkg/client/fakes.go
sathyanarays: +1

cluster-essentials/pkg/client/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+1. Kapp-controller
+2. secret-gen controller.
+
+## APIs
+
+Install(ctx context.Context, config *rest.Config, imgpkgBundlePath string, timeout time.Duration, bundleDestDir string) error
+
+Install install the cluster essentials packages
+It take kubernetes rest config of cluster (config), cluster essential bundler path (imgpkgBundlePath)
```
yharish991, line 13:
```suggestion
It takes kubernetes rest config of cluster (config), cluster essentials bundle path (imgpkgBundlePath)
```

cluster-essentials/pkg/client/install.go
```diff
@@ -0,0 +1,85 @@
...
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	"k8s.io/client-go/rest"
+)
+
+var downloadBundles = downloadBundle
```
yharish991, line 17: question: curious, why are we doing this? why not call the function directly?

cluster-essentials/pkg/client/install.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return fmt.Errorf("pre install validation is failed with error: %w", err)
+	}
+	packages, err := os.ReadDir(bundleDir)
+	if err != nil || len(packages) == 0 {
+		return fmt.Errorf("error while reading the bundleDir %s", bundleDir)
+	}
+	var generatedManifestBytes []byte
+	for _, item := range packages {
+		packageDir := filepath.Join(bundleDir, item.Name())
```
yharish991, line 64: i'm confused, you are saying this `Install` function, expects a imgpkg bundle path, but then here you say each package contains a .imgpkg directory, how can a bundle contain multiple .imgpkg directories? how does the structure of imgpkg bundle look like? may be tree structure in the documentation helps

Vandy-P: tree bundle/
bundle/
├── kapp-controller
│   └── config
│       ├── overlays
│       │   ├── change-namespace.yaml
│       │   ├── update-clusterrole-with-psp.yaml
│       │   ├── update-configmap.yaml
│       │   ├── update-crds.yaml
│       │   ├── update-deployment.yaml
│       │   └── update-strategy-overlay.yaml
│       ├── upstream
│       │   └── kapp-controller.yaml
│       ├── values.star
│       └── values.yaml
└── secretgen-controller
    └── config
        └── secretgen-controller.yaml

Vandy-P: @yharish991  each package is bundle itself

cluster-essentials/pkg/client/install.go
Vandy-P: using monkey patching to mock out these api for unit test

cluster-essentials/pkg/client/install.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func validatePostInstall(restConfig *rest.Config, timeout time.Duration) error {
+        client, _ := kubernetes.NewForConfig(restConfig)
+        for _, packageName := range clusterEssentialPackages {
+                label := "app="+packageName
+                pods,_ := client.CoreV1().Pods(packageName).List(context.TODO(),
+                     metav1.ListOptions{LabelSelector: label})
+                if err := wait.Poll(time.Second*2, time.Second*20, func() (done bool, err error) {
+                        pod, _ := client.CoreV1().Pods(packageName).Get(context.Background(), pods.Items[0].Name, metav1.GetOptions{})
```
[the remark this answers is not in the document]
sathyanarays: We already got the pods in this line: `pods,_ := client.CoreV1().Pods(packageName).List(context.TODO()`. Why do we need to query them again here?

cluster-essentials/pkg/client/install.go
```diff
@@ -0,0 +1,109 @@
...
+			}
+	                //If above annotation is present, get the annotation for cluster essential version
+			// In case of downgrade - Exit
+			// In case of upgrade - move forward with installation
+		}
+	}
+        return nil
+}
+
+func validatePostInstall(restConfig *rest.Config, timeout time.Duration) error {
```
sathyanarays, line 55: Where is the timeout being used?

cluster-essentials/pkg/client/install.go
```diff
@@ -0,0 +1,109 @@
...
+func Install(ctx context.Context, config *rest.Config, imgpkgBundlePath string, timeout time.Duration) error {
+	if timeout == time.Duration(0) {
+		timeout = defaultTimeout
+	}
+	bundleDir, err := os.MkdirTemp(defaultBundleDir, "cluster-essential")
+	defer os.RemoveAll(bundleDir)
+	if err != nil {
+		return err
+	}
+	err = downloadBundles(imgpkgBundlePath, bundleDir)
```
sathyanarays, line 90: can we re-order download and preinstall so that we don't need to download if pre-install fails.

cluster-essentials/pkg/client/install.go
```diff
@@ -20,28 +24,65 @@ var appplyResourcesFromManifests = appplyResourcesFromManifest
 const defaultBundleDir = "/tmp"
 const defaultTimeout = 15 * time.Minute
 
+var clusterEssentialPackages = []string{"kapp-controller", "secretgen-controller"}
+
 func validatePreInstall(restConfig *rest.Config) error {
-	// TODO implementation
-	return nil
+	clientset, _ := kubernetes.NewForConfig(restConfig)
+	nsList, _ := clientset.CoreV1().Namespaces().List(context.Background(), metav1.ListOptions{})
```
rajathagasthya, line 31: Please take context as an arg to the function.

rajathagasthya: Also use controller-runtime's `Client` for CRUD operations on Kubernetes objects. client-go will bring in a bunch of unnecessary dependencies.

cluster-essentials/pkg/client/install.go
```diff
@@ -20,28 +24,65 @@ var appplyResourcesFromManifests = appplyResourcesFromManifest
...
+		for _, packageName := range clusterEssentialPackages {
+			deploymentsClient := clientset.AppsV1().Deployments(n.Name)
+			result, getErr := deploymentsClient.Get(context.TODO(), packageName, metav1.GetOptions{})
+			if getErr != nil {
+				//TODO check for CRD if deployment is not present
+				// if Deplyoment and CRD is not present, continue the loop to check in other namespace 
+				continue
+			}
+			// If deployment is present, check for annotation
+			_, ok := result.Annotations["tanzu.vmware.cluster.essentia/package-name"]
```
rajathagasthya, line 42: `tanzu.vmware.cluster.essentia/package-name`

This is not the right format for annotations. This should be `runtime.tanzu.vmware.com/<something>`.

cluster-essentials/pkg/client/install.go
```diff
@@ -20,28 +24,65 @@ var appplyResourcesFromManifests = appplyResourcesFromManifest
...
+			//If annotation is not present then return error
+			if !ok {
+				return fmt.Errorf("Pre validation check failed for package %s, package is not a part of cluster essential", packageName)
```
rajathagasthya, line 45: Error messages should start with lower case letter.

cluster-essentials/pkg/client/install.go
```diff
@@ -20,28 +24,65 @@ var appplyResourcesFromManifests = appplyResourcesFromManifest
...
+	}
+        return nil
 }
 
 func validatePostInstall(restConfig *rest.Config, timeout time.Duration) error {
-	// TODO implementation
+        client, _ := kubernetes.NewForConfig(restConfig)
+        for _, packageName := range clusterEssentialPackages {
+                label := "app="+packageName
+                pods,_ := client.CoreV1().Pods(packageName).List(context.TODO(),
```
rajathagasthya, line 59: Function needs to take in a context.

cluster-essentials/pkg/client/install.go
Vandy-P: We need bundle beforehand to find out version

cluster-essentials/pkg/client/install.go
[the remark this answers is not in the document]
Vandy-P: we need to query after every interval to get the latest status

yharish991 (changes requested):

cluster-essentials/pkg/client/install.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	// Next: Get all the resource from manifest and check if resource is present or not
+	return nil
+}
+
+// Install installs Cluster Essentials packages.
+// This function attempts to install packages from the given imgpkg bundle path
+// and waits until install is complete or the given timeout is reached, whichever
+// occurs first.
+func Install(ctx context.Context, config *rest.Config, clusterEssentialRepo, clusterEssentialVersion string, timeout time.Duration) error {
```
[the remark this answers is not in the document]
rajathagasthya: Aren't we already doing it by accepting `rest.Config`?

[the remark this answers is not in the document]
yharish991: ah, you are right..

cluster-essentials/pkg/client/install.go
```diff
@@ -0,0 +1,151 @@
...
+				// TODO check for CRD if deployment is not present
+				// if deployment and CRD is not present, continue the loop to check in other namespace
+			}
+			// If deployment is present, check for annotation
+			_, ok := deployment.Annotations[packageNameAnnotation]
+			// If annotation is not present then return error
+			if !ok {
+				return fmt.Errorf("pre validation check failed for package %s, package is not a part of cluster essential", packageName)
+			}
+			// Get the existing cluster essential version
```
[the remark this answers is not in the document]
Vandy-P: for non-cluster-essential , prevalidation will fail at line 65
_, ok := deployment.Annotations[packageNameAnnotation]
    // If annotation is not present then return error
  if !ok {
       return fmt.Errorf("pre validation check failed for package %s, package is not a part of cluster essential", packageName)
 }

PR #4506, opened 2023-03-20
Update autoscaler related annotations and Add filter to node-group-auto-discovery arg

zhouke1991:
1. Add a filter "namespace" for autoscaler arg "--node-group-auto-discovery". Otherwise, in the case of creating a cluster but not in the default namespace, autoscaler can't find the node group.
2. These two annotations have been deprecated since v1.25.0, use new annotations.
```
deprecated:
cluster.k8s.io/cluster-api-autoscaler-node-group-max-size: "5"
cluster.k8s.io/cluster-api-autoscaler-node-group-min-size: "0"

new:
cluster.x-k8s.io/cluster-api-autoscaler-node-group-max-size: "5"
cluster.x-k8s.io/cluster-api-autoscaler-node-group-min-size: "0"
```
3. Since scale-from-zero feature checked in: https://github.com/kubernetes/autoscaler/pull/4840/files we need to add get/list privilege into ClusterRole.

### DIFF
--- a/providers/ytt/03_customizations/autoscaler/autoscaler_overlay.yaml
+++ b/providers/ytt/03_customizations/autoscaler/autoscaler_overlay.yaml
@@ -10,8 +10,8 @@ kind: MachineDeployment
 metadata:
   #@overlay/match missing_ok=True
   annotations:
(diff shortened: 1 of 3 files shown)


zhouke1991:
//usebom:v2.2.0-zshippable/tkg-compatibility:v16

wjun (approved):
LGTM.

PR #4511, opened 2023-03-22
enhance output instuction when creating cc cluster with legacy config

hongtaiwang:

When creating cluster using cluster yaml, the validation does not read configurations from the yaml context. So if none custom flags are explicitly set in this step, tanzu cli will validate with defualt value

ALL customized flags are required to set when creating cluster with a cc cluster yaml.

Fixes #

### Release note

```release-note

```

### DIFF
--- a/tkg/client/cluster.go
+++ b/tkg/client/cluster.go
@@ -180,7 +180,7 @@ func (c *TkgClient) CreateCluster(options *CreateClusterOptions, waitForCluster 
 			// If `features.cluster.auto-apply-generated-clusterclass-based-configuration` feature-flag is not activated
 			// log command to use to create cluster using ClusterClass based config file and return
 			if !c.IsFeatureActivated(constants.FeatureFlagAutoApplyGeneratedClusterClassBasedConfiguration) {
-				log.Warningf("\nTo create a cluster with it, use")
+				log.Warningf("\nTo create a cluster with it, please use the following command together with all customized flags set in the last step. e.g. --tkr, -v")
 				log.Warningf("    tanzu cluster create --file %v", configFilePath)
 				return false, nil
 			}
--- a/tkg/client/cluster_test.go
+++ b/tkg/client/cluster_test.go
@@ -403,7 +403,7 @@ ova: []
 
 				result := captureOutput(*tkgClient, options, false)
 				Expect(result).To(ContainSubstring("Legacy configuration file detected. The inputs from said file have been converted into the new Cluster configuration"))
-				Expect(result).To(ContainSubstring("To create a cluster with it, use"))
+				Expect(result).To(ContainSubstring("To create a cluster with it, please use the following command together with all customized flags set in the last step. e.g. --tkr, -v"))
 			})
 
 		})


tkg/client/cluster.go
```diff
@@ -180,7 +180,7 @@ func (c *TkgClient) CreateCluster(options *CreateClusterOptions, waitForCluster
 			// If `features.cluster.auto-apply-generated-clusterclass-based-configuration` feature-flag is not activated
 			// log command to use to create cluster using ClusterClass based config file and return
 			if !c.IsFeatureActivated(constants.FeatureFlagAutoApplyGeneratedClusterClassBasedConfiguration) {
-				log.Warningf("\nTo create a cluster with it, use")
+				log.Warningf("\nTo create a cluster with it, please use the following command together with customize flags like --tkr, -v, etc.")
```
wjun, line 183: Please provide more context why these customize flags are required.

hongtaiwang: Indeed, all customized flags are required.
When creating cluster using cluster yaml, the validation does not read configurations from the yaml context. So if none custom flags are explicitly set in this step, tanzu cli will validate with defualt value

hongtaiwang: updated the wording to make it more clear

tkg/client/cluster.go
```diff
@@ -180,7 +180,7 @@ func (c *TkgClient) CreateCluster(options *CreateClusterOptions, waitForCluster
 			// If `features.cluster.auto-apply-generated-clusterclass-based-configuration` feature-flag is not activated
 			// log command to use to create cluster using ClusterClass based config file and return
 			if !c.IsFeatureActivated(constants.FeatureFlagAutoApplyGeneratedClusterClassBasedConfiguration) {
-				log.Warningf("\nTo create a cluster with it, use")
+				log.Warningf("\nTo create a cluster with it, please use the following command together with all customized flags set in the last step. e.g. --tkr, -v")
```
xiujuanx, line 183: @hongtaiwang may I know why the feature gate `auto-apply-generated-clusterclass-based-configuration` must to to be enabled when create a cluster with a kubernetes-style yaml file? 
As far as I know, when create a classy workload-cluster with k8s-style yaml file, code logic will not go here. It should be go into here: https://github.com/vmware-tanzu/tanzu-framework/blob/main/tkg/client/cluster.go#L147-L156. So enable this feature gate with a k8s-style yaml file may not make sense.

Also the error is reported by the user is more like a mapping error for `VSPHERE_TEMPLATE` var. At the first step, user specifies `--tkr v1.23.16---vmware.1-tkg.2`, but why the value of this var is parsed to be `/TSV_DC/vm/ubuntu_temp_v1.23.16`? What's the correct template name for v1.23.16---vmware.1-tkg.2 tkr?

BTW, the `-v` is used for debugging AFAIK, so may it's not necessary. Or does it do something else?

hongtaiwang: better to have a zoom call to carify this:)

wjun (approved):
LGTM.

PR #4514, opened 2023-03-22
Extract common PinnipedInfo struct so that it can be used everywhere

joshuatcasey:

Extract common PinnipedInfo struct so that it can be used everywhere

Ran unit tests for changed modules. Not all unit tests can be run locally due to some configuration required.

### DIFF
--- a/cli/core/go.mod
+++ b/cli/core/go.mod
@@ -7,6 +7,7 @@ replace (
 	github.com/vmware-tanzu/tanzu-framework/apis/run => ../../apis/run
 	github.com/vmware-tanzu/tanzu-framework/capabilities/client => ../../capabilities/client
 	github.com/vmware-tanzu/tanzu-framework/cli/runtime => ../runtime
+	github.com/vmware-tanzu/tanzu-framework/pinniped-components/common => ../../pinniped-components/common
 )
 
 require (
@@ -38,6 +39,7 @@ require (
 	github.com/vmware-tanzu/tanzu-framework/apis/cli v0.0.0-00010101000000-000000000000
 	github.com/vmware-tanzu/tanzu-framework/capabilities/client v0.0.0-00010101000000-000000000000
 	github.com/vmware-tanzu/tanzu-framework/cli/runtime v0.0.0-00010101000000-000000000000
+	github.com/vmware-tanzu/tanzu-framework/pinniped-components/common v0.0.0-00010101000000-000000000000
 	go.uber.org/multierr v1.6.0
 	golang.org/x/mod v0.8.0
 	golang.org/x/oauth2 v0.3.0
--- a/cli/core/pkg/auth/tkg/cluster_pinniped_info.go
+++ b/cli/core/pkg/auth/tkg/cluster_pinniped_info.go
@@ -19,24 +19,12 @@ import (
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 
 	netutil "github.com/vmware-tanzu/tanzu-framework/cli/core/pkg/auth/utils/net"
+	"github.com/vmware-tanzu/tanzu-framework/pinniped-components/common/pkg/pinnipedinfo"
 )
 
 const (
 	KubePublicNamespace = "kube-public"
 )
-
-// PinnipedConfigMapInfo defines the fields of pinniped-info configMap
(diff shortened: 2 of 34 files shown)


benjaminapetersen (approved):
So far so good. One step toward de-duplication of some of the Pinniped bits scattered around.

benjaminapetersen:
In addition to extracting the PinnipedInfo struct, this is really setting the stage for further refactors, removing other duplicate code, and avoiding future code duplication as we add additional features to the various CLI components (cluster, managementcluster, login and context) to support future IAM work.  This might be helpful context, possibly to add to the description.

cli/core/go.mod
```diff
@@ -38,6 +39,7 @@ require (
 	github.com/vmware-tanzu/tanzu-framework/apis/cli v0.0.0-00010101000000-000000000000
 	github.com/vmware-tanzu/tanzu-framework/capabilities/client v0.0.0-00010101000000-000000000000
 	github.com/vmware-tanzu/tanzu-framework/cli/runtime v0.0.0-00010101000000-000000000000
+	github.com/vmware-tanzu/tanzu-framework/pinniped-components/common v0.0.0-00010101000000-000000000000
```
benjaminapetersen, line 42: Fake `version v0.0.0-00010101000000-000000000000` in our `require` so the `replace` can use the local.  We do intend to pick up the extraction of the whole `/pinniped-components` directory at some point. Not sure what the timeline is for eviction from TF at this point.

benjaminapetersen: I'll tag @prkalle since we have been discussing this topic as of late.

benjaminapetersen: I believe @jmoroski is overseeing the whole TF refactor and eviction effort, perhaps also interested in what we are doing here.  (see description & one of my other comments for summary).

cli/core/pkg/auth/tkg/cluster_pinniped_info.go
```diff
@@ -19,25 +19,13 @@ import (
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 
 	netutil "github.com/vmware-tanzu/tanzu-framework/cli/core/pkg/auth/utils/net"
+	"github.com/vmware-tanzu/tanzu-framework/pinniped-components/common/pkg/pinnipedinfo"
```
benjaminapetersen, line 22: I'll be honest, with our `require/replace` and definition of a new module in `pinniped-components/common` I find the full path required in these other files odd.  I'm sure this will be resolved when the time comes to move this code to a separate repo.

cli/core/go.mod
benjaminapetersen: If other code (`cli plugins` or `cli core commands`, etc) move out of TF before `/pinniped-components` does, but then references our `<TF>/pinniped-components` I wonder if we have a risk of circular dependency.

cli/core/pkg/auth/tkg/cluster_pinniped_info.go
```diff
@@ -114,7 +102,7 @@ func GetClusterInfoFromCluster(clusterAPIServerURL, configmapName string) (*clie
 // 'discoveryPort' is used to optionally override the port used for discovery. This may be needed on setups that expose
 // discovery information to unauthenticated users on a different port (for instance, to avoid the need to anonymous auth
 // on the apiserver). By default, the endpoint from the cluster-info is used.
-func GetPinnipedInfoFromCluster(clusterInfo *clientcmdapi.Cluster, discoveryPort *int) (*PinnipedConfigMapInfo, error) {
+func GetPinnipedInfoFromCluster(clusterInfo *clientcmdapi.Cluster, discoveryPort *int) (*pinnipedinfo.PinnipedInfo, error) {
```
benjaminapetersen, line 105: This function is a candidate for our next step in refactoring as `GetPinnipedInfoFromCluster()` is duplicated.

pinniped-components/common/doc.go
```diff
@@ -0,0 +1,5 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
```
rajathagasthya, line 1: `s/2022/2023`

rajathagasthya (approved):
Minor comment, but LGTM!

PR #4531, opened 2023-03-29
Combine the zeroes-out AWS credential action for both legacy and cc

12345lcr:

This is to ensures that the Cluster API Provider AWS controller is pinned to control plane nodes and is running without static credentials such that Cluster API AWS runs using the EC2 instance profile attached to the control plane node.

This is done by zeroing out the credentials secret for CAPA, causing the AWS SDK to fall back to the default credential provider chain.

This was implemented for mgmt cluster creation and legacy mgmt cluster upgrade, but missing in classy mgmt cluster upgrade. Moving this patch right after provider upgrade will cover both legacy and classy mgmt cluster upgrades.

Additionally extend the k8s versionpatch waiting time to allow corner cases.

Fixes #

### Release note

```release-note

```

### DIFF
--- a/tkg/client/upgrade_cluster_clusterclass.go
+++ b/tkg/client/upgrade_cluster_clusterclass.go
@@ -22,9 +22,10 @@ func (c *TkgClient) DoClassyClusterUpgrade(regionalClusterClient clusterclient.C
 	log.Infof("Upgrading kubernetes cluster to `%v` version, tkr version: `%s`", kubernetesVersion, tkrVersion)
 	patchJSONString := fmt.Sprintf(`{"spec": {"topology": {"version": "%v"}}}`, tkrVersion)
 
-	// Timeout set to 15 minutes because the continuousTKRDiscoverFreq for tkr-source-controller's fetcher is 10 minutes.
-	// Wait time should be longer than the fetcher's frequency of pulling tkrs.
-	pollOptions := &clusterclient.PollOptions{Interval: upgradePatchInterval, Timeout: 15 * time.Minute}
+	// Timeout set to 30 minutes because the continuousTKRDiscoverFreq for tkr-source-controller's fetcher is 10 minutes.
+	// And kapp package reconcile frequency is 10 minutes.
+	// Wait time should be longer than the fetcher's frequency of pulling tkrs plus the frequency of kapp package reconciliation.
+	pollOptions := &clusterclient.PollOptions{Interval: upgradePatchInterval, Timeout: 30 * time.Minute}
 	err := regionalClusterClient.PatchClusterObjectWithPollOptions(options.ClusterName, options.Namespace, patchJSONString, pollOptions)
 	if err != nil {
 		return errors.Wrap(err, "unable to patch kubernetes version to cluster")
--- a/tkg/client/upgrade_region.go
+++ b/tkg/client/upgrade_region.go
@@ -136,6 +136,11 @@ func (c *TkgClient) UpgradeManagementCluster(options *UpgradeClusterOptions) err
 		return errors.Wrap(err, "failed to upgrade management cluster providers")
 	}
 
+	// Ensure Cluster API Provider AWS is running on the control plane before continuing with EC2 instance profile
+	if err := regionalClusterClient.PatchClusterAPIAWSControllersToUseEC2Credentials(); err != nil {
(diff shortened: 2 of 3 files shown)


alfredthenarwhal:
> @12345lcr:  /test upgrade-aws
Commit: 85d5177

Tests failed! Build no: 4610

tkg/client/upgrade_region.go
```diff
@@ -147,6 +147,11 @@ func (c *TkgClient) UpgradeManagementCluster(options *UpgradeClusterOptions) err
 	if err != nil {
 		return errors.Wrap(err, "error waiting for provider components to be up and running after upgrading them")
 	}
+
+	// Ensure Cluster API Provider AWS is running on the control plane before continuing with EC2 instance profile
+	if err := regionalClusterClient.PatchClusterAPIAWSControllersToUseEC2Credentials(); err != nil {
```
DanielXiao, line 152: I think this should execute before `c.WaitForProviders(regionalClusterClient, waitOptions)`

12345lcr: Should I modify the mgmt cluster creation as well? Since on creation the patch is updated after provider is ready
https://github.com/vmware-tanzu/tanzu-framework/blob/fe8dd986a4313e34518dcea3d7025c3b8c2e84e0/tkg/client/init.go#L378-L384
https://github.com/vmware-tanzu/tanzu-framework/blob/fe8dd986a4313e34518dcea3d7025c3b8c2e84e0/tkg/client/init.go#L739

I added here because I suspect that upgrade goes into multi goroutine and cannot guarantee that the provider is ready  to be patched after `DoProvidersUpgrade`. Like the patched deployment got flushed out by the provider upgrade. If it can be ready to be patched, then it make no sense to have `WaitForProviders` here.

DanielXiao: We need to wait til deployment replica is greater than 1 when we patch CAPA deployment. @randomvariable What do you think?

tkg/client/upgrade_cluster_clusterclass.go
```diff
@@ -22,9 +22,10 @@ func (c *TkgClient) DoClassyClusterUpgrade(regionalClusterClient clusterclient.C
 	log.Infof("Upgrading kubernetes cluster to `%v` version, tkr version: `%s`", kubernetesVersion, tkrVersion)
 	patchJSONString := fmt.Sprintf(`{"spec": {"topology": {"version": "%v"}}}`, tkrVersion)
 
-	// Timeout set to 15 minutes because the continuousTKRDiscoverFreq for tkr-source-controller's fetcher is 10 minutes.
-	// Wait time should be longer than the fetcher's frequency of pulling tkrs.
-	pollOptions := &clusterclient.PollOptions{Interval: upgradePatchInterval, Timeout: 15 * time.Minute}
+	// Timeout set to 30 minutes because the continuousTKRDiscoverFreq for tkr-source-controller's fetcher is 10 minutes.
+	// And kapp package reconcile frequency is 10 minutes.
+	// Wait time should be longer than the fetcher's frequency of pulling tkrs plus the frequency of kapp package reconciliation.
+	pollOptions := &clusterclient.PollOptions{Interval: upgradePatchInterval, Timeout: 30 * time.Minute}
```
DanielXiao, line 28: Why do we need to increase timeout? We don't wait after we patch CAPA deploy, does this cause longer wait time here?

12345lcr: This is not related to the CAPA deploy, I just add a quick fix for another issue that we found from some corner cases that previously the increased time is not enough. 

Since the fix is very simple and I want it to be in the release also, I just combined that fix into this single mr.

tkg/client/upgrade_region.go
12345lcr: After sync with @DanielXiao , we found that there is no goroutine leak after `DoProvidersUpgrade` is finished. And `DoProvidersUpgrade` itself will make sure the provider is deployed successfully. So we decide to move `PatchClusterAPIAWSControllersToUseEC2Credentials` before `WaitForProviders` to reuse `WaitForProviders` for checking the patch result.

DanielXiao (approved):

alfredthenarwhal:
> @12345lcr:  /test upgrade-aws
Commit: 385fe30

Tests failed! Build no: 4622

PR #4555, opened 2023-04-07
Add missing labels for antreaconfig from clusterbootstrap

hangyan:

In the last version bump, antrea controller relies on a package label to working normally. But in cert cases(cb and antreaconfig created from kubectl, the label is missing). This patch added the missing label from clusterboostrap clone

Fixes # TKG-18782

This is also an alternative solution compared to https://github.com/vmware-tanzu/tanzu-framework/pull/4544

Test done in a Live tanzu env

### Release note

```release-note

```

### DIFF
--- a/addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go
+++ b/addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go
@@ -444,7 +444,7 @@ func (h *Helper) HandleExistingClusterBootstrap(clusterBootstrap *runtanzuv1alph
 		h.Logger.Error(err, fmt.Sprintf("unable to add cluster %s/%s as owner reference to providers", cluster.Namespace, cluster.Name))
 	}
 
-	if err := h.addPackageLabelForCNIProvider(cluster.Name, clusterBootstrap); err != nil {
+	if err := h.AddPackageLabelForCNIProvider(cluster.Name, clusterBootstrap); err != nil {
 		h.Logger.Error(err, fmt.Sprintf("unable to add labels to cni provider"))
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


addons/pkg/util/clusterbootstrapclone/clusterbootstrapclone.go
```diff
@@ -995,6 +1026,32 @@ func ownedByDifferentCluster(k8SObject *unstructured.Unstructured, cluster *clus
...
+		// We need to get and update, otherwise there could have concurrency issue: ["the object has been modified; please
+		// apply your changes to the latest version and try again"]
+		newChild, errGetProvider := h.DynamicClient.Resource(*gvr).Namespace(child.GetNamespace()).Get(h.Ctx, child.GetName(), metav1.GetOptions{})
+		if errGetProvider != nil {
+			h.Logger.Error(errGetProvider, fmt.Sprintf("unable to get %s %s/%s", child.GetKind(), child.GetNamespace(), child.GetName()))
+			return errGetProvider
+		}
+		newChild = newChild.DeepCopy()
+		newChild.SetLabels(labels)
+		_, errUpdateProvider := h.DynamicClient.Resource(*gvr).Namespace(newChild.GetNamespace()).Update(h.Ctx, newChild, metav1.UpdateOptions{})
```
wjun, line 1045: How about use patch to avoid update conflict?

hangyan: > 

Sounds better, although multiple similar methods use the same approach to update the resources in  the same file , maybe better to do a whole refactor later.

wjun: Will only this place update the labels? If so, patch will simplify and avoid conflict check.

hangyan: Sure, only in this controller. i will update to use patch method

hangyan: updated.

hangyan:
@wjun Another issue is that this is an alternative solution as https://github.com/vmware-tanzu/tanzu-framework/pull/4544. 4544 is a more general solution, it will update all the providers.

shyaamsn (approved):
LGTM. Please ensure all the tests in CAPA pipeline passes. Then this can be merged.

12345lcr (approved):
LGTM

shyaamsn:
We also need to update the label in the upgrade case. Need to call this same method to update label during upgrade here https://github.com/vmware-tanzu/tanzu-framework/blob/main/addons/controllers/clusterbootstrap_controller.go#L431

hangyan:
> We also need to update the label in the upgrade case. Need to call this same method to update label during upgrade here https://github.com/vmware-tanzu/tanzu-framework/blob/main/addons/controllers/clusterbootstrap_controller.go#L431

updated.

shyaamsn (approved):
LGTM

alfredthenarwhal:
> @shyaamsn:  /test install-vc7,upgrade-vc7
Commit: 82e8782

Tests failed! Build no: 4728

PR #4559, opened 2023-04-10
Add webhook certificate manager package to utils and use it for Featuregates controller

rajathagasthya:

This PR adds Certificate Manager which used to rotate certificates for a controller manager's webhooks. This in intended to replicate the functionality of certmanager for controllers who do not want to use an external dependency to manager self-signed certs.

TODO

- [ ] Tests
- [ ] Updated README

Fixes #

### Release note

```release-note

```


featuregates/controller/main.go
```diff
@@ -114,6 +136,40 @@ func main() {
 
 	//+kubebuilder:scaffold:builder
 
+	signalHanlder := ctrl.SetupSignalHandler()
```
codegold79, line 139: typo. Should be `signalHandler`

util/webhook/certs/certmanager.go
```diff
@@ -0,0 +1,265 @@
...
+	if !nextRotationTime.IsZero() && !nextRotationTime.Before(now) {
+		if !scheduledNextRotationTime.Equal(nextRotationTime) {
+			cm.opts.Logger.Info("Rescheduling next rotation", "nextRotationTime", nextRotationTime.String())
+			return nextRotationTime, nil
+		}
+	}
+
+	// Determine the rotation interval.
+	var rotationInterval time.Duration
+	if value := secret.Annotations[cm.opts.RotationIntervalAnnotationKey]; value != "" {
```
avi-08, line 109: If annotation value is empty, should we set nextRotationTime as now + defaultRotationInterval ?

packages/featuregates/bundle/config/values.yaml
```diff
@@ -1,6 +1,6 @@
 #@data/values
 ---
-namespace: tkg-system
+namespace: default
```
avi-08, line 3: IIUC, this change reflects our divergence from "TKG". In that case, should we have our own namespace, like "tanzu-runtime" maybe?

util/webhook/certs/README.md
avi-08, line 1: Nice doc 👍

packages/featuregates/bundle/config/values.yaml
rajathagasthya: I was just using this change to test locally. I should revert this. But it's not necessary to have our own namespace. We want product teams to define which namespace this should be installed in and set appropriate RBAC.

rajathagasthya:
Closing this in favor of a new PR against`feature/runtime-core`.

PR #4563, opened 2023-04-11
[WIP]Skip vip validation when vip is empty

chenlin07:

Skip vip validation when we enable avi ha provider and set empty vip
Since when we create many workload clusters at same time, it's expected to have many clusters whose control plane endpoint are empty at some time

Fixes #

### Release note

```release-note

```


chenlin07:
move PR to downstream

PR #4569, opened 2023-04-12
Add webhook certificate manager package to utils and use it for Featuregates controller

### LINKED ISSUE #4580: Allow webhook certs to be managed by the controllers themselves

This is part of an effort for removing cert-manager dependency by having utilities controller managers can use to manage webhook certs by themselves. This enables self contained releases of components such as Runtime Core.

Use an external dependency like cert-manager just to manage self signed certs. This is an overkill.

- (x) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Additional context**

rajathagasthya:

This PR adds Certificate Manager which used to rotate certificates for a controller manager's webhooks. This in intended to replicate the functionality of certmanager for controllers who do not want to use an external dependency to manager self-signed certs.

Fixes #4580 

Render `ytt` templates for installing featuregates controller.
```
$ ytt -f packages/featuregates/bundle/config/upstream -f packages/featuregates/bundle/config/values.yaml > featuregates.yaml
```

Apply the manifest to install featuregate controller.
```
$ kubectl apply -f featuregates.yaml
```

Watch featuregates controller logs for cert rotation messages and check if secrets and webhook configs are updated:
```
$ watch -n 5 "kubectl get secret tanzu-featuregates-webhook-server-cert -o jsonpath='{.data.ca\.crt}'
$ watch -n 5 "kubectl get validatingwebhookconfiguration tanzu-featuregates-validating-webhook-core -o jsonpath='{.webhooks[0].clientConfig.caBundle}'
```

### Release note

```release-note

```


util/webhook/certs/README.md
```diff
@@ -0,0 +1,189 @@
+# Certificate Manager Lite
```
sathyanarays, line 1: Is this `Certificate Manager Lite` being used in some other components?

util/webhook/certs/certmanager.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	for _, webhookConfig := range validatingWebhookList.Items {
+		for i := range webhookConfig.Webhooks {
+			webhookConfig.Webhooks[i].ClientConfig.CABundle = caCertData
+		}
+
+		if err := cm.opts.Client.Update(ctx, &webhookConfig); err != nil {
+			if !apierrors.IsConflict(err) {
+				return fmt.Errorf("failed to update validating webhook configuration %s/%s: %w", webhookConfig.Namespace, webhookConfig.Name, err)
```
sathyanarays, line 189: Why are we not continuing with other remaining webhooks even though the current one fails?

util/webhook/certs/certmanager.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	for _, webhookConfig := range mutatingWebhookList.Items {
+		for i := range webhookConfig.Webhooks {
+			webhookConfig.Webhooks[i].ClientConfig.CABundle = caCertData
+		}
+
+		if err := cm.opts.Client.Update(ctx, &webhookConfig); err != nil {
+			if !apierrors.IsConflict(err) {
+				return fmt.Errorf("failed to update mutating webhook configuration %s/%s: %w", webhookConfig.Namespace, webhookConfig.Name, err)
```
sathyanarays, line 207: Same as previous comment: Why are we not continuing with other remaining mutating webhooks even though the current one fails?

util/webhook/certs/certmanager.go
```diff
@@ -0,0 +1,275 @@
...
+	// Determine the rotation interval.
+	var rotationInterval time.Duration
+	if value := secret.Annotations[cm.opts.RotationIntervalAnnotationKey]; value != "" {
+		if rotationInterval, _ = time.ParseDuration(value); rotationInterval == 0 {
+			rotationInterval = defaultRotationInterval
+		}
+	}
+
+	// Update next rotation time and write it to the NextRotationAnnotationKey in the secret.
+	nextRotationTime = now.Add(rotationInterval)
```
avi-08, line 116: If annotation value is empty, should we set nextRotationTime as now + defaultRotationInterval ? Looking at L109 only executes if annotation has some value, otherwise L108 would render rotationInterval as 0s.

util/webhook/certs/README.md
rajathagasthya: This will be most likely used by addons manager as they have their own cert management solution currently. This is an attempt at unifying the code.

util/webhook/certs/certmanager.go
rajathagasthya: Good catch. I fixed this by removing the `value != ""` check.

util/webhook/certs/certmanager.go
```diff
@@ -0,0 +1,275 @@
...
+	opts Options
+}
+
+// Start starts certificate management for a controller manager's webhooks.
+// This method calls os.Exit when an error is encountered.
+func (cm *CertificateManager) Start(ctx context.Context) error {
+	go func() {
+		if err := cm.start(ctx); err != nil {
+			cm.opts.Logger.Error(err, "unexpected failure in starting certificate manager")
+			os.Exit(1)
```
sathyanarays, line 42: Is it okay to stop the controller even for transient network failures such as transient failure to reach the API server?

util/webhook/certs/certmanager.go
```diff
@@ -0,0 +1,275 @@
...
+	if len(secret.Annotations) == 0 {
+		secret.Annotations = map[string]string{}
+	}
+
+	// If this is the first time rotate is called since starting certificate manager, then initialize the webhooks with
+	// any CA data (if any) in the secret. This is for the case when a package is redeployed (for any reason) and the
+	// webhook config properties are reset.
+	if scheduledNextRotationTime.IsZero() {
+		if caCertData := secret.Data[CACertName]; len(caCertData) > 0 {
+			if err := cm.updateWebhookConfigs(ctx, caCertData); err != nil {
```
sathyanarays, line 90: Is this okay to blindly update the webhook config with the cert-data? What if the contents of the secret has proper fields defined but not a valid certificate? What is the certificate present has expired?

util/webhook/certs/certmanager.go
sathyanarays: Avi's argument holds true if the annotation itself is not present?

avi-08 (approved):

util/webhook/certs/certmanager.go
rajathagasthya: That is also taken care of by the way Go handles missing keys in maps.

```go
value := secret.Annotations[cm.opts.RotationIntervalAnnotationKey]
```

`value` will have a zero-value (`""`) if the key is missing and `rotationInterval` will be parsed to `0`.

util/webhook/certs/certmanager.go
rajathagasthya: Yes, it's okay because the deployment will make sure pod is restarted when it exits.

util/webhook/certs/certmanager.go
rajathagasthya: The chances of expiry are low between rotation intervals and restarts. But maybe an additional validity check here is useful. Let me see how to do that.

sathyanarays (approved):

util/webhook/certs/certmanager.go
rajathagasthya: Thinking about this more, I don't believe it's necessary to add checks here. Here's why:

1. We don't expect anyone to manually edit the webhook secret. So we can be fairly sure that whatever cert is in there was created by certificate manager.
2. Because of 1), the webhook secret will always contains the next rotation time annotation. Even if the certificate has expired, the subsequent code will check the next rotation and rotate the certs.

featuregates/controller/main.go
```diff
@@ -25,11 +25,18 @@ import (
...
+	"github.com/vmware-tanzu/tanzu-framework/util/webhook/certs"
 )
 
 var (
-	scheme   = runtime.NewScheme()
-	setupLog = ctrl.Log.WithName("setup")
+	scheme                              = runtime.NewScheme()
+	setupLog                            = ctrl.Log.WithName("setup")
+	defaultWebhookConfigLabel           = "tanzu.vmware.com/featuregates-webhook-managed-certs=true"
+	defaultWebhookServiceNamespace      = "tkg-system"
```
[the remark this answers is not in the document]
rajathagasthya: It's just a default for the flag. I can change to `default` if that makes more sense.

util/webhook/certs/constants.go
```diff
@@ -0,0 +1,24 @@
+// Copyright 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package certs
+
+import "time"
+
+const (
+	// CACertName is the name of the CA certificate.
+	CACertName = "ca.crt"
```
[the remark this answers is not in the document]
rajathagasthya: Yes, in case a consumer needs to access secret data they can use this key.

yharish991 (approved):
LGTM!! can you attach an issue and add testing done in the pr description?

PR #4582, opened 2023-04-19
Readiness controller re-conciliation

sathyanarays:
This PR has adds the logic to re-concile the readiness resources that is triggered by changes in the readiness and readiness provider resources.

Fixes #

Added envtest which providers maximum possible coverage

### Release note

```release-note

```

### DIFF
--- a/readiness/controller/pkg/readiness/readiness_controller.go
+++ b/readiness/controller/pkg/readiness/readiness_controller.go
@@ -9,12 +9,19 @@ import (
 
 	"github.com/go-logr/logr"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	corev1alpha2 "github.com/vmware-tanzu/tanzu-framework/apis/core/v1alpha2"
 )
+
+const contextTimeout = 30 * time.Second
 
 // ReadinessReconciler reconciles a Readiness object
 type ReadinessReconciler struct {
@@ -28,26 +35,21 @@ type ReadinessReconciler struct {
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
-// TODO(user): Modify the Reconcile function to compare the state specified by
-// the Readiness object against the actual cluster state, and then
-// perform operations to make the cluster state reflect the state specified by
-// the user.
-//
-// For more details, check Reconcile and its Result here:
-// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.6.4/pkg/reconcile
 func (r *ReadinessReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	ctxCancel, cancel := context.WithTimeout(ctx, contextTimeout)
+	defer cancel()
+
 	readiness := &corev1alpha2.Readiness{}
-	err := r.Client.Get(ctx, req.NamespacedName, readiness)
+	err := r.Client.Get(ctxCancel, req.NamespacedName, readiness)
 	if err != nil {
-		r.Log.Error(err, "Error while fetching readiness")
 		return ctrl.Result{}, err
 	}
 
 	updateRequired := false
-	oldStatus := readiness.Status
+	currentStatus := readiness.Status
 
 	if len(readiness.Spec.Checks) == 0 {
-		if !oldStatus.Ready {
+		if !currentStatus.Ready {
 			readiness.Status.Ready = true
 			readiness.Status.CheckStatus = []corev1alpha2.CheckStatus{}
 			updateRequired = true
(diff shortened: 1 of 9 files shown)


readiness/controller/pkg/readiness/readiness_controller.go
```diff
@@ -34,14 +36,142 @@ type ReadinessReconciler struct {
...
+		if err != nil {
+			r.Log.Error(err, "Error while updating readiness status")
+			return ctrl.Result{}, err
+		}
+	}
 
 	return ctrl.Result{}, nil
 }
 
+func isUpdateRequired(oldStatus, newStatus corev1alpha2.ReadinessStatus) bool {
```
avi-08, line 123: Why do we need this? Even if there is no change in the status, lastComputedTime can still be changed to this iteration since `lastComputedTime` indicates the latest instance when we "refreshed" the readiness resource.

readiness/controller/pkg/readiness/readiness_controller.go
```diff
@@ -34,14 +36,142 @@ type ReadinessReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.6.4/pkg/reconcile
 func (r *ReadinessReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = context.Background()
```
avi-08, line 37: Context should be initialized with timeout here and passed along wherever needed. This helps in avoiding network calls that may take forever.

```
ctxCancel, cancel := context.WithTimeout(ctx, constants.ContextTimeout)
defer cancel()
```

readiness/controller/pkg/readiness/readiness_controller.go
```diff
@@ -34,14 +36,142 @@ type ReadinessReconciler struct {
...
+
+		for _, check := range readiness.Spec.Checks {
+			checkStatus := corev1alpha2.CheckStatus{
+				Name:      check.Name,
+				Providers: make([]corev1alpha2.Provider, 0),
+				Ready:     false,
+			}
+
+			indices, ok := allChecks[check.Name]
+			if ok {
```
avi-08, line 81: Can be combined with L81
```suggestion
			if indices, ok := allChecks[check.Name]; ok {
```

readiness/controller/pkg/readiness/readiness_controller.go
sathyanarays: We need to make the re-conciliation converge. Otherwise, it keeps on re-conciling till the controller process exits. So, we need this condition to converge.

readiness/controller/pkg/readiness/readiness_controller.go
```diff
@@ -34,14 +36,142 @@ type ReadinessReconciler struct {
...
+	readiness := &corev1alpha2.Readiness{}
+	err := r.Client.Get(ctx, req.NamespacedName, readiness)
+	if err != nil {
+		r.Log.Error(err, "Error while fetching readiness")
+		return ctrl.Result{}, err
+	}
 
-	// TODO(user): your logic here
+	updateRequired := false
+	oldStatus := readiness.Status
```
rajathagasthya, line 47: What's the reason behind naming this `oldStatus`? This is the object that's current.

rajathagasthya (changes requested):

readiness/controller/pkg/readiness/readiness_controller.go
sathyanarays: Exactly, we need this function to make the reconciliation idempotent. If we unconditionally update the `lastComputedTime` in every reconciliation cycle, that will trigger another reconiliation and it in turn will trigger another status update and so on. We need to stop updating the status if we don't have any diff between the previous status and the current computed status.

sathyanarays: The job controller in controller manager does similar. It checks the newly computed state and the current state to decide if an update is required. https://github.com/kubernetes/kubernetes/blob/0afa60086a0fc0b5bf6386e16732a373ad9a10b6/pkg/controller/job/job_controller.go#L871

readiness/controller/pkg/readiness/readiness_controller.go
```diff
@@ -34,14 +36,142 @@ type ReadinessReconciler struct {
...
+		for _, checkStatus := range readiness.Status.CheckStatus {
+			readiness.Status.Ready = readiness.Status.Ready && checkStatus.Ready
+		}
+
+		updateRequired = isUpdateRequired(oldStatus, readiness.Status)
+
+	}
+
+	if updateRequired {
+		time := metav1.Time{Time: time.Now()}
```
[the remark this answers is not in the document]
sathyanarays: Done

readiness/controller/pkg/readiness/readiness_controller.go
sathyanarays: Done

readiness/controller/pkg/readiness/readiness_controller.go
```diff
@@ -34,14 +36,142 @@ type ReadinessReconciler struct {
...
+		if !oldStatus.Ready {
+			readiness.Status.Ready = true
+			readiness.Status.CheckStatus = []corev1alpha2.CheckStatus{}
+			updateRequired = true
+		}
+	} else {
+		allChecks := make(map[string][]int)
+		readiness.Status.CheckStatus = []corev1alpha2.CheckStatus{}
+		providersList := &corev1alpha2.ReadinessProviderList{}
+		err = r.Client.List(ctx, providersList)
```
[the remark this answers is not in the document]
sathyanarays: Done. But, this can be improved further. I will take improvement as a separate change.

readiness/controller/pkg/readiness/readiness_controller.go
```diff
@@ -34,14 +36,142 @@ type ReadinessReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.6.4/pkg/reconcile
 func (r *ReadinessReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = context.Background()
-	_ = r.Log.WithValues("readiness", req.NamespacedName)
+	readiness := &corev1alpha2.Readiness{}
+	err := r.Client.Get(ctx, req.NamespacedName, readiness)
+	if err != nil {
+		r.Log.Error(err, "Error while fetching readiness")
```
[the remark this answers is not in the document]
sathyanarays: Done

readiness/controller/pkg/readiness/readiness_controller.go
sathyanarays: Changed

readiness/controller/pkg/readiness/readiness_controller.go
sathyanarays: Done

readiness/controller/pkg/readiness/readiness_controller.go
```diff
@@ -34,14 +36,142 @@ type ReadinessReconciler struct {
...
+
+	}
+
+	if updateRequired {
+		time := metav1.Time{Time: time.Now()}
+		readiness.Status.LastComputedTime = &time
+
+		err = r.Client.Status().Update(ctx, readiness)
+		if err != nil {
+			r.Log.Error(err, "Error while updating readiness status")
```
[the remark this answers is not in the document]
sathyanarays: Done

readiness/controller/pkg/readiness/readiness_controller.go
rajathagasthya: You can ignore status updates in your reconciliation using predicates, if that's only thing you're worried about.

rajathagasthya: For future discussion: I don't know if `lastComputedTime` is really that useful. Instead, we could requeue the request every 1m (or pick a good number) and reconcile. That would guarantee it's not very stale.

I don't think any consumer would check the status and check if `lastComputedTime` was recent.

PR #4587, opened 2023-04-20
feat(AzureEnvironments): use upstream func for validation

### LINKED ISSUE #3162: Dynamically assign AD endpoints for Azure, support AzureStack
By default, `go-autorest` allows downstream consumers, such as `tanzu-framework` to lookup various endpoints. Presently, `tanzu-framework` [explicitly sets the Active Directory endpoints for a subset of Azure clouds](https://github.com/vmware-tanzu/tanzu-framework/blob/main/pkg/v1/tkg/azure/client.go#L109) (referencing `go-autorest`) while inadvertently crippling the ability to deploy to AzureStack, which mandates the setting of endpoints via an additional configuration file. 

By dynamically assigning the endpoints, rather than defining the endpoints for a subset of clouds, we forgo the need to contain that logic in `tanzu-framework`. We also could eliminate the constants that represent the Azure cloud monikers, which are also present in `go-autorest`. 

The current set of unit tests compare the local Active Directory values in `tanzu-framework` against `go-autorest`, so those would ultimately need to be amended or removed if eliminating the additional logic from `azure/client.go`. 

Currently folks that are attempting to build in secret/top-secret regions are incapable of doing so with the Tanzu CLI because of the limitations that have been mentioned here. Those Azure clouds follow the same pattern as AzureStack, which mandates custom endpoints (and thus the ability to supply an Azure configuration file). [Supplying a custom configuration file is already a first-class citizen in `go-autorest`](https://github.com/Azure/go-autorest/blob/main/autorest/azure/environments.go#L301), so changes would simply need to be made to `tanzu-framework` to support this capability.

Several customers have forked `tanzu-framework` and built a custom CLI to support their POC efforts. Not a reasonable alternative (from a UX perspective) for licensed users, but an alternative nonetheless. 

- ( ) APIs
- ( ) Addons
- (X) CLI
- (X) Docs
- ( ) IAM
- (X) Installation
- ( ) Plugin
- ( ) Security
- ( ) Test and Release
- (X) User Experience
- ( ) Developer Experience

**Additional context**
Please let me know if this is something that you'd be open to a pull request for.

JefeDavis:

removes hard-coded maps for Azure environments and
calls the EnvironmentFromName function from go-autorest/azure library to leverage upstream provider to validate Azure Environments. Also enables initial ability to specify custom endpoints for AzureStackCloud by using the Environment 'AzureStackCloud', specifying AZURE_ENVIRONMENT_FILEPATH, and providing an AzureEnvironment json file

Fixes #3162

Unit tests have been added to cover this change

### Release note

```release-note
Fixes validation handling when attempting to deploy to AzureStackCloud custom endpoints
```


JefeDavis:
@mminges @conzetti

PR #4609, opened 2023-05-02
Update Build Tooling for Integration build script and Bump Go Version to 1.19

### LINKED ISSUE #4597: Copy over latest Build Tooling (v0.0.3) for Integration build script
 Copy over latest Build Tooling for Integration build script because it is outdated and doesn't currently match what is in [Build tooling for Integrations](https://github.com/vmware-tanzu/build-tooling-for-integrations).

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- (x) Plugin
- ( ) Security
- (x) Test and Release
- ( ) User Experience
- (x) Developer Experience

**Additional context**

### LINKED ISSUE #4598: Update/Bump Go Version to 1.19
Go is currently at 1.18 and should be updated to 1.19.

- (x) APIs
- (x) Addons
- (x) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- (x) Plugin
- ( ) Security
- ( ) Test and Release
- ( ) User Experience
- (x) Developer Experience

**Additional context**

codegold79:
Copy over latest Build Tooling for Integration build script and Update/Bump Go Version to 1.19
- includes building, testing, and linting Tanzu CLI plugins
- improvements to existing build targets
- upgrade Go version to 1.19 in all modules

We need this PR because the [Build Tooling for Integrations templates/Makefile](https://github.com/vmware-tanzu/build-tooling-for-integrations/blob/main/templates/Makefile) was updated and therefore needs to be updated in this repo.

Fixes #4597
Fixes #4598 

### Release note

```release-note
Update Go version to 1.19 and build script
```


.golangci.yaml
```diff
@@ -113,6 +113,12 @@ issues:
     - '(Expect directory permissions to be 0750 or less|Expect (WriteFile|file) permissions to be 0600 or less)'
 
   exclude-rules:
+    - path: pinniped/internal/mocks
+      linters:
+        - goimports
+    - path: apis/run/v1alpha1
+      linters:
+        - goimports
```
rajathagasthya, line 121: These are not valid paths anymore.

.golangci.yaml
```diff
@@ -124,6 +130,11 @@ issues:
         - gocritic
       text: "unnecessaryDefer:"
 
+    - path: pkg/v1/tkr/controllers/source/suite_test.go
```
rajathagasthya, line 133: Ditto.

.golangci.yaml
codegold79: Good catch. Sorry about the rebase. I shoud've put the change in a separate commit, but got carried away with clean up. These paths are no longer in the rebased PR.

.golangci.yaml
codegold79: I got rid of this in rebase (I should've done a separate commit, made a mistake and rebased).

featuregates/controller/pkg/featuregate/featuregate_controller.go
```diff
@@ -82,7 +82,7 @@ func (r *FeatureGateReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
-func (r *FeatureGateReconciler) toFeatureGateRequests(o client.Object) []reconcile.Request {
+func (r *FeatureGateReconciler) toFeatureGateRequests(_ client.Object) []reconcile.Request {
```
avi-08, line 85: If we're not using this parameter, it can be removed from the method signature.

hack/verify-dirty.sh
```diff
@@ -29,6 +29,7 @@ echo "Done vendir sync!"
 # because the package-vendir-sync operation renders
 # the pakcage directory contents inaccessible to the
 # host machine user.
+GITHUB_ACTIONS="${GITHUB_ACTIONS:-}"
```
avi-08, line 32: Thanks for catching this. This can be done in line 34 as well:
`if [[ -n "${GITHUB_ACTIONS:-}" ]]; then`

featuregates/controller/pkg/featuregate/featuregate_controller.go
codegold79: @rajathagasthya, do you think I can just delete the client.Object variable? I'm not that familiar with how these watches are registered and I think it will break:

```
// SetupWithManager sets up the controller with the Manager.
func (r *FeatureGateReconciler) SetupWithManager(mgr ctrl.Manager) error {
	return ctrl.NewControllerManagedBy(mgr).
		For(&configv1alpha1.FeatureGate{}).
		Watches(
			&source.Kind{Type: &configv1alpha1.Feature{}},
			handler.EnqueueRequestsFromMapFunc(r.toFeatureGateRequests)).
		Watches(
			&source.Kind{Type: &corev1.Namespace{}},
			handler.EnqueueRequestsFromMapFunc(r.toFeatureGateRequests)).
		Complete(r)
}
```

hack/verify-dirty.sh
codegold79: I like that suggestion. I'll put it in a commit.

featuregates/controller/pkg/featuregate/featuregate_controller.go
codegold79: I think messing with featuregate controller code is outside the scope of this PR. I only changed `o` to `_` to pass lint. To remove the parameter, I'd need to dig into the code and actually do controller tests manually. I suggest someone catch it and change it next time they work on featuregate since they'll have to do testing at that time, anyway.

yharish991: this function needs to be of type [MapFunc](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/handler/enqueue_mapped.go#L30) and it expects client.object as the parameter

yharish991 (approved):
lgtm

featuregates/controller/pkg/featuregate/featuregate_controller.go
codegold79: Thanks, @yharish991. I'll keep the method signature the same, then. No action.

rajathagasthya (approved):

avi-08 (approved):

PR #4614, opened 2023-05-05
Enable verify action

### LINKED ISSUE #4608: `make package-vendir-sync` does not work in Github Actions
**Bug description**

`make verify` fails in Github `verify` action. It calls `make package-vendir-sync` which seems to have trouble with mounted volumes on Github runners (probably because it's doing docker-in-docker stuff).

Logs from one of the failed jobs: https://github.com/vmware-tanzu/tanzu-framework/actions/runs/4833779855/jobs/8614195531

This failure must be fixed and the `verify` action needs to be restored.

- ( ) APIs
- ( ) Addons
- ( ) CLI
- ( ) Docs
- ( ) IAM
- ( ) Installation
- ( ) Plugin
- ( ) Security
- (x) Test and Release
- ( ) User Experience
- ( ) Developer Experience

**Expected behavior**

No failures and verification of uncommitted files is done in actions.

avi-08:
Fixes `permission denied` failures when running `git diff` after `make package-vendir-sync` target.

Fixes #4608 

`Verify` action in this PR is completed successfully.

### Release note

```release-note

```

### DIFF
--- a/hack/verify-dirty.sh
+++ b/hack/verify-dirty.sh
@@ -8,36 +8,39 @@ echo "#############################"
 echo "Verify uncommitted files..."
 echo "#############################"
 
-# Option to ignore changes to compatibility image path in
-# tkg/tkgconfigpaths/zz_bundled_default_bom_files_configdata.go and plugin runtime version in
-# cli/runtime/version/zz_generated_plugin_runtime_version.go. This is to support `usebom` directive
-# in tests.
-ignore_files=(':(exclude)tkg/tkgconfigpaths/zz_bundled_default_bom_files_configdata.go' ':(exclude)cli/runtime/version/zz_generated_plugin_runtime_version.go')
-# Temporarily excluding UI generated bindata file from verify (zz_generated.bindata.go). Currently running into issues
-# blocking the CI main build. CI generated bindata is different from bindata file generated on local
-# developer machines, causing this failure. Need to root cause and then remove this exclusion.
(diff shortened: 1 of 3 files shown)


hack/verify-dirty.sh
```diff
@@ -8,36 +8,39 @@ echo "#############################"
...
-if ! (git diff --quiet HEAD -- . "${ignore_files[@]}" "${ignore_file_ui_bindata}"); then
+echo "Done vendir sync!"
+
+# This is required for github actions execution
+# because the package-vendir-sync operation renders
+# the pakcage directory contents inaccessible to the
+# host machine user.
+if [[ -n "${GITHUB_ACTIONS}" ]]; then
+  echo "Resetting directory ownership"
+  sudo chown -R $(id -nu):$(id -ng) .;
```
rajathagasthya, line 34: This is a great find! Is it possible to do this step in the workflow yaml instead of here? I found https://github.com/marketplace/actions/reset-workspace-ownership-action, but we could do this ourselves in a `step`?

I also found this: https://github.com/marketplace/actions/reset-workspace-ownership-action

avi-08: It'd require an amount of refactoring effort in the verify target flow, since this step needs to be done exactly after `package-vendir-sync` and before running `git diff`. Having it this way is simpler and also provides consistent behaviour if run on developer machines.

yharish991: I'm okay with the current approach..

yharish991 (approved):

